PR #442, opened 2012-08-19
Fix saving PDFs in Wiley Online Library when translation occurs via connectors

simonster:
On each abstract/full text page, Wiley Online Library includes a link that points to a PDF. This URL is the same across browsers, but the content served from it is not. In most browsers, Wiley serves an HTML page with an iframe that contains the PDF. However, for the browsers whose User-Agent string contains "Safari/", Wiley instead serves the PDF directly, with no iframe. The same URL thus points to different content depending on the browser.

This patch takes this behavior into account by inspecting the browser's User-Agent string. If it contains "Safari/", it directly attaches the PDF link URL. Otherwise, it loads the iframe and pulls the PDF link out of it.

To work properly with Zotero Standalone, this requires a current 3.0 branch build (post- zotero/zotero@c81c9478d9fc3cb71b482edcd89e5ec6ce19d923), so that Zotero Standalone uses the browser's User-Agent when retrieving the PDF.

First reported at http://forums.zotero.org/discussion/22409/wiley-online/#Item_9

(New pull request because the commit message was truncated in the previous one.)


adam3smith:
@dstillman @adomasven I have no way of testing if this is still an issue (in which case we should fix) or if Wiley has stopped doing this. Could you check?

adomasven:
Testing on http://onlinelibrary.wiley.com/doi/10.1002/anie.201400837/abstract works on a Mac with Safari and Chrome

adam3smith:
Thanks for confirming. Closing

PR #477, opened 2012-10-08
Preserve capitalized words in BibTeX export

andreas-h:
Following the discussion at http://forums.zotero.org/discussion/25362/capitalize-title-in-bibtex-export/, I made this quick'n dirty patch. Seems to work fine for me.


aurimasv:
Fixed in #813 available now

PR #612, opened 2013-08-23
Zotero Translators for medical library's website from Lyon hospital

FredRiondet:
Translators work for these sites:
http://rbh.docchu-lyon.fr/
http://documentationcentrale.docchu-lyon.fr/
http://www.nosobase-biblio.cclin-france.fr/

The "HospicesCivilsdeLyon" file is a fork of the translator "Library Catalog (Koha)" because our database uses Koha but to specific way:
- specific itemtypes
- different default value for type
- specific targets.

The MARC schema describe into the "MARC_HospicesCivilsdeLyon" file  is specific to this database:
- UNIMARC schema only
- New fields Added
- Specific items types
- Specific "extra"

### DIFF
--- a/HospicesCivilsdeLyon
+++ b/HospicesCivilsdeLyon
@@ -1,19 +1,19 @@
 {
-	"translatorID": "6618c8fb-5419-403d-90ab-1836ccc74905",
+	"translatorID": "26ab1a2b-8740-4f79-a3e7-e0df62818839",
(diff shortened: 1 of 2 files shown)


adam3smith:
I'm really sorry this never got properly reviewed, but in any case it looks like the sites changed, so this is now moot.

Issue #285, opened 2012-04-05
DataCite DOI search translator


aurimasv:
We can use this to supplement CrossRef lookups that fail. I'm also thinking that we can add a search function to the DOI translator (i.e. DOI.js) that will iterate through all the DOI Registration Agencies that we can handle and use that for DOI lookups.

The API is available at http://oai.datacite.org

I will probably implement this in the next few days if no one has anything to say about it.

PR #632, opened 2013-10-24
[Do not merge] DOI search overhaul

aurimasv:
**This still needs some polishing and it will require a couple patches to the Zotero client as well as fixing any translators that are currently using DOI searches** Before proceeding, I would like to know if this will be an acceptable solution.

We want to add support for as many DOIs as we can, which means that we will need 7-8 different search translators to cover all DOIs. Using the current system, we would set priorities for each of these translators and have Zotero query each one in order. Most of our DOIs are for CrossRef and, from the discussion on the forums, the second largest set (though probably quite small compared to CrossRef) would come from DataCite. We could probably do an ok job ranking these with priorities. The biggest problem is that for invalid DOIs, we would pretty much end up querying every single one of the translators, which means that it would go through at least 7-8 http requests for invalid DOIs.

In comes [CrossRef's RA lookup tool](http://www.crossref.org/crweblog/2013/05/find_the_registration_agency_f.html). Using this tool we can intelligently decide which DOIs should be queried through which DOI search translator, thus avoiding all the unnecessary HTTP requests.

Bonus points: with this patch, we can (in theory so far) query CrossRef with up to 10 DOIs in a single HTTP request.

For more details, I'm copying this from the [DOI] translator commit message:

The DOI search translator serves as a dispatcher for other DOI translators:
1. First, it uses CrossRef to look for a DOI, since most DOIs we deal with are from CrossRef
2. If not found, it queries CrossRef's RA lookup tool to determine which RA is responsible for the DOI
3. If available, it uses the appropriate DOI search translator to query the DOI

DOI search translator takes a string (corresponding to a DOI), an item containing a DOI property, or an array containing a combination of these.

The search translator will call `item.complete()` for each DOI queried. `item._query` will be set to the original query corresponding to the resulting item. `item._status` will be a string corresponding to the status of the query. If successful, the status will be 'success'. If the search fails for one of a number of reasons, the status will indicate the reason. Some possible values for `item._status` are 'not found', 'invalid', 'unimplemented', and 'fail'. These are not set in stone.

DOI search translator will pass an array containing the original queries (filtered depending on RA) to the downstream DOI translators. The queries will contain valid DOIs (except those passed to CrossRef) either as literal strings or objects containing DOI properties (or a combination thereof). However, DOI search translators should not assume queries to be valid, because they can be invoked directly from other translators.

DOI search translator should server as the only DOI translator responding to `detectSearch` for DOI queries. All other DOI search translators should not return `true` for these queries. This behavior is overridden (inverted) by setting `directSearch` property to `true` on the query object.

These changes depend on the ability to pass objects associated with items between translators. Implemented in zotero/zotero#401

Replaces #452 Implements #285

Edit: some sample DOIs http://doi.crossref.org/doiRA/10.2353/jmoldx.2009.090037,10.6220/joq.2012.19(1).01,10.5240/6F7E-EF59-329B-1F0A-8440-2,10.12763/ONA1045

### DIFF
--- a/Airiti.js
+++ b/Airiti.js
@@ -2,30 +2,19 @@
 	"translatorID": "5f0ca39b-898a-4b1e-b98d-8cd0d6ce9801",
 	"label": "Airiti",
 	"creator": "Aurimas Vinckevicius",
-	"target": "https?://([^/]*\\.)?airitilibrary.com/searchdetail.aspx",
+	"target": "https?://(?:[^/]+\\.)?airitilibrary\\.com/Publication/alDetailedMesh",
 	"minVersion": "3.0",
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 12,
 	"browserSupport": "gcsib",
-	"lastUpdated": "2013-03-21 05:23:28"
+	"lastUpdated": "2013-10-23 06:56:26"
 }
 
 function detectWeb(doc, url) {
-	var icon = ZU.xpathText(doc, '//div[@id="main_gcs7"]//tbody/tr[2]\
-									//div[starts-with(@class, "icon_")]/@class');
-	if(!icon) return;
-
-	switch(icon) {
-		case "icon_T2":
-			return "thesis";
-		case "icon_J2":
-			return "journalArticle";
-		case "icon_C2":
-			return "conferencePaper";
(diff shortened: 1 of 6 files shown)


adam3smith:
We're effectively doing this now, so closing

PR #680, opened 2014-02-24
adding Google Play for apps

avram:
Does not cover the books & music side of the business.


avram:
Thanks for wrapping up the work on this one!

PR #789, opened 2014-10-26
Target regexp and priority cleanup

aurimasv:
- Make sure periods are always escaped
- Where appropriate, always start with "^https?://"
- For optional subdomains, always use "([^/]*\\.)?", for mandatory subdomains, use "[^/]+\\."
- Remove all instances of non-capturing parentheses and replace with capturing parentheses. This doesn't matter in target regexp and makes it a bit cleaner.
- Adjust priorities based on #765 (probably not exhaustive)

Possibly unnecessary and I'm not sure how great such a huge amount of translator updates is on Zotero servers. I'm going to do a general cleanup of `processDocuments`, `doGet`, `doPost` calls to use relative or at least protocol-relative URLs because of breakage in Chrome, so this should probably wait until the next commit.


zuphilip:
Not all of the points might be already dealt in #1211 or other commits.

I didn't look at:
> * For optional subdomains, always use "([^/]*\.)?", for mandatory subdomains, use "[^/]+\."

> use relative or at least protocol-relative URLs

> adjust all translators to convert titles to sentence case instead of title case if they are in all-upper case.

If anyone wants to continue on something here, then I would simply suggest to open a new issue.

PR #800, opened 2014-11-10
translator for Dryad Digital Repository

rnathanday:
This translator for Dryad Digital Repository biases the translator toward the embedded metadata.

### DIFF
--- a/Dryad Digital Repository.js
+++ b/Dryad Digital Repository.js
@@ -1,0 +1,534 @@
+{
+	"translatorID": "7a81d945-7d9c-4f8c-bd7b-4226c1cab40e",
+	"label": "Dryad Digital Repository",
+	"creator": "Nathan Day",
+	"target": "^https?://(?:www\\.)?datadryad\\.org/",
+	"minVersion": "1.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "g",
+	"lastUpdated": "2015-06-22 20:33:22"
+}
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2014 Dryad Digital Repository
+
+	This file is part of Zotero.
+
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+function detectWeb(doc, url) {
+	// Dryad search page
+	var multiples = ZU.xpath(doc, '//li[contains(@class,"ds-artifact-item")]/div/a');
+	if ((new RegExp(/^https?:\/\/(www\.)?datadryad\.org\/discover/)).test(url) &&
+		multiples !== null && multiples.length > 0)
+	{
+		return 'multiple';
+	} else {
+		var result = ZU.xpathText(doc,'//meta[@name="DC.type"][1]/@content');
+		// Dryad data package
+		if (result === 'Article') {
+			return 'journalArticle';
+		// Dryad data file
+		} else if (result === 'Dataset') {
+			//return 'dataset';
+			return 'journalArticle';
+		}
+	}
+	return false;
+}
+
+function doWeb(doc, url) {
+	var itemType = detectWeb(doc, url);
+	if (itemType === 'journalArticle') {
+		var package_doi = ZU.xpathText(doc, '//meta[@name="DCTERMS.isPartOf"]/@content');
+		var translator = Zotero.loadTranslator('import');
+		// use the Embedded Metadata translator
+		translator.setTranslator('951c027d-74ac-47d4-a107-9c3069ab7b48');
+		translator.setHandler('itemDone', function(doc, item) {
+			var itemDoi   	= item.url.replace(/^.*?doi:/, '');
+			item.DOI      	= itemDoi;
+			item.itemID   	= itemDoi;      // internal value for seeAlso relation
+			item.url      	= '';           // the doi value is to be used in lieu of the url
+			item.itemType 	= itemType;
+			// Add a .seeAlso value if this is a page for a single data file,
+			// rather than for a data package collection.
+			// This is a data file if there is a DCTERMS.isPartOf property.
+			if (package_doi && package_doi !== 'doi:'.concat(itemDoi)) {
+				item.seeAlso = ['http://dx.doi.org/' + package_doi.substring(4)];
+			}
+			// signal a dataset until supported per
+			// https://www.zotero.org/support/dev/translators/datasets
+			item.extra = '{:itemType: dataset}';
+			item.archive = 'Dryad Digital Repository';
+			item.attachments = [];
+			item.shortTitle = '';
+			item.complete();
+		});
+		translator.getTranslatorObject(function (obj) {
+			obj.doWeb(doc, url);
+		});
+		// Value for page's possible DCTERMS.isPartOf <meta> tag,
+		// minus the initial 'doi:'.
+		// This value is used to set the seeAlso value for a data-file record.
+		if (package_doi) {
+			ZU.processDocuments('http://datadryad.org/resource/' + package_doi, doWeb);
+		}
+	} else if (itemType === 'multiple') {
+		var urls = ZU.xpath(doc, '//li[contains(@class,"ds-artifact-item")]/div/a');
+		if (urls.length > 0) {
+			for (var i = 0; i < urls.length; i++) {
+				urls[i] = urls[i].href;
+			}
+			ZU.processDocuments(urls, function (d) {
+				doWeb(d, d.location.href);
+			});
+		}
+	}
+}
+/** BEGIN TEST CASES **/
+var testCases = [
+	{
+		"type": "web",
+		"url": "http://datadryad.org/resource/doi:10.5061/dryad.9025",
+		"items": [
+			{
+				"itemType": "journalArticle",
+				"title": "Data from: Phylogenetic analyses of mitochondrial and nuclear data in haematophagous flies support the paraphyly of the genus Stomoxys (Diptera: Muscidae)",
+				"creators": [
+					{
+						"firstName": "Najla",
+						"lastName": "Dsouli",
+						"creatorType": "author"
+					},
+					{
+						"firstName": "Frédéric",
+						"lastName": "Delsuc",
+						"creatorType": "author"
+					},
+					{
+						"firstName": "Johan",
+						"lastName": "Michaux",
+						"creatorType": "author"
+					},
+					{
+						"firstName": "Eric",
+						"lastName": "De Stordeur",
+						"creatorType": "author"
+					},
+					{
+						"firstName": "Arnaud",
+						"lastName": "Couloux",
+						"creatorType": "author"
+					},
+					{
+						"firstName": "Michel",
+						"lastName": "Veuille",
+						"creatorType": "author"
+					},
+					{
+						"firstName": "Gérard",
+						"lastName": "Duvallet",
+						"creatorType": "author"
+					}
+				],
+				"date": "2011-02-13",
+				"DOI": "10.5061/dryad.9025",
+				"abstractNote": "The genus Stomoxys Geoffroy (Diptera; Muscidae) contains species of parasitic flies that are of medical and economic importance. We conducted a phylogenetic analysis including 10 representative species of the genus including multiple exemplars, together with the closely related genera Prostomoxys Zumpt, Haematobosca Bezzi, and Haematobia Lepeletier & Serville. Phylogenetic relationships were inferred using maximum likelihood and Bayesian methods from DNA fragments from the cytochrome c oxidase subunit I (COI, 753 bp) and cytochrome b (CytB, 587 bp) mitochondrial genes, and the nuclear ribosomal internal transcribed spacer 2 (ITS2, 426 bp). The combination of mitochondrial and nuclear data strongly supports the paraphyly of the genus Stomoxys because of the inclusion of Prostomoxys saegerae Zumpt. This unexpected result suggests that Prostomoxys should be renamed into Stomoxys. Also, the deep molecular divergence observed between the subspecies Stomoxys niger niger Macquart and S. niger bilineatus Grünbreg led us to propose that they should rather be considered as distinct species, in agreement with ecological data. Bayesian phylogenetic analyses support three distinct lineages within the genus Stomoxys with a strong biogeographical component. The first lineage consists solely of the divergent Asian species S. indicus Picard which appears as the sister-group to all remaining Stomoxys species. The second clade groups the strictly African species Stomoxys inornatus Grünbreg, Stomoxys transvittatus Villeneuve, Stomoxys omega Newstead, and Stomoxys pallidus Roubaud. Finally, the third clade includes both African occurring and more widespread species such as the livestock pest Stomoxys calcitrans Linnaeus. Divergence time estimates indicate that the genus Stomoxys originated in the late Oligocene around 30 million years ago, with the major lineages diversifying in the Early Miocene between 20 and 15 million years ago at a time when temperate forests developed in the Northern Hemisphere.",
+				"archive": "Dryad Digital Repository",
+				"extra": "{:itemType: dataset}",
+				"itemID": "10.5061/dryad.9025",
+				"libraryCatalog": "datadryad.org",
+				"attachments": [],
+				"tags": [
+					"Molecular dating",
+					"Phylogenetic relationship",
+					"Stomoxys flies"
+				],
+				"notes": [],
+				"seeAlso": []
+			}
+		]
+	},
+	{
+		"type": "web",
+		"url": "http://datadryad.org/resource/doi:10.5061/dryad.9025/2",
+		"items": [
+			{
+				"itemType": "journalArticle",
+				"title": "Dsouli-InfectGenetEvol11 phylogram",
+				"creators": [
+					{
+						"firstName": "Najla",
+						"lastName": "Dsouli",
+						"creatorType": "author"
+					},
+					{
+						"firstName": "Frédéric",
+						"lastName": "Delsuc",
+						"creatorType": "author"
+					},
+					{
+						"firstName": "Johan",
+						"lastName": "Michaux",
+						"creatorType": "author"
+					},
+					{
+						"firstName": "Eric",
+						"lastName": "De Stordeur",
+						"creatorType": "author"
+					},
+					{
+						"firstName": "Arnaud",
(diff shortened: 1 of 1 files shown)


rnathanday:
We will be updating/resubmitting this translator soon after internal review is complete -- my apologies for the delay.

One question about the updates: where are the values for the item.seeAlso list stored? The seeAlso list is being populated in Scaffold when I run doWeb(), but I am not seeing the data reflected in the Zotero record anywhere (plugin or standalone). Any pointers?

avram:
On `seeAlso`: https://www.zotero.org/support/dev/translators/coding#related

For it to work, both items in the relationship must be imported at the same time.

rnathanday:
This import translator has been updated and squashed into a single commit. To my knowledge, it accurately reflects both Zotero and Dryad metadata practice. My apologies for the delay in this request.

Dryad Digital Repository.js
```diff
@@ -0,0 +1,707 @@
...
+		translator.setTranslator('951c027d-74ac-47d4-a107-9c3069ab7b48');
+		// note: this handler function is called for a data package
+		// or a data file. If called for a data package as part of setting
+		// the seeAlso value, then the 'doc' variable is still aliasing
+		// the data file's DOM
+		translator.setHandler('itemDone', function (obj, item) {
+			var itemUrl = item.url;  // scoped to this handler
+			var doi = itemUrl.replace(/^.*?doi:/, 'doi:');
+			// <meta content="doi:10.5061/dryad.3sh6d" name="DC.identifier">
+			item.doi = doi;
```
zuphilip: I don't see that field `doi` in the tests below. Did you updated those? 

Moreover, I would suggest to replace the prefix completely, i.e.
```
var doi = itemUrl.replace(/^.*?doi:/, '');
```
Then, you can safe the `substring(4)` functions below.

rnathanday:
Philipp, thanks for catching that error. I had neglected to capitalize correctly in 'item.DOI'. I have updated the code and tests accordingly.

There appears to be one outstanding issue that I am unable to make progress on. The second test reports the following error in Scaffold, although the item type is correct when I have it printed in the debugger. Any thoughts?

> TranslatorTester: Running Dryad Digital Repository Test 2
> 09:29:39 TranslatorTester: Translating http://datadryad.org/resource/doi:10.5061/dryad.9025/2
> 09:29:39 TranslatorTester: Dryad Digital Repository Test 2: failed (Detection returned wrong item type)

aurimasv:
There's no good way to avoid that error. What happens is that detectWeb thinks that the item on that page is book (as an example, idk what it actually detects), but the doWeb ends up importing journalArticle (again, as an example). Unless detectWeb is wrong and there's a way to improve it (there might not be), the best course of action would be to find another test case that has matching item types in detectWeb and doWeb.

Dryad Digital Repository.js
```diff
@@ -0,0 +1,227 @@
...
+		// use the Embedded Metadata translator
+		translator.setTranslator("951c027d-74ac-47d4-a107-9c3069ab7b48");
+		translator.setHandler("itemDone", function (obj, item) {
+			item.abstractNote = item.extra;
+			item.extra = '';
+			item.complete();
+		});
+		translator.getTranslatorObject(function (obj) {
+			obj.doWeb(doc, url);
+		});
```
aurimasv: I'd like you to address this. While it works, `.getTranslatorObject` is rather complex due to sandboxing and we should avoid it as much as possible.

Dryad Digital Repository.js
```diff
@@ -0,0 +1,715 @@
...
+		// note: this handler function is called for a data package
+		// or a data file. If called for a data package as part of setting
+		// the seeAlso value, then the 'doc' variable is still aliasing
+		// the data file's DOM
+		translator.setHandler('itemDone', function (obj, item) {
+			var itemUrl = item.url;  // scoped to this handler
+			var doi = itemUrl.replace(/^.*?doi:/, '');
+			item.DOI = doi;
+			item.itemId = doi;       // internal value for seeAlso relation
+			item.url = doiUrl + doi;
```
aurimasv: don't set URL if it's just the DOI. That's what the DOI field is for (which is also hyperlinked)

rnathanday: This was done at the request of the architect at Dryad, to override the URL for a resource (e.g., http://datadryad.org/resource/doi:10.5061/dryad.xyz) with the dx.doi.org url.

aurimasv: If they don't want us to pull their URL, you can just delete the URL, but there's no reason to duplicate data.

Dryad Digital Repository.js
```diff
@@ -0,0 +1,715 @@
...
+			item.itemType = itemType;
+			// add a seeAlso value if this is a data file 'doc' is
+			// scoped to the outer doWeb call, and seeAlso is only set
+			var pdoi = ZU.xpathText(doc, packageXpath);
+			var docdoi = doc.location.href.replace(/^.*?doi:/, '');
```
aurimasv: Don't use `doc.location.href`, use `url` that's passed into `doWeb`, which will be deproxified.

Dryad Digital Repository.js
```diff
@@ -0,0 +1,715 @@
...
+			if (pdoi && pdoi !== doi) {
+				item.seeAlso = [doiUrl + pdoi];  // dx.doi.org url for seeAlso
```
aurimasv: this doesn't match the `itemId` you set above, which is just the DOI. Same below

Dryad Digital Repository.js
```diff
@@ -0,0 +1,715 @@
...
+		// the seeAlso value, then the 'doc' variable is still aliasing
+		// the data file's DOM
+		translator.setHandler('itemDone', function (obj, item) {
+			var itemUrl = item.url;  // scoped to this handler
+			var doi = itemUrl.replace(/^.*?doi:/, '');
+			item.DOI = doi;
+			item.itemId = doi;       // internal value for seeAlso relation
+			item.url = doiUrl + doi;
+			item.itemType = itemType;
+			// add a seeAlso value if this is a data file 'doc' is
```
aurimasv: based on how you call `doWeb` on line 111, that doesn't seem correct.

Dryad Digital Repository.js
```diff
@@ -0,0 +1,715 @@
...
+			// scoped to the outer doWeb call, and seeAlso is only set
+			var pdoi = ZU.xpathText(doc, packageXpath);
+			var docdoi = doc.location.href.replace(/^.*?doi:/, '');
+			if (pdoi && pdoi !== doi) {
+				item.seeAlso = [doiUrl + pdoi];  // dx.doi.org url for seeAlso
+			// import called on data file, and we are in the callback for
+			// the data package being associated with it
+			} else if (docdoi !== doi) {
```
aurimasv: I'm not really following the logic when this is true.

Dryad Digital Repository.js
```diff
@@ -0,0 +1,715 @@
...
+				item.seeAlso = [];
+			}
+			// signal a dataset until supported per
+			// https://www.zotero.org/support/dev/translators/datasets
+			item.extra = '{:itemType: dataset}';
+			item.archive = 'Dryad Digital Repository';
+			var upd = [ "reportType", "letterType", "manuscriptType",
+				"mapType", "thesisType", "websiteType",
+				"presentationType", "postType", "audioFileType"];
+			for (var i = 0; i < upd.length; i++) item[upd[i]] = itemType;
```
aurimasv: you don't need to do this. Just set 'type'

Dryad Digital Repository.js
```diff
@@ -0,0 +1,715 @@
...
+			item.shortTitle = '';
+			item.complete();
+		});
+		translator.getTranslatorObject(function (obj) {
+			obj.doWeb(doc, url);
+		});
+		var package_doi = ZU.xpathText(doc, packageXpath);
+		if (package_doi) {
+			package_doi = package_doi.substring(4); // omit leading 'doi:'
+			ZU.processDocuments([doiUrl + package_doi], function (d) {
```
aurimasv: Your callback should just be `doWeb`, since `ZU.processDocuments` passes back document and url as parameters.

nit: the URL can be a string. Doesn't have to be an array.

Dryad Digital Repository.js
```diff
@@ -0,0 +1,715 @@
...
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+function detectWeb(doc, url) {
+	// Dryad search page
+	if (url.match('^https?://(?:www\\.)?datadryad\\.org/discover')) {
```
aurimasv: here you may check the URL, but you should also check that `multiplesXpath` actually returns nodes. (so `ZU.xpath(doc, multiplesXpath).length`)

Dryad Digital Repository.js
```diff
@@ -0,0 +1,715 @@
...
+			});
+		}
+	} else if (itemType === 'multiple') {
+		var urls = ZU.xpath(doc, multiplesXpath);
+		if (urls.length > 0) {
+			for (var i = 0; i < urls.length; i++) {
+				urls[i] = doc.location.origin + urls[i].textContent;
+			}
+			ZU.processDocuments(urls, function (d) {
+				doWeb(d, d.location.href);
```
aurimasv: again, just use `doWeb` as the callback

Dryad Digital Repository.js
```diff
@@ -0,0 +1,715 @@
...
+		} else if (result === 'Dataset') {
+			//return 'dataset';
+			return 'journalArticle';
+		}
+	}
+	return false;
+}
+
+var packageXpath    = '//meta[@name="DCTERMS.isPartOf"]/@content';
+var multiplesXpath  = '//li[contains(@class,"ds-artifact-item")]/div/a/@href';
```
aurimasv: don't include the `/@href` part. Instead, see my comment below where you handle multiples.

Dryad Digital Repository.js
```diff
@@ -0,0 +1,715 @@
...
+			package_doi = package_doi.substring(4); // omit leading 'doi:'
+			ZU.processDocuments([doiUrl + package_doi], function (d) {
+				doWeb(d, d.location.href);
+			});
+		}
+	} else if (itemType === 'multiple') {
+		var urls = ZU.xpath(doc, multiplesXpath);
+		if (urls.length > 0) {
+			for (var i = 0; i < urls.length; i++) {
+				urls[i] = doc.location.origin + urls[i].textContent;
```
aurimasv: after you fix the `multiplesXpath` above, do `urls[i] = urls[i].href` here.

Dryad Digital Repository.js
rnathanday: In the most recent commit, the URL is unset.

Dryad Digital Repository.js
rnathanday: Fixed in most recent commit.

Dryad Digital Repository.js
rnathanday: The callback is used here to import metadata for a related item from a different page (and different DOM), so `processDocuments` is the right thing here.

Dryad Digital Repository.js
rnathanday: The call `var result = ZU.xpathText(doc,dcTypeXpath);` with the xpath `//meta[@name="DC.type"][1]/@content` (note the [1]) will yield a string or null, so there is no length property to check in this case.

Dryad Digital Repository.js
rnathanday: Same as above: the intention here is to retrieve metadata found at different urls, so `doc` cannot be reused.

Dryad Digital Repository.js
rnathanday: Fixed in most recent commit.

rnathanday:
@aurimasv : I have updated per your comments in a number of cases, and have clarified the data-package/data-file part of the code.

I believe that the usage of `ZU.processDocuments` rather than a recursive call do `doWeb(...)`, as you suggested, is the right thing here given the requirement to retrieve metadata from multiple pages.

Dryad Digital Repository.js
```diff
@@ -0,0 +1,686 @@
...
+	***** END LICENSE BLOCK *****
+*/
+
+var doiUrlBase      = 'http://dx.doi.org/';
+var packageXpath    = '//meta[@name="DCTERMS.isPartOf"]/@content';
+var multiplesXpath  = '//li[contains(@class,"ds-artifact-item")]/div/a';
+var doiXpath        = '//meta[@name="DC.identifier"][starts-with(@content,"doi:10.5061/dryad.")]/@content';
+var dcTypeXpath     = '//meta[@name="DC.type"][1]/@content';
+var discoverPage    = /^https?:\/\/(www\.)?datadryad\.org\/discover/;
+var urlToDoi        = /^.*?doi:/;
```
aurimasv: Please don't declare/set variables far from where they are used. In most cases you should just use these strings directly. The only time you would want to assign these to variables is if you need to use them more than once and the only time these should be in global scope is if you need to use them in multiple functions.

Dryad Digital Repository.js
```diff
@@ -0,0 +1,686 @@
...
+
+function detectWeb(doc, url) {
+	// Dryad search page
+	if (url.match(discoverPage)) {
```
aurimasv: `RegExp.prototype.test` is preferred where you don't actually need the results of the match

Dryad Digital Repository.js
```diff
@@ -0,0 +1,686 @@
...
+		return 'multiple';
```
aurimasv: We've been testing to make sure that there actually are some results on the page before returning 'multiple'. E.g. https://github.com/zotero/translators/blob/80831a0637fc74b4e16f287c3238c72810b556b5/Deutsche%20Nationalbibliothek.js This way, we don't erroneously return 'multiple' for searches that yield no results.

Dryad Digital Repository.js
```diff
@@ -0,0 +1,686 @@
...
+	} else {
+		var result = ZU.xpathText(doc,dcTypeXpath);
+        // Dryad data package
+        if (result === 'Article') {
+            return 'journalArticle';
+        // Dryad data file
+        } else if (result === 'Dataset') {
+            //return 'dataset';
+            return 'journalArticle';
+        }
```
aurimasv: nit: indents

Dryad Digital Repository.js
```diff
@@ -0,0 +1,686 @@
...
+			item.DOI      = itemDoi;
+			item.itemId   = itemDoi;      // internal value for seeAlso relation
+			item.url      = '';           // the doi value is to be used in lieu of the url
+			item.itemType = itemType;
+			// Add a .seeAlso value if this is a page for a single data file,
+			// rather than for a data package collection.
+			// This is a data file if there is a DCTERMS.isPartOf property.
+			item.seeAlso = (package_doi && package_doi !== itemDoi)
+                ? [doiUrlBase + package_doi]    // data file 
+                : [];                           // data package
```
aurimasv: nit: indent, but also it seems more appropriate here to simply use an `if` statement. The `else` part here is basically a no-op.

aurimasv: Or maybe it's not a no-op.

In either case, the `doiUrlBase + package_doi` does not match the `item.itemId   = itemDoi` part above, so this won't work.

Dryad Digital Repository.js
```diff
@@ -0,0 +1,686 @@
...
+		var package_doi = ZU.xpathText(doc, packageXpath);
+		if (package_doi) package_doi = package_doi.substring('doi:'.length);
+		// note: this handler function is called for a data package
+		// or a data file. If called for a data package as part of setting
+		// the seeAlso value, then the 'doc' variable is still aliasing
+		// the data file's DOM
+		translator.setHandler('itemDone', function (obj, item) {
+			var itemDoi   = item.url.replace(urlToDoi, '');
+			item.DOI      = itemDoi;
+			item.itemId   = itemDoi;      // internal value for seeAlso relation
```
aurimasv: `itemID` (case sensitive)

Dryad Digital Repository.js
```diff
@@ -0,0 +1,686 @@
...
+			// https://www.zotero.org/support/dev/translators/datasets
+			item.extra = '{:itemType: dataset}';
+			item.archive = 'Dryad Digital Repository';
+			item.attachments = [];
+			item.shortTitle = '';
+			item.complete();
+		}); // translator.setHandler()
+		translator.getTranslatorObject(function (obj) {
+			obj.doWeb(doc, url);
+		});
```
aurimasv: Why do you need this? Why not just `translator.setDocument(doc); translator.translate()`?

Dryad Digital Repository.js
```diff
@@ -0,0 +1,686 @@
...
+		if (package_doi) {
+			ZU.processDocuments(doiUrlBase + package_doi, function (d) {
```
aurimasv: If you could manually map these between DOI and Dryad URL that would be a lot better. This way the request is cross-origin, so it will not work with bookmarklet and possibly some other browsers over proxies. Seems like it should be possible to construct a Dryad URL just from DOI.

Dryad Digital Repository.js
```diff
@@ -0,0 +1,686 @@
...
+				doWeb(d, d.location.href);
```
aurimasv: following up from https://github.com/zotero/translators/pull/800#discussion_r32957491 I don't think I was clear enough. What you want is `ZU.processDocuments(someURL, doWeb);`

Dryad Digital Repository.js
```diff
@@ -0,0 +1,686 @@
...
+			});
+		}
+	} else if (itemType === 'multiple') {
+		var urls = ZU.xpath(doc, multiplesXpath);
+		if (urls.length > 0) {
+			for (var i = 0; i < urls.length; i++) {
+                urls[i] = urls[i].href;
+            }
```
aurimasv: nit: indents

Dryad Digital Repository.js
```diff
@@ -0,0 +1,686 @@
...
+					"character evolution",
+					"homoplasy",
+					"parsimony",
+					"phylogenetic inference"
+				],
+				"notes": [],
+				"seeAlso": []
+			}
+		]
+	},
```
aurimasv: did you enter this test case manually? Test cases for 'multiples' should not have any actual item data.

Dryad Digital Repository.js
```diff
@@ -0,0 +1,686 @@
...
+					"homoplasy",
+					"parsimony",
+					"phylogenetic inference"
+				],
+				"notes": [],
+				"seeAlso": []
+			}
+		]
+	}
+]
```
aurimasv: Are you testing different things with all of these cases? Based on your code, there shouldn't be that many variations to test, so if we can trim this down, that would be better.

rnathanday: Slimmed the tests down to the three page types accounted for in the code.

Dryad Digital Repository.js
rnathanday: The tests in the most recent commit were added by Scaffold, and the 'multiples' test does have item data added for each individual item.

aurimasv: ok, I'll take a look. That shouldn't be happening

Dryad Digital Repository.js
rnathanday: I updated this call to processDocuments, but I am seeing a failure when processDocuments is called with an array of urls rather than a single url. When called with an array, the callback is executed only once, so I have left the other processDocuments occurrence with an anonymous function.

Dryad Digital Repository.js
rnathanday: I have left this as-is. When I used the calls you suggested with this translator, the doWeb() run in Scaffold erred out and printed a stacktrace.

PR #813, opened 2014-11-22
[BibTeX] Protect words with initial uppercase letters on export

aurimasv:
Introduces extensions.zotero.translators.BibTeX.export.dontProtectInitialCase hidden preference. If set to `true`, will (mostly*) restore previous behavior of only protecting words with internal upper case letters.
*previously all fields except for "pages" were protected. With this patch only a select list of fields gets protected.

Supersedes #477
Depends on zotero/zotero@eb7959a11dc1970804a10af7d6055ba351e0c6be (zotero/zotero#583)

Re https://forums.zotero.org/discussion/42348/export-for-pandoccapitalization-in-bibtexbiblatex/


katrinleinweber:
Is anyone else noticing that `<i>` is preserved in the BibTeX key, if that tag surrounds the first word(s) of an item's title? A few days ago, I thought it was only one auto-export addon's problem](https://github.com/rokondo/Zotero-autoexporting/issues/8), but now it also occurs upon manual export through Zotero 4.0.26 in Firefox 35.0.1 on Mac OS X 10.9.5.

adam3smith:
ticket created https://github.com/zotero/translators/issues/853
patches welcome.

PR #822, opened 2014-12-08
Primo: make data extraction slightly more general

nemobis:
Now tested locally. Also added some comments etc.

### DIFF
--- a/Primo.js
+++ b/Primo.js
@@ -211,7 +211,7 @@ function importPNX(text) {
 	
 	var item = new Zotero.Item();
 	
-	var itemType = ZU.xpathText(doc, '//display/type');
+	var itemType = ZU.xpathText(doc, '//display/type|//search/rsrctype|//facets/rsrctype');
 	if(!itemType) {
 		throw new Error('Could not locate item type');
 	}
@@ -258,10 +258,10 @@ function importPNX(text) {
 		creators = ZU.xpath(doc, '//display/creator'); 
 	}
 	
-	if(ZU.xpathText(doc, '//display/contributor')) {
-		contributors = ZU.xpath(doc, '//display/contributor'); 
+	if(ZU.xpathText(doc, '//display/contributor|//search/creatorcontrib')) {
+		contributors = ZU.xpath(doc, '//display/contributor|//search/creatorcontrib'); 
 	}
-	
+
 	if(!creators && contributors) { // <creator> not available using <contributor> as author instead
 		creators = contributors;
 		contributors = null;
@@ -316,11 +316,12 @@ function importPNX(text) {
 	}
 	
 	// the three letter ISO codes that should be in the language field work well:
-	item.language = ZU.xpathText(doc, '//display/language');
+	item.language = ZU.xpathText(doc, '//display/language|//facets/language');
 	
 	var pages = ZU.xpathText(doc, '//display/format');
 	if(pages && pages.search(/[0-9]+/) != -1) {
 		pages = pages.replace(/[\(\)\[\]]/g, "").match(/[0-9]+/);
+		// And then we just hope pages are the only number mentioned in the format
 		item.pages = item.numPages = pages[0];
 	}
 
@@ -335,7 +336,7 @@ function importPNX(text) {
 	
 	item.edition = ZU.xpathText(doc, '//display/edition');
 	
-	var subjects = ZU.xpath(doc, '//search/subject');
+	var subjects = ZU.xpath(doc, '//display/display|//search/subject');
 	for(var i=0, n=subjects.length; i<n; i++) {
 		item.tags.push(ZU.trimInternal(subjects[i].textContent));
 	}
--- a/Primo.js
+++ b/Primo.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsb",
-	"lastUpdated": "2014-09-23 21:03:41"
+	"lastUpdated": "2014-12-31 18:43:27"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


Primo.js
```diff
@@ -212,7 +213,13 @@ function importPNX(text) {
 	
 	var itemType = ZU.xpathText(doc, '//display/type');
 	if(!itemType) {
-		throw new Error('Could not locate item type');
+		if ( ZU.xpathText(doc, '//facets/rsrctype') ) {
+			var itemType = ZU.xpathText(doc, '//facets/rsrctype');
+		} else if ( ZU.xpathText(doc, '//search/rsrctype') ) {
+			var itemType = ZU.xpathText(doc, '//search/rsrctype');
+		} else {
```
aurimasv: Is this actually necessary? Every catalog I looked at had //display/type. Do you have an example? I also don't like taking metadata from facets/search for the reasons @zuphilip outlined previously.

Primo.js
```diff
@@ -274,6 +281,7 @@ function importPNX(text) {
 		if(creators[i]) {
 			var creator  = ZU.unescapeHTML(creators[i].textContent).split(/\s*;\s*/);
 			for(j in creator){
+				creator[j] = creator[j].replace(/\([\d\sabcefilr.? ]*[-–][\d\abcefilr.? ]*\)/gi, '');
```
aurimasv: I thought I mentioned this previously, but I don't see it now. That regexp is very obscure. You need to comment on what you're doing here. Chances are you don't want to be replacing random letters, but actual words.

Primo.js
```diff
@@ -333,7 +353,7 @@ function importPNX(text) {
 	
 	item.edition = ZU.xpathText(doc, '//display/edition');
 	
-	var subjects = ZU.xpath(doc, '//search/subject');
+	var subjects = ZU.xpath(doc, '//display/display|//search/subject');
```
zuphilip: I don't think that the xml-node `display` can have a child `display` again.

Primo.js
```diff
@@ -25,6 +25,7 @@ Primos with showPNX.jsp installed:
 (3) http://limo.libis.be/primo_library/libweb/action/search.do?vid=LIBISnet&fromLogin=true
 (4.a) http://virtuose.uqam.ca/primo_library/libweb/action/search.do?vid=UQAM
 (5) http://searchit.princeton.edu/primo_library/libweb/action/dlDisplay.do?docId=PRN_VOYAGER2778598&vid=PRINCETON&institution=PRN
+(6) http://digitale.beic.it/primo_library/libweb/action/search.do?vid=beic
```
zuphilip: Your primo/catalogue does not have `showPNX.jsp` installed.

Primo.js
zuphilip: I have seen in your example that you want to replace for example "circa" and some date range. However, the regexp could potentially match some other stuff as well and I feel it would never be complete (if we think about different languages etc.). I could imagine to delete everything in parenthesis for the creator's name (and we should treat the contibutor then the same), but I haven't made my mind if that would be always a good idea...

Primo.js
```diff
@@ -314,12 +324,22 @@ function importPNX(text) {
...
 	if(pages && pages.search(/[0-9]+/) != -1) {
-		pages = pages.replace(/[\(\)\[\]]/g, "").match(/[0-9]+/);
-		item.pages = item.numPages = pages[0];
+		pages = pages.replace(/[\(\)\[\]]/g, "");
+		// And then we'd hope pages are the only number mentioned in the format
+		// Except that they aren't, in BEIC.it, so we focus the match a bit
+		if(pages.search(/ p+[. ]/) != -1) {
+			var dirt = pages;
+			pages = null;
+			pages = dirt.match(/(?:p+[. ]*)([0-9]+) *(?:p+[. ]*)/);
```
zuphilip: The regexp requires a `p` in the beginning <b>and</b> in the end? That seems wrong for me and it seems not to work in your test case.

zuphilip:
Would it make sense to make a seperate translator for digitale.beic.it, which then calls the general Primo translator and then make the individual changes? Then it would also be possible to use your local fields (lds and lsr).

aurimasv:
If you feel that other changes need to be included from above, please start a new pull request.

Primo.js
nemobis: showPnx=true works though.

Primo.js
nemobis: Yes, the purpose is to remove pretty much anything between parentheses

Primo.js
nemobis: I don't have an example handy, but there are cases where the type identification failed for me. In fact you already account for those cases and throw an error. Can you elaborate on why throwing an error is better than using facets/search?

Primo.js
zuphilip: That is not the same, see http://www.exlibrisgroup.org/display/PrimoCC/showPNX+Revisited

nemobis: Philipp Zumstein, 10/02/2015 12:44:
> That is not the same, see
> http://www.exlibrisgroup.org/display/PrimoCC/showPNX+Revisited

Sure, I'm moving the comment elsewhere in a followup

Primo.js
nemobis: Philipp Zumstein, 01/01/2015 23:15:
> I don't think that the xml-node |display| can have a child |display| again.
>

Thanks. This looks like a typo for "search"

nemobis:
@zuphilip said:
> Would it make sense to make a seperate translator for digitale.beic.it, which then calls the general Primo translator and then make the individual changes? Then it would also be possible to use your local fields (lds and lsr).

Me might end up needing to, but I prefer to avoid it if possible. Library records ought to be standard: either those fields are useless copies or they must correspond to something "real". I'm trying to figure it out with the librarians. 

Something truly hacky, which would require a separate translators, is https://github.com/nemobis/translators/commit/c1f2eae34c71d3903bcd358f3282112efdb547c5 . But I think there are alternatives even for that.

@aurimasv said:
> If you feel that other changes need to be included from above, please start a new pull request.

Thanks, done at https://github.com/zotero/translators/pull/850 and https://github.com/zotero/translators/pull/835.

Everyone, thanks for your helpfulness and patience. Sorry for the slow responses and the stop-and-gos, but I develop this while it's actively used and tested ["on the field" at BEIC](https://it.wikipedia.org/wiki/Progetto:GLAM/BEIC). We had [other priorities](https://outreach.wikimedia.org/wiki/GLAM/Newsletter/January_2015/Contents/Italy_report) in January so we set this aside. In the coming weeks we'll work with and on Zotero much more.

PR #835, opened 2014-12-31
Wikipedia: add language parameter, converting to ISO 639-1/3

nemobis:
Tested locally by exporting some works from Primo (beic.it).


Wikipedia Citation Templates.js
```diff
@@ -62,10 +65,14 @@ var typeMap = {
...
 };
 
+// Most Wikipedias use the "First Last" name format in titles and citations.
+// This isn't universal, see e.g. Russian and Chinese.
 function formatAuthors(authors, useTypes) {
 	var text = "";
 	for each(var author in authors) {
-		text += ", "+author.firstName;
+		// The full "continuous" name uses no separators, which need be removed
+		// cf. "Luc, Jean André : de (1727-1817)"
```
zuphilip: The date range should already be handled in the <a href="https://github.com/zotero/translators/blob/master/Primo.js#L277">Primo translator</a>. I tried this also in your primo instance and it worked there too. We could delete the remaining parenthesis on import. However, the ":" as seperation for the "de" looks for me quite unusual, cf. the <a href="http://viaf.org/viaf/103105362/">names in viaf</a>

Wikipedia Citation Templates.js
```diff
@@ -309,6 +316,22 @@ function doExport() {
...
+		if(item.publisher && properties.date && properties.date < '1831') {
+			// Remove excess space and secondary disambiguations
+			var declutter = item.publisher.replace(/ *\& */g, "; ").replace(/ *\([^)]+\)/g, "");
+			// Shuffle names with disambiguation and remove numbering
+			var shuffle3 = declutter.replace(/ *([^,;]+), +([^,;0-9]+) *[0-9]*, +([^,;]+) */g, "$2 $1 ($3)");
+			// Same, other names
+			var shuffle2 = shuffle3.replace(/ *([^,;]+), +([^,;0-9]+) *[0-9]* */g, "$2 $1");
+			// Use comma list, get rid of space buildup for numbered names
+			properties.publisher = shuffle2.replace(/ *[;] */g, ", ").replace(/  /g, " ");
+		}
```
zuphilip: Maybe ZU.trimInternal() would do all the spacing problems. However, I don't think we should do the reorder for all data. If I search for "&" or "," as publisher in my current Zotero library I find for example
 * Hodges, Foster & Co.
 * The Minerals, Metals & Materials Society
 * ProQuest Information & Learning
 * O'Reilly Vlg. Gmbh & Co.
 * Hessisches Ministerium für Landwirtschaft, Forsten und Naturschutz, Landentwicklung
 * Office of Plans and Policy, Federal Communications Commission
 * printed by Henry Hills, printer to the King's most Excellent Majesty, for His Household and Chappel
 * printed for Samuel Smith, at the Princes Arms in St. Paul's Church-yard

The reordering in these example would be wrong. The publication date (<1831) seems for me to be too weak for a condition.

nemobis: Ok, so I'll move this to Primo.js.

Philipp Zumstein, 01/01/2015 15:28:
> The reordering in these example would be wrong. The publication date
> (<1831) seems for me to be too weak for a condition.

The cases with "&" can easily be fixed by checking that there is a comma 
after it. For the commas, something similar.

We are trying to extract semantics from non-semantic data, there will 
always be errors. The question is at what level this should happen (on 
import, I understand).

AbeJellinek:
I'm not sure if this should stay open - it's not exactly common for Zotero items to have ISO 639-2 language codes these days.

That said, it's also unreasonable to expect that every export translator will know how to interpret every standard. Before we can handle the language field in translators like this, we need a `cleanLanguage` function that converts spelled-out language names, ISO 639-2, etc., into ISO 639-1.

PR #836, opened 2015-01-02
Primo

aurimasv:
Re https://forums.zotero.org/discussion/43372/individual-book-authors-display-as-corporate-authors-in-ex-libris-primo/ and #822

This mostly covers author parsing and improved numPages fetching.

CC @zuphilip

Should cover most changes in #822 but I couldn't find the test case

### DIFF
--- a/Primo.js
+++ b/Primo.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsb",
-	"lastUpdated": "2014-09-23 21:03:41"
+	"lastUpdated": "2015-01-02 04:40:01"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


Primo.js
```diff
@@ -251,48 +253,36 @@ function importPNX(text) {
...
-		if(creators[i]) {
-			var creator  = ZU.unescapeHTML(creators[i].textContent).split(/\s*;\s*/);
-			for(j in creator){
-				creator[j] = creator[j].replace(/\d{4}-(\d{4})?/g, '');
-				item.creators.push(Zotero.Utilities.cleanAuthor(creator[j], "author", true));
-			}			
+	// //addata/addau is great because it lists authors in last, first format,
+	// but it can also have a bunch of junk. We'll use it to help split authors
+	var splitGuidance = {};
+	var addau = ZU.xpath(doc, '//addata/addau');
```
zuphilip: We need both here:
```
ZU.xpath(doc, '//addata/addau|//addata/au')
```
AFAIK: The `au` is refering to the creators (authors) and the `addau` is refering to the contributors (additional authors). However, we shouldn't count on that and it should be IMO safe to store both kind of creators/contributors in the same data structure, because we will use it only as a split guidance (lookup table).

Primo.js
```diff
@@ -25,6 +25,7 @@ Primos with showPNX.jsp installed:
 (3) http://limo.libis.be/primo_library/libweb/action/search.do?vid=LIBISnet&fromLogin=true
 (4.a) http://virtuose.uqam.ca/primo_library/libweb/action/search.do?vid=UQAM
 (5) http://searchit.princeton.edu/primo_library/libweb/action/dlDisplay.do?docId=PRN_VOYAGER2778598&vid=PRINCETON&institution=PRN
+(6) http://digitale.beic.it/primo_library/libweb/action/search.do?vid=beic
```
zuphilip: Either move this line up or change the headline. They don't have the `showPNX.jsp` installed. Maybe it is not anymore that much important to list the different kind of primos? Moroever, I think most primos are now on some version 4.x.

Primo.js
```diff
@@ -363,37 +358,184 @@ function importPNX(text) {
 	item.callNumber = ZU.xpathText(doc, '//enrichment/classificationlcc');
 	
 	item.complete();
+}
+
+function cleanAuthor(str) {
```
zuphilip: Can we rename that function because there is a function `ZU.cleanAuthor()` with the same name in the Zotero utilities? Maybe just `clean`...

Primo.js
```diff
@@ -251,48 +253,36 @@ function importPNX(text) {
...
-			}			
+	// //addata/addau is great because it lists authors in last, first format,
+	// but it can also have a bunch of junk. We'll use it to help split authors
+	var splitGuidance = {};
+	var addau = ZU.xpath(doc, '//addata/addau');
+	for (var i=0; i<addau.length; i++) {
+		var author = cleanAuthor(addau[i].textContent);
+		if (author.indexOf(',') != -1) {
+			var splitAu = author.split(',');
+			if (splitAu.length > 2) continue;
```
zuphilip: Is this necessary?

Primo.js
```diff
@@ -251,48 +253,36 @@ function importPNX(text) {
...
+	fetchCreators(item, creators, 'author', splitGuidance);
+	fetchCreators(item, contributors, 'contributor', splitGuidance);
+	
 	//PNX doesn't do well with institutional authors; This isn't perfect, but it helps:
-	for(i in item.creators){
-		if(!item.creators[i].firstName){
-			item.creators[i].fieldMode=1;
+	for(var i=0; i<item.creators.length; i++) {
+		if(!item.creators[i].firstName) {
+			item.creators[i].fieldMode = 1;
```
zuphilip: You are doing this in the `fetchCreators` function already, i.e. delete it here.

Primo.js
```diff
@@ -314,12 +304,16 @@ function importPNX(text) {
...
-	if(pages && pages.search(/[0-9]+/) != -1) {
-		pages = pages.replace(/[\(\)\[\]]/g, "").match(/[0-9]+/);
-		item.pages = item.numPages = pages[0];
+	if(pages && pages.search(/\d/) != -1) {
+		if (item.itemType == 'book') {
+			item.numPages = extractNumPages(pages)
+		} else {
+			pages = pages.replace(/[\(\)\[\]]/g, "").match(/[0-9]+/);
+			item.pages = item.numPages = pages[0];
+		}
```
zuphilip: Do you have an example for the `else` part? (For articles there is also the utility function <a href="https://github.com/zotero/zotero/blob/4.0/chrome/content/zotero/xpcom/utilities.js#L644">getPageRange</a>, however the metadata for articles should be better handled with the addata part from below.)

Primo.js
```diff
@@ -363,37 +358,184 @@ function importPNX(text) {
...
+	}
+}
+
+function extractNumPages(str) {
+	// Borrowed from Library Catalog (PICA). See #756
+	//make sure things like 2 partition don't match, but 2 p at the end of the field do
+	// f., p., and S. are "pages" in various languages
+	// For multi-volume works, we expect formats like:
+	//   x-109 p., 510 p. and X, 106 S.; 123 S.
+	var numPagesRE = /\[?((?:[ivxlcdm\d]+[ ,\-]*)+)\]?\s+[fps]\b/ig,
```
zuphilip: I suggest to add a `\b` in the beginning of this regexp, i.e.
```
var numPagesRE = /\b\[?((?:[ivxlcdm\d]+[ ,\-]*)+)\]?\s+[fps]\b/ig,
```
Currently it will match (highlighted bold) in the string "an<b>d x, 106 s.</b>"

Primo.js
```diff
@@ -363,37 +358,184 @@ function importPNX(text) {
...
+		numPages = [], m;
+	while(m = numPagesRE.exec(str)) {
+		numPages.push(m[1].replace(/ /g, '')
+			.replace(/[\-,]/g,'+')
+			.toLowerCase() // for Roman numerals
+		);
+	}
+	return numPages.join('; ');
```
zuphilip: Does it make sense to add this function as an utility function in ZU?

Primo.js
```diff
@@ -363,37 +358,184 @@ function importPNX(text) {
...
-				"shortTitle": "In Zweifelsfällen entscheidet die Wahrheit"
+				"publisher": "Boston Boston Book and Art Chop",
+				"attachments": [],
+				"tags": [
+					"ART",
+					"GUERRE",
+					"WAR"
+				],
+				"notes": [],
+				"seeAlso": []
```
zuphilip: This book is part of a book series. Maybe we could add that information as well from:
```
<prim:seriestitle>Art and society 1</prim:seriestitle> 
```

Primo.js
```diff
@@ -363,37 +358,184 @@ function importPNX(text) {
...
+				"callNumber": "U21.2",
```
zuphilip: This is just the LCC, which is not necessarily the call number in that particular library. We could try to extract the call number (or at least one of them if several copies in a library exists) from:
```
<prim:availlibrary>$$IPURDUE$$LHSSE$$1HSSE - 2nd floor$$2(355.02 W1945 1994 )$$Savailable$$X01PURDUE_PUWL$$Yhsse$$Zhss2$$P1</prim:availlibrary> 
```
Subfield 2 is the call number. However, shouldn't we then also try to save the name of the library instead of just "Primo"?

zuphilip:
@aurimasv Thank you for manually combining the pull requests for Primo together and doing the "cherry picking". I can confirm that your code does the same (and more) than my pull request and therefore we could close #831 afterwards. However, my first comment here about `addata/au` is essential.

Primo.js
aurimasv: Mostly because I wouldn't know what to do with more than 1 comma

Primo.js
aurimasv: Not generic enough, but we'll come up with something like Scaffold modules (similar to framework)

Primo.js
aurimasv: Let's take this to another PR. I'm not entirely clear on how we're supposed to be treating the call number field. I want to say that it should correspond to the Library Catalog, but LCCN has been pretty much the standard value for most translators. If we're going with Call Number corresponding to current library, then we probably still want to store LCCN as one of the identifiers in the Extra field.

I'm also not totally sure how Library Catalog/Call Number differ from Archive/Location in Archive. https://www.zotero.org/support/kb/item_types_and_fields talks about it a bit, but if you're working with archival material, what would you put in the Library Catalog/Call Number fields? Wouldn't the two sets be almost exclusive? Never worked with archival material myself, so I could be way off base here.

In any case, we need to define these better and provide clear guidance (for both translator developers and Zotero users). CC @adam3smith

aurimasv:
Thank you for your review. To be perfectly honest, I missed #831, but I'm glad that we arrived at the same exact solution :-) I'll try to pull in your commit to give you credit for the code.

aurimasv:
@zuphilip, could you give this one final look? Also, I included your commit from the other PR (with some edits in the commit message). Let me know if you're ok with that.

Primo.js
zuphilip: A seperate PR/issue is a good idea. One remark/reminder (such that we don't forget it): We have here in primo the <a href="http://en.wikipedia.org/wiki/Library_of_Congress_Classification">LCC</a> (classification) which is not the same as the <a href="http://en.wikipedia.org/wiki/Library_of_Congress_Control_Number">LCCN</a> (id).

zuphilip:
Everything looks fine. Thank you very much for considering all my comments and including my commit into this PR by some git magic:)

zuphilip:
<blockquote>To be perfectly honest, I missed [your pull request]</blockquote>
@aurimasv : Did you notice my other pull requests (SORA and JRC)?

aurimasv:
Catching up on PRs now. Can't promise today, but I'll clear the queue by the end of the weekend

PR #838, opened 2015-01-04
Fixes for Google Scholar cases

fbennett:
Patch includes:
  - A fix for the error reported on [the forums](https://forums.zotero.org/discussion/44849/zotero-wont-save-google-scholar-case-but-no-error-report-id-949790485); and
  - MLZ support code blocked in conditionals.

### DIFF
--- a/Google Scholar.js
+++ b/Google Scholar.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2015-01-02 20:25:25"
+	"lastUpdated": "2015-01-06 12:25:05"
 }
 
 /*
@@ -685,7 +685,14 @@ var scrapeCase = function (doc, url) {
 		// citelet looks kind of like this
 		// Powell v. McCormack, 395 US 486 - Supreme Court 1969
 		var item = new Zotero.Item("case");
-		var factory = new ItemFactory(doc, refFrag.textContent, [url]);
+		var attachmentPointer = url;
+		if (Zotero.isMLZ) {
+			var block = doc.getElementById("gs_opinion_wrapper");
+			if (block) {
+				attachmentPointer = block;
+			}
+		}
+		var factory = new ItemFactory(doc, refFrag.textContent, [attachmentPointer]);
 		factory.repairCitelet();
 		factory.getDate();
 		factory.getCourt();
@@ -742,8 +749,8 @@ ItemFactory.prototype.repairCitelet = function () {
 ItemFactory.prototype.repairTitle = function () {
 	// All-caps words of four or more characters probably need fixing.
 	if (this.v.title.match(/(?:[^a-z]|^)[A-Z]{4,}(?:[^a-z]|$)/)) {
-		this.v.title = ZU.capitalizeTitle(this.v.title.toLowerCase())
-						.replace(/([^0-9a-z])V([^0-9a-z])/, "$1v$2");
+		this.v.title = ZU.capitalizeTitle(this.v.title.toLowerCase(), true)
+								.replace(/([^0-9a-z])V([^0-9a-z])/, "$1v$2");	
 	}
 };
 
@@ -779,7 +786,16 @@ ItemFactory.prototype.getDate = function () {
 	var i, m;
 	// Citelet parsing, step (1)
 	if (!this.hyphenSplit) {
-		this.hyphenSplit = this.citelet.split(/\s+-\s+/);
+		if (this.citelet.match(/\s+-\s+/)) {
+			this.hyphenSplit = this.citelet.split(/\s+-\s+/);
+		} else {
+			m = this.citelet.match(/^(.*),\s+([^,]+Court,\s+[^,]+)$/);
+			if (m) {
+				this.hyphenSplit = [m[1], m[2]];
+			} else {
+				this.hyphenSplit = [this.citelet];
+			}
+		}
 		this.trailingInfo = this.hyphenSplit.slice(-1);
 	}
 	if (!this.v.date && this.v.date !== false) {
@@ -824,14 +840,26 @@ ItemFactory.prototype.getCourt = function () {
 	var s, m;
 	// Citelet parsing, step (2)
 	s = this.hyphenSplit.pop().replace(/,\s*$/, "").replace(/\u2026\s*$/, "Court");
-	m = s.match(/(?:([a-zA-Z]+):\s*)*(.*)/);
+	var court = null;
+	var jurisdiction = null;
+	m = s.match(/(.* Court),\s+(.*)/);
 	if (m) {
-		this.v.court = m[2].replace(/_/g, " ");
-		if (m[1]) {
-			this.v.extra = "{:jurisdiction: " + m[1] + "}";
+		court = m[1];
+		jurisdiction = m[2];
+	}
+	if (!court) {
+		m = s.match(/(?:([a-zA-Z]+):\s*)*(.*)/);
+		if (m) {
+			court = m[2].replace(/_/g, " ");
+			jurisdiction = m[1];
 		}
 	}
-	return this.v.court;
+	if (court) {
+		this.v.court = court;
+	}
+	if (jurisdiction) {
+		this.v.extra = "{:jurisdiction: " + jurisdiction + "}";
+	}
 };
 
 
@@ -884,14 +912,40 @@ ItemFactory.prototype.getDocketNumber = function (doc) {
 
 ItemFactory.prototype.getAttachments = function (doctype) {
 	var i, ilen, attachments;
+	var attachmentTitle = "Google Scholar " + doctype;
 	attachments = [];
 	for (i = 0, ilen = this.attachmentLinks.length; i < ilen; i += 1) {
 		if (!this.attachmentLinks[i]) continue;
-		attachments.push({
-			title:"Google Scholar Linked " + doctype,
-			url:this.attachmentLinks[i],
-			type:"text/html"
-		});
+		if ("string" === typeof this.attachmentLinks[i]) {
+			attachments.push({
+				title: attachmentTitle,
+				url:this.attachmentLinks[i],
+				type:"text/html"
+			});
+		} else {
+			// DOM fragment and parent doc
+			var block = this.attachmentLinks[i];
+			var doc = block.ownerDocument;
+
+			// String content (title, url, css)
+			var title = doc.getElementsByTagName("title")[0].textContent;
+			var url = doc.documentURI;
+			var css = "*{margin:0;padding:0;}div.mlz-outer{width: 60em;margin:0 auto;text-align:left;}body{text-align:center;}p{margin-top:0.75em;margin-bottom:0.75em;}div.mlz-link-button a{text-decoration:none;background:#cccccc;color:white;border-radius:1em;font-family:sans;padding:0.2em 0.8em 0.2em 0.8em;}div.mlz-link-button a:hover{background:#bbbbbb;}div.mlz-link-button{margin: 0.7em 0 0.8em 0;}";
+
+			// head element
+			var head = doc.createElement("head");
+			head.innerHTML = '<title>' + title + '</title>';
+			head.innerHTML += '<style type="text/css">' + css + '</style>'; 
+
+			var attachmentdoc = Zotero.Utilities.composeDoc(doc, head, block);
+			attachments.push({
+				title: attachmentTitle,
+				document:attachmentdoc
+			});
+
+			// URL for this item
+			this.item.url = url;
+		}
 	}
 	return attachments;
 };
@@ -941,6 +995,9 @@ ItemFactory.prototype.saveItem = function () {
 				bogusItemID += 1;
 				completed_items.push(this.item);
 			}
+			if (completed_items.length === 0) {
+				throw new Error("Failed to parse \"" + this.citelet + "\"");
+			}
 			for (i = 0, ilen = completed_items.length; i < ilen; i += 1) {
 				for (j = 0, jlen = completed_items.length; j < jlen; j += 1) {
 					if (i === j) {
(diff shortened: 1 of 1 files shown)


Google Scholar.js
```diff
@@ -1,5 +1,6 @@
 {
 	"translatorID": "57a00950-f0d1-4b41-b6ba-44ff0fc30289",
+	"translatorType": 4,
```
aurimasv: Duplicated?

Google Scholar.js
```diff
@@ -685,7 +686,13 @@ var scrapeCase = function (doc, url) {
 		// citelet looks kind of like this
 		// Powell v. McCormack, 395 US 486 - Supreme Court 1969
 		var item = new Zotero.Item("case");
-		var factory = new ItemFactory(doc, refFrag.textContent, [url]);
+		var attachmentPointer = url;
+		if (Zotero.Utilities.setMultiField) {
+			var block = doc.getElementById("gs_opinion_wrapper");
+			if (block) {
+				attachmentPointer = block;
```
aurimasv: Something is whacky with the braces

Google Scholar.js
```diff
@@ -828,7 +844,11 @@ ItemFactory.prototype.getCourt = function () {
 	if (m) {
 		this.v.court = m[2].replace(/_/g, " ");
 		if (m[1]) {
-			this.v.extra = "{:jurisdiction: " + m[1] + "}";
+            if (!Zotero.Utilities.setMultiField) {
+			    this.v.jurisdiction = m[1];
+            } else {
+                this.v.extra = "{:jurisdiction: " + m[1] + "}";
+            }
```
aurimasv: Nit: indents

Google Scholar.js
```diff
@@ -887,11 +907,43 @@ ItemFactory.prototype.getAttachments = function (doctype) {
...
 		if (!this.attachmentLinks[i]) continue;
-		attachments.push({
-			title:"Google Scholar Linked " + doctype,
-			url:this.attachmentLinks[i],
-			type:"text/html"
-		});
+		if (Zotero.Utilities.setMultiField) {
+			if ("string" === typeof this.attachmentLinks[i]) {
+				// URL (string)
+				if (this.attachmentLinks[i].slice(0, 1) === "/") {
```
aurimasv: `.charAt(0)`? But also, Zotero does resolve relative links, so I'm curious why this is necessary

Google Scholar.js
```diff
@@ -887,11 +907,43 @@ ItemFactory.prototype.getAttachments = function (doctype) {
...
+					this.attachmentLinks[i] = "http://scholar.google.com"+this.attachmentLinks[i];
+				}
+				this.attachmentLinks[i] = this.attachmentLinks[i].replace(/&q=[^&]*/g,"");
+				attachments.push({title:"Google Scholar: " + doctype, type:"text/html",
+								  url:this.attachmentLinks[i]});
```
aurimasv: Looks messy. Shouldn't this be the same for Zotero and MLZ?

aurimasv:
I thought you had a separate fork for MLZ translators.

Google Scholar.js
fbennett: Oops, sorry about that. I thought I had run this iteration, but obviously not.

Google Scholar.js
fbennett: Fixed, sorry for the oversight.

Google Scholar.js
fbennett: Still learning JavaScript ... Good catch on the (lack of) need for this. Where the attachment is a string, the code should just fall over to the existing Zotero-orthodox URL attachment block below. Will revise.

Google Scholar.js
fbennett: MLZ has mods to domsaver that allow a DOM element to be extracted and rendered in a custom wrapper. Giving that treatment to the text makes it more readable, and makes for consistency across judgments obtained from other sources. In these pages, we're only interested in the public domain text of the judgements, so there isn't a (legal) problem with the refactoring.

fbennett:
There is a separate stable of translators, but it's updated periodically from upstream. Resolving conflicts takes a bit of time, but the bigger problem is that it's a headache to recast things for pull requests back to Zotero main. To the extent that we can get compatible translator code in place, it will lower the barrier to upstream submissions out of the MLZ suite (which I hope is a good thing ...).

fbennett:
I've cleaned up that logic a little, see what you think.

There is one additional change in the latest commit. Case names in GS are often a cluttered mess of mixed uppercase and lowercase. The translator was originally fixing them by forcing lowercase, then applying capitalizeTitle(), then applying some sort of regexp sub to quash a capitalized "V". capitalizeTitle() is savvy to "v" now, so the last ugly step is not needed - but it's also shy, so the second step wasn't happening, and the case name was coming out in all lowercase. I've applied the "force" toggle, which has things pretty again.

fbennett:
The OP's report noted that the translator failed without error. Would it be better to throw an error when parsing of the cite elements of a case fails?

adam3smith:
> Would it be better to throw an error when parsing of the cite elements of a case fails?
yes, please. a) because otherwise users may not notice
b) errors get logged to https://repo.zotero.org/errors

fbennett:
Done.

aurimasv:
Well, generally, we need to fix this in Zotero so that (1) on single item pages, if no items are saved, Zotero throws an error and informs the user to report the issue and (2) for multiples, we import as many items as possible and somehow inform the user that some items failed (we do know the title at this point, so it shouldn't be too awkward). (1) is easy, (2) is hard and has been a long term goal of mine, and I haven't figured out exactly how we should do it.

For the time being, throwing an informative error for something that you know will fail is ok.

Google Scholar.js
aurimasv: Still need to fix this. How is this even happening though? Are you using Scaffold?

aurimasv:
Frank, instead of relying on some obscure ZU functions to check for MLZ, can you instead introduce a `Zotero.isMLZ` property? That would go along well with `Zotero.isFx`, Zotero.isChrome`, etc.

fbennett:
I had thought of that, happy to introduce it.

Google Scholar.js
```diff
@@ -742,8 +750,7 @@ ItemFactory.prototype.repairCitelet = function () {
 ItemFactory.prototype.repairTitle = function () {
 	// All-caps words of four or more characters probably need fixing.
 	if (this.v.title.match(/(?:[^a-z]|^)[A-Z]{4,}(?:[^a-z]|$)/)) {
-		this.v.title = ZU.capitalizeTitle(this.v.title.toLowerCase())
-						.replace(/([^0-9a-z])V([^0-9a-z])/, "$1v$2");
+		this.v.title = ZU.capitalizeTitle(this.v.title.toLowerCase(), true);
```
aurimasv: Idk about cases, but I thought we were going to switch over to sentence case when fixing all-caps. Though in this case you're not really checking if everything is all-caps and it might be more correct to convert to title case for cases since they tend to contain a lot of proper nouns. Anyway, you'll have to make a call on that.

Google Scholar.js
```diff
@@ -828,7 +844,11 @@ ItemFactory.prototype.getCourt = function () {
 	if (m) {
 		this.v.court = m[2].replace(/_/g, " ");
 		if (m[1]) {
-			this.v.extra = "{:jurisdiction: " + m[1] + "}";
+			if (!Zotero.Utilities.setMultiField) {
+				this.v.jurisdiction = m[1];
```
aurimasv: Isn't this backwards? `jurisdiction` is MLZ-specific, not the other way around. But furthermore, doesn't MLZ store this data in `extra` anyway and then interpret it as a field (for syncing reasons)? And if no, then shouldn't MLZ be parsing out data from `extra` on import and moving it to appropriate field? I'm not terribly against this in GS (this is your baby), but I don't think we'll be introducing MLZ-specific logic in other translators, so it seems like a broader solution for MLZ would be worthwhile.

Google Scholar.js
```diff
@@ -866,6 +886,8 @@ ItemFactory.prototype.getTitle = function () {
 	// Citelet parsing, step (4) [optional]
 	if (this.commaSplit) {
 		this.v.title = this.commaSplit.join(", ");
+	} else {
+		throw new Error("Parsing failed for case item \"" + this.citelet + "\"");
 	}
```
aurimasv: Are you sure this is where the issue comes from? If title was never assigned, then Zotero would throw an error. Current behavior suggests that the translator is returning without calling `item.complete`

Google Scholar.js
```diff
@@ -887,11 +909,33 @@ ItemFactory.prototype.getAttachments = function (doctype) {
 	attachments = [];
 	for (i = 0, ilen = this.attachmentLinks.length; i < ilen; i += 1) {
 		if (!this.attachmentLinks[i]) continue;
-		attachments.push({
-			title:"Google Scholar Linked " + doctype,
-			url:this.attachmentLinks[i],
-			type:"text/html"
-		});
+		if (!Zotero.Utilities.setMultiField || "string" === typeof this.attachmentLinks[i]) {
```
aurimasv: `!Zotero.Utilities.setMultiField` is sort of redundant here, since, in non-MLZ environment, `.attachmentLinks[i]` is always a string. It also makes it seem like you could be entering this block where `.attachmentLinks[i]` is _not_ a string and that would be bad.

Google Scholar.js
fbennett: No, just cruft and negligence.

Google Scholar.js
fbennett: Title-case will be closest to user expectations for this one.

Google Scholar.js
```diff
@@ -887,11 +909,33 @@ ItemFactory.prototype.getAttachments = function (doctype) {
...
+			var url = doc.documentURI;
+			var css = "*{margin:0;padding:0;}div.mlz-outer{width: 60em;margin:0 auto;text-align:left;}body{text-align:center;}p{margin-top:0.75em;margin-bottom:0.75em;}div.mlz-link-button a{text-decoration:none;background:#cccccc;color:white;border-radius:1em;font-family:sans;padding:0.2em 0.8em 0.2em 0.8em;}div.mlz-link-button a:hover{background:#bbbbbb;}div.mlz-link-button{margin: 0.7em 0 0.8em 0;}";
+
+			// head element
+			var head = doc.createElement("head");
+			head.innerHTML = '<title>' + title + '</title>';
+			head.innerHTML += '<style type="text/css">' + css + '</style>'; 
+
+			var attachmentdoc = Zotero.Utilities.composeDoc(doc, head, block);
+			attachments.push({title:"Google Scholar: "+doctype, type:"text/html", document:attachmentdoc});
```
aurimasv: (1) Can we make the title the same across attachments via URL or via this (unless they're actually different content-wise) and (2) `type:"text/html"` is superfluous here because you're using `document` property.

Google Scholar.js
fbennett: D'oh.

For the rest, good suggestions as always. I think the second option should work, and anything that brings MLZ and Zotero play nicer together is a  good thing.

MLZ stores weird field data in `extra` in an encoded form during sync (JSON, with a header to signal the offset), and that rigamarole shouldn't be replicated in the translator. Parsing out field content on translator import should work fine. It's a little tricky, though, since the field name in the `{:field-name: XXX}` markup is a CSL variable, and may need to be reverse-mapped into the database schema in some cases. MLZ has a function for that already, though (for reconstructing missing items from embedded document data), and it should be easy enough to tie it into the translator processing chain. I'll take a look.

Google Scholar.js
fbennett: Understood. I'll go back and test with the original translator code, and pin down the cause.

fbennett:
I think I've covered everything raised so far. I have some changes ready for MLZ that will parse out the {:jurisdiction: XXX} field syntax and map the variable correctly for the target Zotero item type. When this passes muster, I'll make a fresh MLZ release, and we should be squared away.

fbennett:
(I went ahead with the MLZ release, after a little more testing.)

fbennett:
I've been through the tests is Scaffold under Zotero with Firefox 34, confirmed that previous tests all pass, and added a test for the case that triggered this revision.

There is one other small change in the last checkin: I was mistaken when I said that capitalizeTitle() now accounts for the case name "v" - I was looking at MLZ code, not official Zotero - so I've restored the little cleanup regexp at the end of capitalizeTitle().

If there are other issues, just let me know. Otherwise, this seems ready to go.

fbennett:
I'll squash the commits.

fbennett:
If there is anything in this that you would like to see done in a different way, I can dig into it again (I assume there are good reasons why this hasn't been merged, but I'm kind of stuck for ideas).

If there have been second thoughts about permitting the `{:jurisdiction: XX}` markers in the `extra field`, do let me know; I have a stable set of jurisdiction IDs going [here](http://fbennett.github.io/legal-resource-registry/), and there are a couple of other translators in the works that will want to make use of them.

aurimasv:
Sorry, just haven't had time to look at this since last time. Will get through it this weekend.

Google Scholar.js
```diff
@@ -742,8 +749,8 @@ ItemFactory.prototype.repairCitelet = function () {
 ItemFactory.prototype.repairTitle = function () {
 	// All-caps words of four or more characters probably need fixing.
 	if (this.v.title.match(/(?:[^a-z]|^)[A-Z]{4,}(?:[^a-z]|$)/)) {
-		this.v.title = ZU.capitalizeTitle(this.v.title.toLowerCase())
-						.replace(/([^0-9a-z])V([^0-9a-z])/, "$1v$2");
+		this.v.title = ZU.capitalizeTitle(this.v.title.toLowerCase(), true)
+								.replace(/([^0-9a-z])V([^0-9a-z])/, "$1v$2");
```
aurimasv: Could you trim down on the indents (single indent is fine)

aurimasv:
Looks good. nvm about the white space. Not worth the hassle.

Thanks!

fbennett:
Thanks!

The other legal translators pending will come along after I've remapped MLZ client data to align with the new, more systematic identifiers in the [LRR](http://fbennett.github.io/legal-resource-registry/index.html). Been a long road, but the full requirements for legal support finally seem to be coming into focus.

PR #839, opened 2015-01-10
Add translator for WorldCat Discovery.

apcshields:
See this [mailing list conversation](https://groups.google.com/forum/#!topic/zotero-dev/iMmQF-traLU).

### DIFF
--- a/WorldCat Discovery.js
+++ b/WorldCat Discovery.js
@@ -1,0 +1,185 @@
+{
+	"translatorID": "159c65e6-329f-455e-9caa-e0c80bf8435e",
+	"label": "WorldCat Discovery",
+	"creator": "Andrew Shields",
+	"target": "^https?://[^/]+\\.on\\.worldcat\\.org",
+	"minVersion": "3.0.9",
+	"maxVersion": "",
+	"priority": 90,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcs",
+	"lastUpdated": "2015-01-09 21:18:25"
+}
+
+/*
+WorldCat Discovery translator for Zotero.
+Copyright (C) 2015 Andrew Shields
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU Affero General Public License as
+published by the Free Software Foundation, either version 3 of the
+License, or (at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+var detectWeb, doWeb, pageIsActionRecord, parseOpenURLData;
+
+detectWeb = function(doc, url) {
+  var activeRecord, isActionRecord, numberOfRecords, record, records, _i, _len;
+  isActionRecord = pageIsActionRecord(url);
+  records = doc.getElementsByClassName('record');
+  numberOfRecords = records.length;
+  activeRecord = doc.getElementsByClassName('record active')[0];
+  if (!(isActionRecord || numberOfRecords)) {
+	return false;
+  } else {
+	if (isActionRecord) {
+	  activeRecord = doc.getElementById('main');
+	  return parseOpenURLData(activeRecord).type;
+	} else {
+	  for (_i = 0, _len = records.length; _i < _len; _i++) {
+		record = records[_i];
+		Zotero.monitorDOMChanges(record, {
+		  attributes: true,
+		  attributeFilter: ['class']
+		});
+	  }
+	  if (activeRecord) {
+		return parseOpenURLData(activeRecord).type;
+	  } else {
+		return 'multiple';
+	  }
+	}
+  }
+};
+
+doWeb = function(doc, url) {
+  var activeRecord, isActionRecord, items, numberOfRecords, oclcNum, record, records, saveItems, _i, _len;
+  isActionRecord = pageIsActionRecord(url);
+  records = doc.getElementsByClassName('record');
+  numberOfRecords = records.length;
+  activeRecord = doc.getElementsByClassName('record active')[0];
+  if (!(isActionRecord || numberOfRecords)) {
+	return false;
+  } else {
+	saveItems = function(items) {
+	  var oclcNum, _results;
+	  _results = [];
+	  for (oclcNum in items) {
+		url = "//www.worldcat.org/oclc/" + oclcNum + "?page=endnotealt&client=worldcat.org-detailed_record";
+		_results.push(Zotero.Utilities.doGet(url, function(responseDocument) {
+		  var translator;
+		  responseDocument = responseDocument.trim();
+		  if (responseDocument.match(/^TY\s+-\s+CONF/)) {
+			responseDocument.replace(/^TY\s+-\s+CONF/, 'TY  - BOOK');
+			responseDocument.replace(/^AU\s+-\s+/gm, 'A3  - ');
+		  }
+		  translator = Zotero.loadTranslator('import');
+		  translator.setTranslator('32d59d2d-b65a-4da4-b0a3-bdd3cfb979e7');
+		  translator.setString(responseDocument);
+		  translator.setHandler('itemDone', function() {
+			var item;
+			item = arguments[arguments.length - 1];
+			item.extra = void 0;
+			item.archive = void 0;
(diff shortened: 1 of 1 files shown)


aurimasv:
I would use this translator to fetch the appropriate OCLC ID and then ask Open WorldCat translator to retrieve metadata. There's no need to replicate that part of the code unless it actually needs to behave differently, but since you're requesting RIS directly from www.worldcat.org, I don't see how the handling could be different.

I'll make other comments in the code.

apcshields:
As far as I could tell, the Open WorldCat translator doesn't expose that function to other translators, which is why I rewrote it.

—

WorldCat Discovery.js
```diff
@@ -0,0 +1,185 @@
...
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+var detectWeb, doWeb, pageIsActionRecord, parseOpenURLData;
+
+detectWeb = function(doc, url) {
```
aurimasv: Please follow the general conventions in other Zotero translators:
* Use `function name(args) {` to declare functions unless other methods make more sense. (there's a bit more reason for this than just convention)
* Declare and instantiate variables as needed, not towards the top of scope block.

aurimasv:
Correct, it's not currently exposed, but that's a very trivial change.

WorldCat Discovery.js
```diff
@@ -0,0 +1,185 @@
...
+};
+
+doWeb = function(doc, url) {
+  var activeRecord, isActionRecord, items, numberOfRecords, oclcNum, record, records, saveItems, _i, _len;
+  isActionRecord = pageIsActionRecord(url);
+  records = doc.getElementsByClassName('record');
+  numberOfRecords = records.length;
+  activeRecord = doc.getElementsByClassName('record active')[0];
+  if (!(isActionRecord || numberOfRecords)) {
+	return false;
```
aurimasv: you don't need to check this again. `doWeb` is not called unless `detectWeb` passes

WorldCat Discovery.js
apcshields: Obviously, I'm willing to comply with Zotero translator conventions, or I wouldn't have submitted this, but I'd prefer to do my maintenance of the code in CoffeeScript, which doesn't allow that kind of configuration (for good reasons). I'm reluctant to do that kind of rewriting by hand.

—

WorldCat Discovery.js
```diff
@@ -0,0 +1,185 @@
...
+		"type": "web",
+		"url": "http://goshen.on.worldcat.org/search?sortKey=LIBRARY_PLUS_RELEVANCE&databaseList=638&queryString=monkeys&changedFacet=format&scope=&format=all&format=Compfile&subformat=Compfile%3A%3Acompfile_digital&database=all&author=all&year=all&yearFrom=&yearTo=&language=all",
+		"items": "multiple"
+	},
+	{
+		"type": "web",
+		"url": "http://goshen.on.worldcat.org/oclc/9323519",
+		"items": [
+			{
+				"itemType": "conferencePaper",
```
zuphilip: This is a conference proceeding and should therefore map to the itemType `book`.

WorldCat Discovery.js
```diff
@@ -0,0 +1,185 @@
...
+				"title": "Microcomputers in K-12 education: second annual conference proceedings",
+				"creators": [
+					{
+						"lastName": "Conference of Microcomputers in K-12 Education",
+						"creatorType": "author",
+						"fieldMode": 1
+					},
+					{
+						"lastName": "Barrette",
+						"firstName": "Pierre.",
```
zuphilip: This looks like it needs a cleaning. However, AFAIK we shouldn't delete points if only initials are present.

WorldCat Discovery.js
```diff
@@ -0,0 +1,185 @@
...
+	  var oclcNum, _results;
+	  _results = [];
+	  for (oclcNum in items) {
+		url = "//www.worldcat.org/oclc/" + oclcNum + "?page=endnotealt&client=worldcat.org-detailed_record";
+		_results.push(Zotero.Utilities.doGet(url, function(responseDocument) {
+		  var translator;
+		  responseDocument = responseDocument.trim();
+		  if (responseDocument.match(/^TY\s+-\s+CONF/)) {
+			responseDocument.replace(/^TY\s+-\s+CONF/, 'TY  - BOOK');
+			responseDocument.replace(/^AU\s+-\s+/gm, 'A3  - ');
```
aurimasv: regarding @zuphilip's comments below, this doesn't actually do anything. You need to assign the result back to `responseDocument`. But as I said in my other comment, I think all of this processing should be left up to Open WorldCat translator.

WorldCat Discovery.js
```diff
@@ -0,0 +1,185 @@
...
+		  translator.setHandler('itemDone', function() {
+			var item;
+			item = arguments[arguments.length - 1];
+			item.extra = void 0;
+			item.archive = void 0;
+			if (item.libraryCatalog === 'http://worldcat.org') {
+			  item.libraryCatalog = 'Open WorldCat';
+			}
+			item.title = item.title.replace(/\s+:/, ':');
+			return item.complete();
```
aurimasv: "itemDone" handler doesn't expect a return

WorldCat Discovery.js
```diff
@@ -0,0 +1,185 @@
...
+		  });
+		  return translator.getTranslatorObject(function(translator) {
+			translator.options.defaultItemType = 'book';
+			translator.options.typeMap = {
+			  'ELEC': 'book'
+			};
+			return translator.doImport();
```
aurimasv: also no return here

WorldCat Discovery.js
```diff
@@ -0,0 +1,185 @@
...
+		  });
+		}));
+	  }
+	  return _results;
```
aurimasv: no return for selectItems handler

WorldCat Discovery.js
```diff
@@ -0,0 +1,185 @@
...
+  return url.indexOf('/oclc/') !== -1 && url.indexOf('/search?') === -1;
+};
+
+parseOpenURLData = function(baseItemElement) {
+  var contextObject, item, itemType;
+  contextObject = Zotero.Utilities.xpathText(baseItemElement, './/span[@class="Z3988"]/@title');
+  item = new Zotero.Item();
+  Zotero.Utilities.parseContextObject(contextObject, item);
+  itemType = Zotero.Utilities.xpathText(baseItemElement, './/div[@class="basicbib"]//i[contains(@class, "icon")]/@class');
+  item.type = itemType.indexOf('icon-snd') !== -1 ? 'audioRecording' : itemType.indexOf('icon-compfile') !== -1 || itemType.indexOf('icon-intmm') !== -1 ? 'computerProgram' : itemType.indexOf('icon-map') !== -1 ? 'map' : item.itemType;
```
aurimasv: split into multiple lines. Better yet, make it a switch statement.

aurimasv:
In any case, what I said at first holds. This needs to be cleaned up to follow Zotero conventions and then also passed through JSBeautifier to fix indents.

I'd also encourage you to offload RIS parsing to Open WorldCat. I'll update it to accept OCLC IDs as `{oclcID: "123456"}` You can even assume it takes an array of such objects.

zuphilip:
I am just wondering whether we really want two seperate translators here? It looks for me that we could also do the tests and different processing for "Discovery" in the Open WorldCat translator. The advantage I see, is that we don't have to adjust the priorities for a preferences of the two translators.

aurimasv:
IIRC, the interfaces are quite different and would effectively require
separate code up until OCLC ID is retrieved. If it can be done without much
rewriting, then I agree.

apcshields:
So, all in all, would you like me to make these changes to this translator, or would you prefer to rewrite the Open WorldCat translator to handle both?

aurimasv:
I'll have to take a closer look. Will let you know by the end of the
weekend.

aurimasv:
After taking a closer look, I think that this translator should deal with the Discovery interface and pass on the OCLC IDs to Open WorldCat. You can pass them all at the same time in a single array. For an example, see https://github.com/zotero/translators/blob/master/ScienceDirect.js#L295

As of #840 (once it goes through), the search query should be in the form `{identifiers: {oclc: "12345"}}`.

apcshields:
Ok. I'll work on making these corrections and using the Open WorldCat translator instead of parsing the RIS here.

aurimasv:
#840 was merged, so you can search using OCLC ID now

apcshields:
Sorry I never got back to this. Thanks @aurimasv and @zuphilip for helping with it!

PR #840, opened 2015-01-13
[Open WorldCat] Support searching by OCLC ID (from other translators)

aurimasv:
Minor tweaks based on test cases to clean up authors.

The test case shows the kind of input the translator takes (also takes multiples in a form of an array). In short, `{identifiers: {oclc: "12345"}}` to search by OCLC ID. This might need to be adjusted depending on how general identifier handling evolves in Zotero.

This is related to #839

### DIFF
--- a/Open WorldCat.js
+++ b/Open WorldCat.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 12,
 	"browserSupport": "gcsbv",
-	"lastUpdated": "2014-04-10 23:30:53"
+	"lastUpdated": "2015-01-13 17:09:49"
 }
 
 /**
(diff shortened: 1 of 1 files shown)


zuphilip:
I would like to comment on your format `{identifiers: {oclc: "12345"}}`: This still looks like the classical key-value pairs. Did you consider also URIs for identifiers? Personally, I really like the <a href="http://en.wikipedia.org/wiki/Linked_data#Principles">linked data principles</a> outlined by Tim Berners-Lee and AFAIK oclc has for example 5-star Linked Data.
<blockquote>how general identifier handling evolves in Zotero</blockquote>
If the (general) discussions are on some other place, can you then please point me there?

aurimasv:
I'm not aware of any discussions specifically about the format of
identifiers, just that Zotero should support a number of schemas.

URIs are ok, but they're also long and annoying to work with. In the end,
they end up getting shortened into a more convenient namespace. You can
think of oclc being a namespace for the oclc URI. If we export these
identifiers out of Zotero, we can convert them to the fully qualified form.

I didn't invest a lot of time coming up with this structure and I'm not
against changing it (part of the reason why I created the PR), but I don't
think it's worth it for us to bother with URIs internally.

zuphilip:
Okay, I see. Maybe it is worth to look at http://json-ld.org/ for inspiration.

Issue #829, opened 2014-12-19
Use CrossRef's text mining API to retrieve PDFs by DOI


aurimasv:
http://tdmsupport.crossref.org/researchers/

E.g. http://api.crossref.org/works/10.1155/2014/413629 returns a link to the PDF (application/pdf) http://downloads.hindawi.com/journals/bmri/2014/413629.pdf

Not everything works 100% though. E.g. http://api.crossref.org/works/10.5555/515151 could lead you to 404s or http://api.crossref.org/works/10.7554/eLife.05244 might not even lead you to the PDF

The instructions for researchers (link at the top) talks about click-through service, so there's plenty to consider, but overall this is excellent news.

edit: also have to think about licensing

adam3smith:
Zotero PDF retrieval does some version of this now

PR #841, opened 2015-01-15
Move CrossRef to REST API

aurimasv:
Doesn't look like we're losing a lot and it may improve over time as this is being actively developed, but we are gaining tags and file attachments.

Re https://forums.zotero.org/discussion/45598/identifier-not-working-word-plugin-not-downloading/

### DIFF
--- a/CrossRef.js
+++ b/CrossRef.js
@@ -2,343 +2,252 @@
 	"translatorID": "11645bd1-0420-45c1-badb-53fb41eeb753",
 	"translatorType": 8,
 	"label": "CrossRef",
-	"creator": "Simon Kornblith",
-	"target": "^https?://partneraccess\\.oclc\\.org/",
-	"minVersion": "2.1.9",
+	"creator": "Aurimas Vinckevicius",
+	"target": "",
+	"minVersion": "3.0.9",
 	"maxVersion": null,
 	"priority": 90,
 	"inRepository": true,
-	"browserSupport": "gcsv",
+	"browserSupport": "gcsibv",
 	"lastUpdated": "2014-09-09 20:05:00"
 }
 
-/* CrossRef uses unixref; documentation at http://www.crossref.org/schema/documentation/unixref1.0/unixref.html */
-var ns;
-
-/**********************
- * Utilitiy Functions *
- **********************/
-var xmlSerializer = new XMLSerializer();
-function innerXML(n) {
-	return xmlSerializer.serializeToString(n) //outer XML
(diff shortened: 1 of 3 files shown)


adam3smith:
if you add .xml at the end of the API request, you get, essentially unixref, which still does look more complete

CrossRef.js
```diff
@@ -410,14 +307,52 @@ var testCases = [
 				"bookTitle": "Heimat, Region, and Empire",
 				"ISBN": "9780230391116",
 				"publisher": "Palgrave Macmillan",
-				"language": "en",
 				"date": "2012-10-17",
 				"DOI": "10.1017/CCOL0521858429.016",
```
adam3smith: something is off here - are we not starting with an empty item? This is the Hitchens DOI, not the one belonging to this chapter.

aurimasv: That's quite confusing, but it's actually just a dirty test. I had to hand-craft these, since Scaffold doesn't make tests for search translators (yet). I was copy-pasting the first test and then adjusting based on Translator Tester output, which doesn't pick up on invalid fields ( we don't have DOI for bookSection).. Will fix it.

zuphilip:
How is this exactly connected to #829 ?

aurimasv:
CrossRef API sometimes provides links to PDFs, so this patch attaches those links.

zuphilip:
You mention in the other issue, that this api has some licence. Did you check if that is applicable for our purposes?

aurimasv:
The API provides licenses that accompany the PDFs and the idea is that the user agrees to the license when using the PDF. We don't really have a way for the user to agree to the license ahead of time (or on-the-fly) so the license information is stored in the "rights" fields.  So far I've only seen open licenses (creative commons, etc.) so this shouldn't be an issue. I'd imagine that if the user can't access the resource, he will be denied access whether he agrees with the license or not.

You can read more about the API here: http://tdmsupport.crossref.org/

aurimasv:
>if you add .xml at the end of the API request, you get, essentially unixref, which still does look more complete

Nice find! Based on the test cases we mostly just seem to be missing a couple editors and publisher place (though more testing may turn up other faults). I filed issues upstream. Maybe they'll get resolved soon. I'd like to avoid UNIXREF, since it seems to be having [issues with special characters in the URLs](https://github.com/CrossRef/rest-api-doc/issues/38) currently.

adam3smith:
I don't think we'll do this -- I hope/expect us to use content negotiation for DOIs across the board.

PR #844, opened 2015-01-22
Add new translator for pressian.com - a KR news site

mete0r:
Add new translator for http://pressian.com, which is a online news site in South Korea.

This uses the translator framework.


pressian.com.js
```diff
@@ -0,0 +1,32 @@
+{
+	"translatorID": "3f7b7b18-1152-40b6-b9e4-e397f913641b",
+	"label": "pressian.com",
+	"creator": "mete0r@sarangbang.or.kr",
+	"target": "^https?://www.pressian.com/",
```
aurimasv: Escape periods. Also, based on your detection code below (`detect`), you should just include that part in the target regexp. In Scaffold, it should be `^https:?//www\.pressian\.com/news/article\.html`

The `detect` part can then just be `function() { return true; }` (I think)

pressian.com.js
```diff
@@ -0,0 +1,32 @@
...
+	"minVersion": "1.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "g",
```
aurimasv: As @zuphilip says, you can tick all the boxes here.

pressian.com.js
```diff
@@ -0,0 +1,32 @@
...
+function detectWeb(doc, url) { return FW.detectWeb(doc, url); }
+function doWeb(doc, url) { return FW.doWeb(doc, url); }
+
+
+FW.Scraper({
+itemType	: 'newspaperArticle',
+detect		: FW.Url().match(/\/news\/article\.html/),
+title		: FW.Xpath("//title/text()").text(),
+date		: FW.Xpath('//p[@class="arvdate"]/text()').text(),
+creators	: FW.Xpath('//p[@class="arvdate"]/a/span').text().replace(/ .*/,'').cleanAuthor('author'),
```
aurimasv: Are you sure this is right? You're effectively only taking the first word in the span. Also, if the span starts with a space (not uncommon to have extra whitespace in HTML) you'd be clearing everything.

pressian.com.js
```diff
@@ -0,0 +1,32 @@
...
+function detectWeb(doc, url) { return FW.detectWeb(doc, url); }
+function doWeb(doc, url) { return FW.doWeb(doc, url); }
+
+
+FW.Scraper({
+itemType	: 'newspaperArticle',
+detect		: FW.Url().match(/\/news\/article\.html/),
+title		: FW.Xpath("//title/text()").text(),
+date		: FW.Xpath('//p[@class="arvdate"]/text()').text(),
```
aurimasv: nit: don't need `text()` inside the xpath

pressian.com.js
```diff
@@ -0,0 +1,32 @@
...
+creators	: FW.Xpath('//p[@class="arvdate"]/a/span').text().replace(/ .*/,'').cleanAuthor('author'),
+publicationTitle: "프레시안",
+attachments	: [{
+	url: FW.Url().replace(/article/, 'article_print'),
+	title: 'Printable',
+	type: 'text/html'
+}]
+});
```
aurimasv: Please add a test case (see last tab in Scaffold)

pressian.com.js
```diff
@@ -0,0 +1,32 @@
...
+FW.Scraper({
+itemType	: 'newspaperArticle',
+detect		: FW.Url().match(/\/news\/article\.html/),
+title		: FW.Xpath("//title/text()").text(),
+date		: FW.Xpath('//p[@class="arvdate"]/text()').text(),
+creators	: FW.Xpath('//p[@class="arvdate"]/a/span').text().replace(/ .*/,'').cleanAuthor('author'),
+publicationTitle: "프레시안",
+attachments	: [{
+	url: FW.Url().replace(/article/, 'article_print'),
+	title: 'Printable',
```
aurimasv: Call this "Snapshot"

AbeJellinek:
Closing this for now because it uses FW, unfortunately.

PR #850, opened 2015-02-11
Primo: Correct author, place, publisher and page scraping

nemobis:
Avoids current errors in master, like:
* "First ()" or "Andrea da Barberinocirca" as author
* "Meurs, Jacob" as place
* 4 pages instead of 16+228+4
and others.

### DIFF
--- a/Primo.js
+++ b/Primo.js
@@ -1,7 +1,7 @@
 {
 	"translatorID": "1300cd65-d23a-4bbf-93e5-a3c9e00d1066",
 	"label": "Primo",
-	"creator": "Matt Burton, Avram Lyon, Etienne Cavalié, Rintze Zelle, Philipp Zumstein, Sebastian Karcher, Aurimas Vinckevicius",
+	"creator": "Matt Burton, Avram Lyon, Etienne Cavalié, Rintze Zelle, Philipp Zumstein, Sebastian Karcher, Aurimas Vinckevicius, Fondazione BEIC",
 	"target": "/primo_library/|/nebis/|^https?://www\\.recherche-portal\\.ch/zbz/",
 	"minVersion": "2.1.9",
 	"maxVersion": "",
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsb",
-	"lastUpdated": "2015-04-02 18:40:40"
+	"lastUpdated": "2015-05-04 14:37:00"
 }
 
 /*
@@ -25,6 +25,9 @@ Primos with showPNX.jsp installed:
 (3) http://limo.libis.be/primo_library/libweb/action/search.do?vid=LIBISnet&fromLogin=true
 (4.a) http://virtuose.uqam.ca/primo_library/libweb/action/search.do?vid=UQAM
 (5) http://searchit.princeton.edu/primo_library/libweb/action/dlDisplay.do?docId=PRN_VOYAGER2778598&vid=PRINCETON&institution=PRN
+
+Other Primos where showPnx=true is available:
+(6) http://digitale.beic.it/primo_library/libweb/action/search.do?vid=beic
 */
 
 function getSearchResults(doc) {
@@ -212,7 +215,13 @@ function importPNX(text) {
 	
 	var itemType = ZU.xpathText(doc, '//display/type');
 	if(!itemType) {
-		throw new Error('Could not locate item type');
+		if ( ZU.xpathText(doc, '//facets/rsrctype') ) {
+			var itemType = ZU.xpathText(doc, '//facets/rsrctype');
+		} else if ( ZU.xpathText(doc, '//search/rsrctype') ) {
+			var itemType = ZU.xpathText(doc, '//search/rsrctype');
+		} else {
+			throw new Error('Could not locate item type');
+		}
 	}
 	
 	switch(itemType.toLowerCase()) {
@@ -260,7 +269,7 @@ function importPNX(text) {
 		creators = contributors;
 		contributors = [];
 	}
-	
+
 	// //addata/au is great because it lists authors in last, first format,
 	// but it can also have a bunch of junk. We'll use it to help split authors
 	var splitGuidance = {};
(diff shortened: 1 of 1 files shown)


nemobis:
See individual commits for details. Some of the errors corrected in the previous pull requests were still present and have been re-corrected in a different way on top of the new code.

nemobis:
Ok, we've been using this for dozens conversions from Primo to Wikipedia citation templates and it's working well as far as we can see. Let me know what you'd like me to do.

nemobis:
Is there anything you'd like to see happen here? We would like to make it easier for users to install  the improved Primo.js, as requested by some users outside BEIC.

If you'd prefer to split these changes to a separate translator, please let me know what parts you want to be where. Thanks.

Primo.js
```diff
@@ -277,32 +286,89 @@ function importPNX(text) {
...
+	// "The Minerals, Metals & Materials Society"
+	// "Hodges, Foster & Co."
+	// Also ignore annotations which are not proper names, like:
+	// "printed by Henry Hills, printer to the King's most Excellent Majesty, for His Household and Chappel"
+	// Still to handle:
+	// "Hessisches Ministerium für Landwirtschaft, Forsten und Naturschutz, Landentwicklung"
+	if( item.publisher
+		&& item.date
+		&& item.date < '1831'
+		&& item.publisher[0].toUpperCase() == item.publisher[0]
```
adam3smith: what's the purpose of the test for entirely uppercased publishes here? That's not clear from the comments.

adam3smith:
apart from the one question, I don't have any concerns. This would need to be rebased so we can merge it, and we also request that translator commits are squashed into a single commit, so that we don't clutter the commit history too much.

zuphilip:
I think there are more changes which are because of the different base (nemobis/master) not visible. Is that possible?

Primo.js
nemobis: Good point. It was an attempt at identifying the "names" which are actually phrases and contain uncapitalised common nouns, like "printed by". With the further testing we did, I think however it's doing more harm than good: I'll remove it.

nemobis:
Thanks for the comments a and sorry for the wait, I've been abroad for a while. I'll rebase and squash the commits. I'm not sure how the commit message will turn out to be, let's see.

nemobis:
Should be ok now. My repo was indeed outdated, but the upstream changes did not conflict (they were automatically merged, AFAICS).

aurimasv:
I'm getting "Exception parsing JSON" in Scaffold when trying to open the translator. Please double-check that everything is working correctly.

Primo.js
```diff
@@ -530,6 +613,63 @@ var testCases = [
...
+				"attachments": [],
+				"tags": [
+					"Italia - Storia medioevale",
+					"Leggi"
+				],
+				"notes": [],
+				"seeAlso": []
+			}
+		]
+	},
```
aurimasv: More specifically, this is your problem, but that tells me that you didn't run through tests before submitting the commit. Please always run all tests.

nemobis: Aurimas Vinckevicius, 04/05/2015 17:19:
> More specifically, this is your problem, but that tells me that you
> didn't run through tests before submitting the commit. Please always run
> all tests.

Sorry, I had only run them before all the rebases. They now all succeed, 
except libis.be and princeton.edu which don't respond.

Primo.js
```diff
@@ -278,32 +287,87 @@ function importPNX(text) {
...
+	if(!item.place && possibleplace && pubplace[1]) {
+		// Splitting happened and we need a place, we hope this is it
+		item.place = possibleplace;
+	}
+
+	// //addata/date is sometimes available: ever needed?
+	// //addata/risdate can contain the date as written on the manifestation and
+	// be the only date, e.g. "Anno domini MCCCCLXXXXV die XXVII februarii" in
+	// http://digitale.beic.it/primo_library/libweb/action/display.do?doc=39bei_digitool2394625&showPnx=true
 	var date = ZU.xpathText(doc, '//display/creationdate|//search/creationdate');
```
zuphilip: The comment is about `//addata/date` and `addata/risdate` but the next line is not considering them. I think it is very strange if you cannot see the creation date in the display, but it can be searched for...

Primo.js
```diff
@@ -278,32 +287,87 @@ function importPNX(text) {
...
+		item.publisher = stripAuthor( shuffle2.replace(/ *[;] */g, ", ").replace(/  /g, " ") );
+	}
+
+	// The three letter codes, that should be in the language field, usually work well
+	// (although they may be in MARC21, or ISO 639-2, rather than ISO 639-3 or others)
 	item.language = ZU.xpathText(doc, '(//display/language|//facets/language)[1]');
-	
+
+	// BEIC.it uses //dedup/f9, unclear what that is
 	var pages = ZU.xpathText(doc, '//display/format');
```
zuphilip: Delete that comment

Primo.js
```diff
@@ -278,32 +287,87 @@ function importPNX(text) {
...
+		var shuffle3 = declutter.replace(/ *([^,;]+), +([^,;0-9]+) *[0-9]*, +([^,;]+) */g, "$2 $1 ($3)");
+		// Same, other names
+		var shuffle2 = shuffle3.replace(/ *([^,;]+), +([^,;0-9]+) *[0-9]* */g, "$2 $1");
+		// Use comma list, get rid of space buildup for numbered names, clean like authors
+		item.publisher = stripAuthor( shuffle2.replace(/ *[;] */g, ", ").replace(/  /g, " ") );
+	}
+
+	// The three letter codes, that should be in the language field, usually work well
+	// (although they may be in MARC21, or ISO 639-2, rather than ISO 639-3 or others)
 	item.language = ZU.xpathText(doc, '(//display/language|//facets/language)[1]');
```
zuphilip: I don't understand the two comment lines above here. Maybe, just delete them.

Primo.js
nemobis: > The comment is about //addata/date and addata/risdate but the next line is not considering them.

These are left to a later patch to avoid making this harder than it already is to review.

>  I think it is very strange if you cannot see the creation date in the display, but it can be searched for...

If you mean that `//display/creationdate|//search/creationdate` may be redundant, that's not touched by this patch AFAICS. As for the example in the comment, it might be because it's a data in Roman numbers.

Primo.js
nemobis: Can I move todos to the issue tracker, then?

Primo.js
nemobis: Philipp Zumstein, 04/05/2015 21:41:
> I don't understand the two comment lines above here.

The language codes are not in ISO 639-3. That's subject of my other open 
PR, https://github.com/zotero/translators/pull/835

Primo.js
```diff
@@ -318,6 +382,9 @@ function importPNX(text) {
 	item.edition = ZU.xpathText(doc, '//display/edition');
 	
 	var subjects = ZU.xpath(doc, '//search/subject');
+	if(!subjects.length) {
+		subjects = ZU.xpath(doc, '//display/subject');
+	}
```
zuphilip: I would like to suggest to switch the logic here, i.e.first try to receive the subjects from the display part and just afterwards use the search part. I think it is common to add more writings of the same subject for search improvements, as we do for example in our primo catalogue, see e.g. http://primo.bib.uni-mannheim.de/MAN_UB:MAN_ALEPH001466796 . @adam3smith @aurimasv : What do you think? One part is from a previous commit.

Primo.js
```diff
@@ -278,32 +287,87 @@ function importPNX(text) {
...
+	// BEIC.it also uses //display/lds09, //search/lsr12, //search/lsr03, //facets/lfc03
+	item.place = ZU.xpathText(doc, '//addata/cop');
 	
+	// E.g. "Parma : Viotto", "Ambrogio : da Caponago", "Mantegazza, Filippo"
 	var publisher = ZU.xpathText(doc, '//display/publisher');
-	if(publisher) var pubplace = ZU.unescapeHTML(publisher).split(" : ");
-	if(pubplace && pubplace[1]) {
-		item.place = pubplace[0].replace(/,\s*c?\d+|[\(\)\[\]]|(\.\s*)?/g, "");
+	// E.g. "Viotto", "[Ambrogio da Caponago]", "per magistro Philippo dicto Cassano di Mantegatii"
+	if(!publisher) var publisher = ZU.xpathText(doc, '//addata/pub');
```
zuphilip: You don't need the `var` here.

Primo.js
zuphilip: You are right about the second point. However, I disagree on the first point. I find it harder to review with comments which are not used in the code. @aurimasv What do you think about this?

Primo.js
zuphilip: The comment is specific for BEIC and can be wrong for general Primo instances.

Primo.js
```diff
@@ -278,32 +287,87 @@ function importPNX(text) {
...
-		item.publisher = pubplace[0].replace(/,\s*c?\d+|[\(\)\[\]]|(\.\s*)?/g, "")
-			.replace(/^\s*"|,?"\s*$/g, '');
+		// No splitting, use the whole (cleaned) string
+		item.publisher = possibleplace.replace(/^\s*"|,?"\s*$/g, '');
 	}
-	
+	if(!item.place && possibleplace && pubplace[1]) {
+		// Splitting happened and we need a place, we hope this is it
+		item.place = possibleplace;
+	}
```
zuphilip: The different cases here are very hard to read. Maybe, we can make that cases easier, i.e. something like
```
item.place = ZU.xpathText(doc, '//addata/cop');
var publisher = ZU.xpathText(doc, '//display/publisher');
if(!publisher) publisher = ZU.xpathText(doc, '//addata/pub');
if(publisher) {
	publisher = publisher.replace(/,\s*c?\d+|[\(\)\[\]]|(\.\s*)?/g, "");
	item.publisher = publisher.replace(/^\s*"|,?"\s*$/g, '');
	var pubplace = ZU.unescapeHTML(publisher).split(" : ");
	
	if(pubplace && pubplace[1]) {
		var possibleplace = pubplace[0];
		if(!item.place ) {
			item.publisher = pubplace[1].replace(/^\s*"|,?"\s*$/g, '');
			item.place = possibleplace;
		}
		if(item.place && item.place == possibleplace) {
			item.publisher = pubplace[1].replace(/^\s*"|,?"\s*$/g, '');
		}
	}
}
```
Please not that I haven't tested this code and you might want to add some comments again.

Primo.js
```diff
@@ -351,11 +418,18 @@ function importPNX(text) {
...
 	return str
-		// Remove year
-		.replace(/\s*,?\s*\(?\d{4}-?(\d{4})?\)?/g, '')
 		// Remove things like (illustrator). TODO: use this to assign creator type?
-		.replace(/\s*,?\s*\([^()]*\)$/, '');
+		// Also matches an empty parenthesis
+		.replace(/\s*,?\s*\([^()]*\)$/, '')
+		// Remove almost any remaining parenthesis, starting with numbers (dates) and
+		// letters used for common international abbreviations about dates
+		.replace(/\([\d\sabcefilr.? ]*[-–][\d\abcefilr.? ]*\)/gi, '')
```
zuphilip: Was the comment https://github.com/zotero/translators/pull/822#discussion_r22402268 the last one on this issue?

Primo.js
```diff
@@ -278,32 +287,87 @@ function importPNX(text) {
...
+	// "Marnef, Jérôme de & Cavellat, Guillaume, veuve"
+	// But:
+	// "The Minerals, Metals & Materials Society"
+	// "Hodges, Foster & Co."
+	// Also problematic and still to handle: annotations which are not proper names, like:
+	// "printed by Henry Hills, printer to the King's most Excellent Majesty, for His Household and Chappel"
+	// "Hessisches Ministerium für Landwirtschaft, Forsten und Naturschutz, Landentwicklung"
+	if( item.publisher
+		&& item.date
+		&& item.date < '1831'
```
zuphilip: What was the last comment about this previously?

Primo.js
adam3smith: I'm working on integrating some of this -- decided against this after testing, since '//display/subject' isn't separated into tags, so we'd have to do this with a split, which carries a risk of error.

Primo.js
adam3smith: Not taking this: It's too complex and introduces just as many errors as it fixes.

Primo.js
adam3smith: using @zuphilip's code from here

Primo.js
adam3smith: switched this to prefer `//addata/pub`

Primo.js
```diff
@@ -212,7 +215,13 @@ function importPNX(text) {
...
 	if(!itemType) {
-		throw new Error('Could not locate item type');
+		if ( ZU.xpathText(doc, '//facets/rsrctype') ) {
+			var itemType = ZU.xpathText(doc, '//facets/rsrctype');
+		} else if ( ZU.xpathText(doc, '//search/rsrctype') ) {
+			var itemType = ZU.xpathText(doc, '//search/rsrctype');
+		} else {
+			throw new Error('Could not locate item type');
+		}
 	}
```
adam3smith: Took this.

nemobis:
Thanks for looking into this. The new version is still not good enough for BEIC needs so we'll continue using our own Primo.js in Primo for the foreseeable future, to import thousands of citations in Wikipedia. The code will continue to live in https://github.com/nemobis/BEIC/blob/master/Primo.js

Primo.js
nemobis: It's specific to any library using MARC21 standards (ISO 639-2).

nemobis:
Hi, I just want to confirm that we're still stuck with our own Zotero version since the updates in the official version didn't solve any of our issues, either for the import from Primo or for the export to Wikipedia templates.

I still fail to understand the point of most comments above, sorry. Maybe it's easier if you just test exporting records such as <http://digitale.beic.it/primo_library/libweb/action/search.do?fn=search&ct=search&vid=BEIC&vl(freeText0)=aristoteles> and see yourself whether the current behaviour is expected. I don't know of any context where it would be, so maybe I'm missing something such as radically different conventions outside Italy and Wikimedia on what meaning URLs, parentheses, semicolons, language code etc. are supposed to have.

adam3smith:
The general point of most comments above is that these changes are made based on a single library catalog with very specific data and no broad testing on other catalogs & scenarios, yet affect hundreds of Primo installations worldwide.
Doing too much regex-fu on strings pretty much always breaks as much as it fixes. So what I'm saying is not that the import from BEIC via Primo is perfect, which it clearly isn't, but that we may be hitting limits on what's possible to handle in a broadly, and internationally, used translator.
That's definitely true for the inverted publisher issue, unless this rule is/was used for pre-1831 publications universally, there's too much it can break.

I think there may be most room to improve the handling on parentheses after author names, though, as it's not clear if we'd ever want them (in which case it should actually be possible).

Those are the two only things I don't think I moved over, aren't they?

nemobis:
Sebastian Karcher, 27/09/2017 22:06:
 > unless this rule is/was used for pre-1831 publications universally

That's what I've been told by the cataloguing experts.

 > Those are the two only things I don't think I moved over, aren't they?

I don't know, what I see is that if I export records to Wikipedia 
templates every single field is wrong or missing (or at least names, 
publisher, URL and language), so Zotero master is at the moment 
completely unusable.

Again, I think it's easier for you to check yourself and see if you can 
see anything that actually works for you.

PR #852, opened 2015-02-12
[RIS] Store research notes (RN) as note attachments

aurimasv:
Re https://forums.zotero.org/discussion/46548/importing-research-notes-from-endnote-to-zotero/

Don't see any downside and this makes sense, IMO, but would like someone to second this.

### DIFF
--- a/RIS.js
+++ b/RIS.js
@@ -17,7 +17,7 @@
 	"inRepository": true,
 	"translatorType": 3,
 	"browserSupport": "gcsv",
-	"lastUpdated": "2015-02-11 01:23:23"
+	"lastUpdated": "2015-02-12 02:23:27"
 }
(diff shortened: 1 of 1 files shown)


zuphilip:
Are the examples realistic? I don't see any downsides but I am thinking if a note or a tag is better... BTW you will review #845 further?

aurimasv:
If you're referring to the test case, then that's just a collection of all field labels (as displayed in EndNote). It's only meant to help us figure out if we're mapping things correctly.

zuphilip:
Yes, I was refering to the test cases.

aurimasv:
>BTW you will review #845 further?

Yes, I'm working my way up from some older pull requests.

PR #855, opened 2015-02-19
Declare Unicode encoding in TEI translator.

adunning:
This ensures that XML editors parse the file correctly.

### DIFF
--- a/TEI.js
+++ b/TEI.js
@@ -595,7 +595,7 @@ function doExport() {
     }
 
     // write to file.
-    Zotero.write('<?xml version="1.0"?>'+"\n");
+    Zotero.write('<?xml version="1.0" encoding="UTF-8"?>'+"\n");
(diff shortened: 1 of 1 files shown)


dstillman:
Declaring an XML file as UTF-8 isn't necessary. If an XML parser doesn't interpret it as UTF-8, it's broken.

adunning:
Surely, though, it doesn't do any harm. Not declaring this causes many parsers to change any non-ASCII characters into entities, which is irritating.

dstillman:
That should be a setting of the processor (e.g., ```--encode utf-8``` for xmllint), but all right.

(Note that, for translators to get pushed out to clients before the next release, the timestamp at the top of the file has to be updated too. I've done that for you here.)

adunning:
I always forget about that. Thanks!

PR #861, opened 2015-03-03
fix multiples in WorldCat

adam3smith:
I don't quite understand why the diff doesn't work properly. My indenting looks good to me? The actual change in the translator (apart from tests) is two lines, lines 100 and 177 in the new translator.

@aurimasv - could you have a quick look. This shouldn't, ideally, stay broken long, but I want to be sure I didn't do anything dumb.

### DIFF
--- a/Open WorldCat.js
+++ b/Open WorldCat.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 12,
 	"browserSupport": "gcsbv",
-	"lastUpdated": "2015-01-13 17:09:49"
+	"lastUpdated": "2015-03-03 05:51:41"
 }
 
 /**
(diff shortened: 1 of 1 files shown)


aurimasv:
I'm a little confused why you think the diff is wrong. Everything got indented in the `while`. If you want to ignore white space when comparing, add `?w=1` to the URL.

More importantly, though, what is this in response to? This was written in a recursive manner (and still is, actually, but should be harmless) and I suspect is broken on Firefox only (I haven't actually checked) due to https://github.com/zotero/zotero/issues/647 What behavior were you fixing?

Lastly, It would be good if you could add back the two "search" test cases that got removed by Scaffold (I need to fix that).

adam3smith:
yes, this is in response to: https://forums.zotero.org/discussion/47211/zotero-not-pulling-worldcat-results/ and does look like the same issue.

PR #872, opened 2015-04-03
Added translation for more tags; added diff. date format

rm2342:
1. beck-online sometimes uses upper case letters/different names in their span-tags.
2. for some decisions, beck-online uses date format dd-mm-yyyy instead of dd.mm.yyyy
3. added test case for 1. and 2.


adam3smith:
thanks!

PR #873, opened 2015-04-03
Added different names for type of decision

rm2342:
In some parts of Juris, "Urt." is used instead of "Urteil" and "Beschl." instead of "Beschluss" in the properties of a decision


adam3smith:
looks good but you need to bump the timestamp. Otherwise the translator won't update for users.

rm2342:
Thank you. Just changed the timestamp.

PR #879, opened 2015-04-08
unbreak BiblioCommons translator

gmcharlt:
This patch adjusts the BiblioCommons translator to deal with changes to the markup
of the MARC view in /item/catalogue_info/.


adam3smith:
closed via https://github.com/zotero/translators/commit/3b6e378ca96321e96b040a6ef619f41641616c3e
sorry, we overlapped. I used my patch because it gets the Leader right, which means we can detect item types other than books correctly.

gmcharlt:
Great, thanks!

PR #880, opened 2015-04-10
Changed the Target URL regex to reflect new Host.

ggiordano:
The Legislative Insight application was ported to a new host, Therefore we needed to change the Target URL regex.

### DIFF
--- a/Legislative Insight.js
+++ b/Legislative Insight.js
@@ -2,14 +2,14 @@
 	"translatorID": "2bedae3c-bab5-447f-b127-e9babc0e9cfe",
 	"label": "Legislative Insight",
 	"creator": "Kari Hemdal",
-	"target": "^https?://www\\.(?:conquest-leg-insight-cert|conquest-leg-insight)\\.com/legislativeinsight/LegHistMain\\.jsp",
+	"target": "^https?://((li.proquest|preprod.li.proquest)\.com/legislativeinsight/LegHistMain\.jsp)",
 	"minVersion": "3.0",
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsb",
-	"lastUpdated": "2014-04-03 17:41:39"
+	"lastUpdated": "2015-04-10 10:10:29"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


Legislative Insight.js
```diff
@@ -2,14 +2,14 @@
 	"translatorID": "2bedae3c-bab5-447f-b127-e9babc0e9cfe",
 	"label": "Legislative Insight",
 	"creator": "Kari Hemdal",
-	"target": "^https?://www\\.(?:conquest-leg-insight-cert|conquest-leg-insight)\\.com/legislativeinsight/LegHistMain\\.jsp",
+	"target": "^https?://((li.proquest|preprod.li.proquest)\.com/legislativeinsight/LegHistMain\.jsp)",
```
aurimasv: The double backslashes were correct. `^https?://(preprod\\.)?li\\.proquest\\.com/legislativeinsight/LegHistMain\\.jsp` is what you want.

aurimasv:
So that's the only change that is required to make LI work? I thought it was redesigned

ggiordano:
That's the only change. LI was just ported from server on the Conquest site to ProQuest site. I tested it with firefox using Scaffold and it works.


--Gerard


________________________________

adam3smith:
@ggiordano -- we'll still need the regex fixed as per aurimas's comment inline. Otherwise this is fine.

ggiordano:
I added a single backslash before the dot "." to escape them. If I add a double backslash and test the Target URL by coping it inside scaffold target URL and test the regex, it doesn't match the correct URLs.

adam3smith:
it really does need to be a double backslash, because it's in the JSON header of the translator. 
The Scaffold "Target" field uses normal regex and then converts it (i.e. escapes backslashes) when you save. So copy and paste from Scaffold won't work correctly.

ggiordano:
I added the double backslashes.

adam3smith:
great, thanks. Do you know how to squash this all into one commit? That'd be ideal. Otherwise we'll take care of that on our side.

ggiordano:
I'm not sure how to do that. If you can do it that would be great.

Thanks,

Gerard



________________________________

aurimasv:
Thanks! Sorry for the delay.

PR #888, opened 2015-04-28
Adding a GitHub translator

mfenner:
This works for me, but there is still a problem with parsing author names. Before I fix this (currently trapped in a closure) I want to get some feedback on how to best describe the authors of a github repo:
- username of repo owner
- usernames of contributors (or top 10 contributors)
- real names of contributors

Also, how to describe group accounts? This could be the username (there is no real name associated with them), or left out.

Another issue is the type. Based on similar examples I use `JournalArticle`, but we could also use `ComputerProgram`. Strictly speaking we don't know whether a Github repo contains software.

Lastly, the `DataCreated` and `DateModified` could be grabbed from `/graphs/contributors`, but I need to check for an `onLoad` event, as the relevant information is loaded separately.

### DIFF
--- a/Github.js
+++ b/Github.js
@@ -1,0 +1,183 @@
+{
+	"translatorID": "5e385e77-2f51-41b4-a29b-908e23d5d3e8",
+	"label": "Github",
+	"creator": "Martin Fenner",
+	"target": "^https?://(www\\.)?github\\.com/",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcs",
+	"lastUpdated": "2015-04-27 23:41:42"
+}
+
+/**
+	Copyright (c) 2015 Martin Fenner
+
+	This program is free software: you can redistribute it and/or
+	modify it under the terms of the GNU Affero General Public License
+	as published by the Free Software Foundation, either version 3 of
+	the License, or (at your option) any later version.
+
+	This program is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+	Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public
+	License along with this program. If not, see
+	<http://www.gnu.org/licenses/>.
+*/
+
+function detectWeb(doc, url) {
+	// use item type journalArticle until item type computerProgram is used by citation styles
+	if (getItem(doc, true)) return "journalArticle";
+
+	// search results
+	if (url.indexOf("/search?utf8=✓&q=") != -1 && getSearchResults(doc, true)) return "multiple";
+}
+
+function doWeb(doc, url) {
+	if (detectWeb(doc, url) == "multiple") {
+		getItems(doc);
+	} else {
+		getItem(doc);
+	}
+}
+
+function getSearchResults(doc, checkOnly) {
+	return false;
+}
+
+function getSelectedJSON(doc) {
+	var items = getSearchResults(doc);
+	var ids = [];
+
+	Zotero.selectItems(items, function(selectedItems) {
+		if (!selectedItems) return true;
+
+		for (var i in selectedItems) {
+			ids.push(i.substr(1));
+		}
+		getJSON(ids);
+	});
+}
+
+function getItem(doc, checkOnly) {
+	var url = ZU.xpathText(doc, '/html/head/meta[@property="og:url"]/@content');
+ 
+	if (!url) return false;
+	if (checkOnly) return true;
+	
+	var item = new Z.Item("journalArticle");
+	item.url = url;
+	item.title = ZU.xpathText(doc, '/html/head/meta[@property="og:description"]/@content');
+	
+	// use archive and archive location
+	item.archive = "Github";
+	item.archiveLocation = url;
+	
+	var author = ZU.xpathText(doc, '/html/head/meta[@name="octolytics-dimension-user_login"]/@content');
+	item.creators = [{ "firstName": '', 
+					   "lastName": author, 
+					   "creatorType": "author" }];
+	
+	// indicate that this is in fact a software repository
+	item.extra = "{:itemType: computer_program}";
+	
+	item.language = "en-US";
+	
+	var attachment = ZU.xpathText(doc, '//a[@class="js-directory-link"]');
+	if (attachment) {
+		item.attachments.push({
+			url: url + "/blob/master/" + encodeURIComponent(attachment),
+			title: attachment,
+			mimeType: "text/plain",
+			snapshot: false
+		});
+	}
+	
+	item.complete();
+	
+	return item;
+}
+
+function getItems(ids) {
+	return false;
+}/** BEGIN TEST CASES **/
+var testCases = [
+	{
+		"type": "web",
+		"url": "http://www.ncbi.nlm.nih.gov/nuccore/?term=brca1",
+		"items": "multiple"
+	},
+	{
+		"type": "web",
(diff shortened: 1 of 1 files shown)


Github.js
```diff
@@ -0,0 +1,151 @@
+{
+	"translatorID": "5e385e77-2f51-41b4-a29b-908e23d5d3e8",
+	"label": "Github",
```
aurimasv: GitHub (capital H)

Github.js
```diff
@@ -0,0 +1,151 @@
...
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+	Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public
+	License along with this program. If not, see
+	<http://www.gnu.org/licenses/>.
+*/
+
+function detectWeb(doc, url) {
+	if(url.indexOf("/search?") != -1 && getResults(doc).length) {
```
aurimasv: This should call `getSearchResults`, since that's what's used in `doWeb`

Github.js
```diff
@@ -0,0 +1,151 @@
...
+		return "multiple";
+	} else if(getResult(doc)) {
```
aurimasv: no need to split this off into another function, since it's not used anywhere else

Github.js
```diff
@@ -0,0 +1,151 @@
...
+		return "journalArticle";
+	}
+}
```
aurimasv: The logic should be tweaked a little bit. If the URL has `/search?` in it it should return `false` if there are no search results. Otherwise, a search page with no results that has a `meta[@property="og:description"]` tag (for whatever reason, maybe a future change) will return "journalArticle"

Github.js
```diff
@@ -0,0 +1,151 @@
...
+
+function doWeb(doc, url) {
+	if (detectWeb(doc, url) == "multiple") {
+		getSelectedItems(doc);
```
aurimasv: I would prefer if you pulled the contents of this function into `doWeb`, since it's not reused and it just creates a lot of unnecessary functions

Github.js
```diff
@@ -0,0 +1,151 @@
...
+}
+
+function scrape(doc, url) {	
+	var item = new Z.Item("journalArticle");
+	item.url = ZU.xpathText(doc, '/html/head/meta[@property="og:url"]/@content');
+	item.title = ZU.xpathText(doc, '/html/head/meta[@property="og:description"]/@content');
+	
+	// use archive and archive location
+	item.archive = "Github";
+	item.archiveLocation = item.url;
```
aurimasv: no need to duplicate this metadata. Also, we would generally put "GitHub" in Library Catalog, not Archive

Github.js
```diff
@@ -0,0 +1,151 @@
...
+		var urls = [];
+		for(var i in selectedItems) {
+			urls.push(i);
+		}
+		ZU.processDocuments(urls, scrape);
+	});
+}
+
+function scrape(doc, url) {	
+	var item = new Z.Item("journalArticle");
```
aurimasv: this should be "computerProgram". While strictly speaking, it could be anything, primarily most repos are software

Github.js
```diff
@@ -0,0 +1,151 @@
...
+	item.archive = "Github";
+	item.archiveLocation = item.url;
+	
+	var username = ZU.xpathText(doc, '/html/head/meta[@name="octolytics-dimension-user_login"]/@content');
+	item.creators.push(getAuthor(username));
+	
+	// indicate that this is in fact a software repository
+	item.extra = "{:itemType: computer_program}";
+	
+	item.language = "en-US";
```
aurimasv: there is currently no `language` field for the metadata language in a `computerProgram` item type, but there is a `programmingLanguage` field that could be populated. We could store this in the Extra field.

Github.js
```diff
@@ -0,0 +1,151 @@
...
+	item.attachments.push({
+		mimeType: "text/plain",
```
aurimasv: I'm guessing you want to link to the repo here. `mimeType: "text/html"` in that case. `text/plain` doesn't make much sense here

Github.js
```diff
@@ -0,0 +1,151 @@
...
+		document: doc,
+		snapshot: false
```
aurimasv: This needs a title (otherwise it grabs the page title, which is not something we want). Something like "GitHub Repository" should be good.

Github.js
```diff
@@ -0,0 +1,151 @@
...
+	});
+	
+	item.complete();
+	return item;
```
aurimasv: don't need to return

Github.js
dstillman: Why is there an attached link to the repo? Isn't that already in the URL field (where it should be, since it's the canonical link)?

Github.js
```diff
@@ -0,0 +1,151 @@
...
+}
+
+function getResults(doc) {
+	return doc.getElementsByClassName('repo-list-item');
+}
+
+// get the full name from the author profile page
+function getAuthor(username) {
+    var url = "https://github.com/" + encodeURIComponent(username);	
+	ZU.processDocuments(url, function(text) {
```
aurimasv: You should use GitHub API to retrieve this information https://developer.github.com/v3/users/

Github.js
```diff
@@ -0,0 +1,151 @@
...
+		var author = ZU.xpathText(text, '//span[@class="vcard-fullname"]');
+		if (!author) { author = ZU.xpathText(text, '//span[@class="vcard-username"]'); }
+	    if (!author) { author = ZU.xpathText(text, '/html/head/meta[@property="profile:username"]/@content'); }
+	    author = ZU.cleanAuthor(author, "author");
+	});
+	// temporary, until we get the author string out of the closure
+	return ZU.cleanAuthor(username, "author");
```
aurimasv: Since you need to make an _asynchronous_ HTTP request, you cannot return the result directly. You will have to call a callback function from the `ZU.doGet` handler. Similar to what you do in `getSelectedItems` with `ZU.processDocuments`

aurimasv:
In general, I think most of the data you're scraping should be coming directly from the GitHub API. There are a lot of useful API function available. For info about a particular repo, you have https://developer.github.com/v3/repos/#get. For info about repo contributors, you have https://developer.github.com/v3/repos/statistics/

Regarding who is the author, I don't have a straightforward answer. The repo owner would certainly be an author, I think. Given the statistics available, we could also look at percent contribution from other contributors and use that to determine if the author should be listed. You may also be able to do something with https://developer.github.com/v3/repos/collaborators/

Github.js
aurimasv: Fair point. I agree.

mfenner:
@aurimasv thanks for all the helpful feedback, will update the pull request.

One limitation of the Github APIs is that they require authentication, which limits the usefulness in a web translator. Have you found a workaround for authenticated APIs in other web translators?

aurimasv:
Most read requests do not require authentication. E.g. https://api.github.com/users/aurimasv and https://api.github.com/repos/zotero/zotero

From a quick look, I didn't get stats to work, so that may require authentication (or I'm doing something wrong). I'll take a closer look later.

adam3smith:
@mfenner -- are you interested in taking this up again, given your recent blogpost?
For item type, I'd prefer 'computerProgram' -- you can then add DOIs, where they exist, using 
```DOI: 10.xlkj/asdklf``` in the Extra field.

File: Github.js [hunk 14]
```diff
@@ -0,0 +1,151 @@
+{
+	"translatorID": "5e385e77-2f51-41b4-a29b-908e23d5d3e8",
```
zuphilip: translatorID must be unique and you cannot reuse the one from your other translator https://github.com/zotero/translators/blob/bd311c6725863233f24c8cfd00ae842aa2a38e74/NCBI%20Nucleotide.js#L2

(This is similar to doi's which are unique;-)

zuphilip:
Continued now in #1296

adam3smith:
superseded by https://github.com/zotero/translators/issues/1296

PR #889, opened 2015-05-10
Added translator to export file paths of attachments.

publicus:
Using the CSV.js translator as a template, I've created a translator that exports a list of filepaths of attachments. The translator offers two options that should facilitate the use of attachments with external scripts:

1. Export all filepaths in one line (the default is to export each file on its own line).
1. Export only plaintext files (using a list of file extensions that is easy to change within the translator file) (the default is to export the filepaths of all attachments, whether they are pdf, txt, markdown,etc.).

I've squashed my commits; thus, if I've done this correctly, there should only be one commit in the Pull Request. I appreciate your consideration!

### DIFF
--- a/AttachmentFilepaths.js
+++ b/AttachmentFilepaths.js
@@ -1,0 +1,102 @@
+{
+	"translatorID": "e7eb5c9d-d90d-40c4-98a6-a943529ad23e",
+	"label": "Attachment Filepaths",
+	"creator": "Jacob Levernier, editing work by Philipp Zumstein and Aurimas Vinckevicius",
+	"target": "txt",
+	"minVersion": "1.0",
+	"maxVersion": "",
+	"priority": 100,
+	"displayOptions": {
+		"exportCharset": "UTF-8xBOM",
+		"Only Export Paths of Plaintext Files": false,
+		"Export All Filepaths on One Line": false
+	},
+	"inRepository": true,
+	"translatorType": 2,
+	"browserSupport": "g",
+	"lastUpdated": "2015-04-22 18:40:31"
+}
+
+/*
+	This export translator is based on the CSV.js Zotero translator (v3.0) by Philipp Zumstein and Aurimas Vinckevicius. Like the CSV translator, is it licensed under the AGPLv3 (or any later version).
+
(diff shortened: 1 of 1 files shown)


aurimasv:
Feel free to convince me otherwise, but I don't think this is a good fit for a translator. This may be useful as a Zotero feature (currently seems to be more aligned with what ZotFile does), but I'm also not entirely sure what the use of it would be. I can't recall anyone requesting this in the past (though I may be misremembering) and I have a hard time thinking of when I last needed to know the file path of an attachment.

dstillman:
Yes, this could be implemented as a plugin that added a context menu option to copy file paths to the clipboard. But in most cases I imagine an OS smart folder pointing to the 'storage' directory would be easier and sufficient.

adam3smith:
I believe Zutilo can already show and/or copy filepaths.

publicus:
Whether this exporter is accepted or not, thank you all for your consideration and feedback. Zotero is a great project.

The lack of this feature in Zotero has been a show-stopper for me for the last several months. Being able to quickly export attachment filepaths is useful for anyone who is using a terminal to do any processing steps on attachments. In my own use-case, I like to keep notes on different citations in plain-text files, allowing me to use bash to search them, concatenate them, etc. as different projects require. I could also see this being useful if one wanted to use a tool like `pdftk` to combine PDFs, or even just to quickly get an overview of where files are being stored (if the Zotero storage directory isn't being used).

While Zutilo can already show filepaths, as @adam3smith thought, to my knowledge, it only shows an alert box for each filepath, one at a time, forcing the user to manually copy and paste each into a separate document. This is useful for looking at one or two attachments, but quickly becomes inconvenient for larger numbers of files.

If this doesn't seem like a good match for a translator, I would be very grateful to see it as a feature in Zotero proper, as @aurimasv and @dstillman suggested. Without this feature, I wasn't able to use Zotero; my use case doesn't seem very unreasonable to me, so I suspect that others users may have a similar need, as well. Another user actually asked me about this feature specifically recently, [here](http://www.reddit.com/r/LinuxActionShow/comments/2zel4g/markdown_mapper_autogenerate_concept_maps_from/cqkd52n).

PR #890, opened 2015-05-21
Fixes #703

### LINKED ISSUE #703: IACR eprint translator broken
The IACR translator is not working:

https://eprint.iacr.org/2014/218

FredericJacobs:


adam3smith:
this doesn't work, unfortunately -- while the attachment request come out OK, Zotero doesn't automatically convert the string saved to the URL field into an absolute link (which it should be for printing). You probably need to create a separate variable and then tag on either the correct http/https or just default to https for the URL.

Also, you need to update the date field and re-run/update tests -- strongly recommended doing so in scaffold. https://www.zotero.org/support/dev/translators/scaffold

FredericJacobs:
Thanks, will open a new PR when I have figured it out.

aurimasv:
You're also using Chrome connector without Standalone, right? (I think that's the only way you would get the errors you are getting). You're welcome to continue doing so, but not having Standalone open will sometimes result in poorer metadata quality and may be noticeably slower.

FredericJacobs:
Correct. Just installed Zotero Standalone and it works now.

aurimasv:
The error you're reporting should still not happen though, so if you want to submit a patch for it, we'll accept it. If not, we'll end up fixing this at some point.

PR #891, opened 2015-05-21
Fixes #703

### LINKED ISSUE #703: IACR eprint translator broken

FredericJacobs:
ePrint serves all papers over HTTPS. Prevents unsafe AJAX loading issues in Chrome.
I believe this is the easiest and quickest fix.

### DIFF
--- a/ePrint IACR.js
+++ b/ePrint IACR.js
@@ -55,7 +55,7 @@ function scrape(doc, url) {
 	
 	newItem.date = year;
 	newItem.reportNumber = no;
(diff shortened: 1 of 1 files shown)


aurimasv:
That's fine, but you need to update tests and bump the timestamp (top of file). We really recommend using [Scaffold](https://www.zotero.org/support/dev/translators/scaffold) (Firefox extension) for both of those

PR #895, opened 2015-06-01
Add A5 custom RIS field to handle "contributor"

### LINKED ISSUE #894: RIS export of "contributor" field for Book not handled
While RIS.js includes an inline comment that the "contributor" field is not handled, later on it does handle it for the special cases of conferencePaper and film. This suggests that the RIS import/export should handle it in some way.

I'm working with a bibliography that includes a number of contributors for books that contain chapters co-authored by a number of different contributors, where "contributor" appears to be the most appropriate category for their contribution to the work.

For example, Roth in https://books.google.ca/books?id=bp3_Uv0Ul3MC&lpg=PA320&dq=1551930455&pg=PP8#v=onepage&q=roth&f=false appears as a credited contributor to a chapter in the book, where the primary authors are listed separately.

For now, I'm going to propose adding a custom A5 section for export purposes, as otherwise the data is silently dropped.

dbs:
Rather than silently dropping "contributor" entries for books when exporting to
RIS, export them to a new A5 field.

Fixes #894.

Signed-off-by: Dan Scott <dan@coffeecode.net>

### DIFF
--- a/RIS.js
+++ b/RIS.js
@@ -255,6 +255,9 @@ var fieldMap = {
 		"creators/counsel":["case"],
 		"creators/contributor":["conferencePaper", "film"]	//translator does not fit these
 	},
+	A5: {
+		"creators/contributor":["book", "bookSection"]	//translator does not fit these
+	}
 	C1: {
 		filingDate:["patent"], //not in spec
 		"creators/castMember":["radioBroadcast", "tvBroadcast", "videoRecording"],
@@ -1763,7 +1766,7 @@ function doImport() {
 //RIS files have a certain structure, which is often meaningful
 //Records always start with TY and end with ER. This is hardcoded below
 var exportOrder = {
-	"__default": ["TI", "AU", "T2", "A2", "T3", "A3", "A4", "AB", "C1", "C2", "C3",
+	"__default": ["TI", "AU", "T2", "A2", "T3", "A3", "A4", "A5", "AB", "C1", "C2", "C3",
 	"C4", "C5", "C6", "CN", "CY", "DA", "PY", "DO", "DP", "ET", "VL", "IS", "SP",
 	"J2", "LA", "M1", "M3", "NV", "OP", "PB", "SE", "SN", "ST", "SV", "UR", "AN",
 	"DB", "Y2", "L1", "L2", "L4", "N1", "KW"],


aurimasv:
Sorry, I don't think we will be adding any new unofficial or not widely used tags (is A5 recognized by any other software?) to this poorly designed format. We wouldn't want to contribute to the havoc that already exists. There are generally better alternatives to RIS for transferring data depending on your situation.

You're of course welcome create and use any custom translator you want.

I'll close this tomorrow if there are no objections.

dstillman:
Agreed.

dbs:
Okay. So you're not concerned about "Contributors" just silently being dropped from RIS output, then? As poorly designed as it might be, is a pretty standard format for exchanging with other citation managers--see http://www.refworks.com/rwathens/help/Exporting_from_Bibliographic_Programs_and_Importing_into_RefWorks.htm for example, which almost exclusively relies on RIS.

I'm certainly open to alternative suggestions. For now, I'll use a forked version of the RIS.js translator for my needs.

zuphilip:
I agree here with Aurimas and Dan S.
 * (In my experience) users that change from one reference management system to another by using RIS may complain about file links, urls, notes, collections but not something like contributors.
 * Moreover, I don't think any citation style cares about contributors, for example contributors are not mapped anywhere to CSL (see <a href="http://aurimasv.github.io/z2csl/typeMap.xml#map-bookSection">here</a>).
 * Alternatively, I would suggest to reconsider your example, maybe you can switch the function to author, bookAuthor, editor, seriesEditor, or translator alternatively.

aurimasv:
>As poorly designed as it might be, is a pretty standard format for exchanging with other citation managers

But I'm not aware of any of the reference managers supporting A5 tag, so adding it to Zotero would do no good for metadata interchange. For transferring data between Zotero instances, there are much better options, like the Zotero RDF or group libraries.

adam3smith:
I'm curious what @dbs actual use case here is, maybe that would help and move that somewhere more productive. I agree we don't just want to make up RIS tags.

dbs:
Thanks for the consideration. @adam3smith -- as referenced shortly after opening this issue, my actual use case was linked at https://github.com/dbs/ris2web/issues/18

In short, I'm taking a Zotero group library, where multiple people are contributing to that Zotero group library, and exporting it to RIS so that it can be reingested into a relational database. From there, I'm using it as the basis of a smaller, faster dedicated bibliography.

I did look at the Zotero RDF export format but it's not particularly easy to work with for this purpose. Untangling and normalizing the user-generated cruft from Zotero RDF, even though it loads relatively easily into platforms like Apache Marmotta, is painful.

That said, I might refactor my effort to use the MODS export format for sourcing the relational data instead. I had run into some head-scratching issues my first time around (e.g. a bib:Thesis with an @about="#item_535", while there is no matching @id="#item_535" in the exported file), but those quirks aside, it does seem to be less lossy than RIS.

You can consider this issue closed. Thanks again to everyone for even considering it.

aurimasv:
CSV format may be worth a look as well, though it doesn't deal terribly well with fields that can contain multiple values, e.g. contributors.

adam3smith:
if this is about citations, what @zuphilip says bears repeating -- when people enter someone as a contributor into Zotero, that person will/can never be included in a citation and you should expect data entry (and import translator) behavior along those lines.

In the example you give in the other issue, Reuben Roth is actually an author (as are Livingstone and Harold). The latter two should also be entered as book authors.

PR #899, opened 2015-06-05
Enhanced support for new release of Artstor and Shared Shelf Commons

venturi2k:
This release provide enhanced support for Artstor and SSC (Shared Shelf Commons). It also works with new release of Artstor that provide better item look up support.

### DIFF
--- a/ARTstor.js
+++ b/ARTstor.js
@@ -1,22 +1,25 @@
 {
-	"translatorID": "5278b20c-7c2c-4599-a785-12198ea648bf",
-	"label": "ARTstor",
-	"creator": "Sebastian Karcher",
-	"target": "^https?://library\\.artstor\\.org",
-	"minVersion": "3.1",
-	"maxVersion": "",
-	"priority": 100,
-	"inRepository": true,
-	"translatorType": 4,
-	"browserSupport": "gcs",
-	"lastUpdated": "2014-11-13 15:09:17"
+    "translatorID": "5278b20c-7c2c-4599-a785-12198ea648bf",
+    "label": "ARTstor",
+    "creator": "John Justin, Charles Zeng",
+    "target": "\.artstor|\.sscommons\.org:?\w*\/(open)*library",
+    "minVersion": "3.1",
+    "maxVersion": "",
+    "priority": 100,
+    "inRepository": true,
+    "translatorType": 4,
+    "browserSupport": "gcs",
+    "lastUpdated": "2015-06-02 15:48:45"
 }
 
 /*
+	This translator works for Artstor library site (http://library.artstor.org) and
+	Artstor Shared Shelf Commons (http://www.sscommons.org)
+
+
 	***** BEGIN LICENSE BLOCK *****
 	
-	ARTstor Translator, Copyright © 2012 Sebastian Karcher
-	This file is part of Zotero.
+	Artstor Translator, Copyright © 2015 John Justin, Charles Zeng
 	
 	Zotero is free software: you can redistribute it and/or modify
 	it under the terms of the GNU Affero General Public License as published by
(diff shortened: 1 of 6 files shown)


venturi2k:
Additional Support for Artstor Items
- Can select the items from thumbnail page (browse and search result)
- Can get the item data from image viewer
- Work with both Artstor site (library.artstor.org/library) and SSC site (http://www.sscommons.org/openlibrary)
- More complete metadata capture.

ARTstor.js
```diff
@@ -1,22 +1,25 @@
 {
-	"translatorID": "5278b20c-7c2c-4599-a785-12198ea648bf",
-	"label": "ARTstor",
-	"creator": "Sebastian Karcher",
-	"target": "^https?://library\\.artstor\\.org",
-	"minVersion": "3.1",
-	"maxVersion": "",
```
aurimasv: `"https?://([^/]+\\.)?(artstor|sscommons)\\.org\/(open)?library"` is what this should probably be (for proxy reasons). Is port number really part of the url?

aurimasv: though if there are no additional subdomains `"https?://(library\\.artstor|www\\.sscommons)\\.org\/(open)?library"` is better

ARTstor.js
```diff
@@ -34,98 +37,1043 @@
...
-}
-
 function detectWeb(doc, url) {
-	//monitor changes to body's direct children. That's where the metadata popup is added
-	Zotero.monitorDOMChanges(doc.body, {childList: true});
 
-	if (getData(doc, true)) {
-		return "artwork";
-	}
+    if (url.match(/iv2/)) {
```
aurimasv: `url.indexOf('...') != -1` if just testing for a plain string. Or `/iv2/.test(url)` if you want to use regexp.

this test though is very vague. I can imagine iv2 appearing in a lot of places. Can you give this more context?

ARTstor.js
```diff
@@ -34,98 +37,1043 @@
...
+
+        return "artwork";
+    }
+    if (url.match(/imagegroup/) && doc.getElementById("ssContentWrap").style.display == "inline") {
```
aurimasv: if `doc.getElementById("ssContentWrap")` returns null, this will throw and everything will break (and we will get an error report), so you need to test to make sure that element exists first.

ARTstor.js
aurimasv: Actually more context would be good for all of these tests. E.g. are they surrounded by `/`? Do they have `.php` following them, etc.

ARTstor.js
```diff
@@ -34,98 +37,1043 @@
...
+
+        return false;
+    }
+    if (url.match(/imagegroup/) && doc.getElementById("thumbNavSave1").style.display == "block") {
+        Zotero.debug("Save on");
+        return false;
+    }
+    if (url.match(/(S|s)earch/) || url.match(/categories/) || url.match(/imagegroup/) || url.match(/collections/) || url.match(/collaboratoryfiltering/) || url.match(/cluster/)) {
```
aurimasv: here you might as well use the regexp. `/search|categories|imagegroup|..../i.test(url)` (also note the use of `i` flag)

ARTstor.js
```diff
@@ -34,98 +37,1043 @@
...
+        if (doc.getElementsByClassName('MetaDataWidgetRoot').length > 0) {
+            return "artwork";
+
+        } else
+        if (doc.getElementById("floatingPlaceHolder").style.display == "block") {
```
aurimasv: either
`} else if {`
or
```
}
else if {
```

ARTstor.js
```diff
@@ -34,98 +37,1043 @@
...
+
 }
 
-function associateData(newItem, value, field) {
-	if (!value) return;
-	if (typeof value == 'string') value = ZU.trimInternal(value);
-	if (!value) return;
-	
-	newItem[field] = value;
+function decodeSearchData(str) {
```
aurimasv: Could you add a comment showing what you're trying to extract here?

ARTstor.js
```diff
@@ -34,98 +37,1043 @@
...
+    var searchParam = str;
```
aurimasv: no need to do this. just use `str.replace` below (it returns a copy of the string with replacements)

ARTstor.js
```diff
@@ -34,98 +37,1043 @@
...
+    searchParam = searchParam.replace(/(&gt;)/, ":");
```
aurimasv: FYI, this only replaces the first copy of `&gt;`, not all of them. use `g` flag if you want to replace all (maybe you don't)

ARTstor.js
```diff
@@ -34,98 +37,1043 @@
...
+    var param = searchParam.split('&');
+    var searchData = new Object();
+    var item;
+    var sparam;
+    var id;
+    var value;
```
aurimasv: don't declare variables ahead of time. Declare them as they're being used first.

ARTstor.js
```diff
@@ -34,98 +37,1043 @@
...
+    for (var i = 0; i < param.length; i = i + 1) {
```
aurimasv: FYI, just `i++` is ok for the last parameter here

ARTstor.js
```diff
@@ -34,98 +37,1043 @@
...
+        item = param[i];
+        sparam = item.split('=');
```
aurimasv: feel free to chain things like this. So `var sparam = param[i].split('=');`

ARTstor.js
```diff
@@ -34,98 +37,1043 @@
...
+        id = sparam[0];
+        value = sparam[1];
+
+        if (id === 'type') {
+            searchData[id] = parseInt(value);
+        } else
+            searchData[id] = value;
```
aurimasv: Add braces.
```
} else {
...
}
```

ARTstor.js
```diff
@@ -34,98 +37,1043 @@
...
+    if (url.match(/iv2/)) {
+
+        return "artwork";
+    }
+    if (url.match(/imagegroup/) && doc.getElementById("ssContentWrap").style.display == "inline") {
+
+        return false;
+    }
+    if (url.match(/imagegroup/) && doc.getElementById("thumbNavSave1").style.display == "block") {
+        Zotero.debug("Save on");
```
aurimasv: not really sure what this means

ARTstor.js
```diff
@@ -34,98 +37,1043 @@
...
+    '5E': '^',
+    '5F': '_',
+    '60': '`',
+    '7B': '{',
+    '7C': '|',
+    '7D': '}',
+    '7E': '~'
+};
+
+function decrypt(s) {
```
aurimasv: Can you explain what you're trying to decrypt/encrypt here? Looks to me like at least for part of it you're just doing [`decodeURIComponent`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent)/[`encodeURIComponent`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent)

ARTstor.js
```diff
@@ -34,98 +37,1043 @@
...
+    '_': '5F',
+    '`': '60',
+    '{': '7B',
+    '|': '7C',
+    '}': '7D',
+    '~': '7E'
+};
+
+
+function processIVURL(text1, artItem, serverURL, url) {
```
aurimasv: please comment on what these functions do and give an example.

ARTstor.js
```diff
@@ -34,98 +37,1043 @@
...
+    service = service.substring(service.indexOf("?"));
+    service = service.trim();
+    var ivURL = serverURL + "secure/ViewImages" + service + "&zoomparams=&fs=true";
+
+    if (url == "") {
+        artItem.url = ivURL;
+    } else {
+        //	artItem.url=url;
+        artItem.url = ivURL;
+    }
```
aurimasv: so this does the same thing either way?

ARTstor.js
```diff
@@ -34,98 +37,1043 @@
...
+                    var serviceURL4IVURL = serviceURL4Metadata + item + "?_method=FpHtml";
+                    Zotero.Utilities.HTTP.doGet(serviceURL4IVURL, function(text) {
+                        processIVURL(text, artItem, serverURL, url);
+                    });
+
+                    break;
+                case "11":
+                    var serviceURL4QTVR = serverURL + "secure/imagefpx/" + item + "/7"; //QTVR
+                    Zotero.Utilities.HTTP.doGet(serviceURL4QTVR, function(text) {
+                        var jsonQTVR = eval('(' + text + ')');
```
aurimasv: That's no good. Can't use `eval` just like that (it's not secure). Seems like you just need `JSON.parse()`

ARTstor.js
```diff
@@ -34,98 +37,1043 @@
...
+                        artItem.attachments.push({
+                            title: "Artstor Video File Attachment",
+                            url: urlVID,
+                            mimeType: "text/html"
+
+                        });
+                    });
+                    artItem.url = url;
+                    break;
+            }
```
aurimasv: most of these are almost identical. Couldn't you create a map between `objectTypeId` and the variable parts? something like
```
var typeMap {
  '12': {
    id: 10,
    mimeType: 'audio/mpeg3',
    attachmentTitle: '....'
  }
}
```
then use that with a single piece of code

ARTstor.js
```diff
@@ -34,98 +37,1043 @@
...
+                        artItem.notes.push({
+                            note: comment
+                        });
+                    }
+
+                }
+
+
+            });
+            artItem.complete();
```
aurimasv: This cannot be called synchronously. You will have to call it from "done" handlers for the `doGet` calls you use above. Since there are two of them, this gets more complicated. You'll have to make your commentaries call follow your attachments call and then at the end call `artItem.complete()`

ARTstor.js
```diff
@@ -34,98 +37,1043 @@
...
+            next();
+        })
+    })();
+    Zotero.done();
```
aurimasv: you don't need this

ARTstor.js
```diff
@@ -34,98 +37,1043 @@
...
+    newArtItem.archive = "";
+    newArtItem.rights = "";
+    newArtItem.date = "";
+    newArtItem.artworkMedium = "";
+    newArtItem.artist = "";
+    newArtItem.artworkSize = "";
+    newArtItem.title = "Unknown";
+    newArtItem.abstractNote = "";
+    newArtItem.archiveLocation = "";
+    newArtItem.extra = "";
```
aurimasv: you don't need to initialize these

ARTstor.js
```diff
@@ -34,98 +37,1043 @@
...
+
+    var portal = serverURL.substring(7, serverURL.indexOf("."));
```
aurimasv: better use a regexp to extract this, since this could he an https url as well

ARTstor.js
```diff
@@ -34,98 +37,1043 @@
...
+
+    Zotero.debug("getMetadata  portal  " + portal);
+    switch (portal) {
+        case "archaeology":
+            var siteName = false;
+
+            for (var i = 0; i < json.metaData.length; i++) {
+                switch (json.metaData[i].fieldName) {
+                    case "Site Name":
+                        newArtItem.title = (newArtItem.title == "Unknown") ? json.metaData[i].fieldValue.replace(/<\/?[^>]+(>|$)/g, " ") : newArtItem.title + "; " + json.metaData[i].fieldValue.replace(/<\/?[^>]+(>|$)/g, " ");
```
aurimasv: See [`ZU.cleanTags`](https://github.com/zotero/zotero/blob/4.0/chrome/content/zotero/xpcom/utilities.js#L255)

aurimasv: Also, why would there be multiple titles here?

aurimasv:
Sorry, I can't review this any further. In general, you need to do the following:
* run your code through http://jsbeautifier.org/ and use tabs for indents. With examples where necessary
* add comments explaining what your code does
* figure out how to combine a lot of the repetitive code
* add test cases for all the pages you are dealing with using [Scaffold](https://www.zotero.org/support/dev/translators/scaffold).

ARTstor.js
venturi2k: There could be multiple portals under artstor.org. It supports *.artstor.org

ARTstor.js
venturi2k: The url contains /iv2.html, such as http://www.sscommons.org/openlibrary/iv2.html

ARTstor.js
venturi2k: Good. I have added additional check for valid object before perform further operation.

venturi2k:
Please review the code and add it to the Zotero master/main branch.

ARTstor.js
aurimasv: Like I said, `"https?://([^/]+\\.)?(artstor|sscommons)\\.org\/(open)?library"` is what you want. If that doesn't work, please post a URL that you're trying to test against.

ARTstor.js
```diff
@@ -1,131 +1,899 @@
...
+    '{': '7B',
+    '|': '7C',
+    '}': '7D',
+    '~': '7E'
+};
+
+/**
+    Converts two character number character to special character
+**/
+function decrypt(s) {
```
aurimasv: Can you explain to me what kind of data you're trying to "decrypt" here? Could you show a possible input and and output?

aurimasv: Ad how it's different from `encodeURIComponent`/`decodeURIComponent`

ARTstor.js
```diff
@@ -1,131 +1,899 @@
...
-	but WITHOUT ANY WARRANTY; without even the implied warranty of
-	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-	GNU Affero General Public License for more details.
-	
-	You should have received a copy of the GNU Affero General Public License
-	along with Zotero.  If not, see <http://www.gnu.org/licenses/>.
-	
-	***** END LICENSE BLOCK *****
+    This translator works for Artstor library sites (http://library.artstor.org) and
+    Artstor Shared Shelf Commons (http://www.sscommons.org)
```
aurimasv: Please move this comment below license block into a separate comment

ARTstor.js
```diff
@@ -1,131 +1,899 @@
...
+    The return value of this function should be the detected item type (e.g. “journalArticle”, 
+    see the overview of Zotero item types), or, if multiple items are found, “multiple”. 
+**/
+function detectWeb(doc, url) {
+    if (url.match(/\/iv2/)) {
+        // Image viewer window
+        return "artwork";
+    } else if (url.match(/\#3\|/)) {
+        // Thumbnail window page
+        if ((doc.getElementsByClassName('MetaDataWidgetRoot') != null) && (doc.getElementsByClassName('MetaDataWidgetRoot').length > 0)) {
```
aurimasv: `.getElementsByClassName` always returns an array (well, it's not exactly an array, but it's never `null`), so this test will always pass

ARTstor.js
```diff
@@ -1,131 +1,899 @@
...
+            // Don't capture date if small window is present
+            return false;
+        } else if ((doc.getElementById("thumbNavSave1") != null) && (doc.getElementById("thumbNavSave1").style.display == "block")) {
+            // Don't capture data if image group window is in editing state.
+            return false;
+        } else if ((doc.getElementById("ssContentWrap") != null) && (doc.getElementById("ssContentWrap").style.display == "inline")) {
+            // Don't capture data if slide show window is present
+            return false;
+        } else {
+            if (url.match(/zMode/)) {
```
aurimasv: nit: `} else if (...) {`

ARTstor.js
```diff
@@ -1,131 +1,899 @@
...
+                skipThumbnail = true;
+            }
+            if ((doc.getElementById("thumbNavSave1") != null) && (doc.getElementById("thumbNavSave1").style.display == "block")) {
+                // Don't capture data if image group window is in editing state.
+                skipThumbnail = true;
+            }
+            if ((doc.getElementById("ssContentWrap") != null) && (doc.getElementById("ssContentWrap").style.display == "inline")) {
+                // Don't capture data if slide show window is present
+                skipThumbnail = true;
+            }
```
aurimasv: You're already performing these checks in `detectWeb`. `doWeb` is never called unless `detectWeb` passes.

ARTstor.js
```diff
@@ -1,131 +1,899 @@
...
-	if (!value) return;
-	if (typeof value == 'string') value = ZU.trimInternal(value);
-	if (!value) return;
-	
-	newItem[field] = value;
+function doImageViewer(doc, url) {
+    // get the image id and object type from the page
+    // this contains the objId and object type separate by : as in "AWSS35953_35953_25701160:11"
+    var objID = doc.getElementById("objID");
+    if (objID != null) {
```
aurimasv: You should guarantee that this is present in `detectWeb`. It's better to let `doWeb` throw if it's not able to import something, since we then get a proper error report.

ARTstor.js
```diff
@@ -1,131 +1,899 @@
...
-	
-	newItem.complete();
-}
\ No newline at end of file
+function doMetadataWindow(doc, url) {
+    // get object id from metadata window.
+    var metaWindows = doc.getElementsByClassName('MetaDataWidgetRoot');
+    var objItems = [];
+    for (var i = 0; i < metaWindows.length; i++) {
+        var id = metaWindows[i].id.substring(3);
```
aurimasv: is this supposed to be some sort of number? It would help to explain what's going on here.

ARTstor.js
```diff
@@ -1,131 +1,899 @@
...
+
+function getPortal(url) {
+    var portal = url.substring(7, url.indexOf("."));
+    return portal;
+}
+
+function getServerUrl(url) {
+    var serverUrl;
+    if (url.indexOf('/iv2') > 0) {
+        serverUrl = url.substring(0, url.indexOf('iv2\.'));
```
aurimasv: Shouldn't the `if` test also include the period?

ARTstor.js
```diff
@@ -1,131 +1,899 @@
...
+    var pageNoDOM = doc.getElementById("pageNo");
+    if (pageNoDOM != null) {
+        pageNo = parseInt(pageNoDOM.value);
+    }
+
+    // get page size
+    var imagesPerPage = "1";
+    var pageDOM = doc.getElementById("thumbNavImageButt");
+    if (pageDOM != null)
+        imagesPerPage = pageDOM.innerHTML;
```
aurimasv: nit: use braces

ARTstor.js
```diff
@@ -1,131 +1,899 @@
...
+        var id = metaWindows[i].id.substring(3);
+        objItems.push(id);
+    }
+
+    processSelectedObject(doc, url, objItems, 1);
+}
+
+function processSelectedObject(doc, url, selectedObjs, selectionType) {
+    var serviceURL = getThumbnailServiceURL(doc, url);
+    Zotero.Utilities.HTTP.doGet(serviceURL, function(text) {
```
aurimasv: Is the idea here to replicate the search query from the webpage and get the resulting JSON then assume that the results are displayed identically and ask the user to select from that list?

I'm not a huge fan of this approach, because it tends to be quite unstable, hard to keep up to date, and difficult to detect when it breaks. It's annoying and confusing to the user when this does break. You should be detecting the search results that are actually being displayed on the page and then presenting the user with those choices.

ARTstor.js
```diff
@@ -1,131 +1,899 @@
...
+        for (var i = 0; i < json.thumbnails.length; i++) {
+            var thumbnail = json.thumbnails[i];
+            objDescItem = new Object();
+            objDescItem.id = thumbnail.objectId;
+            objDescItem.type = thumbnail.objectTypeId;
+            objDescItem.title = thumbnail.thumbnail1;
+            objDescItems.push(objDescItem);
+        }
+    });
+    return objDescItems;
```
aurimasv: That's an asynchronous call to `doGet` above, so `objDescItems` is always `undefined` when you return it from this function.

aurimasv: Well, not undefined, since you pass it in. But if you're expecting `objDescItems` to have changed after the function returns, then that's not the case. In general, do not do this. Use callbacks.

ARTstor.js
```diff
@@ -1,131 +1,899 @@
...
+        var json = JSON.parse(text);
+        var portal = getPortal(url);
+        if (!(portal in portalMap)) {
+            portal = 'default';
+        }
+        processPortalData(dataItem, json, portalMap[portal], portal);
+        getNotesDataItem(url, objItem, dataItem);
+    });
+
+    return dataItem;
```
aurimasv: same thing about asynchronous calls.

ARTstor.js
```diff
@@ -1,131 +1,899 @@
...
+    }
+    if (dataItem.title == undefined) {
+        dataItem.title = "Unknown";
+    }
+}
+
+function setItemCreator(dataItem, fieldValue) {
+    fieldValue = fieldValue.replace(/<wbr\/>/g, "");
+    var names = [];
+    if (fieldValue.indexOf(';')) {
```
aurimasv: do you mean `.indexOf(...) != -1`?

ARTstor.js
```diff
@@ -1,131 +1,899 @@
...
+        var str = names[i];
+        var contributor = "author";
+        var name = str;
+
+        if (str.indexOf(':') > 0) {
+            var params = str.split(':');
+            contributor = params[0];
+            name = params[1];
+        }
+        dataItem.creators.push(ZU.cleanAuthor(name.replace(/<\/?[^>]+(>|$)/g, " ").replace(/(&gt;)|(&lt;)/g, ""), contributor, false));
```
aurimasv: nit: please split this into multiple lines

ARTstor.js
```diff
@@ -1,131 +1,899 @@
...
+    }
+}
+
+function setItemLabelValue(dataItem, key, label, value) {
+    var cleanValue = value.replace(/<\/?[^>]+(>|$)/g, " ");
+    cleanValue = cleanValue.replace(/\./, "");
```
aurimasv: no `g` flag?

ARTstor.js
```diff
@@ -1,131 +1,899 @@
...
+                url: mediaUrl,
+                mimeType: mediaInfo[2]
+            });
+            dataItem.url = url;
+            dataItem.complete();
+        });
+    } else {
+        dataItem.complete();
+    }
+    Zotero.done();
```
aurimasv: don't need this

ARTstor.js
```diff
@@ -1,131 +1,899 @@
...
+}
+
+function getPortal(url) {
+    var portal = url.substring(7, url.indexOf("."));
```
aurimasv: what if it's an https url?

ARTstor.js
```diff
@@ -1,131 +1,899 @@
...
+        "defer": true,
+        "items": [
+            {
+                "itemType": "artwork",
+                "title": "Trailer Home; Exterior view",
+                "creators": [
+                    {
+                        "firstName": "Barbara",
+                        "lastName": "Lane",
+                        "creatorType": "Image by"
```
aurimasv: that's not going to work

ARTstor.js
```diff
@@ -1,131 +1,899 @@
...
+                    }
+                ],
+                "date": "Photographed: 2001",
```
aurimasv: Should just be "2001"

ARTstor.js
```diff
@@ -1,131 +1,899 @@
...
+                "extra": "Location: Bradford County, Pennsylvania; Collection: Bryn Mawr College Faculty/Staff/Student Photographs; ID Number: 01-07828; Source: Personal photographs of Professor Barbara Lane, 2001",
+                "libraryCatalog": "ARTstor",
+                "rights": "Copyright is owned by the photographer Questions can be directed to sscommons@brynmawr.edu.; This image has been  selected and made av ailable by a user us ing Artstor's softwa re tools Artstor ha s not screened or se lected this image or  cleared any rights  to it and is acting  as an online service  provider pursuant t o 17 U.S.C. §512. Ar tstor disclaims any  liability associated  with the use of thi s image. Should you  have any legal objec tion to the use of t his image, please vi sit http://www.artst or.org/our-organizat ion/o-html/copyright .shtml for contact i nformation and instr uctions on how to pr oceed.",
```
aurimasv: Take a look at the end of this string. Something doesn't look right.

ARTstor.js
```diff
@@ -1,131 +1,899 @@
...
+                "url": "http://www.sscommons.org/openlibrarysecure/ViewImages?id=4jEkdDElLjUzRkY6fz5%2BRXlDOHkje1x9fg%3D%3D&userId=gDFB&zoomparams=&fs=true",
```
aurimasv: 404 Not Found. These should be persistent URLs

ARTstor.js
aurimasv: should be "author"

ARTstor.js
aurimasv: Might be helpful to take a look at https://github.com/zotero/translators/blob/master/Summon%202.js for a similar kind of translator

ARTstor.js
venturi2k: This is reformatted by beautifier. I had it originally as } else if (...).

ARTstor.js
venturi2k: OK, I guess additional check in doWeb is unnecessary.

ARTstor.js
venturi2k: The image id is extracted from the dom id which is prefixed with additional string to make them unique.

ARTstor.js
venturi2k: This is how it can get the service parameter from the AJAX page URL which has additional encoding that is different from the normal encoding from service parameters.

ARTstor.js
venturi2k: In fact, the returned object is not used due to the reason you described. The return parameter is there just for completion.

ARTstor.js
venturi2k: I have already explained it in the comment. It is different in that the encoding does not have % in front of it.

ARTstor.js
venturi2k: the object may have creator of different roles, which is why it is not author, such as photographer, who is not the author of the painting.

venturi2k:
Added support for external viewer capture
Image viewer link incorrect. 
rights string is not correct due to incorrect <wbr/> processing.
Fix get portal url: fixed getPortal string to include https support
Remove Zotero.done()
DecodeSearchData: - uses g flag for replacement
SetItemCreator: check for index value > 0
Don’t return for anychronous call
Explain how it gets image id in metadata window.
Remove duplicate detect code from doWeb since detectWeb has already done that
getElementsByClassName check for array length instead of null.
document decrypt
Replaced encrypt with encodeURIComponent
Fixed bug with trimming <wbr/>

ARTstor.js
venturi2k: The date is mapped to the field which has value that may not always be a year. E.g. It could have multiple year value as in "created: 1879; photographed: 2014", or "circa 1st century BC".

ARTstor.js
aurimasv: Actually, you [originally](https://github.com/venturi2k/translators/commit/cbe856c399209756434408d8fc496e95d4be615c#diff-2bbd96777b59793a19b2dfc8752323ecR58) had it as
```
} else
if (....) {
```
which JSBeautifier (correctly) identified as
```
} else {
  if (...) {
```
Technically they accomplish the same thing, but we'd like to stick to more typical formatting and not have to worry about some unanticipated behavior.

ARTstor.js
aurimasv: Be that as it may, Zotero can only recognize a select set of creator roles and will discard everything else. For artwork, your choices are listed [here](http://aurimasv.github.io/z2csl/typeMap.xml#map-artwork). You can use "artist" (or "author", which is the base type)

ARTstor.js
aurimasv: It's not critical and we would take this as is, but if you're able to clean up the data to be the date of creation without any additional labels, that would be ideal. Zotero is designed with the idea that Date field contains the date of creation. To obtain correct formatting in citations, the user will have to clean up the data manually otherwise.

ARTstor.js
```diff
@@ -1,131 +1,817 @@
...
+    '7B': '{',
+    '7C': '|',
+    '7D': '}',
+    '7E': '~'
+};
+
+/**
+    Converts two character number character to special character
+    it converts 
+        type3D3626kw3Dairstream20trailer26geoIds3D
```
aurimasv: there's something in your code about "!", which isn't explained here but looks like some sort of odd Unicode escapes in the format "!1234!" as decimal code point(?). I'm still not a fan of keeping mapping tables for something like hex character encodings. Here's a version of the decrypt function that should make this more maintainable (based on logic in your decrypt... I think)
```
function decrypt(s) {
  // Why not allow all hex values? Is the content ambiguously escaped? e.g. "C0"
  // If these kind of escapes are possible then everything falls apart anyway,
  // since something like "C325" would be escaped as "C25"
  return s.replace(/!(\d{1,5})!|(\d[\dA-F])/g, function(m, unicode, hex) {
    // Either unicode or hex are set, not both. unicode has priority over hex in the match
    if (unicode) return String.fromCharCode(parseInt(unicode)); // Always parses because of regexp

    // must be hex
    try {  return decodeURIComponent('%' + hex) }
    catch(e) { /* Some hex character escapes are invalid */ }

    return m; // Fail-safe
  });
}
```
This way you don't have to re-invent the wheel and worry about an incomplete mapping table.

ARTstor.js
```diff
@@ -1,131 +1,817 @@
...
+             "id": "all"
+             "bDate": "-2019"
+             "eDate": "2030"
+             "dExact": "1"
+             "prGeoId": ""
+             "origKW": "stimson hall|all#and,1925|all"
+         }
+**/
+function decodeSearchData(str) {
+    var searchParam = str.replace(/(&gt;)/g, ":");
```
aurimasv: are you intentionally replacing this with incorrect decoding? Is that the only HTML entity that can be present? Maybe it's safer to use `ZU.unescapeHTML`, which unescapes all HTML entities.

ARTstor.js
aurimasv: I'm not sure we're talking about the same thing here.

venturi2k: The DOM in the webpage does not have all the information need for prompting the list to the user (it needs the object id, and object type that is associated with the title, which is not in the DOM). This is why it need to use the service call to get the data in javascript to show it. The issues with AJAX application is that all data are stored in javascript, not in the DOM. This is why we have to have the service call to get the correct data.

ARTstor.js
venturi2k: I see your point. I was trying to expose what is already there, but probably it does not make any difference in the end.

ARTstor.js
venturi2k: I like that. I'll try it out.

venturi2k: I can't seem to find a way to make the test case work in Zotero test framework. Should I scrape it?

venturi2k:
Few more bug fixes with HTML entity handling and removed inaccessible resource URL. Please merge it with the main master branch!

venturi2k:
Please review the changes and merge the code. Thanks

ARTstor.js
```diff
@@ -1,131 +1,713 @@
...
-		
-		found = true;
-		data[field] = value;
-	}
-	
-	return found ? data : false;
+    "translatorID": "5278b20c-7c2c-4599-a785-12198ea648bf",
+    "label": "ARTstor",
+    "creator": "John Justin, Charles Zeng",
+    "target": "\.artstor|\.sscommons\.org:?\w*\/(open)*library",
```
adam3smith: I'll let @aurimasv do the full review, but something went wrong here: you need double backslashes (```\\.```) for escaping. Scaffold does this automatically for you. 
But also, that regex isn't equivalent to what you had above given how the or marker works. It would e.g. match ```www.artstore.com``` (and any url containing ".artstor"). I'm not sure why you didn't just go with the suggestion by aurimas above? Also don't understand using * instead of ? etc.

venturi2k: \.artstor|\.sscommons\.org:?\w*\/(open)*library  is what I tested in Scaffold. Using (\\.) does not work in Scaffold. 
I will change the * to ?.
I am not clear on your comment on markers. I have explained above on what the target should cover.

adam3smith: You never replied to aurimas's comment&suggested target above: https://github.com/zotero/translators/pull/899#discussion_r33820499 which I do think would be the best way forward.

If ```.artstor|.sscommons.org:?\w\/(open)library``` passes and ```\.artstor|\.sscommons.org:?\w\/(open)library``` fails in Scaffold, that means your regex is broken: the first one uses the dot as a metacharacter for any character, the second one escapes it to a literal period. ```\.``` definitely works in Scaffold, you can look at any of the other 300+ translators.

What I was referring to with the "or" marker (|) is that you place it, without any parentheses, between \.artstor and the rest of the regex, which means the regex will evaluate to true with any string containing ".artstor"

But again, instead of fixing&discussing these piecemeal, why don't you just tell us for which URLs ```"https?://([^/]+\\.)?(artstor|sscommons)\\.org\/(open)?library"``` doesn't work and we'll take it from there.

ARTstor.js
venturi2k: The URLs are:
www.sscommons.org/openlibrary
library.artstor.org/library
sahara.artstor.org/library
openlibrary.artstor.org/openlibrary

adam3smith: those all match using the above regex, so I'd just use that.

venturi2k: Yes, the above pattern works fine. 

I put in the pattern (https?://([^/]+\.)?(artstor|sscommons)\.org\/(open)?library) without double slash (\\.) in scaffold, and it works. 

Does that mean I should put in extra (\) in the code?

ARTstor.js
venturi2k: OK.

ARTstor.js
adam3smith: great.

> Does that mean I should put in extra () in the code?

sorry, I don't follow -- extra parentheses where? (or did github eat part of your post?)

venturi2k: Yes it did eat the backslash in the paren.

What I to escape the character dot  (.), should I put two double back slash, or single one in the code? Single back slash works in Scaffold.

ARTstor.js
venturi2k: OK, I will fix that.

ARTstor.js
adam3smith: yes, double backslash in the code (single one in scaffold -- if you save from scaffold directly it automatically adds the second backslash)

venturi2k: Great. I have updated the target string.

ARTstor.js
venturi2k: OK, the target string is updated.

ARTstor.js
adam3smith: thanks! @aurimasv will have another look at the style to make sure it's ready for merging.

venturi2k: Thanks great. Thanks,

venturi2k:
Can someone with write access merge the pull request?

ARTstor.js
venturi2k: If any of you can help me merge it, it will be great.

adam3smith:
sorry, why are you closing this?

venturi2k:
I am going to re-send a pull request.

aurimasv:
Sorry for taking so long to review. I'm really swamped with other work and this is a really large translator. Honestly, I think that it is unnecessarily large and I would really like to encourage you to simplify it. I would offer more concrete guidance on how to do it, but I do not currently have the time to investigate this. Maybe someone else can provide guidance.

venturi2k:
I do try to  simplify that based  what the site provides. If the site keeps the ids of the item in the dom, it will be really helpful. But the data are are stored in site javascript which can not be read by Zotero translator code. I'd sure like to make it simpler.

venturi2k:
Also the app does not keep the associate metadata in the DOM, so it requires separate service call to get the necessary data.

aurimasv:
OK, I looked more closely at the issue.

For others interested in this, the webpage functions by attaching event handlers to various elements on the page that maintain internal state regarding object IDs and whatnot. There is a global registry of element IDs <-> object IDs that is available as a global variable (`window.dijit.registry._hash`). We need to be able to get to the object ID for a given element in order to fetch appropriate metadata, but that is not possible in Firefox due to sandboxing (please do correct me if I'm wrong here). Should be possible in Chrome/Opera/Safari, but I haven't tried.

In the absence of the global object map, we can actually fetch this map from the remote server as JSON, but that requires that we correctly reconstruct the query that was originally sent to the server and, in my experience, that is a pretty fragile step (i.e. it becomes quite difficult to account for all possible search/facet/filter/sorting combinations and make sure that everything is in correct order. You end up re-constructing a large chunk of the website and that is rarely, if ever, accurate). Nonetheless, that is what the code in this PR attempts to do (from my not-in-depth analysis) and it may be the best way forward for Firefox (I'll review the code when I get the chance, but if anyone wants to jump on it, please feel free). I still think that for Chrome/Opera/Safari (and bookmarklet) we could take advantage of access to the global object and avoid fetching the JSON from server.

I also contacted Artstor to see if they would be willing to enhance their website to facilitate more robust metadata scraping by Zotero (i.e. embed objectIDs in the DOM). We'll see what they say.

I do wonder if at some point we could allow translators in Firefox access to the full contents of a given window (perhaps read-only, though I suspect that's asking for too much. Maybe via some ZU method, e.g. `ZU.requestWindowProperty(['dijit', 'registry', '_hash'])`), since that would be quite advantageous.

CC @simonster @dstillman

venturi2k:
I checked with Artstor on the suggestion. This could be done with added attribute/tag in DOM so it can save a service call for rebuilding the data. However this may take a while as this requires a new general release for all platforms. Can we proceed with this translator for now?

aurimasv:
Like I said, I'll review it as soon as I have some time, though someone else is welcome to jump in.

venturi2k:
I am trying to experiment using global javascript object in Firefox, but I don't have any success.

aurimasv:
I got in touch with Artstor. As could have been speculated, this is their official translator. @venturi2k, it does make a slight difference if we know this information (e.g. we have more trust in your ability to replicate search queries), so it may be helpful to mention this next time.

Also, Mary Finer from Artstor said she would forward the request to add objectIds to the DOM to the Artstor developers and let me know if that would be feasible (I asked her to do this on Sep. 28). Not sure if that basically means that the request will end up being submitted to @venturi2k. I'll wait a few more days for them to get back to me before I review this PR.

aurimasv:
>I am trying to experiment using global javascript object in Firefox, but I don't have any success.

Yes, that should not work. If it does, it's a bug in Firefox's sandboxing code and we shouldn't rely on it.

venturi2k:
With the new release of Artstor, this updated translator has removed the code for context detection REST calls to get the items selected. The new release of Artstor application added objectID in the DOM.  Please complete merge the code.

ARTstor.js
```diff
@@ -1,131 +1,484 @@
 {
-	"translatorID": "5278b20c-7c2c-4599-a785-12198ea648bf",
+	"translatorID": "7a690951-ed76-43f5-b761-707fef0345d4",
 	"label": "ARTstor",
-	"creator": "Sebastian Karcher",
-	"target": "^https?://library\\.artstor\\.org",
-	"minVersion": "3.1",
+	"creator": "John Justin, Charles Zeng",
+	"target": "\\.artstor|\\.sscommons\\.org:?\\w*\\/(open)*library",
```
adam3smith: I still don't understand why the regex that @aurimasv originally suggested, i.e.
 "^https?://([^/]+\\.)?(artstor|sscommons)\\.org\/(open)?library" doesn't work. That's signficantly more restrictive and more efficient than what you have (which, again, in its current form matches every URL that contains ".artstor" anwhere in the URL).

ARTstor.js
```diff
@@ -1,131 +1,484 @@
 {
-	"translatorID": "5278b20c-7c2c-4599-a785-12198ea648bf",
+	"translatorID": "7a690951-ed76-43f5-b761-707fef0345d4",
```
adam3smith: no, please don't change the ID. That would leave the old translator in place and we'd have to delete that in a separate operation and there's no reason for that.

ARTstor.js
```diff
@@ -1,131 +1,484 @@
...
 function detectWeb(doc, url) {
-	//monitor changes to body's direct children. That's where the metadata popup is added
-	Zotero.monitorDOMChanges(doc.body, {childList: true});
+    var itemType = false; // default - ignore
+    if (url.match(/\/iv2\.|ExternalIV.jsp/)) {
+        // Image viewer window
+        itemType = "artwork";
+    } else if (url.match(/\#3\|/)) {
+        // Thumbnail window page
+        if ((doc.getElementsByClassName('MetaDataWidgetRoot') != null) &&
```
adam3smith: you can save yourself the first test. I'm pretty sure that function never returns null. It just returns an empty array where the class doesn't exists. Applies to several instances of this across the translator.

ARTstor.js
```diff
@@ -1,131 +1,484 @@
...
+/**
+    detectWeb is run to determine whether item metadata can indeed be retrieved from the webpage. 
+    The return value of this function should be the detected item type (e.g. “journalArticle”, 
+    see the overview of Zotero item types), or, if multiple items are found, “multiple”. 
+**/
 function detectWeb(doc, url) {
-	//monitor changes to body's direct children. That's where the metadata popup is added
-	Zotero.monitorDOMChanges(doc.body, {childList: true});
+    var itemType = false; // default - ignore
+    if (url.match(/\/iv2\.|ExternalIV.jsp/)) {
```
adam3smith: please use `.search(/regex/) != -1` to simply test for a regular expression. More efficient than match. Also escape the period before in `.jsp`.

ARTstor.js
```diff
@@ -1,131 +1,484 @@
...
+        // Image viewer window
+        itemType = "artwork";
+    } else if (url.match(/\#3\|/)) {
+        // Thumbnail window page
+        if ((doc.getElementsByClassName('MetaDataWidgetRoot') != null) &&
+            (doc.getElementsByClassName('MetaDataWidgetRoot').length > 0)) {
+            // There are multiple metadata windows visible
+            itemType = "artwork";
+        } else if ((doc.getElementById("floatingPlaceHolder") != null) &&
```
adam3smith: (just to be clear: here both tests are needed)

ARTstor.js
```diff
@@ -1,131 +1,484 @@
...
+        objItems.push(id + ":10"); // default type to image
+    }
+    processObjects(doc, url, objItems);
+}
+
+/**
+	This functions removes extra format tag from string and also decodes the
+	html entity string.
+**/
+function htmlDecode(doc, input) {
```
adam3smith: Zotero has `ZU.unescapeHTML` https://github.com/zotero/zotero/blob/master/chrome/content/zotero/xpcom/utilities.js#L486 -- that should do most of this (except for the first bit). Don't duplicate.

ARTstor.js
```diff
@@ -1,131 +1,484 @@
...
+}
+
+function getPortal(url) {
+    var portal = url.substring(url.indexOf('://') + 3, url.indexOf('.'));
+    return portal;
+}
+
+function getServerUrl(url) {
+    var serverUrl;
+    if (url.indexOf('/iv2\.') > 0) {
```
adam3smith: is this actually what you mean? I'm guessing should be `/iv2.` or is there actually a backslash in the URL?

ARTstor.js
```diff
@@ -1,131 +1,484 @@
...
+        serverUrl = url.substring(0, url.indexOf('iv2\.'));
```
adam3smith: same here

ARTstor.js
```diff
@@ -1,131 +1,484 @@
...
+    {
+        "type": "artwork",
+        "url": "http://www.sscommons.org/openlibrary/ExternalIV.jsp?objectId=4jEkdDElLjUzRkY6fz5%2BRXlDOHkje1x9fg%3D%3D&fs=true",
+        "items": [
+            {
+                "itemType": "artwork",
+                "title": "Trailer Home; Exterior view",
+                "creators": [
+                    {
+                        "firstName": "Image by: Barbara",
```
adam3smith: could we remove the "Image by:" and similar labels from creators?

adam3smith:
See some smaller issues inline, should be quick fixes. I didn't go line-by-line, but there's nothing that strikes me as particularly problematic, so unless @aurimasv or @zuphilip want to take a closer look, I'll merge this once the remaining fixes are implemented.

venturi2k:
Hello,



I have rewrote the translator for Artstor due to new release of the application.  I sent a pull request, and I have not hear anything on this.  I would like to know if there is anything I need to do to get the translator merged so users can take advantage of that.



Any direction from you would really help. Probably you guys are overwhelmed with all the merge requests.



Thanks a lot.



Charles

________________________________

adam3smith:
I left comments on your code more than a week ago, see above.

ARTstor.js
```diff
@@ -1,131 +1,484 @@
...
+        'archaeology': {
+            'Site Name': 'title',
+            'Artifact Title': 'title',
+            'Artifact Description': 'abstractNote',
+            'Artifact Repository': 'archive',
+            'Rights': 'rights',
+            'Site Date': 'date',
+            'Artifact Materials/Techniques': 'artworkMedium',
+            'Artifact Dimensions': 'artworkSize',
+            'Rights': 'rights'
```
zuphilip: Duplicate key `Rights`

ARTstor.js
```diff
@@ -1,131 +1,484 @@
...
+    }
+    if (dataItem.title == undefined) {
+        dataItem.title = "Unknown";
+    }
+}
+
+function setItemCreator(dataItem, fieldValue) {
+    var names = [];
+    if (fieldValue.indexOf(';') > 0) {
+        names = fieldValue.split(';')
```
zuphilip: Missing semicolon

ARTstor.js
```diff
@@ -1,131 +1,484 @@
...
+        var str = names[i];
+        var contributor = "author";
+        var name = str;
+        var value = name.replace(/<\/?[^>]+(>|$)/g, " ").replace(/(&gt;)|(&lt;)/g, "");
+        dataItem.creators.push(ZU.cleanAuthor(value, contributor, false));
+    }
+}
+
+function cleanStringValue(str) {
+    var cleanValue = str.replace(/\<wbr\/>/g, "");
```
zuphilip: That looks fishy, because you escaped `<` but not `>`. I think you don't have to escape any of these.

zuphilip:
@venturi2k Please make sure that you also the AGPL license and look at the other issues here mentioned on github / CC @adam3smith

ARTstor.js
venturi2k: Sure, I removed anything that is before ":".

ARTstor.js
venturi2k: Good catch!

venturi2k:
Minor fixes per review.


	* Keet original ID string
	* Added license block
	* changed target string
	* used search function instead of match
	* used ZU.unescapeHTML
	* removes \ in /iv2\.
	* Removed duplicated "Rights" field.

venturi2k:
@zuphilip All issues mentioned above are resolved in the code. /CC @adam3smith .

adam3smith:
Looking good by me. Merging on Sunday unless there are any objections

adam3smith:
Thanks!

PR #900, opened 2015-06-05
[ACS] Fix to work with proxies

aurimasv:
Re https://forums.zotero.org/discussion/49735/issues-importing-from-pubsacsorg-using-university-home-proxy/

Haven't tested yet, since I currently don't have proxy access. Will test shortly


aurimasv:
OK, seems to be working fine.

PR #901, opened 2015-06-07
restrict detect on Open WorldCat to not detect on WorldCat Discover p…

adam3smith:
…ages. I looked around, this seems very stable & is super efficient, I don't think we can handle this via the target.

This one is pretty high priority since WorldCat Discovery is currently broken, so I'll merge this later tonight unless there are objections.

### DIFF
--- a/Open WorldCat.js
+++ b/Open WorldCat.js
@@ -2,14 +2,14 @@
 	"translatorID": "c73a4a8c-3ef1-4ec8-8229-7531ee384cc4",
 	"label": "Open WorldCat",
 	"creator": "Simon Kornblith, Sebastian Karcher",
-	"target": "^https?://[^/]+\\.worldcat\\.org",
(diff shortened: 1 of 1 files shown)


Open WorldCat.js
```diff
@@ -67,7 +67,9 @@ function getFirstContextObj(doc) {
 }
 
 function detectWeb(doc, url) {
-	var results = getSearchResults(doc);
+	//distinguish from Worldcat Discovery
+	if (ZU.xpath(doc, '/html/body[@id="worldcat"]').length >0){
```
aurimasv: Nit #1: I think you should be able to just do `doc.body.id == "worldcat"`

Not #2: indents (unless it just doesn't show up here in the diff)

Other than that, fine by me

adam3smith:
OK, thanks, squashed and merged with those corrections. Since most of the comments @aurimasv had for the Discovery translator were actually stuff I re-used from this one, I'll submit those fixes along side.

PR #902, opened 2015-06-07
add new WorldCat Discovery translator. Oddly, the RIS data appears fl…

adam3smith:
…awed in different ways, so I built in a number of additional clean-ups and removed some from OpenWorldcat. For the same reason, it almost certainly makes sense to keep the two separate.

This one is less urgent, so I can wait for review.

### DIFF
--- a/WorldCat Discovery Service.js
+++ b/WorldCat Discovery Service.js
@@ -1,0 +1,282 @@
+{
+	"translatorID": "fd8dc5f6-a6dd-42b2-948f-600f5da844ea",
+	"label": "WorldCat Discovery Service",
+	"creator": "Sebastian Karcher",
+	"target": "^https?://[^/]+\\.worldcat\\.org/",
+	"minVersion": "3.0.9",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsbv",
+	"lastUpdated": "2015-06-14 23:43:23"
+}
+
+function detectWeb(doc, url) {
+
+	var results = getSearchResults(doc);
+	if (results.length) {
+		return "multiple";
+	}
+
+
+	//single result
+	// generate item and return type
+	var co = getFirstContextObj(doc);
+	if (!co) return false;
+
+	return generateItem(doc, co).itemType;
+}
+
+
+/**
+ * Gets Zotero item from a WorldCat icon src
+ */
+function getZoteroType(iconSrc) {
+	// only specify types not specified in COinS
+	if (iconSrc.indexOf("icon-rec") != -1) {
+		return "audioRecording";
+	}
+	if (iconSrc.indexOf("icon-com") != -1) {
+		return "computerProgram";
+	}
+	if (iconSrc.indexOf("icon-map") != -1) {
+		return "map";
+	}
+	return false;
+}
+
+/**
+ * Generates a Zotero item from a single item WorldCat page,
+ * or the first item on a multiple item page
+ */
+function generateItem(doc, co) {
+	var item = new Zotero.Item();
+	ZU.parseContextObject(co, item);
+	// item types not covered by COinS will still need to be covered. Leaving the corresponding code from open Worldcat
+	// here as a reminder.
+
+	/*
+		var type = ZU.xpathText(doc,
+			'//img[@class="icn"][contains(@src, "icon-")][1]/@src');
+		if (type) {
+			type = getZoteroType(type);
+			if (type) item.itemType = type;
+		} */
+
+	return item;
+}
+
+function getSearchResults(doc) {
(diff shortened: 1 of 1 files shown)


WorldCat Discovery Service.js
```diff
@@ -0,0 +1,287 @@
+{
+	"translatorID": "fd8dc5f6-a6dd-42b2-948f-600f5da844ea",
+	"label": "WorldCat Discovery Service",
+	"creator": "Sebastian Karcher",
+	"target": "^https?://[^/]+\\.worldcat\\.org",
```
aurimasv: Terminal / for proxy detection

WorldCat Discovery Service.js
```diff
@@ -0,0 +1,287 @@
...
+	if(!id) return false;
+	return id[1];
+}
+
+
+/**
+ * Given an item URL, extract database ID
+ */
+function extractDatabaseID(url) {
+	var id = url.match(/databaseList=[^&#]+/);
```
aurimasv: I'm pretty sure you want to capture the ID here (in parentheses) and return just that part below, instead of the full match

aurimasv: Also, just for completeness and to make sure we don't mess up later, `decodeURIComponent` here then encode it back when building url

aurimasv: Eh, actually, doesn't matter. Both IDs come from url. Hopefully that doesn't change later.

WorldCat Discovery Service.js
```diff
@@ -0,0 +1,287 @@
...
+ * RIS Scraper Function
+ *
+ */
+
+function scrape(risURL) {
+	
+	ZU.doGet(risURL, function (text) {
+		//Z.debug(text);
+		//conference proceedings exported as CONF, but fields match BOOK better
+		text = text.replace(/TY\s+-\s+CONF\s+[\s\S]+\n\s*ER\s+-/g, function(m) {
```
aurimasv: Nit, use lazy + `[\s\S]+?` I know that we probably will never have multiple RIS entries per file, but we're using `g` flag here anyway, so might as well

WorldCat Discovery Service.js
```diff
@@ -0,0 +1,287 @@
...
+			//clean up ISBNs
+			if (item.ISBN){
+				var ISBN = item.ISBN.split(/\s/);
+				var ISBNarray = [];
+				for (var i=0; i<ISBN.length; i++){
+					if (ZU.cleanISBN(ISBN[i])){
+						ISBNarray.push(ZU.cleanISBN(ISBN[i]));
+					} 
+				}
+				item.ISBN = ISBNarray.join("\n");
```
aurimasv: Cleaning up should be unnecessary as of next version, but if we're doing it, let's join using space, not newline (though it's probably getting converted anyway)

WorldCat Discovery Service.js
```diff
@@ -0,0 +1,287 @@
...
+				item.place = item.place.replace(/:\s*$/, "");
+			}
+
+
+			//remove trailing commar after publisher
+			if (item.publisher){
+				item.publisher = item.publisher.replace(/,\s*$/, "");
+			}
+			//correct field mode for corporate authors
+			for (i in item.creators) {
```
aurimasv: Nit: `for (var i=0; ....) {` for arrays

WorldCat Discovery Service.js
aurimasv: I guess that's not true for multiples, so I stand by my suggestion above

WorldCat Discovery Service.js
```diff
@@ -0,0 +1,287 @@
...
+						"creatorType": "author"
+					}
+				],
+				"date": "2012",
+				"ISSN": "0360-3032",
+				"issue": "1",
+				"libraryCatalog": "WorldCat Discovery Service",
+				"pages": "129-132",
+				"publicationTitle": "Trinity journal.",
+				"volume": "33",
```
aurimasv: Nit: we could trim off the period

adam3smith:
@aurimasv -- done. Let me know if that looks good to you.

WorldCat Discovery Service.js
```diff
@@ -0,0 +1,291 @@
...
+			Z.debug(risURL)
+			items[risURL] = title.trim();
+
+		}
+
+		Zotero.selectItems(items, function(items) {
+			if (!items) return true;
+
+			for (var i in items) {
+				scrape(i);
```
aurimasv: I'd rather we collect these in an array and call `scrape` with that. The HTTP requests would then be sequential rather than concurrent, which is not very nice towards WorldCat.

aurimasv:
Other than that one nit, this looks fine. Feel free to merge.

adam3smith:
thanks @aurimasv

apcshields:
Thanks @adam3smith and @aurimasv, for making this happen!

PR #903, opened 2015-06-08
Various fixes for National Archives of Australia translator to deal with new file image viewer

wragge:
The National Archives of Australia made some major changes to the way file images are viewed. This broke the translator's ability to save a single digitised page, and also the ability to save an individual file's details as it would look at the file viewer to get the number of digitised pages. 

I think it should all be working now.

### DIFF
--- a/National Archives of Australia.js
+++ b/National Archives of Australia.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsbv",
-	"lastUpdated": "2014-04-03 17:48:28"
+	"lastUpdated": "2015-06-08 22:01:00"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


aurimasv:
Could you post some URLs that this translator is supposed to work with? If there are no persistent URLs, can you describe how to navigate to the pages of interest?

wragge:
Some examples below (no persistent urls, but these links should redirect you through the guest authentication).

Series: http://www.naa.gov.au/cgi-bin/Search?Number=A1
Item: http://www.naa.gov.au/cgi-bin/Search?O=I&Number=12048
Photo: http://www.naa.gov.au/cgi-bin/Search?O=PSI&Number=4746236

This one goes direct (ie no authentication), but there's no way to link to a specific page within the file. So to test properly you need to follow this link, use the navigation to go to a page, and then fire the translator to make sure you're saving the page you want:

http://recordsearch.naa.gov.au/SearchNRetrieve/Interface/ViewImage.aspx?B=12048

aurimasv:
Thanks! I ended up reorganizing the whole translator and cleaning up the code. Please update your translators via Preferences -> General -> Update Now and play around with it. Let us know if there is something we're not capturing right. (there was some metadata that got dropped)

PR #915, opened 2015-07-04
HeinOnline translator

fbennett:
HeinOnline currently uses the COinS translator. That doesn't fetch full-text attachments, so I did some digging around and came up with this.


zuphilip:
The usual licence block is missing.

(HeinOnline is connected to https://forums.zotero.org/discussion/50413/no-zotero-icon-in-address-bar/#Item_13 )

HeinOnline.js
```diff
@@ -0,0 +1,145 @@
+{
+	"translatorID": "3dcbb947-f7e3-4bbd-a4e5-717f3701d624",
+	"label": "HeinOnline",
+	"creator": "Frank Bennett",
+	"target": "https?://heinonline.org/HOL/(?:LuceneSearch|Page)\\?",
```
adam3smith: start target with ^ for efficiency and escape periods.

HeinOnline.js
```diff
@@ -0,0 +1,145 @@
...
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "g",
+	"lastUpdated": "2015-07-04 10:32:10"
+}
+
+function detectWeb (doc, url) {
+	if (url.match(/\/LuceneSearch\?/)) {
```
adam3smith: use indexOf("/LuceneSearch?/")!=-1, which is much more efficient.

HeinOnline.js
```diff
@@ -0,0 +1,145 @@
...
+		return "multiple";
+	} else {
+		return "journalArticle";
+	}
+}
+
+function Data(doc) {
+	this.node = ZU.xpath(doc, '//form[@id="Print1"]');
+	this.urlbase = "http://heinonline.org/HOL/PDFsearchable?";
```
adam3smith: are we worried about cross http/https requests (which would break this in connectors and potentially otherwise)? In other words, should we make sure we take http/https from the URL?

HeinOnline.js
```diff
@@ -0,0 +1,145 @@
...
+		url = url.replace(/Print/, "Page");
+		url = url.replace(/&terms=[^&]*/, '');
+		
+		items[url] = title;
+		found = true;
+	}
+	return found ? items : false;
+}
+
+function scrapePage(doc, isSingle) {
```
adam3smith: you're not using the isSingle condition/variable anywhere, are you?

adam3smith:
I don't have access to Hein right now, but could we also complement the metadata? E.g. at least a starting page number in the case of the test?

fbennett:
The item shows page 26 as the starting page. Do you mean the ending page number?

fbennett:
@zuphilip What does the usual license block look like? I checked a few other files and didn't see anything. Happy to comply, but I'll need some more concrete guidance.

The post at https://forums.zotero.org/discussion/50413/no-zotero-icon-in-address-bar/#Item_13 mentions Hein but I don't recognize the "International Law Association Database." The entries that I've seen from Hein so far show correct page numbers.

HeinOnline.js
fbennett: Done.

HeinOnline.js
fbennett: Done (dropped the trailing slash, which would yield a non-match always).

HeinOnline.js
fbennett: Done.

HeinOnline.js
fbennett: Removed, thanks.

zuphilip:
@fbennett Well, most important is to mention the GPL licence (explicitely). Moreover, I guess it makes sense to mention the copyright and that the translator are part of zotero. I guess that is not everything is needed to be mentioned explicitely from a legal point of view, but it might help others to reuse the code easier and to store its provenance. You can find my <a href="https://github.com/zuphilip/translators/wiki">template on my wiki pages</a>.

HeinOnline.js
```diff
@@ -0,0 +1,158 @@
+{
+	"translatorID": "3dcbb947-f7e3-4bbd-a4e5-717f3701d624",
+	"label": "HeinOnline",
+	"creator": "Frank Bennett",
+	"target": "^https?://heinonline\.org/HOL/(?:LuceneSearch|Page)\\?",
```
aurimasv: Double backslashes, otherwise you're just escaping the period for the string, which is still an unescaped period in RegExp. Use single if you're typing in Scaffold.

aurimasv: Also you can drop the `?:` for brevity, since these RegExp's are never used for `.match` only `.test`.

HeinOnline.js
```diff
@@ -0,0 +1,158 @@
...
+	if (url.indexOf("/LuceneSearch?") > -1) {
+		return "multiple";
+	} else {
+		return "journalArticle";
+	}
+}
+
+function Data(doc) {
+	this.node = ZU.xpath(doc, '//form[@id="Print1"]');
+	var urlLst = doc.location.href.split('/');
```
aurimasv: Don't use `doc.location.href`, use URL passed into doWeb or detectWeb. Those are deproxified and proxified as needed

HeinOnline.js
```diff
@@ -0,0 +1,158 @@
...
+	this.urlbase = urlLst.slice(0, 4).join('/') + "/PDFsearchable?";
```
aurimasv: What's the point of this? Isn't it always "/HOL/"? Also, use relative URLs

HeinOnline.js
```diff
@@ -0,0 +1,158 @@
...
+function detectWeb (doc, url) {
+	if (url.indexOf("/LuceneSearch?") > -1) {
+		return "multiple";
+	} else {
+		return "journalArticle";
+	}
+}
+
+function Data(doc) {
+	this.node = ZU.xpath(doc, '//form[@id="Print1"]');
```
aurimasv: This returns an array, which actually does work with `.xpath` below, but it can be confusing. You pretty much just want `.getElementById` here anyway.

HeinOnline.js
```diff
@@ -0,0 +1,158 @@
...
+	var urlLst = doc.location.href.split('/');
+	this.urlbase = urlLst.slice(0, 4).join('/') + "/PDFsearchable?";
+	this.queryElems = [];
+	this.tail = "&sectioncount=1&ext=.pdf&nocover=";
```
aurimasv: Just FYI, the position of url arguments generally does not matter (unless you're dealing with an array), so you could just tack these onto your `.urlbase`

HeinOnline.js
```diff
@@ -0,0 +1,158 @@
...
+}
+
+Data.prototype.getval = function(name, returnOnly) {
```
aurimasv: Nit: camel case (`getVal`)

HeinOnline.js
```diff
@@ -0,0 +1,158 @@
...
+	var val = '';
+	var input = ZU.xpath(this.node, './/input[@name="' + name + '"]');
+	if (input && input.length) {
+		val = input[0].value ? encodeURIComponent(input[0].value) : '';
+	}
+	if (!returnOnly) {
+		this.queryElems.push(name + "=" + val);
```
aurimasv: Technically, name may also have to be URL encoded, though it's rare in the wild.

HeinOnline.js
```diff
@@ -0,0 +1,158 @@
...
+	}
+	return val;
+}
+
+Data.prototype.dump = function() {
+	return this.urlbase + this.queryElems.join("&") + this.tail;
+}
+
+function getSearchResults(doc) {
+	var results = ZU.xpath(doc, '//div[contains(@class, "lucene_search_result_b")]'),
```
aurimasv: Up to you, but `.getElementsByClassName("lucene_search_result_b")` may be more robust (assuming that class is not overused)

HeinOnline.js
```diff
@@ -0,0 +1,158 @@
...
+		items = {},
+		found = false
+	for (var i=0; i<results.length; i++) {
+		
+		var url = ZU.xpath(results[i], './/a[contains(@href, "Print")]');
+		url = (url && url.length) ? url[0].href : false;
```
aurimasv: FYI `.xpath` always returns an array, so you don't have to check it.

HeinOnline.js
```diff
@@ -0,0 +1,158 @@
...
+		this.queryElems.push(name + "=" + val);
+	}
+	return val;
+}
+
+Data.prototype.dump = function() {
+	return this.urlbase + this.queryElems.join("&") + this.tail;
+}
+
+function getSearchResults(doc) {
```
aurimasv: Should use this in detectWeb

HeinOnline.js
```diff
@@ -0,0 +1,158 @@
...
+		items[url] = title;
+		found = true;
+	}
+	return found ? items : false;
+}
+
+function scrapePage(doc) {
+	var pdfPageURL = doc.location.href.replace(/\/Page\?/, "/Print?");
+	var item = new Zotero.Item();
+	var spans = ZU.xpath(doc, '//span[contains(@class, " Z3988") or contains(@class, "Z3988 ") or @class="Z3988"][@title]');
```
aurimasv: How about just `.getElementsByClassName`

HeinOnline.js
```diff
@@ -0,0 +1,158 @@
...
+	if (spans && spans.length) {
+		ZU.parseContextObject(spans[0].title, item);
+	}
+
+	ZU.processDocuments([pdfPageURL], 
+		function(pdoc){
```
aurimasv: You get a url here as well

HeinOnline.js
```diff
@@ -0,0 +1,158 @@
+{
+	"translatorID": "3dcbb947-f7e3-4bbd-a4e5-717f3701d624",
+	"label": "HeinOnline",
+	"creator": "Frank Bennett",
+	"target": "^https?://heinonline\.org/HOL/(?:LuceneSearch|Page)\\?",
+	"minVersion": "1.0",
```
aurimasv: 3.0

HeinOnline.js
```diff
@@ -0,0 +1,158 @@
...
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "g",
```
aurimasv: Looks like it should be ok for "gcsbiv"

HeinOnline.js
fbennett: Okay to leave this as is, since the values of "name" are known to not to be ones that require encoding? Or would you prefer to have a wrapper on it for completeness?

HeinOnline.js
fbennett: Good to know.

HeinOnline.js
aurimasv: I'm sure it will be ok as is.

HeinOnline.js
fbennett: The expression was lifted from the COinS translator. I assumed that xpath was used there because of the tag-name and title-attribute constraints.

Could get all elements with the Z3988 class, then discard any that are not spans or that have no title value. Not sure which is less attractive - leaving it like this for now.

fbennett:
@zuphilip Thanks for the licence link.

HeinOnline.js
fbennett: I tried both getElementById() and the first node reported from the xpath statement, and both returned an empty array from a known-working xpath statement against this form node. I must be missing something, but I'm not sure what.

Leading the statement with "//" instead of ".//" yields a result, but in other contexts I have seen that grab everything in the document, which would not be good. If I'm missing something basic about xpath syntax, I can change this over. Otherwise, I would feel more comfortable leaving as it is.

fbennett:
@aurimasv @adam3smith @zuphilip Changes done, ready for another round of rotten tomatoes.

zuphilip:
I just checked the licence info, and that is fine.

No :tomato: but a :+1:  from me.

fbennett:
@adam3smith @aurimasv I've added some code to handle the case of items that report an available PDF in the search listing, but redirect to a not-yet-available placeholder page with no metadata.

HeinOnline.js
```diff
@@ -0,0 +1,210 @@
...
+		} else {
+			item.title = getXPathStr("textContent", doc, '//div[@id="content-container"]//a[1]');
+		}
+		var notAvailableURL = url.replace("/Page?", "/NotAvailable?").replace("handle=", "handle_bad=");
+		item.attachments.push({
+			url:notAvailableURL,
+			mimetype:"text/html",
+			snapshot:true,
+			title:"HeinOnline page placeholder"
+		});
```
aurimasv: what is the purpose of this?

HeinOnline.js
```diff
@@ -0,0 +1,210 @@
...
+		item.url = url;
+		item.abstract = "CAUTION: Resource not yet available at HeinOnline";
```
aurimasv: Should be `.abstractNote`, but also, this should be unnecessary.

HeinOnline.js
```diff
@@ -0,0 +1,210 @@
...
+	return found ? items : false;
+}
+
+function scrapePage(doc, url) {
+	var pdfPageURL = url.replace(/\/Page\?/, "/Print?");
+	var item = new Zotero.Item();
+	var z3988title = getXPathStr("title", doc, '//span[contains(@class, " Z3988") or contains(@class, "Z3988 ") or @class="Z3988"][@title]');
+	ZU.parseContextObject(z3988title, item);
+	
+	if (!item.itemType) {
```
aurimasv: can we also check this in detectWeb? maybe something not as involved as parsing context object (thought that should be ok too)

HeinOnline.js
aurimasv: >redirect to a not-yet-available placeholder page with no metadata

Wouldn't the URL stored in the URL field do that?

fbennett: It's worth capturing the attachment. It doesn't contain structured metadata, but there are useful details in the free text, and we might as well make that available to the user when viewing offline. I think you're right that the URL doesn't need to be rewritten.

aurimasv: In that case, just use `document: doc`

HeinOnline.js
fbennett: Structured metadata is acquired after successful resolution to the article's home page. Placeholder pages don't contain structured metadata, so the details will need manual editing. The purpose of this note is to flag such entries so the user can easily revisit them.

HeinOnline.js
fbennett: Yep, that would be better.

HeinOnline.js
adam3smith: You can convince me of this (if fixed to abstractNote), but right now I'm with Aurimas: wouldn't an empty abstract field be just as easily searchable? Or if need be, add a tag? Putting bad text in the abstract field doesn't strike me as a good solution.

adam3smith:
@fbennett -- the remaining comments should be easy/quick to address, would be happy to merge this.

HeinOnline.js
zuphilip: I agree that `abstractNote` is the wrong field to save any dummy text which can be used as a flag. Maybe, a note might be better?

zuphilip:
@fbennett Any news? There is a demand in the forum for your translator: https://forums.zotero.org/discussion/55105/heinonline-translator-broken/#Item_1

fbennett:
My desk has settled down a bit. I should be able to work out remaining issues this week:

* Dan Epps reports on Twitter that some items pull in incorrect page numbers;
* The translator is failing on PDF attachments when pulling a page in view (PDF downloads from a search listing are OK)

fbennett:
I labelled the checkin "Fall back to COinS," but it's a more or less complete rewrite of the previous proposal. Here are the highlights:

- Content hosted directly by HeinOnline has RIS in the item pages, which we can fetch by building its URL. The RIS is of better quality than the COinS that we were using before.
- Content from the Index to Foreign Legal Periodicals has embedded COinS, so we use that if RIS is not available. 
- PDFs are fetched from item pages if available.

It seems to work well, ready for review and amendments.

HeinOnline.js
```diff
@@ -0,0 +1,240 @@
+{
+	"translatorID": "3dcbb947-f7e3-4bbd-a4e5-717f3701d624",
+	"label": "HeinOnline",
+	"creator": "Frank Bennett",
+	"target": "https?://heinonline.org/HOL/(?:LuceneSearch|Page|IFLPMetaData)\\?",
```
adam3smith: escape periods (`\\.org``). Ad `^` at the beginning. And we're sure they never have `www\\.`?

HeinOnline.js
```diff
@@ -0,0 +1,240 @@
...
+	"minVersion": "1.0",
```
adam3smith: since it's using `ZU.xpath`, this should be 3.0. I don't think we do anything with that info, but might as well do it right.

HeinOnline.js
```diff
@@ -0,0 +1,240 @@
...
+*/
+
+// Get any search results from current page
+// Used in detectWeb() and doWeb()
+function getSearchResults(doc, url) {
+	var results = doc.getElementsByClassName("lucene_search_result_b"),
+		items = {},
+		found = false
+	for (var i=0, ilen=results.length; i<ilen; i++) {
+		var url = getXPathStr("href", results[i], './/a[1]');
```
adam3smith: the way you're using this, any reason to not just do `ZU.xpathText(results[i], './/a[i]/@href')` Just seems odd that there'd be a separate function. But I won't insist here.

HeinOnline.js
```diff
@@ -0,0 +1,240 @@
...
+function getXPathStr(attr, elem, path) {
+	var res = ZU.xpath(elem, path);
+	res = res.length ? res[0][attr] : '';
+	return res ? res : '';
+}
+
+// Extract query values to keys on an object
+function extractQueryValues(url) {
+	var ret = {};
+	ret.base = url.replace(/(.*?)[a-zA-Z]+\?.*/, "$1");
```
adam3smith: maybe add comment with example?

HeinOnline.js
```diff
@@ -0,0 +1,240 @@
...
+		    if (pdfPageURL) {
+			    ZU.doGet(pdfPageURL, function(pdfPage) {
+				    // Call to pdfPageURL prepares PDF for download via META refresh URL
+				    pdfURL = null;
+				    var m = pdfPage.match(/<META.*URL=([^\"]+)/);
+				    if (m) {
+					    var pdfURL = m[1];
+				    }
+				    translateRIS(ris, pdfURL);
+			    } , null , "UTF-8");
```
adam3smith: why do we need to specify utf-8? That should just be the default, no?

adam3smith:
looks good. See comments inline. Apart from the target-related ones (those you should definitely do) see these as questions and/or optional

HeinOnline.js
aurimasv: And it's generally best to get the value of `.href` from the anchor element than the value of the href html attribute via xpath.  Usually it works out the same, but sometimes it may differ.

aurimasv: (That's what you do in your helper function, but I agree with @adam3smith that this is not really necessary)

HeinOnline.js
```diff
@@ -0,0 +1,240 @@
...
+
+// Remangle HeinOnline RIS to form expected by translator
+function fixRIS (ris) {
+	var lines = ris.split("\n");
+	for (var i=0,ilen=lines.length;i<ilen;i++) {
+		lines[i] = lines[i].replace(/^([A-Z0-9][A-Z0-9]) /, "$1  - ");
+		lines[i] = lines[i].replace(/^VO  /, "VL  ");
+		lines[i] = lines[i].replace(/^YR  /, "DA  ");
+		lines[i] = lines[i].replace(/^OP  /, "EP  ");
+	}
```
aurimasv: You don't need to split up everything into single line. You can run the replacements on the whole text with `mg` flags (m makes ^ match start of line).

HeinOnline.js
```diff
@@ -0,0 +1,240 @@
...
+    ** Utilities **
+    ***************
+*/
+
+// Get any search results from current page
+// Used in detectWeb() and doWeb()
+function getSearchResults(doc, url) {
+	var results = doc.getElementsByClassName("lucene_search_result_b"),
+		items = {},
+		found = false
```
aurimasv: Missing semicolon

HeinOnline.js
```diff
@@ -0,0 +1,240 @@
...
+	ris = lines.join("\n");
+	return ris;
+}
+
+// Not all pages have a downloadable PDF
+function translateRIS(ris, pdfURL) {
+	var trans = Zotero.loadTranslator('import');
+	trans.setTranslator('32d59d2d-b65a-4da4-b0a3-bdd3cfb979e7');//https://github.com/zotero/translators/blob/master/RIS.js
+	trans.setString(ris);
+	if (pdfURL) {
```
aurimasv: The if statement should be inside the itemDone handler

HeinOnline.js
```diff
@@ -0,0 +1,240 @@
...
+}
+
+function translateCOinS(COinS) {
+    var item = new Zotero.Item();
+    try {
+        Zotero.Utilities.parseContextObject(COinS, item);
+        item.complete();
+    } catch(e) {
+        Zotero.debug("error parsing COinS");
+    }
```
aurimasv: Let this throw, so the user knows that there was an error and so that we get a report

HeinOnline.js
```diff
@@ -0,0 +1,240 @@
...
+}
+
+// Build URL for RIS, and for PDF if available
+function scrapePage(doc, url) {
+	var risPopupURL = getXPathStr("href", doc, '//form[@id="pagepicker"]//a[contains(@href, "PrintRequest")][1]');
```
aurimasv: Nit: indent

HeinOnline.js
```diff
@@ -0,0 +1,240 @@
...
+    if (risPopupURL) {
+        // If page has RIS, use that
+	    var docParams = extractQueryValues(risPopupURL);
+	    var risURL = docParams.base 
+            + "CitationFile?kind=ris&handle=" + docParams.handle 
+            + "&div=" + docParams.div 
+            + "&id=" + docParams.id 
+            + "&base=js";
+	    var pdfPageURL = getXPathStr("href", doc, '//a[contains(@class, "updatediv")]');
```
aurimasv: `doc.getElementsByClassName` is a better choice here

HeinOnline.js
```diff
@@ -0,0 +1,240 @@
...
+	    ZU.doGet(risURL, function(ris) {
+		    ris = fixRIS(ris);
+		    if (pdfPageURL) {
+			    ZU.doGet(pdfPageURL, function(pdfPage) {
+				    // Call to pdfPageURL prepares PDF for download via META refresh URL
+				    pdfURL = null;
+				    var m = pdfPage.match(/<META.*URL=([^\"]+)/);
+				    if (m) {
+					    var pdfURL = m[1];
+				    }
```
aurimasv: No var

HeinOnline.js
```diff
@@ -0,0 +1,240 @@
...
+            + "&id=" + docParams.id 
+            + "&base=js";
+	    var pdfPageURL = getXPathStr("href", doc, '//a[contains(@class, "updatediv")]');
+	    ZU.doGet(risURL, function(ris) {
+		    ris = fixRIS(ris);
+		    if (pdfPageURL) {
+			    ZU.doGet(pdfPageURL, function(pdfPage) {
+				    // Call to pdfPageURL prepares PDF for download via META refresh URL
+				    pdfURL = null;
+				    var m = pdfPage.match(/<META.*URL=([^\"]+)/);
```
aurimasv: No \ inside regexp. It's not necessary

HeinOnline.js
aurimasv: Please fix indents throughout the file

HeinOnline.js
aurimasv: Seems to me like you can drop the `(.*?)` and use `""` in replace

aurimasv: Also, it appears that "base" may actually be a possible parameter, so I would make sure not to clobber it below

HeinOnline.js
```diff
@@ -0,0 +1,240 @@
...
+}
+
+// Extract query values to keys on an object
+function extractQueryValues(url) {
+	var ret = {};
+	ret.base = url.replace(/(.*?)[a-zA-Z]+\?.*/, "$1");
+	var query = url.replace(/.*?\?/, "");
+	query = query.split("&");
+	for (var i=0,ilen=query.length;i<ilen;i++) {
+		let pair = query[i].split("=");
```
aurimasv: No `let` in translators

HeinOnline.js
```diff
@@ -0,0 +1,240 @@
...
+    *********
+*/
+
+function detectWeb (doc, url) {
+	if (url.indexOf("/LuceneSearch?") > -1) {
+		if (getSearchResults(doc)) {
+			return "multiple";
+		}
+	} else {
+		return "journalArticle";
```
aurimasv: Make sure either RIS out COinS is available

HeinOnline.js
```diff
@@ -0,0 +1,240 @@
...
+
+/*
+    ***************
+    ** Utilities **
+    ***************
+*/
+
+// Get any search results from current page
+// Used in detectWeb() and doWeb()
+function getSearchResults(doc, url) {
```
aurimasv: You don't use this `url` in the function. You also don't always pass it in.

HeinOnline.js
```diff
@@ -0,0 +1,240 @@
...
+			ZU.processDocuments(urls, scrapePage);
+		});
+	} else {
+		scrapePage(doc, url);
+	}
+}
+
+/** BEGIN TEST CASES **/
+var testCases = [
+	{
```
aurimasv: Is it possible to add more tests for different cases above (multiples, two url patterns, ris, CoinS)

adam3smith:
superseded by #1082

PR #920, opened 2015-07-14
ScienceDirect is compatible with translation-server

wetneb:
It seems to me that this scraper (ScienceDirect.js) is compatible with translation-server, based on the tests provided in the scraper (I have not checked that all fields match, though). At least it provides useful information.

Note: this scraper does not work when the client is affiliated to multiple institutions, where it is redirected to a landing page where they have to choose between the affiliations.

### DIFF
--- a/ScienceDirect.js
+++ b/ScienceDirect.js
@@ -8,7 +8,7 @@
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
-	"browserSupport": "gcsib",
+	"browserSupport": "gcsibv",
 	"lastUpdated": "2014-12-04 23:08:27"
 }
 


wetneb:
This comment is actually true for other scrapers as well. Is there a simple way to run all the tests within translation-server to check whether the scrapers work and update compatibility accordingly?

aurimasv:
I'm actually not entirely clear whether there is anything in terms of code compatibility that would make a translator not compatible with server translation. AFAIK, the only reason a page would not be translatable on the server is if the server does not have access to the resource (e.g. proxied URL, or gated resource). @simonster knows this better.

kba:
I can second that, the 'v' flag seems unnecessary, since the translation server runs the same Gecko engine as Firefox. In fact, forcing the translation-server to accept all browser-enabled translators (hack: https://github.com/infolis/translation-server/commit/0f4d49ca2efd4d1f1c0f66b7c13bbea668ec5484#diff-dffa17606df6f2a44db139b3e7c2fcd5) worked for all the examples we tested.

But I'm sure @simonster had good reasons to differentiate between `g` and `v` (https://github.com/zotero/translation-server/commit/699ded0aa636ffbeda77eb476aea503ca4cc25ac).

wetneb:
I use the same hack and it works very well indeed (but it took me a while to write it independently, sadly).

zuphilip:
The `v` flag for Science Direct was set in https://github.com/zotero/translators/commit/26cc99d0562732bd616a2d1c0b043c23eb4dc70a and therefore we can close this issue.

_For your information (as far as I understand it)_: The translation server looks different when calling urls, e.g. we have to wrap the document such that it looks like to have similar properties than in the browser context. Not everything might work, especially one has to look for functions like `document.location` or `document.head`, calling another translator inside a translator and passing some values...

PR #923, opened 2015-07-26
rewrite youTube based on a mix of screenscraping (abstracts) and JSON…

adam3smith:
… embedded on the page.

Issues (I think I picked the best solution for all of these but just to have a record):
1. There's no good solution for usernames. I'm now letting clean author parse them, I think that's the best we can do.
2. The keywords are entered like this, i.e. as individual keywords per word ("for" "and") etc. Nothing we can do about that
3. Anyone remember why/whether we decided to make uploaders "contributors" not "authors"? Might be APA etc. related?

### DIFF
--- a/YouTube.js
+++ b/YouTube.js
@@ -9,16 +9,16 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsbv",
-	"lastUpdated": "2015-06-10 10:46:40"
+	"lastUpdated": "2015-07-31 15:41:17"
 }
 
 function detectWeb(doc, url) {
-	if (getVideoId(url)) {
+	if (url.search(/\/watch\?(?:.*)\bv=[0-9a-zA-Z_-]+/) != -1) {
 		return "videoRecording";
 	}
-	
+
 	//Search results
-	if ( getSearchResults(doc, true) ) {	
+	if (getSearchResults(doc, true)) {
 		return "multiple";
 	}
 }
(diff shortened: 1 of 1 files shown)


YouTube.js
```diff
@@ -26,97 +26,84 @@ function detectWeb(doc, url) {
...
-	ZU.doGet(uris, function(text) {	
-		var ns = {"default":"http://www.w3.org/2005/Atom", "media":"http://search.yahoo.com/mrss/", "yt":"http://gdata.youtube.com/schemas/2007"};
-		
-		var parser = new DOMParser();
-		var doc = parser.parseFromString(text, "text/xml");
+function scrape(doc, url) {
+	var newItem = new Zotero.Item("videoRecording");
+	//grab the JSON in the header of the page and remove JS code
+	var data = ZU.xpathText(doc, '//script[contains(text(), "ytplayer.config")]');
+	data = data.match(/ytplayer.config =(.+);\s*ytplayer.load/)[1];
```
aurimasv: `/ytplayer\.config\s*=(.+);\s*ytplayer\.load/`

YouTube.js
```diff
@@ -26,97 +26,84 @@ function detectWeb(doc, url) {
...
+	//Z.debug(data)
+	try {
+		var obj = JSON.parse(data);
+	} catch (e) {
+		Zotero.debug("JSON parse error");
```
aurimasv: I'd log the `data` string, I think

YouTube.js
```diff
@@ -139,12 +127,28 @@ var testCases = [
...
+					"New",
+					"Reference",
+					"Research",
+					"University",
+					"Web",
+					"Zotero",
+					"and",
+					"bibliography",
+					"for"
+				],
```
aurimasv: do you think it's worth getting keywords at all, given the mess?

YouTube.js
```diff
@@ -166,22 +170,40 @@ var testCases = [
...
-				"title": "HD Charlie Chaplin - The Great Dictator: Speech to Humanity | HD",
+				"title": "The Great Dictator - Speech to Humanity | HD",
 				"creators": [
 					{
-						"lastName": "EMS co",
-						"creatorType": "contributor",
-						"fieldMode": 1
+						"firstName": "E. M. S.",
+						"lastName": "co",
+						"creatorType": "contributor"
```
aurimasv: I'd vote for treating all author names as institutional authors without attempting to parse them. I'm not sure how often  people actually use their full names on YouTube

YouTube.js
```diff
@@ -166,22 +170,40 @@ var testCases = [
...
-				"tags": [],
+				"tags": [
+					"1940",
+					"Chaplin",
+					"Charlie Chaplin",
+					"Charlie Chaplin (Author)",
+					"Dictator",
+					"Great",
+					"HD",
+					"High-definition Video (Film Format)",
```
aurimasv: though it's interesting that these keywords didn't get split up by word... Maybe the Zotero video is the exception, not the rule.

aurimasv:
>Anyone remember why/whether we decided to make uploaders "contributors" not "authors"? Might be APA etc. related?

That seems to date back to before 2008, so I have no idea why "contributor" was chosen. I think "author" would be more consistent with everything else we scrape. I guess the issue is that frequently the videos posted on YouTube are not produced by the poster. I think we should just let the Zotero user correct this.

YouTube.js
adam3smith: yeah, because I think this is an entry error, not a YouTube one as per below.

YouTube.js
adam3smith: OK, makes sense. One of the vids I tried this on was a real name, but I agree that's rare

adam3smith:
OK, I'll switch it to author then. That's in line with at least APA: http://blog.apastyle.org/apastyle/2011/10/how-to-create-a-reference-for-a-youtube-video.html

adam3smith:
OK, should be all good now

zuphilip:
( Reference to French blog describing this problem: http://zotero.hypotheses.org/696#commentlist-container )

YouTube.js
```diff
@@ -26,97 +26,81 @@ function detectWeb(doc, url) {
...
-	ZU.doGet(uris, function(text) {	
-		var ns = {"default":"http://www.w3.org/2005/Atom", "media":"http://search.yahoo.com/mrss/", "yt":"http://gdata.youtube.com/schemas/2007"};
-		
-		var parser = new DOMParser();
-		var doc = parser.parseFromString(text, "text/xml");
+function scrape(doc, url) {
+	var newItem = new Zotero.Item("videoRecording");
+	//grab the JSON in the header of the page and remove JS code
+	var data = ZU.xpathText(doc, '//script[contains(text(), "ytplayer.config")]');
+	data = data.match(/ytplayer\.config\s*=(.+);\s*ytplayer\.load/)[1];
```
aurimasv: nit: Just in case can you make that `(.+?)`. The xpath can grab multiple script elements in theory and there's some chance that ytplayer.load could be mentioned somewhere else, so lazy operator is a bit safer.

YouTube.js
```diff
@@ -166,22 +166,40 @@ var testCases = [
...
...he people have the power to make this life free and beautiful, to make this life a wonderful adventure, then in the name of democracy let's use that power, let us all unite! Let us fight for a new world, a decent world that will give men a chance to work, that will give you the future and old age and security. By the promise of these things, brutes have risen to power, but they lie. They do not fulfil their promise, they never will.Dictators free themselves but they enslave the people.Now let us fight to fulfill that promise, let us fight to free the world, to do away with national barriers, do away with greed, with hate and intolerance.Let us fight for a world of reason, a world where science and progress will lead to all men's happiness.Soldiers! In the name of democracy, let us all unite!The Great Dictator is a 1940 American comedy-drama film starring, written, produced, scored, and directed by Charlie Chaplin, following the tradition of many of his other films. Having been the only Hollywood filmmaker to continue to make silent films well into the period of sound films, this was Chaplin's first true talking picture as well as his most commercially successful film. (Wikipedia)",
```
aurimasv: Not a big deal, but this may be closer to the original if we used `.innerHTML` and then `ZU.cleanTags` Now, the newlines are missing causing some sentences to start without any white space after a period. (may also require a pass through `ZU.unescapeHTML` afterwards if there are any HTML entities... and maybe some additional white space cleanup)

aurimasv:
Two nits, but otherwise this is fine with me. Feel free to merge.

adam3smith:
thanks for the help.

PR #928, opened 2015-08-02
Fix book -> mvbook (multi-volume book)

gracile-fr:
A book with number of volumes is a BibLaTeX <code>@mvbook</code>;
A book with volume is a BibLaTeX <code>@book</code>
See https://forums.zotero.org/discussion/50928/biblatex-translator-and-mvbook-entry-type/
@andersjohansson

### DIFF
--- a/BibLaTeX.js
+++ b/BibLaTeX.js
@@ -15,7 +15,7 @@
 		"exportFileData": false,
 		"useJournalAbbreviation": false
 	},
-	"lastUpdated": "2015-07-02 02:03:17"
+	"lastUpdated": "2015-08-03 08:45:20"
 }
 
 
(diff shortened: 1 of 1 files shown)


BibLaTeX.js
```diff
@@ -471,8 +471,9 @@ function encodeFilePathComponent(value) {
 			if(item.itemType == "book" && !creatorCheck(item,"author") &&
 			   creatorCheck(item, "editor")) type = "collection";
 
-			//biblatex recommends us to use mvbook for multi-volume books
-			if (type == "book" && item.volume) type = "mvbook"
+			//biblatex recommends us to use mvbook for multi-volume book
+			//i.e. a book with "# of vols" filled
+			if (type == "book" && item.numberOfVolumes) type = "mvbook"
```
aurimasv: Logically, we should check that `item.numberOfVolumes > 1`, since we do sometimes import 1 into that field. However, that would break if someone entered something like "2 vols". So this is ok.

aurimasv:
You need to bump the timestamp at the top

gracile-fr:
manually?

adam3smith:
yes, for export translators you need to bump the timestamp manually since there's no scaffold, so this is great.
I don't see how you addressed @aurimasv's concern above, though?

gracile-fr:
<blockquote>However, that would break if someone entered something like "2 vols". So this is ok.</blockquote>

adam3smith:
yeah, we have code in place to remove the 1 from num volumes from SUDOC imports. All good then, thanks.

PR #929, opened 2015-08-04
New translator: HCSP

jomuller:
As suggested in the [zotero dev Google group](https://groups.google.com/forum/#!topic/zotero-dev/WXIVULPDYVc), I send a pull request for a new translator.

This translator is for the website ["Haut Conseil de la Santé Publique"](http://www.hcsp.fr/explore.cgi/avisrapports) widely used in France for healthcare. It's the first translator I wrote, then there is most likely some defect in this script. I tested it on a lot of pages (like this one) in this website, and it worked well. Please, feel free to criticize if there is something bad.

If it reaches the quality standards, should it be possible to make available this translator for the others Zotero users?

### DIFF
--- a/HCSP.js
+++ b/HCSP.js
@@ -1,0 +1,312 @@
+{
+	"translatorID": "3e375db7-5f6e-4861-ab8b-3ddb0f4accc9",
+	"label": "HCSP",
+	"creator": "Joris Muller",
+	"target": "^https?://www\\.hcsp\\.fr/explore\\.cgi/",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcs",
+	"lastUpdated": "2015-08-05 09:11:04"
+}
+
+/*
+HCSP Translator
+A Zotero's Translator for the website of "Haut Conseil de Santé Publique"
+Copyright (C) 2015 Joris Muller
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+// Usage of Zotero.Utilities.xpath() function. Exist only since Zotero 2.1
+// https://www.zotero.org/support/dev/client_coding/changes_in_zotero_2.1#xpath_utility_functions
+// Zotero.selctItems used only since 3.0
+// Then minimal version of Zotero needed is 3.0
+
+function detectWeb(doc, url) {
+	/* The URLs have upper or lower case on the website
+	depending the way you access to a page */
+	if (url.indexOf("avisrapportsdomaine") != -1) {
+		return 'report';
+	} else if (url.indexOf("avisrapports") != -1 || url.indexOf("AvisRapports") != -1  ) {
+		// Detect if there is at last one result
+		return getSearchResults(doc) ? 'multiple' : false;
+	}
+}
+
+function doWeb(doc, url) {
+
+	if (detectWeb(doc, url) == "multiple") {
+
+		var allItems = getSearchResults(doc);
+
+		Zotero.selectItems(allItems, function(items) {
+			if (!items) return true;
+
+			var articlesURLs = [];
+			
+			for (var itemURL in items) {
+				articlesURLs.push(itemURL);
+			}
+
+			ZU.processDocuments(articlesURLs, scrape);
+		  
+		})
+
+	} else {
+		scrape(doc, url);
+	}
+}
+
+// For multiple articles
+function getSearchResults(doc) {
+	// Flag to check if something was found
+	var found = false;
+
+	// Titles and URLs of each item are on the third column of the only table
+	// This column is the only one with links (<a>)
+	var xpath_titles = '//tbody/tr/td/a';
+	var titles = ZU.xpath(doc, xpath_titles);
+
+	// Put all the titles and their URLs in the items object
+	var allItems = {};
+
+	for (var i = 0; i< titles.length; i++) {
+		var nextTitle = titles[i];
+		allItems[nextTitle.href] = nextTitle.textContent;
+		found = true;
+	}
+
+	// if no item found (no result in search), return false, else the items
+	return found ? allItems : false;
+}
+
+function scrape(doc, url) {
+
+	// init a new Zotero's item
+	var newItem = new Zotero.Item("report");
+
+	/*---------- Constants items ----------*/
+	newItem.publicationTitle = "Rapport de l'HCSP";
+	newItem.language = "fr-FR";
+	newItem.institution = "Haut Conseil de la Santé Publique";
+	newItem.place = "Paris";
+
+	// Creator is the institution (HCSP)
+	// Names of the author are inside the PDFs (if there is one)
+	newItem.creators.push({"lastName": "HCSP", "creatorType": "author", "fieldMode": 1});
+
+	/*---------- URL ----------*/
+	newItem.url = url;
+	
+	/*---------- Title ----------*/
+	var xpathTitle = '//li[@class="active"]';
+	var title = ZU.xpathText(doc, xpathTitle);
+	if (title) {
+		newItem.title = ZU.trimInternal(title);
+	}
+
+	/*---------- Date ----------*/
+	// Date is not easy to find because it is not in a named bloc. It is
+	// somewhere in one of the <p> in the div 'doc-donnees'
+	// Then to find it, use a Regex on each <p> bloc
+	var xpathDate = "//div[@id='doc-donnees']/p[contains(text(),'Date du document')]";
+	var dateString = ZU.xpathText(doc, xpathDate);
+	if (dateString) {
+		var regexDate = /\b(\d{2})\/(\d{2})\/(\d{4})\b/;
+		var date = dateString.match(regexDate);
+		if (date) {
+			var dateISO8601 = date[3] + "-" + date[2] + "-" + date[1];
+			newItem.date = dateISO8601;
+		}
+	}
+	
+	/*---------- Tags ----------*/
+	var xpathTags = '//div[@id="doc-donnees"]/span/a';
+	var xresultTags = ZU.xpath(doc, xpathTags);
+
+	for (var i = 0; i < xresultTags.length; i++) {
+		var actualTag = xresultTags[i].textContent;
+		var actualTagTrimmed = actualTag.trim();
+		newItem.tags.push(actualTagTrimmed);
+	}
+
+	/* --------- abstract -----------*/
+	var xpath_abstract = '//div[@class="avistexte"]';
+	var abstract_note = ZU.xpathText(doc, xpath_abstract, null, "\n");
+	if (abstract_note) {
+		newItem.abstractNote = ZU.trimInternal(abstract_note);
+	}
+
+	/* ----------- attachment --------- */
+	// PDFs are in a div on the right part of the page.
+	var xpath_pdf = '//div[@id="doc-donnees"]/p/a[@target="_blank"]';
+	// ZU.xpath produce Arrays of HTMLAnchorElement
+	var xresultsPdf = ZU.xpath(doc, xpath_pdf);
+
+	// For each PDF anchor get the url and push it
+	for (var i = 0; i < xresultsPdf.length; i++) {
+		
+		var nextResult = xresultsPdf[i];
+		var pdfUrl = nextResult.href;
+
+		// Get the first word
+		var typeOfDoc = nextResult.textContent.match(/^\w+/i);
+		var pdfTitle = typeOfDoc + " HCSP Full Text PDF";
+
+		var one_pdf_attachement = {
+			url: pdfUrl,
+			title: pdfTitle,
+			mimeType: "application/pdf"
+		}
+
+		newItem.attachments.push(one_pdf_attachement);
+	}
+
+	newItem.complete();
+}
+
+/** BEGIN TEST CASES **/
+var testCases = [
+	{
+		"type": "web",
+		"url": "http://www.hcsp.fr/Explore.cgi/avisrapportsdomaine?clefr=474",
+		"items": [
+			{
+				"itemType": "report",
+				"title": "Vaccination contre les infections invasives à méningocoque C au-delà de 24 ans, notamment chez les hommes ayant des relations sexuelles avec d’autres hommes (HSH)",
+				"creators": [
+					{
+						"lastName": "HCSP",
+						"creatorType": "author",
+						"fieldMode": 1
+					}
+				],
+				"date": "2014-11-07",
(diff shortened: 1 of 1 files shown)


HCSP.js
```diff
@@ -0,0 +1,228 @@
+{
+    "translatorID": "3e375db7-5f6e-4861-ab8b-3ddb0f4accc9",
+    "label": "HCSP",
+    "creator": "Joris Muller",
+    "target": "http://www.hcsp.fr",
```
adam3smith: ```^https?//www\.hcsp\.fr/```
Make more efficient, allow for a current or future https version, escape ```.``` and end target with a ```/``` to work with proxies.

HCSP.js
```diff
@@ -0,0 +1,228 @@
...
+// Usage of Zotero.Utilities.xpath() function. Exist only since Zotero 2.1
+// https://www.zotero.org/support/dev/client_coding/changes_in_zotero_2.1#xpath_utility_functions
+// Then minimal version of Zotero needed is 2.1
+
+// TODO: Use camelStyle for variables to be consistent with Zotero coding style
+
+function detectWeb(doc, url) {
+    /* The URLs have upper or lower case on the website
+    depending the way you access to a page */
+    if (url.match("[Ee]xplore.cgi/avisrapportsdomaine")) {
```
adam3smith: if you're matching on a literal string, just use
```.indexOf("myString") != -1``` which is much more efficient -- particularly important for the detect function which runs a lot.

File: HCSP.js [hunk 3]
```diff
@@ -0,0 +1,228 @@
...
+        return 'report';
+    } else if (url.match("[Ee]xplore.cgi/avisrapports")) {
```
adam3smith: ditto

HCSP.js
```diff
@@ -0,0 +1,228 @@
...
+        // NOTE: "for of loop" are quite new and works only with newest browsers
+        // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of
+        var allItems = {};
+        for (var nextTitle of titles) {
+            allItems[nextTitle.href] = nextTitle.textContent;
+        }
+        
+        // Let the user choose the items he wants
+        var itemsSelected;
+        itemsSelected = Zotero.selectItems(allItems);
```
adam3smith: ```Zotero.selectItems``` should be run as a callback function. See e.g. https://github.com/zotero/translators/blob/d82c1efd2ca75af0caa0d94884e38d2e6afbc210/Dryad%20Digital%20Repository.js#L60 
for an example that otherwise does the same you do.

HCSP.js
```diff
@@ -0,0 +1,228 @@
...
+            articles.push(one_item)
+        }
+
+    } else {
+        // Page of the article : get the URL
+        articles = [url]
+    }
+
+    Zotero.Utilities.processDocuments(articles, scrape, function(){Zotero.done();});
+    Zotero.wait();
```
adam3smith: also see the translator above for how to simplify this: ```Zotero.done``` and ```Zotero.wait``` are no longer needed.

HCSP.js
```diff
@@ -0,0 +1,228 @@
...
+
+    var namespace = doc.documentElement.namespaceURI;
+    // Resolver for XPaths
+
+    // init a new Zotero's item
+    var newItem = new Zotero.Item("report");
+
+    /*---------- Constants items ----------*/
+    newItem.publicationTitle = "Rapport de l'HCSP";
+    newItem.language = "fr";
```
adam3smith: ```fr-FR``` (presumably)

HCSP.js
```diff
@@ -0,0 +1,228 @@
...
+        // Titles and URLs of each item are on the third column of the only table
+        // This column is the only one with links (<a>)
+        var xpath_titles = '//tbody/tr/td/a';
+        var titles = ZU.xpath(doc, xpath_titles);
+        
+        // Put all the titles and their URLs in the items object
+        // NOTE: "for of loop" are quite new and works only with newest browsers
+        // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of
+        var allItems = {};
+        for (var nextTitle of titles) {
```
adam3smith: I'm not sure, but given that ```for...of``` doesn't run in IE, I think we might prefer a different loop here. We usually just do
```for(var i = 0; i<titles.length;i++)``` which has the added benefit of being super efficient.

File: HCSP.js [hunk 8]
```diff
@@ -0,0 +1,228 @@
...
+
+    for (var nextResult of xresultTags) {
+        var actualTag = nextResult.textContent;
+        var actualTagTrimmed = actualTag.trim();
+        newItem.tags.push(actualTagTrimmed);
+    }
+
+    /* --------- abstract -----------*/
+    var xpath_abstract = '//div[@class="avistexte"]';
+    var abstract_note = ZU.trimInternal(ZU.xpathText(doc, xpath_abstract, namespace, "\n"));
```
adam3smith: why do you need the namespace here?
Also, you should test for the existence of ```ZU.xpathText(doc, xpath_abstract, namespace, "\n")``` before running ZU.trimInternal, which will throw an error when run on ```null```.
I'd do 
```
var abstract_note = ZU.xpathText(doc, xpath_abstract, namespace, "\n");
if (abstract_note){
  newItem.abstractNote = ZU.trimInternal(abstract_note);
}
```

HCSP.js
```diff
@@ -0,0 +1,228 @@
...
+    var pdf_filename_regex = /=(.*)/
+
+    // For each PDF anchor get the url and title and push it
+    for (one_pdf_anchor of xresult_pdfs) {
+        var pdf_url = one_pdf_anchor.href;
+        var pdf_title = pdf_url.match(pdf_filename_regex)[1];
+
+        var one_pdf_attachement = {
+            url: pdf_url,
+            title: pdf_title,
```
adam3smith: As per Zotero conventions, this should be "HCSP Full Text PDF" not the pdf title

HCSP.js
```diff
@@ -0,0 +1,228 @@
...
+    newItem.publicationTitle = "Rapport de l'HCSP";
+    newItem.language = "fr";
+    newItem.institution = "Haut Conseil de la Santé Publique";
+    newItem.place = "Paris";
+
+    // Creator is the institution (HCSP)
+    // Names of the author are inside the PDFs (if there is one)
+    // TODO: Find a way to have a single field instead of two
+    var creator = ZU.cleanAuthor('HCSP', 'author', false);
+    newItem.creators.push(creator);
```
adam3smith: Use
```newItem.creators.push({"lastName": "HCSP", "creatorType": "author", "fieldMode": 1})```;

File: HCSP.js [hunk 11]
```diff
@@ -0,0 +1,228 @@
...
+        }
+    }
+    
+    newItem.date = String(date);
+
+    /*---------- Tags ----------*/
+    var xpathTags = '//div[@id="doc-donnees"]/span/a';
+    var xresultTags = ZU.xpath(doc, xpathTags);
+
+    for (var nextResult of xresultTags) {
```
adam3smith: again, use simpler loop.

HCSP.js
```diff
@@ -0,0 +1,228 @@
...
+
+/** BEGIN TEST CASES **/
+var testCases = [
+    {
+        "type": "web",
+        "url": "http://www.hcsp.fr/explore.cgi/avisrapportsdomaine?clefr=509",
+        "items": [
+            {
+                "itemType": "report",
+                "title": "Évaluation des besoins des personnes handicapées et  élaboration des réponses au sein des MDPH",
```
adam3smith: looks like this needs cleaning up using ```ZU.trimInternal()``` -- there's a double space between "et" and "elaboration"

adam3smith:
I think that should cover most issues -- might need another round of reviews after you fix these. Let us know if you have any questions, and, of course, thanks for putting this together.

HCSP.js
aurimasv: That is actually a RegExp (strings are converted to RegExp in `match`), but explore.cgi part, and even both "avisrapports..." parts should simply be moved to the target RegExp above.

Also, iirc, we do case insensitive matching with the target RegExp

HCSP.js
```diff
@@ -0,0 +1,228 @@
...
+    newItem.place = "Paris";
+
+    // Creator is the institution (HCSP)
+    // Names of the author are inside the PDFs (if there is one)
+    // TODO: Find a way to have a single field instead of two
+    var creator = ZU.cleanAuthor('HCSP', 'author', false);
+    newItem.creators.push(creator);
+
+    /*---------- URL ----------*/
+    newItem.url = doc.location.href;
```
aurimasv: Use url passed to `scrape`as the argument

HCSP.js
```diff
@@ -0,0 +1,228 @@
...
+    for (var nextP of docDonnees) {
+        var pContent = nextP.textContent;
+
+        if (pContent.match("Date du document")){
+            var date = pContent.match("[0-9]{2}/[0-9]{2}/[0-9]{4}");
+            break;
+        }
+    }
+    
+    newItem.date = String(date);
```
aurimasv: (1) while this works (sometimes), please keep the scope of the variables sane. i.e. above, `date`should be declared outside the `for` loop.
(2) because of the for loop and the if statement, `date` could remain `undefined` (and because of (1), undeclared, which iirc throws an error), do you need to check to make sure. Or simply assign `newItem.date` inside the `if`
(3) you're converting an array of strings to string, which probably works out somehow, but you should just retrieve the first (only) element of the array

HCSP.js
zuphilip: This means AFAIK
```
^https?://www\.hcsp\.fr/
```

HCSP.js
```diff
@@ -0,0 +1,228 @@
...
+                "itemType": "report",
+                "title": "Évaluation des besoins des personnes handicapées et  élaboration des réponses au sein des MDPH",
+                "creators": [
+                    {
+                        "firstName": "",
+                        "lastName": "HCSP",
+                        "creatorType": "author"
+                    }
+                ],
+                "date": "11/05/2015",
```
aurimasv: Ambiguous date. Since you're using a RegExp to extract it anyway, please convert it to `YYYY-MM-DD` format

HCSP.js
jomuller: Fixed. Thank you!

HCSP.js
jomuller: The `Explore.cgi` is not necessary I think. I will add it to the _target_ of the header. This way, no needs to keep a regex and I can use _.indexOf_ if it's more efficient. What do you think?

adam3smith: yes, perfect

HCSP.js
jomuller: Fixed

HCSP.js
jomuller: I'm not used with this syntax. Is this a trik to use named argument in javascript? By the way thank you for the advice! Fixed

HCSP.js
jomuller: I used the argument `namespace` because I did't find how to use named argument in javascript and I wanted to access to the last argument (the separator). But there must be a more elegant solution.

HCSP.js
adam3smith: so ```ZU.cleanAuthor()``` just generates a Zotero creator object -- the same one you can see down in the translator test -- from a string (and some optional arguments like the creator type). But instead, you can just write that object out yourself, which is what I suggested to you here. Among other things, that allows you to just specify the fieldMode rather than adding it later on.

HCSP.js
jomuller: Thank you for yours advises. I followed all + turn the date to the ISO 8601 format.

HCSP.js
```diff
@@ -0,0 +1,231 @@
...
+    // Names of the author are inside the PDFs (if there is one)
+    newItem.creators.push({"lastName": "HCSP", "creatorType": "author", "fieldMode": 1});
+
+    /*---------- URL ----------*/
+    newItem.url = url;
+    
+    /*---------- Title ----------*/
+    var xpathTitle = '//li[@class="active"]';
+    var title = ZU.xpathText(doc, xpathTitle);
+    newItem.title = title;
```
adam3smith: could you still use ```ZU.trimInternal()``` on the title to prevent double spaces (see below).

HCSP.js
adam3smith: this is still not fixed.

HCSP.js
```diff
@@ -0,0 +1,231 @@
...
+            mimeType: "application/pdf"
+        }
+
+        newItem.attachments.push(one_pdf_attachement);
+    }
+
+    newItem.complete();
+}
+
+/** BEGIN TEST CASES **/
```
adam3smith: update your test when you're done with everything else.

HCSP.js
jomuller: Just working on it :-). There is a little difficulty: there should be 2 PDFs in some cases: the advise and the repport. Then I will search for the text in the `<a>` and then produce "Avis HCSP Full Text PDF" or "Rapport HSCP Full Text PDF". Is it ok?

HCSP.js
```diff
@@ -0,0 +1,231 @@
+{
+    "translatorID": "3e375db7-5f6e-4861-ab8b-3ddb0f4accc9",
+    "label": "HCSP",
+    "creator": "Joris Muller",
+    "target": "^https?//www\.hcsp\.fr/[Ee]xplore.cgi/",
```
aurimasv: Double backslashes if you're escaping inside the string. I'd recommend using Scaffold, which takes care of this. Also, you don't need to worry about case, since we do case-insensitive matching

aurimasv: and the last period is not escaped

HCSP.js
```diff
@@ -0,0 +1,231 @@
...
+    "minVersion": "2.1",
```
aurimasv: This should be 3.0

HCSP.js
adam3smith: that's perfect yes.

HCSP.js
```diff
@@ -0,0 +1,231 @@
...
+    } else if (url.indexOf("avisrapports") != -1) {
+        return 'multiple';
+    }
+}
+
+function doWeb(doc, url) {
+
+    if (detectWeb(doc, url) == "multiple") {
+
+        var allItems = getSearchResults(doc);
```
aurimasv: Since you have it in a function anyway, might as well use it in `detectWeb`. Generally, something like `return getSearchResults(doc) ? 'multiple' : false;`

HCSP.js
```diff
@@ -0,0 +1,231 @@
...
+
+        Zotero.selectItems(allItems, function(items) {
+            if (!items) return true;
+
+            var articlesURLs = [];
+            for (var itemURL in items) {
+                articlesURLs.push(itemURL);
+                }
```
aurimasv: nit: indents

HCSP.js
```diff
@@ -0,0 +1,231 @@
...
+    
+    /*---------- Title ----------*/
+    var xpathTitle = '//li[@class="active"]';
+    var title = ZU.xpathText(doc, xpathTitle);
+    newItem.title = title;
+ 
+    /*---------- Date ----------*/
+    // Date is not easy to find because it is not in a named bloc. It is
+    // somewhere in one of the <p> in the div 'doc-donnees'
+    // Then to find it, use a Regex on each <p> bloc
```
aurimasv: Since you're not doing anything complicated with your regexp, you can actually incorporate it into your xpath:
```
var date = ZU.xpath(doc, "//div[@id='doc-donnees']/p[contains(text(),'Date du document')][0]");
if (date) {
  date = date.textContent.match(/\b(\d{2})\/(\d{2})\/(\d{4})\b/); // note the \b
  if (date) {
    newItem.date = date[3] + "-" + date[2] + "-" + date[1];
  }
}

HCSP.js
jomuller: Thank you for this precision. Fixed.

HCSP.js
jomuller: I followed your proposition and used Scaffold. Fixed

HCSP.js
jomuller: Fixed

HCSP.js
jomuller: Good idea. Added

HCSP.js
jomuller: A little bit harder to read for me (then harder to maintain). Does this improve performances?

jomuller:
Thank you very much for your reviews all! I learned a lot. I hope now the translator to be closer to the quality standards of Zotero. I tried to add all your suggestions and added two test with corner cases.

adam3smith:
nothing more from my side, thanks! @aurimasv -- anything from you?

aurimasv:
All looks ok. Could you squash the commits?

adam3smith:
let us know if you need help with rebase&squash. There are instructions e.g. here:
http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

jomuller:
I'm a little bit lost in this process. I never did it before, and I'm not sure to understand the goal. Have I to merge all my commits into a single one? Is that right?

aurimasv:
Yes. If you don't want to spend time figuring it out, we can actually do it
during the merge. So up to you

jomuller:
Thank you. I will try to do it. Another occasion to learn something about GIT!

jomuller:
I tried to rebase and squash. First time I did it. Is it what you expected?

aurimasv:
Yeah, that's great. One not though. Could you amend that commit and clean up the commit message? It sound be short and simple (pretty much your first line) http://stackoverflow.com/questions/179123/edit-an-incorrect-commit-message-in-git

aurimasv:
Or... Not

adam3smith:
oops, sorry, too late.

jomuller:
May I fix it on my side or is it really to late ? Whatever, thank you to merge it!

aurimasv:
Yeah, once it's merged into the master branch, it's permanent (for the most part). Thanks for working on this.

adam3smith:
I mean - if it's important you can always create another pull request.

jomuller:
May I fix it on my side by changing the commit message or is it really to late?

adam3smith:
no, not really -- we can technically mess with the commit history in master the same way you did to squash your commits, but I don't think we'd want to do that and you wouldn't be able to do it at all.

PR #931, opened 2015-08-09
be better about bylines

adam3smith:
See the added test. There might be a better way of doing this, but the current translators mangles this -- presumably not very unusual -- byline format horribly.

### DIFF
--- a/Embedded Metadata.js
+++ b/Embedded Metadata.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2015-07-12 15:17:36"
+	"lastUpdated": "2015-08-09 05:27:45"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


Embedded Metadata.js
```diff
@@ -680,11 +680,26 @@ function getAuthorFromByline(doc, newItem) {
...
 	if(actualByline) {
+		//Z.debug(actualByline.innerHTML)
 		byline = ZU.trimInternal(actualByline.textContent);
-		Z.debug("Extracting author(s) from byline: " + byline);
-		byline = byline.split(/\bby[:\s]+/i);
-		byline = byline[byline.length-1].replace(/\s*[[(].+?[)\]]\s*/g, '');
-		var authors = byline.split(/\s*(?:(?:,\s*)?and|,|&)\s*/i);
+		var authors = [];
+		//make sure we're separating multiple authors on same line in byline
+		if (ZU.xpath(actualByline, './/*[string-length(text())>1]').length > 2){
```
aurimasv: wouldn't this be a problem if, say first name and last name for each author are styled differently? (E.g. Last name in all-caps) this would require name parts to be in separate elements and that shouldn't work with this code 

IIRC, the original reasoning was that if there are going to be multiple authors per byline, they would be separated somehow (comma or and). I didn't take into account implicit separators via li elements. I think we should use that structure for splitting up authors, since that also makes sense semantically, instead of something on random elements.

adam3smith: so you mean look for and split at ```<li>``` elements specifically?

aurimasv: Well, not split per se, but we can see if the byline contains li elements and iterate through those.

adam3smith: yeah, that's what I meant. I'll do that.

PR #934, opened 2015-08-12
Changes to the TEI Translator

FranziskaHorn:
The changes refer to the export of the Zotero key, the export 
of the short title and the tei-conform export of formatting in titles.

Here is a more detailed description of the changes:

1. the export of the Zotero key
If you enable "Generate XML IDs", an XML ID is generated for
each entry. The current translator uses the author's last name
and the date of publication to generate the XML ID (e.g.
xml:id="McDermott1934"). These values are not stable, they can
change for some reason and a different XML ID can be generated
with the next export (e.g. xml:id="McDermott1934a"). 

In some cases, a unique and stable value for an item is needed.
We propose to export the unique item key for each entry.

The item key does not replace the XML ID. Zotero item keys can
start with a number but valid values for the attribute @xml:id
have to begin with a letter. Therefore we propose to export the
item key in addition to the XML ID.

2. export of short title
Content of the field "Short Title" is now exported as the following
<title type="short">Short Title</title>.

3. tei-conform export of formatting in titles
Zotero's tags for formatting (e.g. <i></i> and <sup></sup>) are
replaced by the TEI tags (e.g. <hi rend="italics"></hi> and
<hi rend="sup"></hi>). For example, the text <i>italic text</i>
is exported as follows: &lt;hi rend="italics&gt;italic text&lt;/hi&gt;

Unfortunately, the less-than sign and the greater-than sign are shown
as entities similar to the export result of the current TEI translator
(&lt;i&gt;italic text&lt;/i&gt;). Up to now it was not possible for
us to solve this.

Here is an example of an exported item:

                [...]
                 <biblStruct type="conferencePaper" xml:id="Sundström2009" corresp="IZECEB9P">
                    <analytic>
                        <title level="a">&lt;hi rend="italics"&gt;Rodskarl&lt;/hi&gt;,
                            &lt;i&gt;Trynta&lt;/i&gt; and &lt;i&gt;Spænneklo&lt;/i&gt;. Bynames in
                            the Town Court Record Book of Arboga from a Name-Semantic Point of
                            View</title>
                        <title type="short">Sundström 2009 – &lt;i&gt;Rodskarl&lt;/i&gt;,
                            &lt;i&gt;Trynta&lt;/i&gt; and &lt;i&gt;Spænneklo&lt;/i&gt;</title>
                        <author>
                            <forename>Agneta</forename>
                            <surname>Sundström</surname>
                        </author>
                    </analytic>
                    <monogr>
                        <title level="m">Names in Multi-Lingual, Multi-Cultural and Multi-Ethnic
                            Contact</title>
                        <editor>
                            <forename>Wolfgang</forename>
                            <surname>Ahrens</surname>
                        </editor>
                        [....]
                    </monogr>
                </biblStruct>
               [...]

Of course, we also changed some metadata (label, translatorID etc.). The basis
is the TEI translator by Stefan Majewski (last update 2015-07-20 06:45:00)

### DIFF
--- a/TEI.js
+++ b/TEI.js
@@ -19,7 +19,7 @@
 		"Full TEI Document": false,
 		"Export Collections": false
 	},
-	"lastUpdated":"2015-09-01 06:39:12"
+	"lastUpdated":"2015-09-22 15:19:00"
 }
 
 // ********************************************************************
@@ -109,6 +109,30 @@ var exportedXMLIds = {};
 var generatedItems = {};
 var allItems = {};
 
+// replace formatting with TEI tags
+function replaceFormatting (title){
+    var titleText = title;
+    // italics
+    titleText = titleText.replace(/<i>/g, '<hi rend="italics">');
+    titleText = titleText.replace(/<\/i>/g, '</hi>');
+    // bold
+    titleText = titleText.replace(/<b>/g, '<hi rend="bold">');
+    titleText = titleText.replace(/<\/b>/g, '</hi>');
+    // subscript
+    titleText = titleText.replace(/<sub>/g, '<hi rend="sub">');
+    titleText = titleText.replace(/<\/sub>/g, '</hi>');
+    // superscript
+    titleText = titleText.replace(/<sup>/g, '<hi rend="sup">');
+    titleText = titleText.replace(/<\/sup>/g, '</hi>');
+    // small caps
+    titleText = titleText.replace(/<span style="font-variant:\s*small-caps;">(.*?)<\/span>/g, '<hi rend="smallcaps">$1</hi>');
+    titleText = titleText.replace(/<sc>/g, '<hi rend="smallcaps">');
+    titleText = titleText.replace(/<\/sc>/g, '</hi>');
+    // no capitalization
+    titleText = titleText.replace(/<span class="nocase">(.*?)<\/span>/g, '<hi rend="nocase">$1</hi>');
+            
+    return titleText;
+}
 
 function genXMLId (item){
     var xmlid = '';
@@ -196,6 +220,8 @@ function generateItem(item, teiDoc) {
             bibl.setAttributeNS(ns.xml, "xml:id", myXmlid);
             exportedXMLIds[myXmlid] = bibl;
         }
+        //create attribute for Zotero item URI
+        bibl.setAttribute("corresp", item.uri);
     }
 
     generatedItems[item.itemID] = bibl;
@@ -214,7 +240,7 @@ function generateItem(item, teiDoc) {
         analyticTitle.setAttribute("level", "a");
         analytic.appendChild(analyticTitle);
         if(item.title){
-            analyticTitle.appendChild(teiDoc.createTextNode(item.title));
+            analyticTitle.appendChild(teiDoc.createTextNode(replaceFormatting(item.title)));
         }
 
         // there should be a publication title!
@@ -226,7 +252,7 @@ function generateItem(item, teiDoc) {
             else{
                 pubTitle.setAttribute("level", "m");
             }
-            pubTitle.appendChild(teiDoc.createTextNode(item.publicationTitle));
+            pubTitle.appendChild(teiDoc.createTextNode(replaceFormatting(item.publicationTitle)));
             monogr.appendChild(pubTitle);
         }
         // nonetheless if the user pleases this has to be possible
@@ -235,26 +261,40 @@ function generateItem(item, teiDoc) {
             pubTitle.setAttribute("level", "m");
             monogr.appendChild(pubTitle);
         }
+        // short title
+        if(item.shortTitle){
+            var shortTitle = teiDoc.createElementNS(ns.tei, "title");
+            shortTitle.setAttribute("type", "short");
+            shortTitle.appendChild(teiDoc.createTextNode(item.shortTitle));
+            analytic.appendChild(shortTitle);		
+        }
     }
     else {
         bibl.appendChild(monogr);
         if(item.title){
             var title = teiDoc.createElementNS(ns.tei, "title");
             title.setAttribute("level", "m");
-            title.appendChild(teiDoc.createTextNode(item.title));
+            title.appendChild(teiDoc.createTextNode(replaceFormatting(item.title)));
             monogr.appendChild(title);
         }
         else if(!item.conferenceName){
             var title = teiDoc.createElementNS(ns.tei, "title");
             monogr.appendChild(title);
         }
+        // short title
+        if(item.shortTitle){
+            var shortTitle = teiDoc.createElementNS(ns.tei, "title");
+            shortTitle.setAttribute("type", "short");
+            shortTitle.appendChild(teiDoc.createTextNode(item.shortTitle));
+            monogr.appendChild(shortTitle);		
+        } 
     }
 
     // add name of conference
     if(item.conferenceName){
         var conferenceName = teiDoc.createElementNS(ns.tei, "title");
         conferenceName.setAttribute("type", "conferenceName");
-        conferenceName.appendChild(teiDoc.createTextNode(item.conferenceName));
+        conferenceName.appendChild(teiDoc.createTextNode(replaceFormatting(item.conferenceName)));
         monogr.appendChild(conferenceName);
     }
 
@@ -267,14 +307,14 @@ function generateItem(item, teiDoc) {
         if(item.series){
             var title = teiDoc.createElementNS(ns.tei, "title");
             title.setAttribute("level", "s");
-            title.appendChild(teiDoc.createTextNode(item.series));
+            title.appendChild(teiDoc.createTextNode(replaceFormatting(item.series)));
             series.appendChild(title);
         }
         if(item.seriesTitle){
             var seriesTitle = teiDoc.createElementNS(ns.tei, "title");
             seriesTitle.setAttribute("level", "s");
             seriesTitle.setAttribute("type", "alternative");
-            seriesTitle.appendChild(teiDoc.createTextNode(item.seriesTitle));
+            seriesTitle.appendChild(teiDoc.createTextNode(replaceFormatting(item.seriesTitle)));
             series.appendChild(seriesTitle);
         }
         if(item.seriesText){


aurimasv:
>Therefore we propose to export the item key in addition to the XML ID.

And you do this via &commat;corresp attribute, but [according to TEI spec](http://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-att.global.linking.html#tei_att.corresp), this is supposed to be a URI. You can actually [retrieve the item URI](https://github.com/zotero/zotero/blob/4.0/test/tests/data/translatorExport.js#L41) by bumping the translator minVersion to 4.0.27. However, the URI (and, I think, item key) is not technically stable, since it will change if item was created in an unsynced library and then synced. After syncing, it should remain stable though (unless you merge items, in which case, only one of the IDs will remain stable).

However, why shouldn't we just use &commat;xml:id? Is there a particular reason to preserve the current ID schema?

>Unfortunately, the less-than sign and the greater-than sign are shown as entities similar to the  export result of the current TEI translator (&lt;i>italic text&lt;/i>). Up to now it was not possible for us to solve this.

This is a bit complicated to solve. I think the easiest way would be to first escape invalid characters to XML entities (i.e. `ZU.htmlSpecialChars(...)`), then replace `&lt;i&gt;` with `<hi>` etc., then insert that text as `.innerHTML` (even though it's an XML document) of the appropriate element.

>Of course, we also changed some metadata (label, translatorID etc.)

But you kept the same file name (and label actually), so you are replacing the current TEI translator. If you are submitting this as an improvement to the current TEI translator, the ID should stay the same. If this is supposed to be a separate translator, then you'd have to convince us that two TEI translators are necessary.

I haven't looked through the actual changes. Once we sort out the conceptual issues, I'll take a look at the technical part.

adam3smith:
@zuphilip and me, who asked @FranziskaHorn to submit this, were certainly assuming this would replace the existing TEI translator.

zuphilip:
Yes. For me the 2) export of short title,and 3) tei-conform export of formatting in titles are clearly an improvement and they should (after the review process) pushed to the official translator. I haven't yet looked closely at 1) export of the Zotero key.

IMO the change of the ID was just a mistake and thus the former creator should also appear in the metadata.

FranziskaHorn:
The changes are meant as an improvement. I will change the metadata including the ID of the translator.

> why shouldn't we just use @xml:id?

Unfortunately, @xml:id has a restriction: valid values have to begin with a letter or an underscore.
A number is not allowed, but some Zotero item keys start with a number. Therefore we proposed to export the item key in addition to the XML ID via @corresp.

> And you do this via @corresp attribute, but according to TEI spec, this is supposed to be a URI. You can actually retrieve the item URI 

Yes, @corresp doesn't fit perfectly according to TEI. It's a broader interpretation. As far as I can see it could be a compromise to use @corresp and the item URI (which ends with the Zotero item key) as you suggested, @aurimasv.

aurimasv:
You also have to consider that two different items in two different libraries (e.g. personal and group) could have the same key, so using a URI, which incorporates the library ID is really the best choice to maintain unique IDs

TEI.js
```diff
@@ -109,6 +109,27 @@ var exportedXMLIds = {};
...
+            titleText = titleText.replace('</b>', '</hi>');
+            // subscript
+            titleText = titleText.replace('<sub>', '<hi rend="sub">');
+            titleText = titleText.replace('</sub>', '</hi>');
+            // superscript
+            titleText = titleText.replace('<sup>', '<hi rend="sup">');
+            titleText = titleText.replace('</sup>', '</hi>');
+            // small caps
+            titleText = titleText.replace('<span style="font-variant:small-caps;">', '<hi rend="smallcaps">');
+            titleText = titleText.replace('</span>', '</hi>');
```
aurimasv: something to consider: not all &lt;/span> may be closing the small-caps span. I think it's ok to ignore that for now.

TEI.js
```diff
@@ -109,6 +109,27 @@ var exportedXMLIds = {};
...
+            titleText = titleText.replace('<b>', '<hi rend="bold">');
+            titleText = titleText.replace('</b>', '</hi>');
+            // subscript
+            titleText = titleText.replace('<sub>', '<hi rend="sub">');
+            titleText = titleText.replace('</sub>', '</hi>');
+            // superscript
+            titleText = titleText.replace('<sup>', '<hi rend="sup">');
+            titleText = titleText.replace('</sup>', '</hi>');
+            // small caps
+            titleText = titleText.replace('<span style="font-variant:small-caps;">', '<hi rend="smallcaps">');
```
aurimasv: this may be a little too strict in terms of possible whitespace. e.g. "font-variant: small-caps;" but I think we would accept this.

TEI.js
```diff
@@ -169,6 +190,12 @@ function genXMLId (item){
     return xmlid;
 }
 
+// create Zotero item URI 
+function genURI (item){
+ var uri = item.uri;
+ return uri;
+ }
+
```
aurimasv: not necessary. use `item.uri` directly.

TEI.js
```diff
@@ -183,7 +210,7 @@ function generateItem(item, teiDoc) {
 
     if(Zotero.getOption("Generate XML IDs")){
         if(!generatedItems[item.itemID]){ 
-            var xmlid =  genXMLId(item);
+            var xmlid = genXMLId(item);
```
aurimasv: nit: undo whitespace change

TEI.js
```diff
@@ -196,6 +223,9 @@ function generateItem(item, teiDoc) {
             bibl.setAttributeNS(ns.xml, "xml:id", myXmlid);
             exportedXMLIds[myXmlid] = bibl;
         }
+        //create attribute for Zotero item URI
+        var itemURI = genURI(item);
+        bibl.setAttribute("corresp", itemURI);
```
aurimasv: you don't need to assign to variables first. `bibl.setAttribute("corresp", item.uri);` would be sufficient.

TEI.js
```diff
@@ -214,6 +244,8 @@ function generateItem(item, teiDoc) {
         analyticTitle.setAttribute("level", "a");
         analytic.appendChild(analyticTitle);
         if(item.title){
+            // replace formatting
+            item.title = replaceFormatting(item.title);
```
aurimasv: same here

TEI.js
```diff
@@ -366,7 +422,6 @@ function generateItem(item, teiDoc) {
         monogr.appendChild(edition);
     }
 
-
```
aurimasv: undo whitespace change

aurimasv:
FYI, you need to let us know when you want us to review changes. We don't get notifications when you push new commits.

In general, please make sure that you use tabs for indents and that everything is indented properly.

>Unfortunately, the less-than sign and the greater-than sign are shown
as entities similar to the export result of the current TEI translator
(<i>italic text</i>). Up to now it was not possible for
us to solve this.

This is still an issue, no? Have you tried my suggestion above? Did it not work?

FranziskaHorn:
The new commit includes your suggestions, @aurimasv's. Furthermore, the replaceFormatting function now replaces more than one occurrence of a matched item. 

The < & > signs are still an issue. We tried @aurimasv's suggestion but the output still had the entities, e.g. &lt;hi rend="italics"&gt;

TEI.js
```diff
@@ -109,6 +109,27 @@ var exportedXMLIds = {};
...
+    titleText = titleText.replace(/<\/b>/g, '</hi>');
+    // subscript
+    titleText = titleText.replace(/<sub>/g, '<hi rend="sub">');
+    titleText = titleText.replace(/<\/sub>/g, '</hi>');
+    // superscript
+    titleText = titleText.replace(/<sup>/g, '<hi rend="sup">');
+    titleText = titleText.replace(/<\/sup>/g, '</hi>');
+    // small caps
+    titleText = titleText.replace(/<span style="font-variant:small-caps;">|<span style="font-variant: small-caps;">/g, '<hi rend="smallcaps">');
+    titleText = titleText.replace(/<\/span>/g, '</hi>');
```
adunning: Two suggestions here:

1. CSL also has a (deprecated?) `<sc>` element that can also be used for small caps.
2. Replacing every `</span>` with `</hi>` is going to cause problems, since e.g. it can also be used for `<span class="nocase">`.

adam3smith: (that's technically Zotero and citeproc-js, not CSL, but yeah, this should deal with ```<sc>``` for small caps, too.)

FranziskaHorn:
So far the function replaceFormatting replaces the tags presented here https://www.zotero.org/support/kb/rich_text_bibliography

Of course, we can update the function in order that it replaces the deprecated tag `<sc>` for small caps, too. Thank you for the hint, @adunning.

As far as I understand the tag `<span class="nocase">` it is used to prevent capitalization. We can update the function replaceFormatting in order that it replaces this `<span>`-Element with `<hi rend="nocase">`. Would that be an appropriate solution?

Please let us know if there are other `<span>`-Elements we should consider.

FranziskaHorn:
The new commit includes an update of the replaceFormatting function, the function now replaces the tag `<sc>` for small caps and `<span class="nocase">`, too.

TEI.js
```diff
@@ -109,6 +109,29 @@ var exportedXMLIds = {};
...
+    titleText = titleText.replace(/<b>/g, '<hi rend="bold">');
+    titleText = titleText.replace(/<\/b>/g, '</hi>');
+    // subscript
+    titleText = titleText.replace(/<sub>/g, '<hi rend="sub">');
+    titleText = titleText.replace(/<\/sub>/g, '</hi>');
+    // superscript
+    titleText = titleText.replace(/<sup>/g, '<hi rend="sup">');
+    titleText = titleText.replace(/<\/sup>/g, '</hi>');
+    // small caps
+    titleText = titleText.replace(/<span style="font-variant:small-caps;">|<span style="font-variant: small-caps;">|<sc>/g, '<hi rend="smallcaps">');
```
aurimasv: let's go with
```
titleText = titleText.replace(/<span style="font-variant:\s*small-caps;">(.*?)<\/span>/g, '<hi rend="smallcaps">$1</hi>');
titleText = titleText.replace(/<sc>/g, '<hi rend="smallcaps">');
titleText = titleText.replace(/<\/sc>/g, '</hi>');
```
to avoid replacing some unrelated `</span>` tags, though that shouldn't be a frequent occurrence.
Same for `nocase` below

TEI.js
```diff
@@ -235,26 +260,40 @@ function generateItem(item, teiDoc) {
...
-            title.appendChild(teiDoc.createTextNode(item.title));
+            title.appendChild(teiDoc.createTextNode(replaceFormatting(item.title)));
             monogr.appendChild(title);
         }
         else if(!item.conferenceName){
             var title = teiDoc.createElementNS(ns.tei, "title");
             monogr.appendChild(title);
         }
+        // short title
+	    if(item.shortTitle){
```
aurimasv: nit: indents

TEI.js
```diff
@@ -235,26 +260,40 @@ function generateItem(item, teiDoc) {
             pubTitle.setAttribute("level", "m");
             monogr.appendChild(pubTitle);
         }
+        // short title
+	    if(item.shortTitle){
```
aurimasv: nit: indent

FranziskaHorn:
The new commit includes your suggestions, @aurimasv's, we updated the replaceFormatting function.

FranziskaHorn:
There were no further questions and change requests in the last two weeks. Is it possible to merge this pull request now? Please let me know if I have to take over the changes made by the closed pull request #944 "Update TEI translator for changes in 2.8.".

aurimasv:
You will have to resolve merge conflicts (by rebasing) and squash all the commits into one (that should make it easier to rebase too, if you squash first) at which point, we can probably merge this. Everything looks technically acceptable, but I don't have any expertise with TEI, so it would be best if someone else merges this.

adunning:
The `replaceFormatting` function is a great addition, and the usage is correct TEI; this will be much appreciated.

FranziskaHorn:
There is a new commit. We have squashed all the commits into one and we did a rebase, too.

adunning:
Since work was started on this, a change was made in TEI on the placement of `<idno>` that is causing a warning when validating the results of this translator against the latest schema. Not your fault; see <http://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-biblStruct.html>. The fix for this is trivial, if you have the time to include it, but if not I will fix it at some point (will do so *after* this is merged to prevent another conflict).

adam3smith:
@adunning -- if you want to, you can submit it as a pull requests to @FranziskaHorn's TEI.js branch of the repository. If she accepts it, the PR here will auto-update. Let me know if you want to go that way (which I'd prefer).

adunning:
Looking at it further, I am realizing that it will take a bit more thought than it first appeared, so it will not be for a another month at least. Basically, the translator does not put `<idno>` in a semantically appropriate place to indicate what part of the bibliographical information it describes, and TEI is now banning that (reasonably enough).

adam3smith:
OK, I'll then give this one last review and merge as appropriate.

adam3smith:
thanks everyone!

PR #938, opened 2015-08-15
Adding scrappers for Haaretz, TheMarker and Ynet

eranroz:
These are major Israeli newspaper websites. 
Those scrappers add very basic details and the creators details of the news.

### DIFF
--- a/Haaretz.js
+++ b/Haaretz.js
@@ -1,0 +1,147 @@
+{
+	"translatorID": "d6f64d96-aa6f-4fd3-816f-bdef842c7088",
+	"label": "Haaretz",
+	"creator": "Eran Rosenthal",
+	"target": "^https?://www\\.haaretz\\.(co\\.il|com)/",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2015-08-17 22:01:10"
+}
+
+/**
+	Copyright (c) 2015 Eran Rosenthal
+	
+	This program is free software: you can redistribute it and/or
+	modify it under the terms of the GNU Affero General Public License
+	as published by the Free Software Foundation, either version 3 of
+	the License, or (at your option) any later version.
+	
+	This program is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+	Affero General Public License for more details.
+	
+	You should have received a copy of the GNU Affero General Public
+	License along with this program. If not, see
+	<http://www.gnu.org/licenses/>.
+*/
+
+function detectWeb(doc, url) {
+	if (ZU.xpathText(doc, '//header//h1')) {
+		return 'newspaperArticle';
+	}
+}
+
+function doWeb(doc, url) {
+	var item = new Zotero.Item('newspaperArticle');
+	item.title = ZU.xpathText(doc, '//header//h1');
+	item.url = url;
+	if(url.indexOf('haaretz.com') != -1) {
+		item.publicationTitle = 'Haaretz';
+		item.language = 'en';
+	} else{
+		item.publicationTitle = 'הארץ';
+		item.language = 'he';
+	}
+
+	var abstract = ZU.xpathText(doc, '//header/p');
+	if (!abstract) abstract = ZU.xpathText(doc, '//meta[@property="og:description"]/@content');
+	item.abstractNote = abstract;
+
+	var authors = ZU.xpath(doc, '//address/a[@rel="author"]');
+	for(var i=0; i<authors.length; i++) {
+		item.creators.push(ZU.cleanAuthor(authors[i].textContent, 'author'));
+	}
+
+	item.date = ZU.xpathText(doc, '//time[@itemprop="datePublished"]/@datetime').split('T')[0];
+	var keywords = ZU.xpathText(doc, '//head/meta[@name="news_keywords"]/@content').split(',');
+	for(var i=0; i<keywords.length; i++) {
+		if(keywords[i].length>0) item.tags.push(keywords[i].trim());
+	}
+	item.complete();
+}
+/** BEGIN TEST CASES **/
+var testCases = [
+	{
+		"type": "web",
+		"url": "http://www.haaretz.com/news/diplomacy-defense/1.671202",
+		"items": [
+			{
+				"itemType": "newspaperArticle",
+				"title": "Islamic Jihad: If Hunger Striker Dies, We'll Respond With Force Against Israel",
+				"creators": [
+					{
+						"firstName": "Jack",
+						"lastName": "Khoury",
+						"creatorType": "author"
+					},
+					{
+						"firstName": "Shirly",
+						"lastName": "Seidler",
+						"creatorType": "author"
+					},
+					{
+						"firstName": "Ido",
+						"lastName": "Efrati",
+						"creatorType": "author"
+					}
+				],
+				"date": "2015-08-14",
+				"abstractNote": "Islamic Jihad says it will no longer be committed to maintaining calm if Mohammed Allaan, who lost consciousness after 60-day hunger strike, dies.",
+				"language": "en",
+				"libraryCatalog": "Haaretz",
+				"publicationTitle": "Haaretz",
+				"shortTitle": "Islamic Jihad",
+				"url": "http://www.haaretz.com/news/diplomacy-defense/1.671202",
+				"attachments": [],
+				"tags": [
+					"Palestinian hunger strike"
+				],
+				"notes": [],
+				"seeAlso": []
+			}
+		]
+	},
+	{
+		"type": "web",
+		"url": "http://www.haaretz.co.il/news/politics/1.2708080",
+		"items": [
+			{
+				"itemType": "newspaperArticle",
+				"title": "פלסטיני דקר חייל ופצע אותו באורח קל בכביש 443 סמוך לבית חורון",
+				"creators": [
+					{
+						"firstName": "גילי",
+						"lastName": "כהן",
+						"creatorType": "author"
+					},
+					{
+						"firstName": "עמירה",
+						"lastName": "הס",
+						"creatorType": "author"
+					}
+				],
+				"date": "2015-08-15",
+				"abstractNote": "כוח צה\"ל שהיה במקום פתח באש לעבר הפלסטיני ופצע אותו באורח קל, והוא נעצר. החייל והדוקר פונו לבית החולים שערי צדק. בתחילת השבוע נדקר באזור צעיר ישראלי נוסף שנפצע בינוני",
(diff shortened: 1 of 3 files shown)


Haaretz.js
```diff
@@ -0,0 +1,43 @@
...
+	item.publicationTitle = 'הארץ';
+	item.url = doc.location.href;
+	item.language = 'he';
+	var abstract = ZU.xpathText(doc, '//header/p');
+	if (!abstract) abstract = ZU.xpathText(doc, '//meta[@property="og:description"]/@content');
+	item.abstractNote = abstract;
+
+	var author = ZU.xpathText(doc, '//address/a[@rel="author"]');
+	if (author) {
+		item.creators.push(author);
```
adam3smith: this won't work. You need
```item.creators.push(ZU.cleanAuthor(author, "author"))``` to first generate a creator object. 

Also, does Haaretz never have multiple authors on an article? I know the English version does.

File: Haaretz.js [hunk 2]
```diff
@@ -0,0 +1,43 @@
+{
+	"translatorID": "d6f64d96-aa6f-4fd3-816f-bdef842c7088",
+	"label": "Haaretz",
+	"creator": "Eran Rosenthal",
+	"target": "http://www\\.haaretz\\.co\\.il/",
```
adam3smith: Talking about the English version: couldn't we make this work for haaretz.com easily, too? Looks like the structure is quite similar.

Ynet.js
```diff
@@ -0,0 +1,36 @@
...
+	item.publicationTitle = 'Ynet';
+	item.url = doc.location.href;
+	item.language = 'he';
+	var abstract = ZU.xpathText(doc, '//div[@class="art_header_sub_title"]');
+	if (!abstract) abstract = ZU.xpathText(doc, '//meta[@property="og:description"]/@content');
+	item.abstractNote = abstract;
+
+	var author = ZU.xpathText(doc, '//div[@class="art_header_footer"]//a');
+	if (author) {
+		item.creators.push(Zotero.Utilities.cleanAuthor(author));
```
adam3smith: closer, but you need to specify the creator type, i.e. ```(author, "author")```

TheMarker.js
```diff
@@ -0,0 +1,48 @@
...
+
+	var abstract = ZU.xpathText(doc, '//header/p');
+	if (!abstract) abstract = ZU.xpathText(doc, '//meta[@property="og:description"]/@content');
+	newItem.abstractNote = abstract;
+	
+	newItem.date = ZU.xpathText(doc, '//time[@itemprop="datePublished"]')
+	
+	var author = ZU.xpathText(doc, '//address/a[@rel="author"]');
+	if (author) {
+		newItem.creators.push(author);
```
adam3smith: same here

adam3smith:
Thanks.

General notes for all 3:
We require a license statement (AGPL3) at the the top, see e.g. https://github.com/zotero/translators/blob/master/sbn.it.js#L15 for an example.

We require translator tests for all three -- you can auto-generate those when using Scaffold, doing them manually is a pain.

We'd also much prefer if you added search results, which shouldn't be too complicated. For that purpose, you'd also want to move the scraping out of doWeb into a separate function (we usually call it scraper) that you can then directly call for multiples from doWeb.

Let us know if any of that is unclear.

Haaretz.js
```diff
@@ -0,0 +1,43 @@
...
+function doWeb(doc, url) {
+	var item = new Zotero.Item('newspaperArticle');
+	
+	var title = ZU.xpathText(doc, '//header/h1');
+	if(!title) {
+		title = ZU.xpathText(doc, '//meta[@property="og:title"]/@content');
+	}
+	item.title = title;
+	item.publicationTitle = 'הארץ';
+	item.url = doc.location.href;
```
adam3smith: no need for this, just use the ```url``` argument passed to doWeb.

TheMarker.js
```diff
@@ -0,0 +1,48 @@
+{
+	"translatorID": "b2d61bb5-5b21-41b7-9c83-1abcbf14639b",
+	"label": "TheMarker",
+	"creator": "Eran Rosenthal",
+	"target": "http://www\\.themarker\\.com/",
+	"minVersion": "1.0",
```
adam3smith: min version should be 3.0 for the ZU.xpath functions

Haaretz.js
adam3smith: Use ^https://``` for efficiency and to make sure this works if they enable https versions of the sites.

eranroz:
Thank you for the good comments. I updated the patch accordingly.

Haaretz.js
```diff
@@ -0,0 +1,137 @@
...
+function doWeb(doc, url) {
+	var item = new Zotero.Item('newspaperArticle');
+	
+	var title = ZU.xpathText(doc, '//header/h1');
+	if(!title) {
+		title = ZU.xpathText(doc, '//meta[@property="og:title"]/@content');
+	}
+	item.title = title;
+	item.url = url;
+	if(/haaretz\.com/.test(url)) {
```
adam3smith: more efficient: 
```if (url.indexOf("haaretz.com")!=-1)```

File: Haaretz.js [hunk 8]
```diff
@@ -0,0 +1,137 @@
...
+	"translatorID": "d6f64d96-aa6f-4fd3-816f-bdef842c7088",
+	"label": "Haaretz",
+	"creator": "Eran Rosenthal",
+	"target": "^https?://www\\.haaretz\\.(co\\.il|com)/",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "g",
```
adam3smith: you can mark this compatible across browsers, too

adam3smith:
Great, a couple of smaller notes still. 
Any way we could convince you to add search results to the extent possible?
@aurimasv -- how do you feel about accepting this w/o multiples?

Haaretz.js
```diff
@@ -0,0 +1,137 @@
...
+	if (ZU.xpathText(doc, '//header/h1')) {
+		return 'newspaperArticle';
+	}
+}
+
+function doWeb(doc, url) {
+	var item = new Zotero.Item('newspaperArticle');
+	
+	var title = ZU.xpathText(doc, '//header/h1');
+	if(!title) {
```
aurimasv: Given the `detectWeb` code, thus would never happen

Haaretz.js
```diff
@@ -0,0 +1,137 @@
...
+					},
+					{
+						"firstName": "Ido",
+						"lastName": "Efrati",
+						"creatorType": "author"
+					}
+				],
+				"abstractNote": "Islamic Jihad says it will no longer be committed to maintaining calm if Mohammed Allaan, who lost consciousness after 60-day hunger strike, dies.",
+				"language": "en",
+				"libraryCatalog": "Haaretz",
```
aurimasv: Date not available?

Haaretz.js
```diff
@@ -0,0 +1,137 @@
...
+				"itemType": "newspaperArticle",
+				"title": "פרשנות || נתניהו מעריך שיפסיד בקונגרס על הסכם הגרעין, ובונה על הרווח מכך",
+				"creators": [
+					{
+						"firstName": "חמי",
+						"lastName": "שלו",
+						"creatorType": "author"
+					}
+				],
+				"date": "Sat Aug 15 17:03:17 IDT 2015",
```
aurimasv: We should clean this up a bit. The double year might screw up Zotero's parsing. I can make more concrete recommendations later

TheMarker.js
```diff
@@ -0,0 +1,103 @@
...
+				"itemType": "magazineArticle",
+				"title": "השופט גרוסקופף מציג: הבובות של נוחי דנקנר",
+				"creators": [
+					{
+						"firstName": "גיא",
+						"lastName": "רולניק",
+						"creatorType": "author"
+					}
+				],
+				"date": "15.08.2015 13:23",
```
aurimasv: Should convert to YYYY-MM-DD format

TheMarker.js
```diff
@@ -0,0 +1,103 @@
...
+	
+	var authors = ZU.xpath(doc, '//address/a[@rel="author"]');
+	for(var i=0; i<authors.length; i++) {
+		newItem.creators.push(ZU.cleanAuthor(authors[i].textContent, 'author'));
+	}
+
+	newItem.attachments = [{
+		document: doc,
+		title: "TheMarker",
+		mimeType: "text/html"
```
aurimasv: mimeType not necessary with document

Ynet.js
```diff
@@ -0,0 +1,86 @@
...
+				"itemType": "newspaperArticle",
+				"title": "תעלומת הקצין מארה\"ב, הסודות והמאהבת",
+				"creators": [
+					{
+						"firstName": "ירון",
+						"lastName": "דרוקמן",
+						"creatorType": "author"
+					}
+				],
+				"date": "08:14 , 15.08.15",
```
aurimasv: Clean this up too

TheMarker.js
eranroz: is YYYY-MM-DDTHH:mm:ss is OK?

adam3smith: it works, but once you have the date in that format, you might as well get rid of the timestamp, too, since Zotero doesn't do anything with it.

Haaretz.js
```diff
@@ -0,0 +1,137 @@
...
+					}
+				],
+				"date": "2015-08-15",
+				"abstractNote": "כוח צה\"ל שהיה במקום פתח באש לעבר הפלסטיני ופצע אותו באורח קל, והוא נעצר. החייל והדוקר פונו לבית החולים שערי צדק. בתחילת השבוע נדקר באזור צעיר ישראלי נוסף שנפצע בינוני",
+				"language": "he",
+				"libraryCatalog": "Haaretz",
+				"publicationTitle": "הארץ",
+				"url": "http://www.haaretz.co.il/news/politics/1.2708080",
+				"attachments": [],
+				"tags": [],
```
aurimasv: you can grab keywords from `//head/meta[@name="news_keywords"]/@content`

TheMarker.js
```diff
@@ -0,0 +1,100 @@
...
+				"publicationTitle": "TheMarker",
+				"shortTitle": "השופט גרוסקופף מציג",
+				"url": "http://www.themarker.com/markerweek/thisweek/1.2707370",
+				"attachments": [
+					{
+						"title": "TheMarker",
+						"mimeType": "text/html"
+					}
+				],
+				"tags": [],
```
aurimasv: same here

aurimasv:
Would be great to grab keywords (not for the last translator, since those keywords are very generic). Other than that, looks good. Squash after you're done and we'll merge (e.g. [here](http://stackoverflow.com/questions/14534397/squash-all-my-commits-into-one-for-github-pull-request) or [here](http://blog.steveklabnik.com/posts/2012-11-08-how-to-squash-commits-in-a-github-pull-request))

Thanks!

P.S. You should post here when you add new commits or squash/rebase commits. We don't get notifications when this happens.

eranroz:
1. Added keywords/tags
2. squash

adam3smith:
excellent, thank you!

PR #941, opened 2015-09-01
Discriminates phdthesis and mastersthesis BibTeX types.

cgranade:
This PR discriminates between the `@phdthesis` and `@mastersthesis` BibTeX types using the type field of Zotero items of type `thesis`. See https://forums.zotero.org/discussion/6276/support-for-masters-thesis-type-and-full-bibtex-format/ for a discussion of the pros and cons to this approach.

### DIFF
--- a/BibTeX.js
+++ b/BibTeX.js
@@ -1173,6 +1173,19 @@ function doExport() {
 		// determine type
 		var type = zotero2bibtexTypeMap[item.itemType];
 		if (typeof(type) == "function") { type = type(item); }
+
+		// The thesis Zotero type doesn't nicely map onto a BibTeX type,
+		// as BibTeX uses both @mastersthesis and @phdthesis. We thus have
+		// to tell them apart by the "Type" field as well as by the item type.
+		if (type == "phdthesis") {
+			// In practice, we just want to separate out masters theses,
+			// and will assume everything else maps to @phdthesis. Better to
+			// err on the side of caution.
+			if (item["type"] && item["type"].toLowerCase() == "masters") {
+				type = "mastersthesis";
+			}
+		}
+
 		if(!type) type = "misc";
 		
 		// create a unique citation key
--- a/BibTeX.js
+++ b/BibTeX.js
@@ -1183,6 +1183,7 @@ function doExport() {
 			// err on the side of caution.
 			if (item["type"] && item["type"].toLowerCase() == "masters") {
 				type = "mastersthesis";
+				item["type"] = "";
 			}
 		}
 


BibTeX.js
```diff
@@ -1173,6 +1173,20 @@ function doExport() {
...
 		if (typeof(type) == "function") { type = type(item); }
+
+		// The thesis Zotero type doesn't nicely map onto a BibTeX type,
+		// as BibTeX uses both @mastersthesis and @phdthesis. We thus have
+		// to tell them apart by the "Type" field as well as by the item type.
+		if (type == "phdthesis") {
+			// In practice, we just want to separate out masters theses,
+			// and will assume everything else maps to @phdthesis. Better to
+			// err on the side of caution.
+			if (item["type"] && item["type"].toLowerCase() == "masters") {
```
adam3smith: I think we can be a little more generous here, no? I'm thinking ```.indexOf("master")!=-1```
@karnesky -- any objections here?

aurimasv: I wouldn't go that far, but maybe something like
```
var type = item.type && item.type.toLowerCase().replace(/[\s.]+|thesis/g, '');
if (type &&  (type == 'master' || type == 'masters' || type == 'ms')) {
```

adam3smith: we'd also want "Master's" covered; also MA and MSc are frequently used. May want to test for those in uppercase, though, to be safer. 
What's the objection to testing against just "master"?  (and then also common abbreviations) I can't really see a case where that wouldn't be a master's thesis.

aurimasv: I'd rather test for something specific, that's all. Much more predictable behavior. We can easily cover what you suggested above.

karnesky: This looks good to me too.  If we're going to be very specific, we should ensure BibTeX imports gets the proper item.type populated by whatever we're trying to prescribe.

I have no idea what people tend to use for the "Type" field in Zotero.  Is there a way to survey usage to make sure that whatever pattern we select is both sane and comprehensive?

cgranade: One way to survey might be to check what existing translators do on common
thesis repositories, such as those based on MIT's dspace.

aurimasv: Existing behavior probably varies quite a lot, so I wouldn't consider it. We could ask @dstillman to run a query on the synced user data for that field and give us a list of top N results. I'm sure he's busy with other things though, so let's just go with best effort.

adam3smith:
closed in https://github.com/zotero/translators/commit/71a55e808983d1f6007fbd550f032265c9423042
thanks!

PR #943, opened 2015-09-01
strip extraneous title info for Ovid.

adam3smith:
cf. https://forums.zotero.org/discussion/51255/ovid-could-not-save-item-zotero-firefox/#Item_14

### DIFF
--- a/Ovid.js
+++ b/Ovid.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcs",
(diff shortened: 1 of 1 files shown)


adam3smith:
@aurimasv doing this as a PR since I can't test, though should be unproblematic.

aurimasv:
I can't reproduce the reported issue, but I also can't access PsycINFO through OVID. Other databases seem to be fine (though some have things like `[Review]` at the end, which... I _think_ we should be keeping)

PR #947, opened 2015-09-20
Fixed OZON.ru translator (actually, rewritten from scratch).

roveo:
Hi there! OZON.ru translator wasn't working, so I took some time to rewrite it.

The only functionality missing from my version compared to the previous one (when it worked anyway) is parsing the edition numbers, which I do have plans to add. New functionality is volume number extraction and overall more readable and maintainable code. Will add test cases soon.

I had some doubts regarding the filename and translator GUID, because, well, the code is completely different, so should the GUID stay the same? Does it identify a translator or a resource? Anyway I can put the old one into the code in a couple of seconds if needed, just tell me.

### DIFF
--- a/OZON.ru.js
+++ b/OZON.ru.js
@@ -1,396 +1,151 @@
 {
 	"translatorID": "73be930f-5773-41b2-a7a1-37c0eeade92f",
 	"label": "OZON.ru",
-	"creator": "Victor Rybynok",
-	"target": "^https?://www\\.ozon\\.ru",
+	"creator": "Mikhail Akimov",
+	"target": "^https?://www\\.ozon\\.ru/context/detail/id/",
 	"minVersion": "2.1.9",
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2014-04-03 18:47:00"
+	"lastUpdated": "2015-10-07 13:03:19"
 }
 
-/**
-	OZON.ru translator. Only supports "Book" "Item Type".
-	Copyright (c) 2012 Victor Rybynok, v.rybynok@gmail.com
+/*
+	OZON.ru Translator
+	Copyright (C) 2015 Mikhail Akimov
 
-	Description
-	===========
-
-	Zotero translator for OZON.ru on-line store.
-
-	Comments
-	========
-
-	This version of OZON.ru translator only supports "Book" "Item Type"
-	for "Книги" and "Цифровые книги" ozon.ru site sections.
-
-	Change log
-	==========
-
-	2013-05-27
-
-	* New page layout parsing.
-
-	2012-08-11
-
-	* Corrected detail description parsing to accommodate the new book
-	  description pages template.
-	* Corrected "edition number" recognition.
-	* Introduced change log and to do comment sections.
-	* Few minor issue fixes.
-
-	2012-04-02
-
-	* Initial release.
-
-	TODO
-	====
-
-	* Develop more universal algorithm for "edition number" recognition.
-	  Currently it only recognises up to 10 editions in Russian verbal
-	  representation.
-
-	Legal notice
-	============
-
-	This program is free software: you can redistribute it and/or
-	modify it under the terms of the GNU Affero General Public License
-	as published by the Free Software Foundation, either version 3 of
-	the License, or (at your option) any later version.
+	This program is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
 
 	This program is distributed in the hope that it will be useful,
 	but WITHOUT ANY WARRANTY; without even the implied warranty of
-	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
-	Affero General Public License for more details.
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU Affero General Public License for more details.
 
-	You should have received a copy of the GNU Affero General Public
-	License along with this program.  If not, see
-	<http://www.gnu.org/licenses/>.
+	You should have received a copy of the GNU Affero General Public License
+	along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
 function detectWeb(doc, url) {
-	var itemTypeXPath = '//ul[@class="navLine"]/li[1]/a';
-	var itemTypeDOMNode =
-	doc.evaluate(itemTypeXPath, doc, null,
-	XPathResult.ANY_TYPE, null).iterateNext();
-	var itemType;
-	if (itemTypeDOMNode) itemType = itemTypeDOMNode.textContent;
-	else return "";
-	if (itemType == "Книги") {
-		var nameXPath = '//div[@class="techDescription"]/div/div[2]/span';
-		var valueXPath = '//div[@class="techDescription"]/div/div[3]/span';
-		var nameXPathRes = doc.evaluate(
-		nameXPath, doc, null, XPathResult.ANY_TYPE, null);
-		var valueXPathRes = doc.evaluate(
-		valueXPath, doc, null, XPathResult.ANY_TYPE, null);
-		if (!IsAudioBook(nameXPathRes, valueXPathRes)) return "book";
-	} else {
-		itemTypeXPath = '//ul[@class="navLine"]/li[2]/a';
-		itemTypeDOMNode =
-		doc.evaluate(itemTypeXPath, doc, null,
-		XPathResult.ANY_TYPE, null).iterateNext();
-		if (itemTypeDOMNode) itemType = itemTypeDOMNode.textContent;
-		else return "";
-		if (itemType == "Цифровые книги") return "book";
-	}
-}
+	var hasIsbn = ZU.xpathText(doc, '//*[@itemprop="isbn"]') !== null;
+	var hasContent = ZU.xpath(doc, '//div[@class="bContentBlock"]').length > 0;
 
-function IsAudioBook(nameXPathRes, valueXPathRes) {
-	var nameDOMNode;
-	var valueDOMNode;
-	while( (nameDOMNode = nameXPathRes.iterateNext()) &&
-	(valueDOMNode = valueXPathRes.iterateNext()) ) {
-		var nameText = CleanText(nameDOMNode.textContent);
-		var valueText = CleanText(valueDOMNode.textContent);
-		if (nameText == "Формат аудиокниги") return true;
-	}
-	return false;
-}
-
-function MergeObjects(obj1, obj2) {
-	var result = new Object();
-	for (var attrName in obj1) {
-		result[attrName] = obj1[attrName];
-	}
-	for (var attrName in obj2) {
-		result[attrName] = obj2[attrName];
-	}
-	return result;
-}
-
-function CleanText(text) {
-	var result = text.replace(/^\s*|\s*$/g, '');
-	result = result.replace(/\s+/g, ' ');
-	return result;
-}
-
-function CleanCreator(creatorStr) {
-	return CleanText(creatorStr);
-}
-
-function GetCreators(creatorsStr, creatorsCategory) {
-	var result = new Array();
-	var creator;
-	if (creatorsStr.match(",")) {
-		// http://www.ozon.ru/context/detail/id/7341440/
-		var authors = creatorsStr.split(",");
-		for (var i in authors) {
-			creator = CleanCreator(authors[i]);
-			creator = Zotero.Utilities.cleanAuthor(creator, creatorsCategory);
-			result.push(creator);
-		}
-	} else {
-		// http://www.ozon.ru/context/detail/id/7586040/
-		creator = CleanCreator(creatorsStr);
-		creator = Zotero.Utilities.cleanAuthor(creator, creatorsCategory);
-		result.push(creator);
-	}
-	return result;
-}
-
-function GetLanguage(languageStr) {
-	switch (languageStr) {
-	case "Русский":
-		return "ru";
-	case "Английский":
-		return "en";
-	default:
-		return languageStr;
-	}
-}
-
-function GetISBN(isbnDateStr) {
-	return isbnDateStr.replace(/;\s.*$/, "");
-}
-
-function GetDate(isbnDateStr) {
-	var dateSlash = isbnDateStr.replace(/^.*;\s|\sг\.$/g, "");
-	var dateArray = dateSlash.split("/");
-	var dateZotero = dateArray.reverse().join("-");
-	return dateZotero;
-}
-
-function GetNumPages(numPagesStr) {
-	var result = numPagesStr.replace(/\sстр\./, "");
-	if (result == "0") return "";
-	return result;
-}
-
-function GetEdition(editionStr) {
-	var numStr;
-
-	// example: 5-е издание, исправленное.
-	// http://www.ozon.ru/context/detail/id/8378725/
-	numStr = editionStr.match(/^\d+-е\sиздание/);
-	if(numStr) return editionStr.replace(/-е\sиздание.*/, "");
-
-	// example: Издание 6-е, исправленное.
-	// http://www.ozon.ru/context/detail/id/92650/
-	numStr = editionStr.match(/^издание\s\d+-е/i);
-	if(numStr) return editionStr.replace(/^издание|-е.*$/ig, "");
-
-	// example: 5-е изд.
-	// http://www.ozon.ru/context/detail/id/4579288/
-	numStr = editionStr.match(/\d+-е\sизд\./);
-	if(numStr) return editionStr.replace(/-е\sизд\..*/, "");
-
-	// example: Издание третье, переработанное и дополненное.
-	// http://www.ozon.ru/context/detail/id/2476706/
-	numStr = editionStr.match(/^издание\s\S+[\.,\,]/i);
-	if(numStr) {
-		numStr = numStr[0].replace(/^издание\s|[\.,\,]$/ig,"");
-		numStr = numStr.toLowerCase();
-		// Zotero.debug("Издание -> " + numStr);
-		switch(numStr) {
-		case "второе":
-			return "2";
-		case "третье":
-			return "3";
-		case "четвертое":
-		case "четвёртое":
-			return "4";
-		case "пятое":
-			return "5";
-		case "шестое":
-			return "6";
-		case "седьмое":
-			return "7";
-		case "восьмое":
-			return "8";
-		case "девятое":
-			return "9";
-		case "десятое":
-			return "10";
-		default:
-			return numStr;
-		}
-	}
-	return "";
-}
-
-function ParseProductDetail(pdXPathRes) {
-	var result = new Object();
-	var creatorsObject = new Object();
-	var pdDOMNode;
-	while (pdDOMNode = pdXPathRes.iterateNext()) {
-		var text = CleanText(pdDOMNode.textContent);
-		var name = text.replace(/[:,\s].*$/, "");
-		var value = text.replace(/^\S*\s+/, "");
-		//Zotero.debug(name + " -> " + value);
-		switch (name) {
-		case "Автор":
-		case "Авторы":
-			creatorsObject["authors"] = GetCreators(value, "author");
-			break;
-		case "Составитель":
-			creatorsObject["compilers"] = GetCreators(value, "editor");
-			break;
-		case "Переводчик":
-		case "Переводчики":
-			creatorsObject["translators"] = GetCreators(value, "translator");
-			break;
-		case "Редактор":
-		case "Редакторы":
-			creatorsObject["editors"] = GetCreators(value, "editor");
-			break;
-		case "Язык":
-		case "Языки":
-			result["language"] = GetLanguage(value);
-			break;
-		case "Издательство":
-			result["publisher"] = value;
-			break;
-		case "Серия":
-			result["series"] = value;
-			break;
-		case "ISBN":
-			result["ISBN"] = GetISBN(value);
-			result["date"] = GetDate(value);
-			break;
-		default:
-			//"Naked" year
-			if (name.match(/^\d\d\d\d$/) && !result["date"]) {
-				result["date"] = name;
-			}
-		}
-	}
-	creatorsObject["all"] = new Array();
-	if(creatorsObject["authors"]) {
-		creatorsObject["all"] =
-		creatorsObject["all"].concat(creatorsObject["authors"]);
-	}
-	if(creatorsObject["compilers"]) {
-		creatorsObject["all"] =
-		creatorsObject["all"].concat(creatorsObject["compilers"]);
-	}
-	if(creatorsObject["translators"]) {
-		creatorsObject["all"] =
-		creatorsObject["all"].concat(creatorsObject["translators"]);
-	}
-	if(creatorsObject["editors"]) {
-		creatorsObject["all"] =
-		creatorsObject["all"].concat(creatorsObject["editors"]);
-	}
-	if(creatorsObject["all"]) {
-		result["creators"] = creatorsObject["all"];
-	}
-	return result;
-}
-
-function ParseTechDesc(nameXPathRes, valueXPathRes) {
-	var result = Object();
-	var nameDOMNode;
-	var valueDOMNode;
-	while( (nameDOMNode = nameXPathRes.iterateNext()) &&
-	(valueDOMNode = valueXPathRes.iterateNext()) ) {
-		var nameText = CleanText(nameDOMNode.textContent);
-		var valueText = CleanText(valueDOMNode.textContent);
-		if (nameText == "Страниц") {
-			result["numPages"] = GetNumPages(valueText);
-			return result;
-		}
-	}
-}
-
-function ParseDetailDesc(nameXPathRes, valueXPathRes) {
-	var result = Object();
-	var nameDOMNode;
-	var valueDOMNode;
-	while( (nameDOMNode = nameXPathRes.iterateNext()) &&
-	(valueDOMNode = valueXPathRes.iterateNext()) ) {
-		var nameText = CleanText(nameDOMNode.textContent);
-		var valueText;
-		switch(nameText) {
-		case "От производителя":
-			valueText =
-			valueDOMNode.innerHTML.replace(/^[\s\S]*ANNOTATION\] -->/, "");
-			//Zotero.debug("valueText -> " + valueText);
-			valueText = Zotero.Utilities.cleanTags(valueText);
-			result["abstractNote"] = valueText;
-			break;
-		case "От OZON.ru":
-			valueText =
-			valueDOMNode.textContent.replace(/^[\s\S]*От OZON.ru/, "");
-			valueText = CleanText(valueText);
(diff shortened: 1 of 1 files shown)


OZON.ru.js
```diff
@@ -9,596 +9,130 @@
...
 
-/**
-	OZON.ru translator. Only supports "Book" "Item Type".
-	Copyright (c) 2012 Victor Rybynok, v.rybynok@gmail.com
 
-	Description
-	===========
+function detectWeb(document, url) {
+	var URLRegexp = /https?:\/\/www\.ozon\.ru\/context\/detail\/id\/\d+\//;
+    var isURLValid = URLRegexp.test(url);
```
aurimasv: * please pass the translator through http://jsbeautifier.org/ (tab for indent)
* if you're not planning to re-use a particular variable, it's probably better to not create it. e.g. `/https?:\/\/www\.ozon\.ru\/context\/detail\/id\/\d+\//.test(url)`
* variable names should start in lower case (unless they reference a class) and should follow camelCase conventions
* In this case, you want to place the URL filter regexp in the `target` property at the top of the translator. `detectWeb` is only called if the url passes that test.

OZON.ru.js
```diff
@@ -9,596 +9,130 @@
...
+    var hasISBN = xpath('//*[@itemprop="isbn"]/text()', document).iterateNext() !== null;
```
aurimasv: Use `ZU.xpathText(doc, '//*[@itemprop="isbn"]`)`, which returns `null` if there is no xpath match

OZON.ru.js
```diff
@@ -9,596 +9,130 @@
...
 
-	2013-05-27
+	var publisher = xpath('.//*[@itemprop="publisher"]/a/text()', content).iterateNext();
+	if (publisher !== null)
+		book.publisher = publisher.textContent;
 
-	* New page layout parsing.
+	var numPages = xpath('.//*[@itemprop="numberOfPages"]/text()', content).iterateNext();
+	if (numPages !== null)
+		book.numPages = numPages.textContent;
```
aurimasv: wrap in braces if placing block on the next line. I.e. 
```
if (test) doThis();
```
but
```
if (test) {
  doThis();
}
```

OZON.ru.js
```diff
@@ -9,596 +9,130 @@
...
 
-	2012-08-11
+	var isbn = xpath('.//*[@itemprop="isbn"]/text()', content).iterateNext();
+	if (isbn !== null) {
+		var isbnMatch = /[\d-]+/.exec(isbn);
```
aurimasv: reference `.textContent` explicitly

OZON.ru.js
```diff
@@ -9,596 +9,130 @@
...
-	Legal notice
-	============
+	book.creators = [];
 
-	This program is free software: you can redistribute it and/or
-	modify it under the terms of the GNU Affero General Public License
-	as published by the Free Software Foundation, either version 3 of
-	the License, or (at your option) any later version.
+	// authors
+	var authors = iteratorToArray(xpath('.//*[@itemprop="author"]/a/text()', content));
```
aurimasv: `ZU.xpath(doc, xpath)` will do that for you

OZON.ru.js
```diff
@@ -9,596 +9,130 @@
...
+	for (i in authors)
```
aurimasv: Iterate arrays using `for (var i=0; i<array.length; i++)` loops

OZON.ru.js
```diff
@@ -9,596 +9,130 @@
...
+		book.creators.push(createCreator('author', authors[i].textContent));
```
aurimasv: you could also just use `.map`

OZON.ru.js
```diff
@@ -9,596 +9,130 @@
...
-	result = result.replace(/\s+/g, ' ');
-	return result;
+	book.complete();
 }
 
-function CleanCreator(creatorStr) {
-	return CleanText(creatorStr);
+function xpath(expr, context) {
+	// executes standard XPath search on a document
+    return document.evaluate(expr, context, null, XPathResult.ANY_TYPE, null);
```
aurimasv: never refer to `document` directly. Use `doc` that is passed into `doWeb`, `detectWeb`, callback to `ZU.processDocuments`

OZON.ru.js
```diff
@@ -9,596 +9,130 @@
...
-	var valueDOMNode;
-	while( (nameDOMNode = nameXPathRes.iterateNext()) &&
-	(valueDOMNode = valueXPathRes.iterateNext()) ) {
-		var nameText = CleanText(nameDOMNode.textContent);
-		var valueText = CleanText(valueDOMNode.textContent);
-		if (nameText == "Страниц") {
-			result["numPages"] = GetNumPages(valueText);
-			return result;
-		}
+function splitFullName(fullName) {
```
aurimasv: Use [`ZU.cleanAuthor`](https://github.com/zotero/zotero/blob/4.0/chrome/content/zotero/xpcom/utilities.js#L148) instead

OZON.ru.js
```diff
@@ -9,596 +9,130 @@
...
-		"url": linkurl,
-		"title": "OZON.ru Link",
-		"mimeType": "text/html",
-		"snapshot": false
-	}];
-
-	newItem.complete();
-}
-/** BEGIN TEST CASES **/
-var testCases = [
```
aurimasv: Would it be possible to keep tests? I.e. are the URLs still the same? If no, then can we fix the URLs to refer to the same articles?

Also, looks like you're not using [Scaffold](https://www.zotero.org/support/dev/translators/scaffold). Please do. It will make your life easier (though you need Firefox)

OZON.ru.js
```diff
@@ -1,7 +1,7 @@
 {
-	"translatorID": "73be930f-5773-41b2-a7a1-37c0eeade92f",
+	"translatorID": "de242517-746f-494c-a442-ce3ce06edcb3",
```
aurimasv: This should stay the same.

OZON.ru.js
```diff
@@ -9,596 +9,130 @@
...
-	"lastUpdated": "2014-04-03 18:47:00"
+	"lastUpdated": "2015-09-20 18:13:00"
 }
 
-/**
-	OZON.ru translator. Only supports "Book" "Item Type".
-	Copyright (c) 2012 Victor Rybynok, v.rybynok@gmail.com
 
-	Description
-	===========
```
aurimasv: We'd like a license block. You can copy paste from https://github.com/zotero/translators/blob/master/ACM%20Digital%20Library.js#L15 and adjust authorship

OZON.ru.js
roveo: I tried, but couldn't get Scaffold running (Mac OS X).

Thanks a lot for your remarks, will look into them soon.

aurimasv: What happened when you tried?

roveo: It opens, but when I push buttons nothing happens. I tries `Load`, `Generate` for ID, `Test Regex`, running the script etc.

aurimasv: Do you also have Zotero installed in Firefox?

roveo: Yeah, missed that. Using Scaffold now, it's much more convenient, thanks!

OZON.ru.js
roveo: It doesn't work for input like "Henry S. Warren jr." (my function doesn't either). I can submit an issue if needed.

zuphilip:
I guess the two commits correspond to the review from @aurimasv ? Please note there are not automatic notifications if you add new commits.

roveo:
@zuphilip yes they do.

OK, I'll make comments after commits in future.

roveo:
@aurimasv pls see the commited code, I think I the issues you pointed out.

OZON.ru.js
```diff
@@ -1,604 +1,103 @@
 {
 	"translatorID": "73be930f-5773-41b2-a7a1-37c0eeade92f",
 	"label": "OZON.ru",
-	"creator": "Victor Rybynok",
-	"target": "^https?://www\\.ozon\\.ru",
-	"minVersion": "2.1.9",
+	"creator": "Mikhail Akimov",
+	"target": "http://www.ozon.ru/context/detail/id/\\d{3,}",
```
aurimasv: "^https?://www\\.ozon\\.ru/context/detail/id/\\d{3}" This is used for test (not match), so there's no point in matching more than needed, like `{3,}`

OZON.ru.js
```diff
@@ -1,604 +1,103 @@
...
-	"lastUpdated": "2014-04-03 18:47:00"
+	"browserSupport": "gcsi",
+	"lastUpdated": "2015-09-22 10:03:30"
 }
 
-/**
-	OZON.ru translator. Only supports "Book" "Item Type".
-	Copyright (c) 2012 Victor Rybynok, v.rybynok@gmail.com
+/*
+	ACM Digital Library Translator
```
aurimasv: This should be changed to say "OZON.ru translator"

OZON.ru.js
```diff
@@ -1,604 +1,103 @@
...
-	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
-	Affero General Public License for more details.
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+	GNU General Public License for more details.
 
-	You should have received a copy of the GNU Affero General Public
-	License along with this program.  If not, see
-	<http://www.gnu.org/licenses/>.
+	You should have received a copy of the GNU General Public License
+	along with this program. If not, see <http://www.gnu.org/licenses/>.
```
aurimasv: Sorry, i didn't notice the license block initially. The original was correct (though removing the log and to-do list was a good choice)

@adam3smith could you re-license ACM under AGPL?

OZON.ru.js
```diff
@@ -1,604 +1,103 @@
...
 */
 
+var ZU = Zotero.Utilities;
```
aurimasv: That is already set

OZON.ru.js
```diff
@@ -1,604 +1,103 @@
...
-	var result = new Object();
-	for (var attrName in obj1) {
-		result[attrName] = obj1[attrName];
-	}
-	for (var attrName in obj2) {
-		result[attrName] = obj2[attrName];
-	}
-	return result;
-}
+	var content = ZU.xpath(doc, '//div[@class="bContentBlock"]')[0];
```
aurimasv: Is this guaranteed to exist? If no, test for it in detectWeb

OZON.ru.js
```diff
@@ -1,604 +1,103 @@
...
-	switch (languageStr) {
-	case "Русский":
-		return "ru";
-	case "Английский":
-		return "en";
-	default:
-		return languageStr;
+	var isbn = ZU.xpathText(content, './/*[@itemprop="isbn"]');
+	if (isbn !== null) {
+		book.isbn = /[\d\-]+/.exec(isbn)[0];
```
aurimasv: `ZU.cleanISBN` instead

OZON.ru.js
```diff
@@ -1,604 +1,103 @@
...
+		book.date = /\d{4}(?= г.)/.exec(isbn)[0];
 	}
-}
+	// volume (parsed from the title)
+	var volume = /(т(\.|ом)|vol(\.|ume))\s+\d+/i.exec(book.title);
```
aurimasv: Turn the parentheses into non-capturing `(?:...)` and use a capturing group against `\d+`, then you won't need to replace below

aurimasv: But also looks like you may want to clean up the title, by capturing everything that precedes the current match. I.e. add `(.+?)` at the beginning

OZON.ru.js
adam3smith: sure. Can treat this as agreement to license change, I'll change it in ACM when I get a chance.

OZON.ru.js
```diff
@@ -1,604 +1,103 @@
...
-		if (nameText == "Страниц") {
-			result["numPages"] = GetNumPages(valueText);
-			return result;
-		}
-	}
-}
+function getAdditionalProperty(element) {
+	var matches = /[А-я]+(?=:)/.exec(element.textContent);
+	if (matches === null)
+		return undefined;
```
aurimasv: Use braces

aurimasv: And everywhere else in these situations (see above)

OZON.ru.js
```diff
@@ -1,604 +1,103 @@
...
-		var nameText = CleanText(nameDOMNode.textContent);
-		var valueText = CleanText(valueDOMNode.textContent);
-		if (nameText == "Страниц") {
-			result["numPages"] = GetNumPages(valueText);
-			return result;
-		}
-	}
-}
+function getAdditionalProperty(element) {
+	var matches = /[А-я]+(?=:)/.exec(element.textContent);
```
aurimasv: The `[А-я]` is probably not what you want. It will create a range based on unicode codepoints, between A and я, which matches a lot of characters.

OZON.ru.js
```diff
@@ -1,604 +1,103 @@
...
-		"url": linkurl,
-		"title": "OZON.ru Link",
-		"mimeType": "text/html",
-		"snapshot": false
-	}];
-
-	newItem.complete();
 }
-/** BEGIN TEST CASES **/
-var testCases = [
```
aurimasv: Is there really no way to maintain test cases?

OZON.ru.js
avram: But `[а-яА-ЯёЁ]` should be fine, so long as we are OK ignoring the non-Russian and archaic Cyrillic characters, which should be fine for Ozon.

avram: That said, `[А-я]` is in fact a reasonable range of code points, albeit insufficient due to it missing `ё`; the initial letter of the range is `CYRILLIC CAPITAL LETTER A`.

OZON.ru.js
zuphilip: Note there are more non AGPL licences in translators: <a href="https://github.com/zotero/translators/search?utf8=%E2%9C%93&q=%22copy+of+the+GNU+General+Public+License%22&type=Code">111 results</a> . I guess it would be preferable if we manage to change them all?

aurimasv: It's hard to do this retro-actively, since we have to have the original licensor agree to this. It was easier here, since @adam3smith is the licensor. In general, we should organize the licensing better. Not sure if we could do this now, but it would be nice to go with something like https://www.clahub.com/

OZON.ru.js
aurimasv: My mistake, did not put in the effort to check on my phone that the first letter was CYRILLIC CAPITAL A. Thanks for that @avram.

When using non-ASCII characters, particularly in ranges or where ambiguous, we should use Unicode escapes, so `[\u0410-\u044F]` (plus what @avram says above about the missing characters, though they don't matter in this case, since we're only looking at a set number of `itemprop` names)

OZON.ru.js
avram: This is not the best forum to discuss this I suppose -- but I agree to the relicensing of any of my translator work under the Affero GPL.

zuphilip: Okay, we can continue here: https://github.com/zotero/translators/issues/951

OZON.ru.js
roveo: That's complicated, because volume number can be in the middle too, like here: http://www.amazon.com/The-Art-Computer-Programming-Searching/dp/0201896850
So I'll have to replace anyway.

roveo: And in this case volume number is an essential part of the title, there's no reason to clean it up I think.

OZON.ru.js
roveo: > we're only looking at a set number of `itemprop` names
That's not the case here, these are not `itemprop`, these are additional properties that don't have an assigned `itemprop` name, so theoretically they can contain `Ё`.

roveo:
@aurimasv 
 - Added tests (5 old + 3 new)
 - Escaped everything cyrillic
 - Added Ё to cyrillic ranges
 - Fixed style issues

OZON.ru.js
```diff
@@ -1,563 +1,346 @@
...
-		itemTypeDOMNode =
-		doc.evaluate(itemTypeXPath, doc, null,
-		XPathResult.ANY_TYPE, null).iterateNext();
-		if (itemTypeDOMNode) itemType = itemTypeDOMNode.textContent;
-		else return "";
-		if (itemType == "Цифровые книги") return "book";
-	}
-}
+	var hasIsbn = ZU.xpathText(doc, '//*[@itemprop="isbn"]') !== null;
+	var hasContent = ZU.xpath(doc, '//div[@class="bContentBlock"]') !== null;
```
aurimasv: This is always true. `ZU.xpath` always returns an array, though it may be empty.

OZON.ru.js
```diff
@@ -1,563 +1,346 @@
...
-	}
-}
 
-function GetISBN(isbnDateStr) {
-	return isbnDateStr.replace(/;\s.*$/, "");
-}
+	var isbn = ZU.xpathText(content, './/*[@itemprop="isbn"]');
+	if (isbn !== null) {
+		book.ISBN = ZU.cleanISBN(isbn);
+		book.date = /\d{4}(?=\s\u0433\.)/.exec(isbn)[0];
```
aurimasv: Sorry if I mislead you, but I did only mean to ask to escape unicode in character ranges and where a character may be ambiguous (e.g. a single Cyrillic capital `A`). So most of what you had was ok. Should also make a note of the unescaped character.

OZON.ru.js
```diff
@@ -1,563 +1,346 @@
...
-		case "От OZON.ru":
-			valueText =
-			valueDOMNode.textContent.replace(/^[\s\S]*От OZON.ru/, "");
-			valueText = CleanText(valueText);
-			result["edition"] = GetEdition(valueText);
-			break;
-		}
+function getAdditionalProperty(element) {
+	// cyrillic alphabet
+	var matches = /[\u0430-\u044f\u0410-\u044f\u0451\u0401]+(?=:)/.exec(element.textContent);
```
aurimasv: The first range is redundant with the second

OZON.ru.js
```diff
@@ -1,563 +1,346 @@
...
-	}];
 
-	newItem.complete();
-}
 /** BEGIN TEST CASES **/
 var testCases = [
 	{
 		"type": "web",
-		"url": "http://www.ozon.ru/context/detail/id/87889/",
+		"url": "http://www.ozon.ru/context/detail/id/1335648/",
```
aurimasv: These should be restored to the same URLs. This is not the same book.

aurimasv: And please make sure they end up in the same order

aurimasv:
Looks good otherwise. Please release this under AGPLv3 license.

roveo:
@aurimasv OK, fixed, license changes. What about that issue with `ZU.cleanAuthor`? Where can I submit a report?

OZON.ru.js
aurimasv: You can submit this as an issue to zotero/zotero. Thanks!

OZON.ru.js
```diff
@@ -1,396 +1,120 @@
...
-			return "9";
-		case "десятое":
-			return "10";
-		default:
-			return numStr;
+	for (var i = 0; i < additionalProperties.length; i++) {
+		var property = getAdditionalProperty(additionalProperties[i]);
+		if (typeof property === 'undefined') continue;
+		if (/Переводчики?/.test(property.name)) {
+			getMultipleCreators(property.value, 'translator').forEach(function (d) { book.creators.push(d); });
```
aurimasv: nit: use `book.creators = book.creators.concat(getMultipleCreators(....))` instead

OZON.ru.js
```diff
@@ -1,396 +1,120 @@
...
-		case "От OZON.ru":
-			valueText =
-			valueDOMNode.textContent.replace(/^[\s\S]*От OZON.ru/, "");
-			valueText = CleanText(valueText);
-			result["edition"] = GetEdition(valueText);
-			break;
-		}
+function getAdditionalProperty(element) {
+	// cyrillic alphabet
+	var matches = /[\u0430-\u044f\u0410-\u042fёЁ]+(?=:)/.exec(element.textContent);
```
aurimasv: nit: order ranges in order of code point values, though in this case the two ranges are adjacent, so they can be just a single range.

OZON.ru.js
```diff
@@ -1,396 +1,120 @@
...
-	pdObject, ParseDetailDesc(nameXPathRes, valueXPathRes));
-
-	var newItem = new Zotero.Item("book");
-	for (var field in pdObject) newItem[field] = pdObject[field];
+function getMultipleCreators(value, creatorType) {
+	return value
+			.split(',')
+			.map(function (d) {
+				return ZU.cleanAuthor(d, creatorType, false);
+			});
```
aurimasv: nit: indents

OZON.ru.js
```diff
@@ -406,24 +131,18 @@ var testCases = [
...
-					}
-				],
-				"title": "Язык программирования C++",
-				"language": "ru",
-				"publisher": "Бином, Невский Диалект",
-				"ISBN": "5-7940-0031-7, 5-7989-0127-0, 0-201-88954-4",
-				"date": "1999-1-1",
-				"numPages": "991",
+				"date": "1999",
+				"ISBN": "9785794000313",
```
aurimasv: the other ISBNs no longer available on page?

OZON.ru.js
```diff
@@ -406,24 +131,18 @@ var testCases = [
...
-				"libraryCatalog": "OZON.ru"
+				"language": "Русский",
```
aurimasv: Let's at least substitute `Русский` with `ru`. That should probably take care of most cases. If you want to substitute some other common languages in the catalog with appropriate IETF language codes, feel free to do so as well.

OZON.ru.js
```diff
@@ -406,24 +131,18 @@ var testCases = [
...
+				"libraryCatalog": "OZON.ru",
+				"numPages": "991",
+				"publisher": "Бином, Невский Диалект",
+				"url": "http://www.ozon.ru/context/detail/id/87889/",
```
aurimasv: Actually, the URL should point to the website with an available full-text resource (hosted online). We do not include URLs to catalog entries.

OZON.ru.js
```diff
@@ -406,24 +131,18 @@ var testCases = [
...
 				],
-				"notes": [],
-				"tags": [],
-				"seeAlso": [],
-				"attachments": [
-					{
-						"title": "OZON.ru Link",
-						"mimeType": "text/html",
-						"snapshot": false
-					}
```
aurimasv: In absence of a link back to OZON.ru in the URL field, we sometimes attach links to items as per above. If you think it is useful for the user to be able to get back to OZON.ru from within Zotero, you can include that as well.

OZON.ru.js
```diff
@@ -484,119 +199,210 @@ var testCases = [
...
-				"seeAlso": [],
-				"attachments": [
+				"notes": [],
+				"seeAlso": []
+			}
+		]
+	},
+	{
+		"type": "web",
+		"url": "http://www.ozon.ru/context/detail/id/1335648/",
```
aurimasv: Could you please swap this one with the one below?

aurimasv: Or wait... hmmm, not sure why the diff is messed up. Nvm, please ignore

OZON.ru.js
roveo: They are, this is just output of `ZU.cleanISBN` - the first of them. There's no mechanism to store multiple ISBNs for one book, and as far as I know ISBN in Zotero is normally stored as a single integer, so just copying the text contents seemed inappropriate.

OZON.ru.js
roveo: Oh, OK, I thought this should just be source URL.

OZON.ru.js
```diff
@@ -484,119 +199,210 @@ var testCases = [
...
+				"itemType": "book",
+				"title": "Искусство программирования. Том 1. Основные алгоритмы",
+				"creators": [
 					{
-						"title": "OZON.ru Link",
-						"mimeType": "text/html",
-						"snapshot": false
+						"firstName": "Дональд Эрвин",
+						"lastName": "Кнут",
+						"creatorType": "author"
```
aurimasv: If I am matching this up correctly, looks like we lost a translator. Is that expected?

aurimasv: Actually a few translators and an editor, no?

OZON.ru.js
aurimasv: Well, yes, `ZU.cleanISBN` will return a single ISBN. You would have to split them up first, then clean them individually and re-join. Actually, you don't even have to worry about cleaning them up in the translator, Zotero should automatically clean all of them up during import (I don't recall if they have to be generally clean from non-ISBN-related text).You can store multiple ISBNs in the same field separated by space, commas, semicolons, something else. Zotero will clean all of that up and de-duplicate any ISBN-10/ISBN-13s. I don't think we take full advantage of this yet, but a full list of ISBNs could be used for detecting duplicates. Otherwise, we just pass the first ISBN to the citation processor.

OZON.ru.js
roveo: Yepp, that's right. This book uses a different template for some reason, let me look into it.

OZON.ru.js
roveo: Looks OK in Sublime and Scaffold. That's exactly why I don't like using tabs for indentation :)

aurimasv: Not sure why it should look right, since there are certainly two tabs for that indent.

roveo: Sometimes the chained methods are aligned by the dot, like this:

```js
var foo = bar.method(something)
             .anotherMethod(somethingElse)
             .oneMore(whatever);
```

aurimasv: Oh, I see. We don't try to align, just indent once for continuation.

roveo: OK.

roveo:
@aurimasv 
- Language now in standard code (basic list of European languages)
- Support for multiple ISBNs
- Layout-independent additional properties (without @itemprop value)

OZON.ru.js
```diff
@@ -1,396 +1,150 @@
...
-			creator = Zotero.Utilities.cleanAuthor(creator, creatorsCategory);
-			result.push(creator);
+
+	var isbnField = ZU.xpathText(content, './/*[@itemprop="isbn"]');
+	var isbnRe = /[\d\-]{9,}/g;
+	var isbn;
+	book.ISBN = [];
+	if (isbnField !== null) {
+		while ((isbn = isbnRe.exec(isbnField)) !== null) {
+			book.ISBN.push(ZU.cleanISBN(isbn[0]));
```
aurimasv: `book.ISBN` needs to be a string. Convert the array to string afterwards, or use something like `book.ISBN += isbn[0] + ' ';` (again, you don't need to clean anything up, since Zotero will do it later anyway)

aurimasv:
Feel free to squash after that last nit is fixed and we can merge.

roveo:
@aurimasv here we go, fixed and squashed

OZON.ru.js
```diff
@@ -1,396 +1,151 @@
...
 		}
-	} else {
-		// http://www.ozon.ru/context/detail/id/7586040/
-		creator = CleanCreator(creatorsStr);
-		creator = Zotero.Utilities.cleanAuthor(creator, creatorsCategory);
-		result.push(creator);
+		book.ISBN = book.ISBN.join(' ');
```
aurimasv: That's the wrong place to do this. You should do it outside the if statement. (Also, squash all commits into one)

roveo:
@aurimasv done

aurimasv:
Thanks for working on this! Sorry it took so long to review.

PR #955, opened 2015-10-07
[Springer] Work when no EM is present

adam3smith:
Changes look bigger than they are (also ran this through jsbeautify because I moved code around a bit, accounts for some changes).
Essentially does 2 things:
1. If we don't have EM, use the RIS data from Springer
2. Move the scraping from the page to a separate function so we can call it in the item handler of both. 

Also: add DOI for non-journal item types.

### DIFF
--- a/Springer Link.js
+++ b/Springer Link.js
@@ -9,18 +9,16 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsbv",
-	"lastUpdated": "2013-08-13 20:03:26"
+	"lastUpdated": "2015-10-11 19:37:31"
 }
 
 function detectWeb(doc, url) {
 	var action = url.match(/^https?:\/\/[^\/]+\/([^\/?#]+)/);
 	if(!action) return;
-	
 	if(!doc.head || !doc.head.getElementsByTagName('meta').length) {
 		Z.debug("Springer Link: No head or meta tags");
 		return;
 	}
-	
 	switch(action[1]) {
 		case "search":
 		case "journal":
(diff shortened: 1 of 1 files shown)


Springer Link.js
```diff
@@ -31,185 +29,191 @@ function detectWeb(doc, url) {
...
+		translator.setTranslator("951c027d-74ac-47d4-a107-9c3069ab7b48");
+		translator.setDocument(doc);
+		translator.setHandler("itemDone", function(obj, item) {
+			//sometimes we get an error about title not being set
+			if(!item.title) {
+				Z.debug("Springer Link: title not found");
+				Z.debug(item);
+				if(doc.head) {
+					//clean up and strip out uninteresting content
+					Z.debug(doc.head.innerHTML.replace(/<style[^<]+(?:<\/style>|\/>)/ig, '').replace(/<link[^>]+>/ig, '').replace(
```
aurimasv: that's a pretty long line

Springer Link.js
```diff
@@ -31,185 +29,191 @@ function detectWeb(doc, url) {
...
+		item.DOI = ZU.xpathText(doc,
+			'//dd[@id="abstract-about-book-chapter-doi"\
 					or @id="abstract-about-doi"][1]');
+	}
+	if(!item.publisher) {
+		item.publisher = ZU.xpathText(doc, '//dd[@id="abstract-about-publisher"]');
+	}
+	if(!item.date) {
+		item.date = ZU.xpathText(doc, '//dd[@id="abstract-about-cover-date"]') || ZU.xpathText(doc,
+			'//dd[@id="abstract-about-book-chapter-copyright-year"]');
```
aurimasv: nit: this line break is rather awkward (a couple other line breaks are also the same way)

Springer Link.js
```diff
@@ -31,185 +29,191 @@ function detectWeb(doc, url) {
...
+			item.series = ZU.xpathText(doc, '//dd[@id="abstract-about-book-series-title"]');
+		}
+		if(!item.seriesNumber) {
+			item.seriesNumber = ZU.xpathText(doc, '//dd[@id="abstract-about-book-series-volume"]');
+		}
+	}
+	//add the DOI to extra for non journal articles
+	if(item.itemType != "journalArticle" && item.DOI) {
+		item.extra = "DOI: " + item.DOI;
+	}
```
aurimasv: Hmm, I feel like we should do this for all translators.

Springer Link.js
```diff
@@ -330,65 +336,59 @@ var testCases = [
 		"items": [
 			{
 				"itemType": "bookSection",
```
aurimasv: Is this one of the cases where we use RIS data?

aurimasv:
Looks ok to me. Feel free to merge. Just a few formatting nits, no blockers.

Springer Link.js
adam3smith: I agree. You're saying that should be moved into zotero code?

Springer Link.js
adam3smith: I relied on jsbeautify (set to 110 chars max) entirely. I personally think that's good practice because that allows future versions to just be passed through the same beautifier w/o messing up the diff. (also applies to "long line" above). 
I'm OK adjusting the char length (I thought this worked well, though), but I'd rather not edit the automatically chosen linebreaks.

Springer Link.js
adam3smith: yes, the only one in the tests currently. why?

Springer Link.js
aurimasv: Not right now, but maybe we can think about doing it. For now, delete `item.DOI` after moving it to `item.extra` so this doesn't break in the future.

Springer Link.js
aurimasv: just wondering if we had a test case that covers that condition.

Springer Link.js
aurimasv: I think the general consensus is 80 character line length and that's where Scaffold draws the line (IIRC). Hmmm, maybe we can configure a linter to produce nicer code directly in Scaffold. It's ok for now, though I think it's easier to follow when the line breaks are more logical.

adam3smith: OK, I'll switch to 80.

adam3smith:
removed the DOI as discussed above, shortened lines to 80, squashed and merged. Thanks @aurimasv

PR #957, opened 2015-10-13
Fix for PDF Download

symac:
Cairn has changed the way they handle the PDF download by adding an intermediate page that will call the real PDF download. So the embedded metadata translator won't be able to download the PDF as mimetype is not the good one.

This fix add extra processing after the import translator to request the actual PDF.

### DIFF
--- a/CAIRN Info.js
+++ b/CAIRN Info.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2014-03-11 22:36:38"
+	"lastUpdated": "2015-10-15 05:27:29"
 }
 
 /*
@@ -84,6 +84,20 @@ function doWeb(doc,url)
 						tags.push(keywords[i])
 					}
 				}
+
+				// The default value for PDF download is on an HTML page that calls the actual download
+				// we need to add this attachment after the import translator has been running
+				// We remove the attached PDF because they won't be working
+				for(var i=item.attachments.length - 1; i>=0; i--) {
+					if(item.attachments[i].mimeType == 'application/pdf') {
+						item.attachments.splice(i, 1);
+					}
+				}
+				
+				var pdfUrl = ZU.xpathText(doc, '//meta[@name="citation_pdf_url"]/@content');
+				pdfUrlDownload = pdfUrl + "&download=1";
+				item.attachments.push({title : "Full Text PDF", mimeType: "application/pdf", url: pdfUrlDownload});
+				
 				item.complete();
 				});
 		translator.getTranslatorObject(function (obj) {


CAIRN Info.js
```diff
@@ -84,6 +84,22 @@ function doWeb(doc,url)
 						tags.push(keywords[i])
 					}
 				}
+				
+				// The default value for PDF download is on an HTML page that calls the actual download
+				// we need to add this attachment after the import translator has been running
+				// We remove the attached PDF because they won't be working
+				for(var i=0, n=item.attachments.length; i<n; i++) {
```
adam3smith: would we ever have two PDFs? because then we'd want to have the for loop count down from attachments.length when using delete, right?

adam3smith:
you'll also need to bump the timestamp.
With those two nits done, you can squash this into a single commit and we can merge.

CAIRN Info.js
symac: @adam3smith I must admin that I have copied/pasted what is done [in Embedded metadata.js](https://github.com/zotero/translators/blob/master/Embedded%20Metadata.js#L480-L484) without thinking about it, I will update that tonight yes, thanks for the input.

CAIRN Info.js
```diff
@@ -84,6 +84,22 @@ function doWeb(doc,url)
...
+					if(item.attachments[i].mimeType == 'application/pdf') {
+						delete item.attachments[i];
```
aurimasv: EM is wrong here, you should use
```
item.attachments.splice(i, 1);
i--;
```

CAIRN Info.js
```diff
@@ -84,6 +84,22 @@ function doWeb(doc,url)
...
+					}
+				}
+				
+				var pdfUrl = ZU.xpathText(doc, '//meta[@name="citation_pdf_url"]/@content');
+				if (pdfUrl) {
+					pdfUrlDownload = pdfUrl + "&download=1";
+					item.attachments = [{title : "Full Text PDF", mimeType: "application/pdf", url: pdfUrlDownload}];
```
aurimasv: This should be a `.push`, not direct assignment.

symac: Thanks, this was a test I had forgotten to put back.

CAIRN Info.js
symac: @aurimasv I have tried to mix your suggestion and @adam3smith's one to fix this part, not sure it was what you were thinking of, otherwise let me know

symac:
@adam3smith I am really bad at git and while trying to squash the previous commits I am just adding extra ones :( Any advice on how to really squash my commits?

zuphilip:
If you are using Tortoise Git (Windows, locally) than you can do this graphically:
image

Afterwards, you need to force the push to github with the `-f` option.

symac:
Thanks @zuphilip but I am on Linux. But I think I will need to find a GUI as I was hoping to really understand Git one day, but I am starting to think it will never happen :)

symac:
Well, after some fight with Git, I think I won and squashed all my commit so @adam3smith and @aurimasv if the latest version of the file is OK for you, feel free to merge!

adam3smith:
looks great.
FWIW, rebase&squash is just
`git rebase -i HEAD~3` where the 3 is the number of commits you want to squash,
then substitutue pick by squash for all but one commit and press ctrl+X to save&exit, 
edit the commit message and press ctrl+X to save&exit, confirm the save location and the rebase is done. Then `git push --force`

adam3smith:
thanks!

Issue #948, opened 2015-09-24
Add ubiquitypress translator


adam3smith:
http://www.ubiquitypress.com/
Use URL pattern and then maybe test for `//a[class="press-logo"]` text matching http://www.ubiquitypress.com/site/publish/
Articles already have good embedded metadata, so this would be mainly about multiples.

PR #960, opened 2015-10-16
add EM-based translator for Ubiquity Journals

adam3smith:
Increasingly common provider of hosting for open access journals. Tests are for the three different schemas in the target regex

### DIFF
--- a/Ubiquity Journals.js
+++ b/Ubiquity Journals.js
@@ -1,0 +1,158 @@
+{
+	"translatorID": "5c11e3bd-caf5-4da6-95d8-e67c57929098",
+	"label": "Ubiquity Journals",
+	"creator": "Sebastian Karcher",
+	"target": "^https?://[^/]+(/articles/10\\.\\d{4,9}/[-._;()/:a-z0-9A-Z]+|/articles/?$|/\\d+/volume/\\d+/issue/\\d+)",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 250,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2015-10-19 05:23:30"
+}
+
+/*
+	Ubiquiy Press Translator
+	Copyright (C) 2015 Sebastian Karcher
+
+	This file is part of Zotero.
+
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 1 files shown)


adam3smith:
closes #948

Ubiquity Journals.js
```diff
@@ -0,0 +1,156 @@
+{
+	"translatorID": "5c11e3bd-caf5-4da6-95d8-e67c57929098",
+	"label": "Ubiquity Journals",
+	"creator": "Sebastian Karcher",
+	"target": "/articles/10\\.\\d{4,9}/[-._;()/:a-z0-9A-Z]+|/articles/?$|/\\d+/volume/\\d+/issue/\\d+",
```
zuphilip: Should we force here also that there is no other substructere before, i.e. something like
```
"target": "^[^/]+(/articles/10\\.\\d{4,9}/[-._;()/:a-z0-9A-Z]+|/articles/?$|/\\d+/volume/\\d+/issue/\\d+)",
```
?

Ubiquity Journals.js
```diff
@@ -0,0 +1,156 @@
...
+				"ISSN": "2050-5833",
+				"abstractNote": "<p class=\"p1\">This paper explores the relation between the ‘DIY’ (‘do-it-yourself’) movement and ‘DIY architecture’, and the notion of social transformation, in examples of DIY manuals and discourse of North America drawn from the 1940s to the 1970s. The DIY movement emerged as a significant phenomenon in North America of the 1950s, where it was associated with a mainstream audience and a residential market. By the 1960s, the DIY approach was embraced by the North American counterculture as a self-sustaining sensibility that could overcome a reliance on the mainstream, consumerist society that spurned it. On the surface, the association of DIY with the counterculture and countercultural architects appears to denote a significant ideological shift from its original association with the beliefs and culture of mainstream North America and the nuclear family. However, one of the key characterisations of the DIY movement identified in the present paper is the way it is bound to the notion of social identity and transformation, regardless of ideology. Particular attention is paid to DIY manuals and discourse of the 1950s.</p>",
```
zuphilip: Maybe clean the `abstractNote` from the tags.

adam3smith: so - - there are actually escaped html tags (think &amp;lt;p&amp;gt;) in the abstract. That seemed to be so clearly just bogus on their part that I didn't bother cleaning them, but I guess I could? Will standard removeTags work?

Ubiquity Journals.js
adam3smith: Let me check some more URLs, but I really like that as a general idea.

Ubiquity Journals.js
zuphilip: I don't think `cleanTags` will work with escaped html tags, cf. https://github.com/zotero/zotero/blob/master/chrome/content/zotero/xpcom/utilities.js#L265 .  However, it should be easy to come up with something similiar, maybe only this case with the `<p>`-Tag. E.g. something like
```
item.abstractNote = abstract.replace(/&lt;p[^&>]*&gt;/g, "").trim();
```

adam3smith:
used both of your suggestions; thanks @zuphilip

PR #963, opened 2015-10-21
Renamed Talis Prism 2 to Capita Prism

philjohn:
Talis's library business was sold to Capita in 2011, so changing the label to reflect this.

The URI structure being looked for is also from the version of Prism that followed 2, which doesn't have a version number.


adam3smith:
Thanks. We'd want the filename to reflect this change. Also, you'll need to bump the timestamp so that the translator actually gets updated for users.

adam3smith:
superseded by #1090

PR #964, opened 2015-10-21
Expand the regex target of Aleph

karnesky:
https://forums.zotero.org/discussion/52977/zotero-finds-false-positives/#Item_10


adam3smith:
thanks. You need to bump the timestamp, still (squash when you're done). I can also do this manually tonight, since it's just one line.

PR #968, opened 2015-11-02
Translator for CONTENTdm archive and library database software

emmareisz:
Translator for CONTENTdm archive and library database software, v6 with limited support for v4 and v5.

### DIFF
--- a/CONTENTdm.js
+++ b/CONTENTdm.js
@@ -1,0 +1,577 @@
+{
+	"translatorID": "f715a5a1-c362-47cf-b736-2cb2c882b852",
+	"label": "CONTENTdm",
+	"creator": "Emma Reisz",
+	"target": "/cdm/|/cdm4/",
+	"minVersion": "3.0.9",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsbv",
+	"lastUpdated": "2015-11-02 10:55:08"
+}
+
+/**
+	***** BEGIN LICENSE BLOCK *****
+
+	CONTENTdm translator; Copyright © 2015 Emma Reisz
+	This file is part of Zotero.
+
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero.  If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+/*
+  ContentDM is an OCLC product used by several hundred libraries, archives and museums.
+  CDM may be hosted by OCLC or by the customer.
+  CDM includes an API which calls metadata as XML or JSON, but customers may disable it.
+  By preference, this translator scrapes XML from an API query (slow)
+  and from a webpage as a fallback (quick but vulnerable to a redesign).
+  XML is preferred to JSON because it allows code to be shared with the webscraper.
+  The translator will also scrape some data from legacy CDM webpages (version < 6).
+  Older CDM versions had no standardised field names so comprehensive scraping is unrealistic.
+*/
+
+function detectWeb( doc, url ) {
+  if ( !url.match( "/cdm/" ) && !url.match( "/cdm4/" ) ) return;
+  if ( url.match( "/cdm/search/" ) ) return "multiple";
+  else if ( url.match( "/cdm/landingpage" ) || !url.substring( url.indexOf( '/cdm/' ) + 5 ) ) return "webpage";
+  else if ( url.match( "/cdm/fullbrowser" ) ) { // Pages without item metadata
+	originatingItem = ZU.xpath( doc, '//link[@rel="canonical"]' );
+	if ( originatingItem && originatingItem[ 0 ].href ) return "document"; // Slow to scrape again for item type
+	else return "";
+  }
+  return ( function() {
+	var itemType = fetchXPathText( doc, '//*[@id="metadata_type"]' );
+	if ( !itemType ) itemType = fetchXPathText( doc, '//*[@id="metadata_object_type"]' );
+	switch ( itemType ) {
+	  case "Image":
+	  case "Photograph":
+	  case "Engraved portrait":
+	  case "Negative":
+		return 'artwork';
+	  case "Text":
+		return 'manuscript';
+	  default:
+		return 'document';
+	}
+  }() );
+}
+
+function fetchXPathText( doc, xpath ) { // Removes excess whitespace & breaks, respaces items overtrimmed by trimInternal.
+  if ( ZU.xpathText( doc, xpath ) ) {
+	return ZU.trimInternal( ZU.xpathText( doc, xpath ).replace( /^\s+|\s+$|\;$|\,$/g, '' ).replace( /([a-z])([A-Z])/g, '$1\. $2' ) );
+  } else return;
+}
+
+function fetchXPathValue( doc, xpath ) {
+  if ( ZU.xpath( doc, xpath )[ 0 ] ) return ZU.xpath( doc, xpath )[ 0 ].value;
+  else return;
+}
+
+function fetchXmlUrl( doc, url ) { // Fetches xml urls if using OCLC server (as constructor)
+  var serverID = new RegExp( "/collection\/p([A-Z0-9]+)coll", "i" );
+  var serverID2 = new RegExp( "//cdm([0-9]+).contentdm.oclc.org" );
+  var collectionID = new RegExp( "/collection\/([A-Z0-9]+)\/", "i" );
+  var objectID = new RegExp( "/id\/([A-Z0-9]+)", "i" );
+  var itemID = new RegExp( "/show\/([A-Z0-9]+)", "i" );
+  var objectUrlHttp = fetchXPathValue( doc, '//input[@id="cdm_objectRefUrl"]' );
+  var itemUrlHttp = fetchXPathValue( doc, '//input[@id="cdm_newRefUrl"]' );
+  var server;
+  var xmlQueryUrl;
+  if ( objectUrlHttp ) url = objectUrlHttp; // Use object page as primary even if item page is present
+  else if ( itemUrlHttp ) {
+	url = itemUrlHttp;
+	itemUrlHttp = "";
+  } // Use item page as primary if no object page is present
+  if ( url.match( serverID ) ) server = url.match( serverID )[ 1 ];
+  else if ( url.match( serverID2 ) ) server = url.match( serverID2 )[ 1 ];
+  if ( server && url.match( collectionID ) && url.match( objectID ) ) {
+	xmlQueryUrl = "https://server" + server + ".contentdm.oclc.org/dmwebservices/index.php?q=dmGetItemInfo/" + url.match( collectionID )[ 1 ] + "/";
+	this.object = xmlQueryUrl + url.match( objectID )[ 1 ] + "/show/xml";
+  } else this.object = "";
+  if ( itemUrlHttp ) this.item = xmlQueryUrl + itemUrlHttp.match( objectID )[ 1 ] + "/show/xml";
+  else if ( this.object && url.match( itemID ) ) this.item = xmlQueryUrl + url.match( itemID )[ 1 ] + "/show/xml";
+  else this.item = "";
+}
+
+function scrape( doc, url ) {
+  var item = new Zotero.Item( detectWeb( doc, url ) );
+  url = doc.documentURI;
+  var xmlUrl = new fetchXmlUrl( doc, url );
+
+  /* Mappings for XML and Web (CDM>5)
+   * The scraper only scrapes fields specified in the mapping table.
+   * Enter mappings as [zoteroField, CDMfield];
+   * note CDM allows customers to configure fields
+   * so these are inconsistently implemented.
+   */
+  var mapping = [
+			[ 'title', 'title' ],
+			[ 'creator', 'creato' ],
+			[ 'creator2', 'creata' ],
+			[ 'contributor', 'contri' ],
+			[ 'abstractNote', 'descri' ],
+			[ 'language', 'langua' ],
+			[ 'language2', 'languag' ],
+			[ 'archiveLocation', 'identi' ], // On CDM this is 'Identifier'
+			[ 'archiveLocation2', 'locala' ],
+			[ 'archive', 'source' ],
+			[ 'collection', 'collec' ],
+			[ 'collection2', 'relati' ],
+			[ 'publisher', 'publis' ], // Omitted by manuscript item type
+			[ 'rights', 'rights' ],
+			[ 'subject', 'subject' ], // Scraped into tags
+			[ 'subject2', 'subjea' ],
+			[ 'subject3', 'subjeb' ],
+			[ 'subject4', 'subjec' ],
+			[ 'person', 'person' ],
+			[ 'medium', 'type' ],
+			[ 'format', 'format' ],
+			[ 'date', 'date' ],
+			[ 'date2', 'dated' ],
+			[ 'place', 'covera' ],
+			[ 'place2', 'coverab' ],
+			[ 'period', 'period' ],
+		  ];
+
+  /* Mappings for Web (CDM<6)
+   * CDM < 6 does not provide standard field names.
+   * The scraper scrapes the whole display table,
+   * and converts displayed fieldnames to camelCase
+   * to maximise the chance of a Zotero match.
+   * Hardcode additional matches here as [zoteroField, displayedField];
+   */
+
+  var legacyMapping = [ // Enter mappings as [zoteroField,camelizedCDMField]
+			[ "place", "town" ],
+			[ "abstractNote", "description" ],
+			[ "subject2", "keywords" ],
+			[ "subject3", "subjectKeywords" ],
+			[ "date2", "workDate" ],
+			[ "creator", "bookCreator" ],
+			[ "rights", "rights--UsageStatement" ],
+			[ "contributor2", "photographer" ],
+			[ "title2", "bookTitle" ],
+		  ];
+
+  var docTitle = ( function() {
+	if ( doc.title == "ContentDM" ) return url.substring( url.indexOf( '://' ) + 3, url.indexOf( '/cdm/' ) ).replace( /^\s+|\s+$/g, '' );
+	else return doc.title.replace( /^\s+|\s+$/g, '' );
+  }() );
+
+  item.libraryCatalog = ( function() {
+	var catalog = fetchXPathText( doc, '//*[@id="breadcrumb_top_content"]/a[2]' );
+	if ( catalog ) return catalog;
+	else if ( docTitle.match( '::' ) ) return docTitle.substring( doc.title.indexOf( '::' ) + 3 ); // Fallback: use second part of webpage title
+	else return docTitle;
+  }() );
+
+  item.websiteType = "ContentDM";
+  item.attachments.push( {
+	title: "Snapshot of CONTENTdm record",
+	mimeType: "text/html",
+	document: doc
+  } );
+  item.url = url;
+  //  item.attachments.push({url:url, title: "Link to Record", mimeType: "text/html", snapshot: false}); // optional; url set instead
+  attachPDF( doc, item );
+
+  ZU.doGet( xmlUrl.object, function( text ) {
+	// Z.debug(text);                                           // View primary XML doc
+	var scrapeMode;
+	if ( !xmlUrl.object || !text || text.match( "The web server is not configured" ) ) scrapeMode = "Web";
+	else { // Scrape from API XML
+	  var docxml = ( new DOMParser() ).parseFromString( text, "text/xml" );
+	  if ( xmlUrl.item ) { // Scraping from two XML docs
+		ZU.doGet( xmlUrl.item, function( text ) {
+		  var doc2xml = ( new DOMParser() ).parseFromString( text, "text/xml" );
+		  // Z.debug(text);                                   // View secondary XML doc
+		  scrapeToItem( docxml, item, url, docTitle, mapping, "xml", doc2xml );
+		  item.complete();
+		} );
+	  } else { // Scraping from a single XML doc
+		scrapeToItem( docxml, item, url, docTitle, mapping, "xml" );
+		item.complete();
+	  }
+	  return; // Return from the ZU.doGet callback after XML scrape
+	}
+	if ( scrapeMode == "Web" ) {
+	  var headerXPath;
+	  var contentXPath;
+	  var offset;
+	  var xhttp = new XMLHttpRequest();
+	  xhttp.open( "GET", url, false );
+	  xhttp.send();
+	  fullSource = xhttp.responseText;
+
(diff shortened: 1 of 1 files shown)


CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
+{
+	"translatorID": "f715a5a1-c362-47cf-b736-2cb2c882b852",
+	"label": "CONTENTdm",
+	"creator": "Emma Reisz",
+	"target": "/cdm/|/cdm4/",
+	"minVersion": "3.0.9",
+	"maxVersion": "",
+	"priority": 100,
```
aurimasv: This should be 270 as per #765

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+  CDM includes an API which calls metadata as XML or JSON, but customers may disable it.
+  By preference, this translator scrapes XML from an API query (slow)
+  and from a webpage as a fallback (quick but vulnerable to a redesign).
+  XML is preferred to JSON because it allows code to be shared with the webscraper.
+  The translator will also scrape some data from legacy CDM webpages (version < 6).
+  Older CDM versions had no standardised field names so comprehensive scraping is unrealistic.
+*/
+
+function detectWeb( doc, url ) {
+  if ( !url.match( "/cdm/" ) && !url.match( "/cdm4/" ) ) return;
```
aurimasv: This is not needed. The target regexp ensures this is the case.

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+  if ( url.match( "/cdm/search/" ) ) return "multiple";
```
aurimasv: Too generic. Make sure that there are actually search results to scrape. See https://github.com/zotero/translators/blob/e9423311b4a492dc7862c7d220f1ac5c23246b82/Le%20Monde.js#L41 for the preferred way to do this.

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+  else if ( url.match( "/cdm/landingpage" ) || !url.substring( url.indexOf( '/cdm/' ) + 5 ) ) return "webpage";
```
aurimasv: Not sure what this is supposed to be. We generally don't match on plain web pages. Also, we prefer to use `url.indexOf("foo") != -1` or, if you need regexp, `/foo?/.test(url)` over `.match()`. Lastly, it's preferable to be more obvious with what you're trying to do. E.g. `!url.substring( url.indexOf( '/cdm/' ) + 5 )` should be `/\/cdm\/$/.test(url)` or `url.indexOf('/cdm/') == url.length - 5`

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+  else if ( url.match( "/cdm/fullbrowser" ) ) { // Pages without item metadata
+	originatingItem = ZU.xpath( doc, '//link[@rel="canonical"]' );
+	if ( originatingItem && originatingItem[ 0 ].href ) return "document"; // Slow to scrape again for item type
```
aurimasv: "document" is not a great type to show. I'd pick something that is most likely to actually be scraped. Looks like "artwork" might be common.

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+	else return "";
```
aurimasv: just `return;`

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+  }
+  return ( function() {
```
aurimasv: any specific reason why you need the `function` wrapper here?

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+	var itemType = fetchXPathText( doc, '//*[@id="metadata_type"]' );
```
aurimasv: when fetching by ID, it's best to use `doc.getElementById()`, rather than using xpaths which iterate through the whole page and is about 5,000 times slower. https://jsperf.com/getelementbyid-vs-queryselector/4 (if you need complicated logic, it's generally ok)

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+	switch ( itemType ) {
+	  case "Image":
+	  case "Photograph":
+	  case "Engraved portrait":
+	  case "Negative":
+		return 'artwork';
+	  case "Text":
+		return 'manuscript';
+	  default:
+		return 'document';
```
aurimasv: Do you have some examples of this? "document" is a bad choice, since it lacks a lot of fields.

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+	}
+  }() );
+}
+
+function fetchXPathText( doc, xpath ) { // Removes excess whitespace & breaks, respaces items overtrimmed by trimInternal.
+  if ( ZU.xpathText( doc, xpath ) ) {
```
aurimasv: avoid calling this twice

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+	return ZU.trimInternal( ZU.xpathText( doc, xpath ).replace( /^\s+|\s+$|\;$|\,$/g, '' ).replace( /([a-z])([A-Z])/g, '$1\. $2' ) );
```
aurimasv: `ZU.trimInternal` also trims leading and trailing whitespace. Escaping the period inside the replacement string is unnecessary. Also, please keep lines under 80 characters when possible.

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+	  }
+	  ZU.processDocuments( articles, scrape );
+	} );
+
+  } else if ( url.match( "/cdm/fullbrowser" ) ) {
+	var originatingItem = ZU.xpath( doc, '//link[@rel="canonical"]' ); // Look for link to originating item
+	if ( originatingItem[ 0 ] && originatingItem[ 0 ].href ) {
+	  url = originatingItem[ 0 ].href;
+	  ZU.processDocuments( url, scrape );
+	} else return false;
```
aurimasv: this must never happen

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+
+  } else if ( ZU.xpath( doc, '//input[@id="cdm_newRefUrl"]' ) ) {
```
aurimasv: `ZU.xpath` always returns an array, so this is always `true`

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+}
+
+function fetchXPathValue( doc, xpath ) {
+  if ( ZU.xpath( doc, xpath )[ 0 ] ) return ZU.xpath( doc, xpath )[ 0 ].value;
+  else return;
+}
+
+function fetchXmlUrl( doc, url ) { // Fetches xml urls if using OCLC server (as constructor)
+  var serverID = new RegExp( "/collection\/p([A-Z0-9]+)coll", "i" );
+  var serverID2 = new RegExp( "//cdm([0-9]+).contentdm.oclc.org" );
```
aurimasv: escape periods (double backslash)

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+	xmlQueryUrl = "https://server" + server + ".contentdm.oclc.org/dmwebservices/index.php?q=dmGetItemInfo/" + url.match( collectionID )[ 1 ] + "/";
+	this.object = xmlQueryUrl + url.match( objectID )[ 1 ] + "/show/xml";
+  } else this.object = "";
+  if ( itemUrlHttp ) this.item = xmlQueryUrl + itemUrlHttp.match( objectID )[ 1 ] + "/show/xml";
+  else if ( this.object && url.match( itemID ) ) this.item = xmlQueryUrl + url.match( itemID )[ 1 ] + "/show/xml";
+  else this.item = "";
+}
+
+function scrape( doc, url ) {
+  var item = new Zotero.Item( detectWeb( doc, url ) );
```
aurimasv: move item type detection into a separate function. Then call it from here and `detectWeb()`. You can't have it return "multiple" here or a `null`.

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+  url = doc.documentURI;
```
aurimasv: There should be no reason to do this. Use `url` as passed into this function when storing it in `item.url` below. It handles proxies.

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+  var xmlUrl = new fetchXmlUrl( doc, url );
```
aurimasv: If you want to return more than one parameter, return an object. Don't do this, it's unconventional and confusing.

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+			[ "subject2", "keywords" ],
+			[ "subject3", "subjectKeywords" ],
+			[ "date2", "workDate" ],
+			[ "creator", "bookCreator" ],
+			[ "rights", "rights--UsageStatement" ],
+			[ "contributor2", "photographer" ],
+			[ "title2", "bookTitle" ],
+		  ];
+
+  var docTitle = ( function() {
```
aurimasv: why?

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+	if ( doc.title == "ContentDM" ) return url.substring( url.indexOf( '://' ) + 3, url.indexOf( '/cdm/' ) ).replace( /^\s+|\s+$/g, '' );
```
aurimasv: split this up into multiple lines

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+	else return doc.title.replace( /^\s+|\s+$/g, '' );
```
aurimasv: `.trim()`

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+  }() );
+
+  item.libraryCatalog = ( function() {
+	var catalog = fetchXPathText( doc, '//*[@id="breadcrumb_top_content"]/a[2]' );
+	if ( catalog ) return catalog;
+	else if ( docTitle.match( '::' ) ) return docTitle.substring( doc.title.indexOf( '::' ) + 3 ); // Fallback: use second part of webpage title
+	else return docTitle;
+  }() );
```
aurimasv: I'll let @adam3smith decide this, but in general, we just let this be the URL host. Given that this is an online catalog, it's both informative for the user and it helps us with debugging.

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+
+  item.websiteType = "ContentDM";
```
aurimasv: That is not the purpose of this field. "ContentDM" should probably end up at the end of `.libraryCatalog` as " (ContentDM)"

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+  item.attachments.push( {
+	title: "Snapshot of CONTENTdm record",
+	mimeType: "text/html",
+	document: doc
+  } );
```
aurimasv: Would this be informative? If so, just "Snapshot" is fine (and you can drop the `mimeType` part when attaching a `document`)

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+  item.url = url;
+  //  item.attachments.push({url:url, title: "Link to Record", mimeType: "text/html", snapshot: false}); // optional; url set instead
+  attachPDF( doc, item );
+
+  ZU.doGet( xmlUrl.object, function( text ) {
+	// Z.debug(text);                                           // View primary XML doc
+	var scrapeMode;
+	if ( !xmlUrl.object || !text || text.match( "The web server is not configured" ) ) scrapeMode = "Web";
```
aurimasv: Not sure what the `!xmlUrl.object` is supposed to mean here. If it's true, `ZU.doGet` would be pointless.

aurimasv: Also, `scrapeMode` is a global variable. Instantiate it first.

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+	  return; // Return from the ZU.doGet callback after XML scrape
+	}
+	if ( scrapeMode == "Web" ) {
+	  var headerXPath;
+	  var contentXPath;
+	  var offset;
+	  var xhttp = new XMLHttpRequest();
+	  xhttp.open( "GET", url, false );
+	  xhttp.send();
+	  fullSource = xhttp.responseText;
```
aurimasv: Use `ZU.doGet`

aurimasv: Though I don't really understand what you're trying to do here.

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+		offset = 0;
+		legacyScrapeToItem( doc, item, headerXPath, contentXPath, offset, legacyMapping, docTitle, url );
+	  }
+
+	  item.complete();
+
+	  return; // Return from the ZU.doGet callback after Web scrape
+	}
+  } );
+  return '';
```
aurimasv: Remove

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+	  else if ( fullSource.match( "CONTENTdm" ) ) { // Web scrape for CDM Version 4 or below
+		headerXPath = '//div/div[3]/table[2]/tbody/tr/td[1]';
+		contentXPath = '//div/div[3]/table[2]/tbody/tr/td[2]';
+		offset = 0;
+		legacyScrapeToItem( doc, item, headerXPath, contentXPath, offset, legacyMapping, docTitle, url );
+	  }
+
+	  item.complete();
+
+	  return; // Return from the ZU.doGet callback after Web scrape
```
aurimasv: remove

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+	}
+  } );
+  return '';
+}
+
+function legacyScrapeToItem( doc, item, headerXPath, contentXPath, offset, mapping, docTitle, url ) {
+  var headers = ZU.xpath( doc, headerXPath ); // Scrapes the whole metadata table and looks for matches after.
+  var contents = ZU.xpath( doc, contentXPath );
+  for ( i = 0; i < contents.length; i++ ) {
```
aurimasv: `var i`

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+	var header;
+	var content;
+	j = i + offset;
```
aurimasv: `var j`

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+	if ( headers[ j ].textContent ) {
+	  header = headers[ j ].textContent;
+	  header = camelize( header );
+	}
+	if ( contents[ i ].textContent ) content = contents[ i ].textContent;
```
aurimasv: what if it's not? (content keeps the value it had in previous loop)

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+	item[ header ] = content;
+  }
+  for ( var i in mapping ) item[ mapping[ i ][ 0 ] ] = item[ mapping[ i ][ 1 ] ];
```
aurimasv: This seems unnecessary, why not assign it correctly in the first place. We also generally use a map for this. So something like `if (map[header]) item[map[header]] = content;`

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+  tidyItem( item );
+  if ( !item.title ) item.title = getTitle( doc, docTitle, url );
+
+  function camelize( str ) {
+	return str.replace( /(?:^\w|[A-Z]|\b\w)/g, function( letter, index ) {
```
aurimasv: I think what you want is `/^[A-Z]|\b[a-z]/g`

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+	  return index === 0 ? letter.toLowerCase() : letter.toUpperCase();
+	} ).replace( /\s+/g, '' );
+  }
+}
+
+function scrapeToItem( doc, item, url, docTitle, mapping, prefixes, doc2 ) {
+  var info = []; // Scrapes only fields named in the mapping array.
+  var xpath;
+  var itemText;
+  for ( var i in mapping ) {
```
aurimasv: use `for (var i=0; i<arr.length; i++) {...` to iterate arrays

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+		  [ 'place', 'place2' ],
+		  [ 'creator', 'creator2' ],
+		  [ 'archiveLocation', 'archiveLocation2' ],
+		  [ 'medium', 'format' ],
+		  [ 'contributor', 'contributor2' ],
+		];
+  for ( var i in merges ) {
+	var mergeArray = [];
+	for ( var j in merges[ i ] ) {
+	  mergeArray.push( item[ merges[ i ][ j ] ] ); // Arrays are easier to filter
```
aurimasv: or you could just not push it in the first place if it's going to be filtered out later anyway

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+function getTitle( doc, docTitle, url ) {
+  var title;
+  if ( docTitle.match( "::" ) ) docTitle = docTitle.substring( 0, docTitle.indexOf( "::" ) );
+  if ( url.match( "/cdm/landingpage" ) ) title = docTitle + ": About this collection";
+  else title = docTitle; // Fallback: first part of webpage title
+  return title.replace( /^\s+|\s+$/g, '' );
+}
+
+function cleanAuthor( item ) {
+  var authorType = [ "creator", "contributor" ]; // Author field names in the mapping table must be Zotero author types
```
aurimasv: "creator" is not a valid author type. You want "author"

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+  for ( var i in authorType ) {
+	if ( item[ authorType[ i ] ] ) {
+	  fetchedAuthor = item[ authorType[ i ] ].replace( /\([^)]*\)|\[[^\]]*\]|\;$/g, '' ).replace( /, Sir |, Dr /g, ', ' ); // Fetch and remove parentheses and honorifics
+	  var nameArray = [];
+	  item[ authorType[ i ] ] = "";
+	  nameArray = fetchedAuthor.split( ";" ); // Split authors at semicolon and push into an array
+	  for ( var j in nameArray ) item.creators.push( ZU.cleanAuthor( nameArray[ j ], authorType[ i ], 1 ) );
```
aurimasv: Use `true`/`false` instead of `1`/`0`

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+}
+
+function attachPDF( doc, item ) {
+  var pdfPath = '//embed';
+  var pdfScrape = ZU.xpath( doc, pdfPath );
+  var pdfObject = pdfScrape[ 0 ]; // Could iterate looking for multiple attachments
+  var pdfTitle;
+  if ( pdfObject && pdfObject.src ) { // Can't simply test for .src
+	var itemTitle = fetchXPathText( doc, '//*[@id="metadata_title"]' );
+	if ( itemTitle ) pdfTitle = "Attachment: " + itemTitle;
```
aurimasv: Generic names, like "Full Text PDF" are preferred.

aurimasv:
Please address the comments above to start and run the code through http://jsbeautifier.org/ using tabs for indenting. Looks like there will be further comments after that.

CONTENTdm.js
adam3smith: ( `ZU.xpath().length`  would be the way to go==> it's 0 for a null object, which evaluates to false)

CONTENTdm.js
adam3smith: yup, agree. The catalog labels as currently provided don't appear terribly useful.

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+
+/** BEGIN TEST CASES **/
+var testCases = [
+	{
+		"type": "web",
+		"url": "http://cdm15979.contentdm.oclc.org/cdm/compoundobject/collection/p15979coll3/id/2419",
+		"items": [
+			{
+				"itemType": "manuscript",
+				"title": "MS.15.1.2.000. Sir Robert Hart Diary: Volume 02: February 1855-July 1855",
```
adam3smith: this looks like the archive location is prepended to the title. Could we clean that up better?

CONTENTdm.js
adam3smith: example is below. Yes, please don't default to document. manuscript is a good default for archival sources.

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+  }
+  tidyItem( item );
+  return item;
+}
+
+function tidyItem( item ) {
+  mergeFields( item );
+  replaceFields( item );
+  cleanAuthor( item );
+  item.tags = []; //Tags must be in an array
```
adam3smith: not needed -- if you're going to use `.split()` below, you always get an array.

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+  item.tags = item.subject.replace( /\([^)]*\)/g, '' ).split( /[.;]/ );
```
adam3smith: I'd check for item.subject here to be safe -- otherwise `.replace()` will trigger an error when it's empty.

CONTENTdm.js
```diff
@@ -0,0 +1,577 @@
...
+  for ( var k in item.tags ) item.tags[ k ] = ZU.trimInternal( item.tags[ k ].replace( /^\s+|\s+$/g, '' ).replace( /\s+\,/g, ',' ) );
```
adam3smith: see above -- use sequential `for` loop to iterate over arrays. See e.g. here for details: http://stackoverflow.com/a/3010848/1483360

CONTENTdm.js
emmareisz: Why do you want to remove the backwards compatibility?

aurimasv: Sorry, I'm not following. Compatibility with what?

emmareisz: I think I'm not following! What do you want me to remove?

aurimasv: The unnecessary `return` statement. (below as well)

emmareisz: Ah. No problem. Thanks.

CONTENTdm.js
emmareisz: There are two points here. 1) In this particular catalogue record, 'MS.15.1.2.000' is given as the item title, hence it is scraped by the translator. On this occasion, the item title is redundant, but in many cases it isn't. 

2) The slightly awkward format is because CDM 6 has both object-level data and item-level data. These are implemented inconsistently by libraries, and it is not possible to say that either can be safely omitted. Because of the way CDM displays, libraries often choose to put essential information at the Object level, and (as in this example) simply dropping the object-level data if the item data is present would lead to some uninformative scrapes. Hence the translator combines the item information and the object level data.

zuphilip: It looks for me that `MS.15.1.2.000` is just a single page and the diary contains several pages, e.g. <a href="http://cdm15979.contentdm.oclc.org/cdm/search/searchterm/MS.15.1.2.005">MS.15.1.2.005</a> stands for page number 5.

emmareisz: Exactly.

adam3smith: OK, great. Leave as is, then; one less thing to fix.

adam3smith:
@emmareisz I think the needed changes are fairly small -- are you interested in finishing this translator up still?

emmareisz:
Hi @adam3smith - yes, I've been using the CONTENTdm translator and sharing it with colleagues, so I'd like to get it incorporated into Zotero. CONTENTdm's also had an update in the meantime. I should have some time over Christmas to finish it.

I'd like to get the [BNA translator](https://groups.google.com/forum/#!topic/zotero-dev/WMvlNHvysJY) finished first, though, if that's ok - could you take a quick look at it and then I'll do a PR?

adam3smith:
Yes of course -- I left a comment there.

AbeJellinek:
@emmareisz, any progress on this or the BNA translator? I would love to get all this merged!

emmareisz:
Just pottering along using them privately! I'll email you @AbeJellinek

PR #972, opened 2015-11-27
remove duplicate data and fix abstract for Hindawi; 

adam3smith:
Also switch test item to a better one.

(The actual data doesn't change apart from the improved abstract, so you can ignore the test changes)

### DIFF
--- a/Hindawi Publishers.js
+++ b/Hindawi Publishers.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsbv",
-	"lastUpdated": "2014-05-01 00:37:52"
+	"lastUpdated": "2015-11-27 22:23:46"
 }
 
 /*
@@ -49,6 +49,11 @@ function detectWeb(doc,url) {
 	return false;
 }
 
+function decodeEntities(doc, encodedString) {
+	var textArea = doc.createElement('textarea');
+	textArea.innerHTML = encodedString;
+	return textArea.value;
+}
 
 function doWeb(doc,url)
 {
@@ -71,6 +76,10 @@ function doWeb(doc,url)
 			ZU.processDocuments(urls, doWeb);
 		});
 	} else {
+		var namespace = {"x"  : "http://www.w3.org/1999/xhtml"}
+	        var abstract = ZU.xpathText(doc, '//x:meta[@name="citation_abstract"][1]/@content', namespace);
+	        //convert html entities in abstract
+		abstract = decodeEntities(doc, abstract);
 		var translator = Zotero.loadTranslator('web');
 		//use Embedded Metadata
 		translator.setTranslator("951c027d-74ac-47d4-a107-9c3069ab7b48");
@@ -80,7 +89,27 @@ function doWeb(doc,url)
 				// use article ID as a page (seems to be the last part of URL/DOI)
 				item.pages = 'e' + item.DOI.substr(item.DOI.lastIndexOf('/') + 1);
 			}
-			item.extra = "";
+			//remove duplicate metadata			
+			for (i in item){
+				if (typeof item[i] == "string" && item[i].match(/^.+,/)){
+					//Z.debug(item[i])
+					if (item[i].match(/^(?:PMID: )?(.+),/)[1] == item[i].match(/, (.+)$/)[1]){
+						item[i] = item[i].replace(/,.+?$/, "")
+					}
+				}
+			}
+			//remove duplicate authors
+			if (item.creators.length%2 == 0){
+				var j=item.creators.length -1
+				for (var i = item.creators.length/2 -1; i>=0; i--, j--){
+					if (item.creators[i].firstName === item.creators[j].firstName &&
+						item.creators[i].lastName === item.creators[j].lastName){
+							item.creators.splice(j, 1);
+						}
+				}
+			}
+			item.abstractNote=abstract;
+			//item.extra = "";
 			item.complete();
 		});
 		translator.translate();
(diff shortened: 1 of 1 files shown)


Hindawi Publishers.js
```diff
@@ -49,6 +49,11 @@ function detectWeb(doc,url) {
 	return false;
 }
 
+function decodeEntities(doc, encodedString) {
```
adam3smith: @aurimasv -- I know you're super busy, but if you could take 2mins to look at this? Seems like this should be save because we're in a sandbox, but still want to be sure. Couldn't find a method of converting the html entitites without this, but maybe overlooking something simple.

aurimasv: https://github.com/zotero/zotero/blob/4.0/chrome/content/zotero/xpcom/utilities.js#L451 should do what you want

Hindawi Publishers.js
```diff
@@ -71,6 +76,10 @@ function doWeb(doc,url)
 			ZU.processDocuments(urls, doWeb);
 		});
 	} else {
+		var namespace = {"x"  : "http://www.w3.org/1999/xhtml"}
```
aurimasv: move `namespace` out of the `if` block above and reuse it here instead of re-declaring.

aurimasv: Actually, just declare it as a global, since it's used everywhere.

Hindawi Publishers.js
```diff
@@ -71,6 +76,10 @@ function doWeb(doc,url)
...
+	        var abstract = ZU.xpathText(doc, '//x:meta[@name="citation_abstract"][1]/@content', namespace);
+	        //convert html entities in abstract
+		abstract = decodeEntities(doc, abstract);
```
aurimasv: `abstract` could be `null`

Hindawi Publishers.js
```diff
@@ -80,7 +89,27 @@ function doWeb(doc,url)
 				// use article ID as a page (seems to be the last part of URL/DOI)
 				item.pages = 'e' + item.DOI.substr(item.DOI.lastIndexOf('/') + 1);
 			}
-			item.extra = "";
+			//remove duplicate metadata
```
aurimasv: I'm not really clear on what you're trying to remove from which fields. How is this duplicate metadata getting in. Is this only touching PMID in the extra field?

adam3smith: No, it's for all fields that don't get overwritten when you call them twice. Their EM has every highwire field duplicated.

Hindawi Publishers.js
```diff
@@ -80,7 +89,27 @@ function doWeb(doc,url)
...
+			//remove duplicate authors
+			if (item.creators.length%2 == 0){
+				var j=item.creators.length -1
+				for (var i = item.creators.length/2 -1; i>=0; i--, j--){
+					if (item.creators[i].firstName === item.creators[j].firstName &&
+						item.creators[i].lastName === item.creators[j].lastName){
+							item.creators.splice(j, 1);
+						}
+				}
+			}
```
aurimasv: I'm guessing that the authors are either all duplicated or none of them are, right? It doesn't make much sense otherwise. If so, I think this should be cleaner.
```
if (item.creators.length && item.creator.length % 2 == 0) {
  var duplicate = true;
  for (var i = 0, j = item.creators.length / 2; duplicate && j < item.creators.length; i++, j++) {
    var a1 = item.creators[i], a2 = item.creators[j];
    duplicate = a1.firstName == a2.firstName && a1.lastName == a2.lastName;
  }

  if (duplicate) {
    item.creators = item.creators.slice(0, item.creators.length / 2);
  }
}
```

Hindawi Publishers.js
aurimasv: Oh, I see. What about the PMID? The PMID label is not duplicated?

aurimasv:
>The actual data doesn't change apart from the improved abstract, so you can ignore the test changes

How come we are changing the test then?

Hindawi Publishers.js
```diff
@@ -80,7 +89,27 @@ function doWeb(doc,url)
...
+			item.abstractNote=abstract;
```
aurimasv: nit: space around `=`

Hindawi Publishers.js
adam3smith: Correct, just the number not the label. That's because the EM import works
with ZU.xpathText
On Mon, Jan 4, 2016 at 8:43 AM, Aurimas Vinckevicius <
notifications@github.com> wrote:

> In Hindawi Publishers.js
> <https://github.com/zotero/translators/pull/972#discussion_r48736252>:
>
> > @@ -80,7 +89,27 @@ function doWeb(doc,url)
> >  				// use article ID as a page (seems to be the last part of URL/DOI)
> >  				item.pages = 'e' + item.DOI.substr(item.DOI.lastIndexOf('/') + 1);
> >  			}
> > -			item.extra = "";
> > +			//remove duplicate metadata			
>
> Oh, I see. What about the PMID? The PMID label is not duplicated?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/zotero/translators/pull/972/files#r48736252>.
>



-- 
Sebastian Karcher, PhD
www.sebastiankarcher.com

Hindawi Publishers.js
aurimasv: Also, might as well move all the `abstract` code here.

adam3smith:
I changed tests because I wanted a better test case, so I switched them out.

Hindawi Publishers.js
aurimasv: ok, let's do the following then:
```
for (var i in item) {
  item[i] = item[i].replace(/^(PMID: )?(.+), \2$/, "$1$2");
}
```

adam3smith: cool. We need the test for type, though, otherwise this breaks on non-string fields.

adam3smith:
OK, should be all good now.

Hindawi Publishers.js
```diff
@@ -80,7 +80,29 @@ function doWeb(doc,url)
 				// use article ID as a page (seems to be the last part of URL/DOI)
 				item.pages = 'e' + item.DOI.substr(item.DOI.lastIndexOf('/') + 1);
 			}
-			item.extra = "";
+			//remove duplicate metadata			
+			for (var i in item){
+				if (typeof item[i] == "string" && item[i].match(/^.+,/)){
```
aurimasv: Was the `.match` part necessary? More curious than anything, since I can't think of a reason why it should be.

aurimasv:
Looks good. Feel free to merge.

adam3smith:
(this was still an issue, so merging)

Issue #797, opened 2014-11-05
Update Wikipedia Citation Template to meet guidelines


aurimasv:
See http://en.wikipedia.org/wiki/Wikipedia:Citation_templates and https://forums.zotero.org/discussion/41819/wikipedia-citation-templates-and-their-limitations/#Item_6 (but there might be more changes that need to be made)

owcz:
@aurimasv: I can help with this but since it's been a year and I don't know what has been updated, might be useful to define the spec

adam3smith:
The main job is actually tracking down the current version of the WP citation template and details about preferences for the various fields, which you are probably in a better position to do than us. 

The actual edits should be fairly light and just keep with the current structure of the translator (note that you can't use Scaffold for export translators, so you'll just have to test this in Zotero proper.

owcz:
Yep, I'm really familiar with the citation templates—my question is just what specifically needs updating (specific fields for specific item types, I'd presume?)

I know, for instance, that the citation template chokes on `YYYY-MM` formats (which JSTOR is wont to give) and prefers `MMMM YYYY`

The other thing here is that some users will prefer having the citation templates formed with ref tags around them (`<ref name="xxxx">{{cite book ... }}</ref>`) rather than the `{{cite book}}` alone. And some users will also prefer either mdy or dmy formatted dates over the `YYYY-MM-DD` notation. It would be nice to let users select these options but it looks like there isn't a mechanism for that... and we'd end up just offering slightly modified styles off of the default

adam3smith:
Yeah, we won't offer customization (it technically is possible to code translator-specific preferences into translators that can be changed eiterh via checkbox or hidden preference, but we discourage it whenever there isn't a very strong case). So what we're after is, in this order of priority:
1) Something that works reliably with WP and
2) Wherever there is something like a standard for WP, follows that.
3) Where 1) and 2) don't apply, follow general standards. E.g. prefer ISO dates (YYYY-MM-DD) over other formats. 

The list at the top of 
https://forums.zotero.org/discussion/41819/wikipedia-citation-templates-and-their-limitations with following discussion would be a good starting point. The discussion also gives a good sense about the types of considerations. What we'd obviously want to avoid is too much of a favor of your (or anoyone else's) personal preferences, though I understand if you'll have to make judgment calls in some places (ideally the template would be welll defined enough so you wouldn't, but the way I understand it, that's not the case).

owcz:
[...]

PR #973, opened 2015-12-05
Create Polygon.js

owcz:
translator for polygon.com—have a series of translators on the way but wanted to make sure I was on the right track (or whether multiple, related, individual sites should be included within the same translator?)

### DIFF
--- a/Polygon.js
+++ b/Polygon.js
@@ -1,0 +1,169 @@
+{
+	"translatorID": "fa7c37b1-fda4-418a-a8b8-2491929411ab",
+	"label": "Polygon",
+	"creator": "czar",
+	"target": "^https?://(www\\.)?polygon\\.com",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2015-12-06 18:45:12"
+}
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2015 czar
+	http://en.wikipedia.org/wiki/User_talk:Czar
+
+	This file is part of Zotero.
+
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+function pageType(doc, url) { // pulls from <script> metadata set by Vox (Article, HubPage, Other)
+	var temp = ZU.xpathText(doc, '//script[contains(text(), "page_type :")]');
+	if (temp) {
+		return temp.match(/page_type\s?:\s?"([^"]*)"/)[1];
+	} else return null;
+}
+
+function detectWeb(doc, url) {
+	switch (pageType(doc, url)) {
+		case "Article":
+			return "webpage";
+		case "HubPage":
+			return "multiple";
+	}
+	if (url.indexOf("/search?q=") != -1) {
+		return "multiple"
+	} else return null;
+}
+
+function scrape(doc, url) {
+	var newItem = new Zotero.Item("webpage");
+	newItem.websiteTitle = "Polygon";
+	newItem.language = "en-US";
+
+	newItem.url = url;
+	newItem.title = ZU.xpathText(doc, '//h2[@class="m-entry__title"]');
+	newItem.date = ZU.xpathText(doc, '//meta[@property="article:published_time"]/@content').split("T")[0];
+
+	newItem.abstractNote = ZU.xpathText(doc, '//meta[@name="description"]/@content');
+	newItem.attachments.push({
+		document: doc,
+		title: "Polygon snapshot",
+		mimeType: "text/html"
+	});
+
+	// Authors
+	var authorMetadata = ZU.xpathText(doc, '//script[contains(text(), "dataLayer =")]');
+	if (authorMetadata) {
+		authorMetadata = authorMetadata.match(/"Author"\s?:\s?"([^"]*)"/)[1].split(":");
+		do {
+			newItem.creators.push(ZU.cleanAuthor(authorMetadata[0], "author"));
+			authorMetadata.shift();
+		}
+		while (authorMetadata.length);
+	}
+
+	newItem.complete();
+}
+
+function doWeb(doc, url) { // news & search pages supported, Feature support to come
+	if (detectWeb(doc, url) == "multiple") {
+		var items = {};
+		articles = [];
+		var titles = ZU.xpath(doc, '//*[contains(concat(" ", normalize-space(@class), " "), "m-block__meta")]/h2/a') //this is needed to catch the class attribute when among others, e.g., "m-block__meta meta" on search pages
+		for (var i = 0; i < titles.length; i++) {
+			items[titles[i].href] = titles[i].textContent;
+		}
+		Zotero.selectItems(items, function(items) {
+			if (!items) {
+				return true;
+			}
+			for (var i in items) {
+				articles.push(i);
+			}
+			Zotero.Utilities.processDocuments(articles, scrape);
+		})
+	} else scrape(doc, url);
+}
+/** BEGIN TEST CASES **/
+var testCases = [{
+		"type": "web",
+		"url": "http://www.polygon.com/2015/6/14/8779193/earthbound-mother-famicom-wii-u-virtual-console-nintendo",
+		"items": [{
+			"itemType": "webpage",
+			"title": "Nintendo brings Mother to Virtual Console as Earthbound Beginnings — it's available today",
+			"creators": [{
+				"firstName": "Michael",
+				"lastName": "McWhertor",
+				"creatorType": "author"
+			}, {
+				"firstName": "Arthur",
+				"lastName": "Gies",
+				"creatorType": "author"
+			}],
+			"date": "2015-06-14",
+			"abstractNote": "At this year's Nintendo World Championships, Nintendo announced that Wii U owners would be able to get their hands on an EarthBound Beginnings, a Virtual Console re-release of the original game in...",
+			"language": "en-US",
+			"url": "http://www.polygon.com/2015/6/14/8779193/earthbound-mother-famicom-wii-u-virtual-console-nintendo",
+			"websiteTitle": "Polygon",
+			"attachments": [{
+				"title": "Polygon snapshot",
+				"mimeType": "text/html"
+			}],
+			"tags": [],
+			"notes": [],
+			"seeAlso": []
+		}]
+	}, {
+		"type": "web",
+		"url": "http://www.polygon.com/2013/11/11/5090912/sega-genesis-retrospective-book-turns-to-kickstarter-for-funding",
+		"items": [{
+			"itemType": "webpage",
+			"title": "Sega Genesis retrospective book turns to Kickstarter for funding",
+			"creators": [{
+				"firstName": "Griffin",
+				"lastName": "McElroy",
+				"creatorType": "author"
+			}],
+			"date": "2013-11-11",
+			"abstractNote": "Sega Mega Drive/Genesis: Collected Works, a \"documentary art book\" chronicling the creation and full history of Sega's classic home console and its iconic software, has turned to Kickstarter to seek...",
+			"language": "en-US",
+			"url": "http://www.polygon.com/2013/11/11/5090912/sega-genesis-retrospective-book-turns-to-kickstarter-for-funding",
+			"websiteTitle": "Polygon",
+			"attachments": [{
+				"title": "Polygon snapshot",
+				"mimeType": "text/html"
+			}],
+			"tags": [],
+			"notes": [],
+			"seeAlso": []
+		}]
+	}, {
+		"type": "web",
+		"url": "http://www.polygon.com/news",
+		"items": "multiple"
+	}, {
+		"type": "web",
+		"url": "http://www.polygon.com/search?q=earthbound",
+		"items": "multiple"
+	}]
+/** END TEST CASES **/


adam3smith:
How similar are the sites? If they all belong to the same family of sites (e.g. same publisher etc.) we'd want them in one translator. If they just happen to be pretty similar, do separate ones. There are a number of (pretty esay to fix) issues in this one, so hold off on submitting the other ones for now, but let's figure the larger question out first.

owcz:
Not very similar, so I'll keep them separate. (Closest sister site is theverge.com—same publisher—and that site itself has different XPaths.) Now that Wikipedia has a few tools (citation fillers) that use the Zotero translators, I'm looking to add support for some of my common references. Any quick suggestions on this one?

Also the Scaffold documentation isn't in great shape—I tried signing up to edit the wiki but the edit option has disappeared since I logged in and there's no guidance in its stead. And I imagine more volunteers would be spurred to write translators if the Framework option replaced the HWZT training in prominence.

adam3smith:
>Also the Scaffold documentation isn't in great shape—I tried signing up to edit the wiki but the edit option has disappeared since I logged in and there's no guidance in its stead. And I imagine more volunteers would be spurred to write translators if the Framework option replaced the HWZT training in prominence.

As for documentation, there's really two different things:
There's actual documentation, which I think is in pretty good shape: https://www.zotero.org/support/dev/translators/coding (i.e. complete&accurate)
and then there are detailed how-tos and instructions like HWZT which are in a terrible state, mostly because they're old and a ton of work to write. About the Wiki, which page in particular are you referring to? Some pages are locked, in particular those with direct links to installable software, because of security concerns. 

Will comment on the translator in-line.

Polygon.js
```diff
@@ -0,0 +1,85 @@
...
+	var soughtXPath = '//*[@id="article-body"]';
+		evalXPath = doc.evaluate(soughtXPath, doc, nsResolver, XPathResult.ANY_TYPE, null).iterateNext().textContent;
+	newItem.notes = [ evalXPath.trim().replace(/\n/g, " ") ];
+	
+	newItem.attachments.push({url:url, title:newItem.title ,mimeType:"text/html"});
+	newItem.complete();
+}
+
+function doWeb(doc, url) {
+	var articles = new Array();
```
adam3smith: if you're not doing search results and other multiples, all you need for doWeb is to call `scrape(doc, url)`

Polygon.js
```diff
@@ -0,0 +1,85 @@
...
+	articles = [url];
+	Zotero.Utilities.processDocuments(articles, scrape, function(){Zotero.done();});
+	Zotero.wait();
+}
+/** BEGIN TEST CASES **/
+var testCases = []
```
adam3smith: we want test cases -- those should be simple to add with scaffold: 
Simply go on a site this work on, call up Scaffold, go to the Testing tab and click "New Web". Then save the test. We'd want tests to cover all important cases -- in this case, one might be enough.

Polygon.js
```diff
@@ -0,0 +1,85 @@
...
+ 	}
+	
+	return null;
+}
+
+function scrape(doc, url) {
+	var namespace = doc.documentElement.namespaceURI; 
+	var nsResolver = namespace ? function(prefix) {
+		if (prefix == 'x') return namespace; else return null; 
+		} : null;
```
adam3smith: don't use the namespace declaration. If you really want to use doc.evaluate, simply replace nsResolver with null (which is all this function does for any regular html page anyway). But more below.

Polygon.js
```diff
@@ -0,0 +1,85 @@
...
+
+function scrape(doc, url) {
+	var namespace = doc.documentElement.namespaceURI; 
+	var nsResolver = namespace ? function(prefix) {
+		if (prefix == 'x') return namespace; else return null; 
+		} : null; 
+	
+	var newItem = new Zotero.Item("webpage");
+	newItem.url = doc.location.href;
+	newItem.language = "en";
```
adam3smith: "en-US" I'd assume?

Polygon.js
```diff
@@ -0,0 +1,85 @@
...
+	newItem.publicationTitle="Polygon";
+	
+		// Article title
+	var soughtXPath = "//h2";
+		evalXPath = doc.evaluate(soughtXPath, doc, nsResolver, XPathResult.ANY_TYPE, null).iterateNext().textContent;
+	newItem.title = evalXPath;
```
adam3smith: I'd really recommend using the built-in functions: These three lines could just be
`newItem.title = ZU.xpathText(doc, '//h2)`

Polygon.js
```diff
@@ -0,0 +1,85 @@
...
+		// Article text
+	/** not sure why this isn't importing
+	var soughtXPath = '//*[@id="article-body"]/p[1]';
+		evalXPath = doc.evaluate(soughtXPath, doc, nsResolver, XPathResult.ANY_TYPE, null).iterateNext().textContent;
+	newItem.abstract = evalXPath+" ..." ; **/
+	var soughtXPath = '//*[@id="article-body"]';
+		evalXPath = doc.evaluate(soughtXPath, doc, nsResolver, XPathResult.ANY_TYPE, null).iterateNext().textContent;
+	newItem.notes = [ evalXPath.trim().replace(/\n/g, " ") ];
+	
+	newItem.attachments.push({url:url, title:newItem.title ,mimeType:"text/html"});
```
adam3smith: more efficient: use `document:doc` instead of `url:url`

Polygon.js
```diff
@@ -0,0 +1,85 @@
...
+	newItem.publicationTitle="Polygon";
+	
+		// Article title
+	var soughtXPath = "//h2";
+		evalXPath = doc.evaluate(soughtXPath, doc, nsResolver, XPathResult.ANY_TYPE, null).iterateNext().textContent;
+	newItem.title = evalXPath;
+	
+		// Authors
+	var authorCounter = 0;
+	var authorArray = new Array();
```
adam3smith: why not grab author and date from the meta header without this ugly (and fragile) scraping? It looks like they provide that nicely:

```
 <meta property="article:published_time" content="2015-11-24T21:48:47Z" />
  <meta name="author" content="Philip Kollar" />
```
(from http://www.polygon.com/2015/11/24/9793694/league-of-legends-crafting-loot-system-toxic-players-hextech-2016-update)

Polygon.js
```diff
@@ -0,0 +1,85 @@
...
+		// Date
+	var soughtXPath = '//*[@id="article-content"]/header/p';
+		evalXPath = doc.evaluate(soughtXPath, doc, nsResolver, XPathResult.ANY_TYPE, null).iterateNext().textContent;
+	newItem.date = evalXPath.split("on ")[1].split(/\n/)[0];
+	
+		// Article text
+	/** not sure why this isn't importing
+	var soughtXPath = '//*[@id="article-body"]/p[1]';
+		evalXPath = doc.evaluate(soughtXPath, doc, nsResolver, XPathResult.ANY_TYPE, null).iterateNext().textContent;
+	newItem.abstract = evalXPath+" ..." ; **/
```
adam3smith: this should be `newItem.abstracNote`
aurimasv.github.io/z2csl/typeMap.xml
has a full field list.

Again, consider if you can't scrape this from the metaheader.

Polygon.js
adam3smith: Also, by convention we don't title snapshots with article titles. This should just by `title: "Polygon Snapshot"`

Polygon.js
```diff
@@ -0,0 +1,85 @@
...
+	"translatorType": 4,
+	"browserSupport": "g",
+	"lastUpdated": "2015-12-05 20:54:17"
+}
+
+// http://en.wikipedia.org/wiki/User_talk:Czar
+
+function detectWeb(doc, url) {
+	if (doc.title.match("| Polygon")) {
+ 		return "webpage";
```
adam3smith: this looks too lax to me. E.g. would trigger on http://www.polygon.com/videos

adam3smith:
See if those make sense to you, let me know if you have any questions. We'd also really like to have basic multiples covered by translators if at all possible. I know Citoid doesn't use that, but for Zotero it's a huge benefit.

owcz:
This is great feedback. Thanks! I'll give it a look. (What are basic multiples?) Main wiki page I wanted to edit was https://www.zotero.org/support/dev/how_to_write_a_zotero_translator_plusplus and also /Scaffold (the Framework option replaced "In Repository", which isn't explained, as well as exporting from Scaffold). I can see what you mean about security re: downloads, but I also see the page staying in the same condition. (I also couldn't find that type list for the life of me.) Really appreciate your help

Polygon.js
owcz: Issue with this is multiple authors, e.g., [Nintendo brings Mother ...](http://www.polygon.com/2015/6/14/8779193/earthbound-mother-famicom-wii-u-virtual-console-nintendo):
`<meta name="author" content="Michael McWhertor">`
Unless I pull from the `<!-- GTM DataLayer -->` section? Thoughts? This is an edge case but I'd like to be able to handle it

owcz:
All right—all fixed! Thanks again for the suggestions. What do you think?
 -  removed doc.evaluate
 - open to suggestions on the `//script[contains(` portions (needed to pull from unlabeled `<script>` tags)
 - many of the (duly noted) fixes you suggested were parts that came verbatim from the HWZT tutorial, FYI
- also don't miss my now collapsed comment on https://github.com/zotero/translators/pull/973#discussion-diff-46764499 if you have any ideas on parsing the names

Polygon.js
```diff
@@ -0,0 +1,129 @@
...
+	newItem.url = doc.location.href;
+	newItem.title = ZU.xpathText(doc, '//h2[@class="m-entry__title"]');
+	newItem.date = ZU.xpathText(doc, '//meta[@property="article:published_time"]/@content').split("T")[0];
+
+	newItem.abstractNote = ZU.xpathText(doc, '//meta[@name="description"]/@content');
+	newItem.notes = [ ZU.xpathText(doc, '//*[@id="article-body"]').trim() ];
+	newItem.attachments.push({document:doc, title:"Polygon snapshot" ,mimeType:"text/html"});
+
+		// Authors
+	var authors = ZU.xpathText(doc, '//script[contains(., "dataLayer =")]').split('\"Author\":\"')[1].split(/\",/)[0].split(":");
```
adam3smith: I like this in general and I totally agree we want all authors. Two suggestions:
1) `contains(text(), "dataLayer")` gives us at least a little bit more efficiency.
2) I'm concerned that this can too easily break in a way that will have the entire translator fail. So I'd suggest first doing just the `ZU.xpathText(...)` and then doing `if (authors)` and your parsing and splitting.

Polygon.js
```diff
@@ -0,0 +1,129 @@
...
+	var newItem = new Zotero.Item("webpage");
+	newItem.websiteTitle = "Polygon";
+	newItem.language = "en-US";
+
+	newItem.url = doc.location.href;
+	newItem.title = ZU.xpathText(doc, '//h2[@class="m-entry__title"]');
+	newItem.date = ZU.xpathText(doc, '//meta[@property="article:published_time"]/@content').split("T")[0];
+
+	newItem.abstractNote = ZU.xpathText(doc, '//meta[@name="description"]/@content');
+	newItem.notes = [ ZU.xpathText(doc, '//*[@id="article-body"]').trim() ];
```
adam3smith: we don't put the whole article in notes. For Zotero users, they get the snapshot. Please delete this line.

Polygon.js
```diff
@@ -0,0 +1,129 @@
...
+	if (pageType == "Article") { return "webpage" } 
+	return null;
+}
+
+function scrape(doc, url) {
+	var newItem = new Zotero.Item("webpage");
+	newItem.websiteTitle = "Polygon";
+	newItem.language = "en-US";
+
+	newItem.url = doc.location.href;
```
adam3smith: `newItem.url = url`

Polygon.js
```diff
@@ -0,0 +1,129 @@
...
+	newItem.title = ZU.xpathText(doc, '//h2[@class="m-entry__title"]');
+	newItem.date = ZU.xpathText(doc, '//meta[@property="article:published_time"]/@content').split("T")[0];
+
+	newItem.abstractNote = ZU.xpathText(doc, '//meta[@name="description"]/@content');
+	newItem.notes = [ ZU.xpathText(doc, '//*[@id="article-body"]').trim() ];
+	newItem.attachments.push({document:doc, title:"Polygon snapshot" ,mimeType:"text/html"});
+
+		// Authors
+	var authors = ZU.xpathText(doc, '//script[contains(., "dataLayer =")]').split('\"Author\":\"')[1].split(/\",/)[0].split(":");
+	newItem.creators = new Array();
```
adam3smith: don't needs this. Every Zotero item already has an emptry creator array. delete.

Polygon.js
```diff
@@ -0,0 +1,129 @@
...
+	do {
+		newItem.creators.push({ 
+			firstName:authors[0].substring(0,authors[0].lastIndexOf(" ")),
```
adam3smith: How about instead of 40-44, using:
`newItem.creators.push(ZU.cleanAuthor(authors[0], "author"));` (really useful utility)

Polygon.js
```diff
@@ -0,0 +1,129 @@
...
+			creatorType:"author" 
+		});
+		authors.shift();
+	}
+	while (authors.length);
+	
+	newItem.complete();
+}
+
+function doWeb(doc, url) {
```
adam3smith: To get you a sense of what I mean by multiples, try something like:
```
	    var items = {};
	    articles = [];
	    var titles = ZU.xpath(doc, '//div[@class="m-block__meta meta"]/h2/a')
	    for (var i = 0; i < titles.length; i++) {
	        items[titles[i].href] = titles[i].textContent;
	    }
	    Zotero.selectItems(items, function(items) {
	        if (!items) {
	            return true;
	        }
	        for (var i in items) {
	            articles.push(i);
	        }
	        Zotero.Utilities.processDocuments(articles, scrape);
	    })
	}
```

(followed by `else` obviously). You'd then have to test for multiples in `detectWeb`, too, something like 
```	
else if (url.indexOf("/search?q=")!=-1){
		return "multiple"
	}
```

This would be for search results. You can see if you also want to include landing pages. Does that make sense? I realize not your core interest since I don't think Citoid ever uses this, but it makes the translator a lot more useful for researchers wanting to ingest large amounts of data.

adam3smith:
Looking much much better already. Hope you don't mind if we go a couple of iterations here. I think that will pay of pretty quickly. Let me know if you have any questions.

I'll check on the doc pages. Those aren't generally locked down, so I'll have to ask what going on there.

If you have a couple of people on WP interested, we could even consider doing a webinar on translator coding. I've done this once before with PLoS, that one was more for people interested in dataset citations, but especially if Wikimedia foundation could host, I'd be happy to do one more focussed on scraping, with and w/o Framework.

Polygon.js
```diff
@@ -0,0 +1,129 @@
+{
+	"translatorID": "fa7c37b1-fda4-418a-a8b8-2491929411ab",
+	"label": "Polygon",
+	"creator": "czar",
+	"target": "^https?://(?:www\\.)polygon\\.com",
```
zuphilip: Any reason you have the `www` in parenthesis here? Is it optional (then you would need to add a question mark afterwards)? Otherwise I guess you can delete the parenthesis around the `www`.

Polygon.js
```diff
@@ -0,0 +1,129 @@
...
+	"minVersion": "1.0",
```
zuphilip: New translators should have `3.0` as `minVersion`. (Actually, the newest scaffold version should have this as default value as well.)

Polygon.js
```diff
@@ -0,0 +1,129 @@
...
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "g",
```
zuphilip: Any reason to restrict this to `g` only? By default you should add all options here.

Polygon.js
```diff
@@ -0,0 +1,129 @@
...
+	"lastUpdated": "2015-12-06 02:04:51"
+}
+
+// http://en.wikipedia.org/wiki/User_talk:Czar
```
zuphilip: You have to add a licence block here similar to https://github.com/zuphilip/translators/wiki/Common-code-blocks-for-translators#licence-block

zuphilip:
What is the difference from the currently available metadata translator for this website? In the two examples I have seen the creators and the date, which are not present in the result from the metadata translator. Actually, you are already using a lot of the `meta` fields here and therefore it might be a better idea to first call the metadata translator. @adam3smith What do you think?

Polygon.js
owcz: My Scaffold 3.2.1 opens with an Untitled file with a `minVersion` of 1.0, FYI

Polygon.js
owcz: I didn't test the other versions, but I can add the options anyway

Polygon.js
owcz: Is this necessary? I saw that many translators did not have such a block. If it is, can I use a freer license than the GPL3 for the translator?

adam3smith:
Thanks @zuphilip -- I did think about the Embeded Metadata translator, but since it would be for three lines of code or so, it didn't seem worth it. Good thing to keep in mind in general, though @owcz, this is good to keep in mind as an option. See e.g. the MDPI Journals translator for a quick implementation.

Polygon.js
adam3smith: yeah, we are making sure all new translators have a license block. Since Zotero is AGPL licensed, it makes sense to use that; I'd rather not have mixed licensing. But if you want to dual license it and add, say, an MIT license, that's fine.

Polygon.js
adam3smith: yes, this is so simple it'll work on all setups. You can generally assume that all pure scrapers (i.e. you're not calling up any other pages/translators) do (as do most other translators).

Polygon.js
owcz: What part do you see breaking? Haven't had any issues in my tests so far. Do you mean an `if` check within or instead of the `do while` loop?

adam3smith: If you have an article without authors and `ZU.xpathText` returns `null`, then the entire translator fails because of `.split()` being called on `null`. We generally want things to be more robust. I didn't see any examples of this in a quick scan of the site, so it's not addressing an urgent issue but rather protecting us against its possibility.

adam3smith: And I mean the if check before the loop. Specifically before calling `.split()` -- since the loop depends on `authors.length`, i.e. like most loops has an implicit `if` that won't cause any issues.

Polygon.js
```diff
@@ -0,0 +1,167 @@
...
+            if (!items) {
+                return true;
+            }
+            for (var i in items) {
+                articles.push(i);
+            }
+            Zotero.Utilities.processDocuments(articles, scrape);
+        })
+    }
+	scrape(doc, url);
```
adam3smith: you want this to be 
```
else {
  scrape(doc, url)
}
```

otherwise you're also calling scrape directly on search results etc. which is causing the break you're seing.

adam3smith: Could you also add a test for one of the multiples? Works the same way, though you'll see the test looks a lot simpler. It just checks that importing the first item returns something.

Polygon.js
```diff
@@ -0,0 +1,167 @@
+{
+	"translatorID": "fa7c37b1-fda4-418a-a8b8-2491929411ab",
+	"label": "Polygon",
+	"creator": "czar",
+	"target": "^https?:\\/\\/(www.)?polygon.com",
```
adam3smith: You've lost a bunch of things that were right before. You want
`^https?://(www\.)?polygon\.com`
in scaffold, which will then give you 
`^https?://(www\\.)?polygon\\.com`
in the actual translator.

Polygon.js
```diff
@@ -0,0 +1,167 @@
...
+    GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
+    along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+    ***** END LICENSE BLOCK *****
+*/
+
+function pageType (doc, url) { 
+	return ZU.xpathText(doc, '//script[contains(text(), "page_type :")]').split('page_type : \"')[1].split(/\"/)[0];
```
adam3smith: kind of the same issue as below. This throws an error when this evaluates to null. I actually did find an example for this:
http://www.polygon.com/rss/index.xml
While this isn't fatal (detectWeb also just returns false when it throws an error) we still want to avoid this, so have the `pageType` function return null when `ZU.xpathText` returns null.

adam3smith:
I think the comments above should be the last of it from me (though not making promises yet ;)

Polygon.js
owcz: Why doesn't Zotero.debug work in detectWeb and is there another option for debugging output there?

Polygon.js
zuphilip: I suggest also to use a regexp to extract the relevant part, something like `match('/"Author"\s?:\s?"([^"]*)"/')`

Polygon.js
owcz: @adam3smith: why does the rest of the script run on the RSS page when it failed `detectWeb`? 

Would it be better to put everything in the `scrape` under an `if` that repeats the function of `detectWeb`, or is there a better method?

Polygon.js
adam3smith: @zuphilip - any particular reason? I personally prefer using `match` over
`split`, too, but I don't see any reason to reject split for this. But I
may be overlooking something?

On Sun, Dec 6, 2015 at 10:18 AM, Philipp Zumstein <notifications@github.com>
wrote:

> In Polygon.js
> <https://github.com/zotero/translators/pull/973#discussion_r46772348>:
>
> > +
> > +function scrape(doc, url) {
> > +	var newItem = new Zotero.Item("webpage");
> > +	newItem.websiteTitle = "Polygon";
> > +	newItem.language = "en-US";
> > +
> > +	newItem.url = doc.location.href;
> > +	newItem.title = ZU.xpathText(doc, '//h2[@class="m-entry__title"]');
> > +	newItem.date = ZU.xpathText(doc, '//meta[@property="article:published_time"]/@content').split("T")[0];
> > +
> > +	newItem.abstractNote = ZU.xpathText(doc, '//meta[@name="description"]/@content');
> > +	newItem.notes = [ ZU.xpathText(doc, '//*[@id="article-body"]').trim() ];
> > +	newItem.attachments.push({document:doc, title:"Polygon snapshot" ,mimeType:"text/html"});
> > +
> > +		// Authors
> > +	var authors = ZU.xpathText(doc, '//script[contains(., "dataLayer =")]').split('\"Author\":\"')[1].split(/\",/)[0].split(":");
>
> I suggest also to use a regexp to extract the relevant part, something
> like match('/"Author"\s?:\s?"([^"]*)"/')
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/zotero/translators/pull/973/files#r46772348>.
>



-- 
Sebastian Karcher, PhD
www.sebastiankarcher.com

Polygon.js
```diff
@@ -0,0 +1,177 @@
...
+
+	newItem.url = url;
+	newItem.title = ZU.xpathText(doc, '//h2[@class="m-entry__title"]');
+	newItem.date = ZU.xpathText(doc, '//meta[@property="article:published_time"]/@content').split("T")[0];
+
+	newItem.abstractNote = ZU.xpathText(doc, '//meta[@name="description"]/@content');
+	newItem.attachments.push({document:doc, title:"Polygon snapshot" ,mimeType:"text/html"});
+
+		// Authors
+	var authors = ZU.xpathText(doc, '//script[contains(text(), "dataLayer =")]').split('\"Author\":\"')[1].split(/\",/)[0].split(":");
```
adam3smith: so I guess I was unclear here. What I mean is something like this:

```
var authors = ZU.xpathText(doc, '//script[contains(text(), "dataLayer =")]');
if (authors){
  authors = authors.split('\"Author\":\"')[1].split(/\",/)[0].split(":");
}
```
That way, if you don't have a data layer this won't break.  Looking at this, though, it would still break if there are no authors in the data layer (because [1] wouldn't exist in the returned array after the first split). I won't insist, but if you can find a way to avoid that scenario, that'd be great.

Polygon.js
```diff
@@ -0,0 +1,177 @@
...
+	var newItem = new Zotero.Item("webpage");
+	newItem.websiteTitle = "Polygon";
+	newItem.language = "en-US";
+
+	newItem.url = url;
+	newItem.title = ZU.xpathText(doc, '//h2[@class="m-entry__title"]');
+	newItem.date = ZU.xpathText(doc, '//meta[@property="article:published_time"]/@content').split("T")[0];
+
+	newItem.abstractNote = ZU.xpathText(doc, '//meta[@name="description"]/@content');
+	newItem.attachments.push({document:doc, title:"Polygon snapshot" ,mimeType:"text/html"});
```
adam3smith: nit: fix punctuation/spacing -- `, mimeType`

Polygon.js
owcz: I think I got it (you were just a few seconds ahead of the commit!) 😜

Polygon.js
```diff
@@ -0,0 +1,185 @@
...
+	        Zotero.selectItems(items, function(items) {
+	            if (!items) {
+	                return true;
+	            }
+	            for (var i in items) {
+	                articles.push(i);
+	            }
+	            Zotero.Utilities.processDocuments(articles, scrape);
+	        })
+		default:
```
adam3smith: there's really no need for this. See above. doWeb will never run (there's in no way to make it run) when detectWeb doesn't return a non-empty string. (unless Citoid does this differently from Zotero, but I don't think it does).
I'd prefer we stick with the if/else syntax we had before, since that's how it's done in all other translators, which makes it much faster to read.

Polygon.js
```diff
@@ -0,0 +1,185 @@
...
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+    GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
+    along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+    ***** END LICENSE BLOCK *****
+*/
+
+function pageType (doc, url) {
```
adam3smith: nit: don't use variables without declaring them. i.e. add
`var temp;` here.

Polygon.js
owcz: When I revert back to `else scrape(doc, url);` in `doWeb`, your RSS example throws an error even though it shouldn't be getting past `detectWeb`

adam3smith: right, but the only way you can run `doWeb` when `detectWeb` returns nothing is in Scaffold (and there, you can also run it on a site outside of the target regex. Like you could run it here on github, in which case you'd also get an error). In Zotero itself, doWeb only becomes available after detectWeb returns a legal itemType. 
And I'm assuming Citoid, too, will stop as soon as detectWeb fails (otherwise it'd be running into way more problems).

adam3smith: or to put it more simply: errors in doWeb that only occur on sites that fail the regex and/or `detectWeb` are no problem.

owcz: That's what I thought but since the above edits had me going for completeness and now that the `switch` protects against those cases, I'd prefer to keep the `switch` over `if/else` (unless you insist?)

adam3smith: I'm fine with that, but then please fix the indenting (you can just run the entire transltor through
http://jsbeautifier.org/ make sure to set it to indent with tab.)

owcz:
Okay! Took a bit longer than I had expected but I'm much more confident with the translator structure now. Thank you both for your help! Anything else?

```
         -   "libraryCatalog": "Polygon"
         -   "accessDate": "CURRENT_TIMESTAMP"
```
Scaffold lists these at the bottom of the `Returned item`. The minus denotes types that aren't actually used in the `newItem`, right? But since I didn't explicitly assign these fields, something is generating them...

My next translator should be much smoother—wish I knew about some of the methods shortcuts in the tut. (I think many of the proposed edits above would be helpful to include in the revised HWZT eventually.) I'll reach out to Wikipedians about a translator workshop too

Polygon.js
adam3smith: actually, come to think of it -- I think the desirable behavior for actually be to fail noisily and not silently: From the Zotero perspective, if a user clicks the URL bar icon, it is much better to throw a visible error than to fail silently, because they may not notice the latter. That's different from partial failures to import, in which case the user experience of maybe missing authors is preferable to the whole thing breaking (though those are sometimes judgment calls: e.g. we have Zotero purpusefully fail when items don't have a title).
So yeah, not liking the precendent of the default: return null, even if it'll never fire in this case, sorry.

avram:
As far guidance goes, the link @adam3smith gave previously, https://www.zotero.org/support/dev/translators/coding, is quite complete and does follow current best practices including all the shortcuts.

adam3smith:
sorry, see my later comment on the default for the switch in doWeb. 

After that -- do you know how to rebase&squash commits? We'd like this all in a single commit with a simple commit message (i.e. don't keep the entire history in the commit message either).
See e.g. http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

adam3smith:
libraryCatalog and accessedDate are automatically added to every item. libraryCatalog is not present for webpages, which is why that's removed. accessedDate is just a display quirk in Scaffold, I think -- as you'll see if you're actually using Zotero, that does get included.

adam3smith:
yup, all perfect now, thanks! 
If you're planning to contribute a bunch, might be worth figuring out rebase&squash. It's something we request very commonly. But if you just want this in, I can commit it for you. Let me know.

owcz:
Squashed 🍆

adam3smith:
perfect thanks! This is live immediately on the Zotero side. You'll have to see what Citoid does.

adam3smith:
For HWZT -- I'd actually just encourage re-writing that entirely (probably as a github page) rather than editing. We'd then remove it from the Zotero Wiki. 
It was an amzing resource at the time, but essentially all of it is outdated now, both in terms of the tools and in terms of Zotero, and can be done much simpler (e.g. no more need for any external tools for xpaths since Firefox's "Inspect Element". It's also entirely for scrapers and so ignores import translators entirely, where with more and more metadata on websites we'll likely use those more rather than less.
It's on my list of things that I want to do, so if someone else wants to start it, I'd be happy to contribute.

@dstillman -- is the dev section of the wiki restricted? I can edit e.g.  https://www.zotero.org/support/dev/translators/scaffold but @owcz reports not seeing an edit option for that after logging in.

owcz:
@adam3smith re: wiki edit access, in case you were curious:

> You have to request editing access by emailing support@zotero.org with your wiki username. ([link](https://github.com/zotero/translators/issues/797#issuecomment-225340480))

PR #976, opened 2015-12-08
Create Rock, Paper, Shotgun.js

owcz:
translator for [Rock, Paper, Shotgun](http://rockpapershotgun.com), written in Framework

### DIFF
--- a/Rock, Paper, Shotgun.js
+++ b/Rock, Paper, Shotgun.js
@@ -1,0 +1,168 @@
+{
+	"translatorID": "5a5b1fcd-8491-4c56-84a7-5dba19fe2c02",
+	"label": "Rock, Paper, Shotgun",
+	"creator": "czar",
+	"target": "^https?://(www\\.)?rockpapershotgun\\.com",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2015-12-08 06:01:10"
+}
+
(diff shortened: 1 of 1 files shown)


adam3smith:
something went wrong here, possibly related to the rebase. You don't want to add the Polygon translator again.

Rock, Paper, Shotgun.js
```diff
@@ -0,0 +1,149 @@
...
+FW.Scraper({
+itemType         : 'blogPost',
+detect           : FW.Url().match(/\/(\d{4})\/(\d{2})\/(\d{2})\//),
+title            : FW.Xpath('//header/h2/a').text().trim(),
+attachments      : [{ url: FW.Url(),
+  title:  "Rock, Paper, Shotgun snapshot",
+  type: "text/html" }],
+creators         : FW.Xpath('//header/aside/p/a').text().cleanAuthor("author"),
+date             : FW.Url().match(/(\d{4})\/(\d{2})\/(\d{2})/).replace(/\//g,"-"),
+tags             : FW.Xpath('//footer/p[@class="tags"]').text(),
```
adam3smith: `.split(/,\s*/)` it would seem from looking at the tests, right?

Rock, Paper, Shotgun.js
```diff
@@ -0,0 +1,149 @@
...
+
+	***** END LICENSE BLOCK *****
+*/
+
+function detectWeb(doc, url) { return FW.detectWeb(doc, url); }
+function doWeb(doc, url) { return FW.doWeb(doc, url); }
+
+FW.Scraper({
+itemType         : 'blogPost',
+detect           : FW.Url().match(/\/(\d{4})\/(\d{2})\/(\d{2})\//),
```
adam3smith: why the parentheses here and below? You're not using the different matched elements, so might as well leave them out (and gain some admittedly minimal efficiency)

adam3smith:
For the multi-scrapers (or really for all scrapers) you can simply have several of them. See e.g. https://github.com/zotero/translators/blob/master/Wikimedia%20Commons.js for a complex version of a FW translator. i.e. in your case

```
FW.MultiScraper({
itemType         : 'multiple',
detect           : FW.Url().match(/\/category\//), // category pages
choices          : {
  titles :  FW.Xpath('//div[@class="post-inner"]/h2/a').text().trim(),
  urls    :  FW.Xpath('//div[@class="post-inner"]/h2/a').key("href")
}
});

FW.MultiScraper({
itemType         : 'multiple',
detect           : FW.Url().match(/\/\?s\=/), // search pages
choices          : {
  titles :  FW.Xpath('//li/div/h3/a').text().trim(),
  urls    :  FW.Xpath('//li/div/h3/a').key("href")
}
});
```

They then detect sequentially (i.e. the first one to detect fires, any later ones are ignored).

adam3smith:
That should be all. Thanks!

owcz:
Squashed 🍆

adam3smith:
That was faster. Nice work & thanks.

owcz:
@adam3smith: What's the easiest way to have a translator (such as this one) default to EM for all fields not specified? It looks like EM can do almost all of the work for this translator besides for the author field. (I'm going to email the pub about supporting full metadata but thought I should know for the future.)

adam3smith:
https://github.com/zotero/translators/blob/master/Le%20Monde.js
is a good example. In other words, call the EM translator, and then fix everything you need to in the item handler, where you can modify what's already in the item as well as add additional fields.

More generally, @zuphilip has collected a very good set of sample codes and translators for specific situations here: https://github.com/zuphilip/translators/wiki 
I very much recommend both his code blocks and the list of common mistakes.

owcz:
Nice. Thanks! So just to clarify, there's no EM fallback option baked into Z Framework?

adam3smith:
correct, you have to explicitly call it.

Issue #932, opened 2015-08-09
[EM] add support for additional OpenGraph formats


adam3smith:
http://ogp.me/
haven't checked in detail what we currently recognize, but we'd want to support ```article:```, ```book:```, ```video:```, and ```music:```

adam3smith:
E.g. this has some of the ```article:``` tags: https://hbr.org/2015/08/how-to-do-walking-meetings-right

adam3smith:
Also add the simple meta name="author": http://www.w3schools.com/tags/att_meta_name.asp

zuphilip:
In the PR #973 there are also examples with `og` as well as `<meta name="author" content="..."/>`.

adam3smith:
closed in https://github.com/zotero/translators/pull/998

owcz:
Not OG but for EM, I think the following should be supported:

 1. `<span itemprop="name" content="Andrew Goldfarb">` ([ex](http://www.ign.com/articles/2012/07/27/limbo-follow-up-project-2-is-at-least-2-years-away)) for author ([spec](https://www.w3.org/TR/2009/WD-html5-20090825/microdata.html))
 2. `<meta name="live_date" value="2012-07-27">` ([ex](http://www.ign.com/articles/2012/07/27/limbo-follow-up-project-2-is-at-least-2-years-away)) for date
 - (then the _IGN_ translator can be deprecated)
 - `<meta name="pub_date" content="2015-12-21T20:15:19+00:00">` ([ex](http://www.pcgamer.com/2015-personal-pick-cibele/)) for date
 -  `<time class="meta__time updated" datetime="2014-08-04T18:00:00-04:00">` ([ex](http://kotaku.com/shes-making-an-honest-game-about-sex-1615923190)) for date
 - Should [_Wired_](http://www.wired.com/2015/05/sex-videogames-make-feel-weird-thats-point/) be using `<meta itemprop="datePublished" content="2015-05-25T05:45:37+00:00">` or `<meta name="DisplayDate" content="2015-05-25">` instead of what it's using right now?
 - [_Eurogamer_](http://www.eurogamer.net/articles/2015-11-02-cibele-review) would use `<span itemprop="datePublished" content="2015-11-05">` too (many of these sites still need custom author support—I'll work on that)
[...]

PR #977, opened 2015-12-08
Create Vox.js

owcz:
translator for vox.com

### DIFF
--- a/Vox.js
+++ b/Vox.js
@@ -1,0 +1,89 @@
+{
+	"translatorID": "797c7c7a-bb43-4b9e-bc6e-63b720fa1bb9",
+	"label": "Vox",
+	"creator": "czar",
+	"target": "^https?://(www\\.)?vox\\.com",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2015-12-08 17:24:33"
+}
+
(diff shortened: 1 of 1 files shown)


owcz:
I was able to pull lots of simple metadata for this one. Would it be be better to merge date and author detection into the standard metadata scraper instead? That would be:
```
creators         : FW.Xpath('//meta[@name="author"]/@content').text().cleanAuthor("author"),
date             : FW.Xpath('//meta[@property="article:published_time"]/@content').text().split("T"),
```
(The metadata scraper only pulled the title and description without this translator.)

Also the multiple looks like it's handled by Google CSE—is there a translator that does that already? I would think that it's a common multiple to detect.

adam3smith:
Yes, I'd like these added to EM. There is an open ticket for article:property type metatags:
https://github.com/zotero/translators/issues/932

The meta[@name="author"] is a basic HTML metatag. I guess concern here might be that data quality is too low, but I think it should be fine. http://www.w3schools.com/tags/att_meta_name.asp.

I don't think we have a recyleable implementation of a google CSE, so we'd need that for sure. 

So let's wait on this translator & I'll see what I can do with EM (or you can, if you're feeling adventurous).

adam3smith:
@owcz -- OK, the EM/RDF changes are now live and the Vox.com data all imports. If you still want to do this, you can simply call the EM translator for pages and would just have to call multiples.

owcz:
@adam3smith – nice! Looks good. Only thing I'd add is that it might be good to support the "language" parameter, and that it might be worth splitting the date (`2016-01-16T16:00:03Z` → `2016-01-16`), though I know Zotero usually does that on export anyway.

adam3smith:
for language you mean the lang attribute of html? Looks like that might be worth it, yes. Not sure how often that turns out to be wrong, though. We'd be particularly concerned about false positive "en" declarations, since those will mess with citation styles. I looked at a couple of larger newspapers and they're all correct (Spiegel, Le Monde, El Pais), but less certain about smaller sites. 

If you look at how the data actually imports, Zotero understands the date including the time internally, so I don't see any reason to chop off the time. Some people may enjoy having it included (e.g. if you study something like online journalism) and as you say, it's getting ignored on export, both to citations and to export formats.

owcz:
The original pull request had `language         : "en-US",`, which doesn't appear to trigger in EM—just wanted to make sure it was accounted if necessary. Re: the latter, sounds good

adam3smith:
oh right. If you were to write a translator, you'd have to explicitly set that in the item handler. The EM translator can't get that at the moment and as per above I'm not sure if we should get it from the only place I currently see it on vox.

PR #979, opened 2015-12-15
DABI translator

jmiba:
translator for http://dabi.ib.hu-berlin.de


adam3smith:
before I get into a detailed review of this, wouldn't it make more sense to just use the bibtex which they usefully provide simply by changing modus=html to modus=bibtex in the URL?

jmiba:
Well, I don't think so, really. The convenience of saving a reference with a single click is what people expect. The bibtex is rather an imposition, especially for the non-techies, don't you think?

adam3smith:
oh, that's not what I meant, sorry. Of course this should still work with one click. 
I just mean have the Zotero translator get the bibtex and import it rather than scraping the data from the page. See e.g. https://github.com/zotero/translators/blob/master/Stanford%20Encyclopedia%20of%20Philosophy.js for a template.

jmiba:
O, I see. Didn't know it was possible. Seems reasonable. I'll have a look into it, thanks.

jmiba:
I just realized that the bitex is incomplete as compared with the database html. So I would have to add bits and pieces from the html anyway. Perhaps it is easier to do a complete scrape of the webpage instead of harvest the bibtex and enrich it with scrapings (if that's possible at all).

adam3smith:
OK, looked good to me, but I only glanced very quickly (what's missing?). We can do this as a pure scrape, then.

For a start, could you run this through http://jsbeautifier.org/ set to indent with tabs (or any comparable tool)? It's really hard to read without any indenting.

jmiba:
I think it should be good now. Do you?

DABI.js
```diff
@@ -0,0 +1,677 @@
+{
+    "translatorID": "5cf8bb21-e350-444f-b9b4-f46d9fab7827",
+    "label": "DABI",
+    "creator": "Jens Mittelbach",
+    "target": "dabi.ib.hu-berlin\\.de/.*?",
```
aurimasv: Use `"^https?://dabi\\.ib\\.hu-berlin\\.de/"` (there's no point in adding optional matches to the end, since they don't restrict anything)

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+    "minVersion": "1.0",
+    "maxVersion": "",
+    "priority": 100,
+    "inRepository": true,
+    "translatorType": 4,
+    "browserSupport": "g",
```
aurimasv: Should be ok to use "gcsvbi" here

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+    "lastUpdated": "2015-12-15 18:59:56"
+}
+
+/* FW LINE 57:6869c32952b1 */
```
aurimasv: I think this happened because of jsbeautify, but we definitely do not want to expand the framework code. I would suggest removing all of the code from this line down to the `function detectWeb...` and then re-saving the translator in Scaffold with the "Use translator framework" checkbox ticked (Metadata tab)

aurimasv: Oh, nvm. Seems like you're not actually using the translator framework. Just remove the lines then.

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+    Zotero.wait()
+};
+
+function detectWeb(doc, url) {
+    //namespace code
+    var namespace = doc.documentElement.namespaceURI;
+    var nsResolver = namespace ? function(prefix) {
+        if (prefix == 'x') return namespace;
+        else return null;
+    } : null;
```
aurimasv: you don't need any of this namespace code any more (you must have been using some very old translator as a template)

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+
+    if (doc.title.match("DABI. Datensatz Vollanzeige")) {
```
aurimasv: `doc.title.indexOf("DABI. Datensatz Vollanzeige") == 0` or even `doc.title == "DABI. Datensatz Vollanzeige"` would be preferred. (avoid regexp when possible)

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+        return "journalArticle";
+    } else if (doc.title.match("DABI: Rechercheergebnis")) {
+        var treffer = doc.evaluate('//*[text()="Keine Treffer für die Suche nach:"]', doc, nsResolver, XPathResult.ANY_TYPE, null).iterateNext();
```
aurimasv: As a general note, use `ZU.xpath(doc, "<your xpath here").length` instead (`.length` is 0 if there is no match).

However, that xpath is quite heavy (it iterates through every node on the page and checks the text content... that could be very very slow). What you should do is use the same code that you use to collect items for the multiples selection dialog in `doWeb` and check here if there are any items to display. See https://github.com/zotero/translators/blob/e9423311b4a492dc7862c7d220f1ac5c23246b82/Engineering%20Village.js for an example.

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+
+    //variable declarations
+    var articles = new Array();
+    var items = new Object();
+    var nextLink;
+
+    //If Statement checks if page is a Search Result, then saves requested Items
+    if (detectWeb(doc, url) == "multiple") {
+        var links = doc.evaluate('//tr/td[1]/a', doc, nsResolver, XPathResult.ANY_TYPE, null);
+        var title = doc.evaluate('//tr/td[3]', doc, nsResolver, XPathResult.ANY_TYPE, null);
```
aurimasv: It would be better to iterate through each row and pick out the 2nd and 4th cells. Use `doc.getElementsByTagName("tr")` to select rows more efficiently. You can use the same for `td` afterwards.

(Also, this is the part of code that you should reuse in `detectWeb`)

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+
+        while (nextLink = links.iterateNext()) {
+            items[nextLink.href] = title.iterateNext().textContent;
+
+        }
+        items = Zotero.selectItems(items);
```
aurimasv: See the Engineering Village translator above for the proper way to use `Zotero.selectItems`. This usage is outdated. `Zotero.selectItems` should now be used asynchronously with a callback.

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+        for (var i in items) {
+            articles.push(i);
+        }
+
+    } else if (detectWeb(doc, url) == "journalArticle") {
+        //saves single page items
+        articles = [url];
+    } else {
+        return false;
```
aurimasv: This should never be the case, because `detectWeb` should always guarantee that there is something to scrape.

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+    }
+
+    //Tells Zotero to process everything. Calls the "scrape" function to do the dirty work.
+    Zotero.Utilities.processDocuments(articles, scrape, function() {
+        Zotero.done();
```
aurimasv: `Zotero.done()` is deprecated. You don't need this callback at all.

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+    });
+    Zotero.wait();
```
aurimasv: `Zotero.wait()` is also deprecated. You do not need it at all.

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+    //Translator is FINISHED after running this line. Note: code doesn't run from top to bottom only.
+}
+
+//The function used to save well formatted data to Zotero
+function associateData(newItem, items, field, zoteroField) {
+    if (items[field]) {
+        newItem[zoteroField] = items[field];
+    }
+}
```
aurimasv: Personally, I wouldn't make a function just to avoid an `if` statement, but if you really want to, I would declare this function inside scrape and then avoid passing in `newItem` and `items`, since those would be in scope anyway.

aurimasv: Or, you could create a mapping between labels and Zotero fields and then loop through the mappings to assign the data. That would be my preference.

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+    //namespace code
+    var namespace = doc.documentElement.namespaceURI;
+    var nsResolver = namespace ? function(prefix) {
+        if (prefix == 'x') return namespace;
+        else return null;
+    } : null;
+
+    //variable declarations
+    var newItem = new Zotero.Item('journalArticle');
+    newItem.url = doc.location.href;
```
aurimasv: Use the `url` passed into function.

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+    newItem.title = "No Title Found";
```
aurimasv: If you have a good reason to believe that a title might be missing from metadata (i.e. you can find examples where that is the case), you should not be assigning anything to the title. Zotero automatically rejects items without titles during import and sends an error report to zotero.org. Missing titles are generally a good indicator that something about the translator is not working (probably the website was updated). This allows us to catch issues quickly.

If you do expect titles to be missing, then you should use something like `"[No Title]"` (@adam3smith do you remember if we had anything specific)

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+
+    var items = new Object();
+    var headers;
+    var contents;
+    var link;
```
aurimasv: Declare variables as they are needed and close to where they are being used.

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+
+    var myXPathObject = doc.evaluate('//tr/th', doc, nsResolver, XPathResult.ANY_TYPE, null);
+    var myXPathObject2 = doc.evaluate('//tr/td', doc, nsResolver, XPathResult.ANY_TYPE, null);
+    var link = doc.evaluate('//tr/td/a', doc, nsResolver, XPathResult.ANY_TYPE, null).iterateNext();
+
+    //While Loop to populate "items" Object and save tags to an Array.
+    while (headers = myXPathObject.iterateNext()) {
+
+        headersTemp = headers.textContent;
+        if (!headersTemp.match(/\w/)) {
```
aurimasv: `/\w/.test(headersTemp)` (match is unnecessary and more expensive)
Wouldn't it make more sense to just skip the row? (i.e. `continue;`)

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+    var items = new Object();
+    var headers;
+    var contents;
+    var link;
+    var blankCell = "temp";
+    var headersTemp;
+    var tagsContent = new Array();
+
+    var myXPathObject = doc.evaluate('//tr/th', doc, nsResolver, XPathResult.ANY_TYPE, null);
+    var myXPathObject2 = doc.evaluate('//tr/td', doc, nsResolver, XPathResult.ANY_TYPE, null);
```
aurimasv: Iterate through each row first, then select `th` and `td` in each. The way you do it is prone to mismatch errors.

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+        }
+
+        items[headersTemp.replace(/\s+/g, '')] = contents.replace(/^\s*|\s*$/g, '');
+
+    }
+
+    //Formatting and saving "title" fields
+    if (items["Titel"]) {
+        newItem.title = items["Titel"].replace(/\*/g, '');
+        var short = newItem.title.replace(/^(Die )|(Der )|(Das )/, '');
```
aurimasv: Capturing parentheses are unnecessary. I think what you actually want is `/^(Die|(Der|Das) /`

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+        short = short.replace(/(?:\S+\s+){2}\S+/);
```
aurimasv: What exactly is the intent here? In general, I wouldn't worry too much about creating a short title.

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+        newItem.shortTitle = short.substring(0, 1).toUpperCase() + short.slice(1);
+    }
+
+    if (items["Untertitel"]) {
```
aurimasv: Also, make sure title has been set already. I would move this to the end of the previous if block.

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+        newItem.title = newItem.title.concat(": ", items["Untertitel"]);
```
aurimasv: No need to use `.concat`. `newItem.title += ": " + items["Untertitel"];`

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+    }
+
+    //Formatting and saving "Author" field
+    if (items["Autoren"]) {
+        var author = items["Autoren"];
+        if (author.match("; ")) {
```
aurimasv: Use '.indexOf(...) != -1`

aurimasv: Though you can ignore this check altogether and just `.split` directly. You will simply get a one-element array if there is no ";"

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+            var authors = author.split("; ");
+            for (var i in authors) {
```
aurimasv: Use `for (var i=0; i<authors.length; i++) {` to iterate arrays

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+                newItem.creators.push(Zotero.Utilities.cleanAuthor(authors[i], "author", true));
+            }
+        } else {
+            newItem.creators.push(Zotero.Utilities.cleanAuthor(author, "author", true));
+        }
+    }
+
+    //Formatting and saving "pages" field
+    if (items["Anfangsseite"] && items["Endseite"]) {
+        newItem.pages = items["Anfangsseite"].concat("-", items["Endseite"]);
```
aurimasv: Use `+` for string concatenation.

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+    } else {
+        newItem.pages = items["Anfangsseite"];
+    }
```
aurimasv: You can make all of this into
```
if (items["Anfangsseite"]) {
  newItem.pages = items["Anfangsseite"] + (items["Endseite"] ? "-" + items["Endseite"] : "");
}
```

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+
+    //Saving the tags to Zotero
+    if (items["Schlagwörter"]) {
+        var tags = items["Schlagwörter"];
+        if (tags.match("; ")) {
+            var tagsContent = tags.split("; ");
+        } else {
+            newItem.tags.push(tagsContent);
+        }
+        for (var i in tagsContent) {
```
aurimasv: This should be inside the if block above.

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+            newItem.tags.push(tagsContent[i]);
+        }
+    }
+
+    //Link
+    newItem.url = link;
```
aurimasv: What is this URL pointing to? I don't see any links in the table on the sample page I am looking at. (also, you should move the relevant code to the same location. You've already assigned `newItem.url` to something else above and the assignment to link is also way above)

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+
+    //Associating and saving the well formatted data to Zotero
+    associateData(newItem, items, "Zeitschrift", "publicationTitle");
+    associateData(newItem, items, "Jahr", "date");
+    associateData(newItem, items, "Heft", "issue");
+    associateData(newItem, items, "Band", "volume");
+    associateData(newItem, items, "Abstract", "abstractNote");
+
+
+    newItem.repository = "DABI";
```
aurimasv: This should be `newItem.libraryCatalog`

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+
+    //Scrape is COMPLETE!
+    newItem.complete();
+}
```
aurimasv: Please include some test cases (covering different metadata). Use Scaffold to generate them.

aurimasv:
I left some comments, but not to be repetitive I only left each comment once. Please take the comments into consideration for the whole translator and fix all related occurrences.

Thanks for contributing!

DABI.js
```diff
@@ -0,0 +1,677 @@
...
+    "minVersion": "1.0",
+    "maxVersion": "",
+    "priority": 100,
+    "inRepository": true,
+    "translatorType": 4,
+    "browserSupport": "gcsvbi",
+    "lastUpdated": "2015-12-15 18:59:56"
+}
+
+/* FW LINE 57:6869c32952b1 */
```
adam3smith: that must have been a misunderstanding: Since you are not using the Framework, you can delete everything from line 15 to (and including) line 504

DABI.js
jmiba: OK, I will do that (having figured out the how-to).

DABI.js
jmiba: I made some styles that rely heavily on short titles. So I want a short short title to be automatically generated. Admittedly, this is not elegant, I will try something else.

DABI.js
```diff
@@ -0,0 +1,321 @@
...
+    "browserSupport": "gcsibv",
+    "lastUpdated": "2016-01-08 11:43:21"
+}
+
+function detectWeb(doc, url) {
+    if (doc.title.trim().indexOf("DABI. Datensatz Vollanzeige") == 0) {
+        return "journalArticle";
+    } else if (doc.title.trim().indexOf("DABI: Rechercheergebnis") == 0) {
+        var keinTreffer = doc.getElementsByTagName("br")[7].nextSibling.data.indexOf("Keine Treffer für die Suche nach:")
+        if (keinTreffer == -1) {
```
aurimasv: Instead, check if `getSearchResults` returns anything

DABI.js
```diff
@@ -0,0 +1,321 @@
...
+            return "multiple";
+        }
+    }
+}
+
+function doWeb(doc, url) {
+    var ids = [];
+
+    //If Statement checks if page is a Search Result, then saves requested Items
```
aurimasv: Drop this comment

DABI.js
```diff
@@ -0,0 +1,321 @@
...
+            if (!items) return true;
+            for (var i in items) {
+                ids.push(i);
+            }
+        });
+    } else if (detectWeb(doc, url) == "journalArticle") {
+        //saves single page items
+        ids = [url];
+    }
+    ZU.processDocuments(ids, scrape);
```
aurimasv: There's no need to "reload" the current page (though, I think we don't actually reload anyway). Generally, the way this is done is: if single page => call scrape directly with doc, url; if multiples, `ZU.processDocuments` on selected URLs, then call scrape on the results (i.e. what you have above, but inside `Z.selectItems`)

aurimasv: Oh yeah, the main reason this would not work is that `Z.selectItems` is asynchronous, so `ids` is empty by the time you call `ZU.processDocuments`

DABI.js
```diff
@@ -0,0 +1,321 @@
...
+
+function getSearchResults(doc) {
+    var trs = doc.getElementsByTagName("tr"),
+        tds = null,
+        items = {};
+
+    for (var i = 1; i < trs.length; i++) {
+        tds = trs[i].getElementsByTagName("td");
+        for (var n = 0; n < tds.length; n++) {
+            var url = doc.location.origin + "/cgi-bin/dabi/" + tds[0].firstChild.getAttribute("href"),
```
aurimasv: use relative URLs, so no need for `doc.location.origin` part. Even better, you could get a reference to the `a` element and get it's `href` property, which will automatically be properly resolved. Feel free to use xpaths to get the elements - seems like that would make things simpler.

DABI.js
```diff
@@ -0,0 +1,321 @@
...
+                author = tds[1].innerHTML,
```
aurimasv: Inner HTML is rarely what you want. You probably want textContent.

DABI.js
```diff
@@ -0,0 +1,321 @@
...
+                title = tds[2].innerHTML.replace(/<br>/g, '. ');
+
+            if (!author == '') {
```
aurimasv: This says `(!author) == ''`, which works, but is quite odd. `if (!author)` would be better.

DABI.js
```diff
@@ -0,0 +1,321 @@
...
+                var item = author.replace(/; <br>.*/, ' et al.') + ": " + title;
```
aurimasv: This is not the format we generally use. I would say title is sufficient (and that's what we typically present). If you feel that there are a lot of same titles on the page (or there can be), then you could qualify with author. Title should be the first part of the string.

DABI.js
```diff
@@ -0,0 +1,321 @@
...
+}
+
+function scrape(doc, url) {
+    var newItem = new Zotero.Item('journalArticle');
+    var trs = doc.getElementsByTagName("tr"),
+        headers,
+        contents,
+        items = {};
+
+    //For Loop to populate "items" Object and save tags to an Array.
```
aurimasv: That seems to be self-explanatory. Though the second part of this statement does not appear to be relevant to this loop.

DABI.js
```diff
@@ -0,0 +1,321 @@
...
+        }
+    }
+    return items;
+}
+
+function scrape(doc, url) {
+    var newItem = new Zotero.Item('journalArticle');
+    var trs = doc.getElementsByTagName("tr"),
+        headers,
+        contents,
```
aurimasv: Declare objects as they are used (i.e. `headers` and `contents` should be inside the for loop)

DABI.js
```diff
@@ -0,0 +1,321 @@
...
+    for (var i = 0; i < trs.length; i++) {
+        headers = trs[i].getElementsByTagName("th")[0].textContent;
+        contents = trs[i].getElementsByTagName("td")[0].innerHTML;
+
+        items[headers.replace(/\s+/g, '')] = contents.trim();
+    }
+
+    //set url to fulltext resource, if present; else to database item
+    if (items["URL"] == '') {
+        newItem.url = url;
```
aurimasv: URL should always point to the full-text resource. It should not point to a DB entry where the full text is not available.

DABI.js
```diff
@@ -0,0 +1,321 @@
...
+    } else {
+        var link = doc.createElement('a');
+        link.innerHTML = items["URL"];
+        newItem.url = link.firstChild.getAttribute("href");
```
aurimasv: What is the purpose of this?

DABI.js
```diff
@@ -0,0 +1,321 @@
...
+
+        items[headers.replace(/\s+/g, '')] = contents.trim();
+    }
+
+    //set url to fulltext resource, if present; else to database item
+    if (items["URL"] == '') {
+        newItem.url = url;
+    } else {
+        var link = doc.createElement('a');
+        link.innerHTML = items["URL"];
```
aurimasv: No, let's not do this. Grab the URL properly as you're filling out the `items` object

DABI.js
aurimasv: nvm, I see it.

DABI.js
aurimasv: @adam3smith, do we want the URLs pointing to the actual PDF? Based on the code below, seems like it sometimes can.

aurimasv: Seems like [some PDFs](http://dabi.ib.hu-berlin.de/cgi-bin/dabi/vollanzeige.pl?artikel_id=18305&modus=html) are hosted in the DB. Do we want to point to the DB entry in that case?

DABI.js
```diff
@@ -0,0 +1,321 @@
...
+            }];
+        }
+    }
+
+
+    //Formatting and saving "title" fields
+    if (items["Titel"]) {
+        newItem.title = items["Titel"].replace(/\*/g, '');
+        var short = newItem.title.replace(/^\W?(?:Die |Der |Das |\.{3}\s?)?/, '');
+        short = short.replace(/\W?(,|:|\?|!|\.|\"|\')/, '').split(' ').slice(0, 6).join(' ');
```
aurimasv: instead of a capturing group (`(,|:|\?|!|\.|\"|\')`), use character group `[,:?!."']`

DABI.js
```diff
@@ -0,0 +1,321 @@
...
+        newItem.shortTitle = short.substring(0, 1).toUpperCase() + short.slice(1);
```
aurimasv: @adam3smith, your call on the short title logic here. I'd say defer to Zotero, but we can keep it if this is better.

DABI.js
```diff
@@ -0,0 +1,321 @@
...
+        }
+    }
+
+    //Sometimes titles are missing
+    if (items["Untertitel"] && !items["Titel"]) {
+        newItem.title = items["Untertitel"].replace(/\*/g, '');
+        var short = newItem.title.replace(/^\W?(?:Die |Der |Das |\.{3}\s?)?/, '');
+        short = short.replace(/\W?(,|:|\?|!|\.|\"|\').*$/, '').split(' ').slice(0, 6).join(' ');
+        newItem.shortTitle = short.substring(0, 1).toUpperCase() + short.slice(1);
+    }
```
aurimasv: Instead of replicating this here, you could just do `if (!items["Titel"]) items["Titel"] = items["Untertitel"];` before the previous if block.

DABI.js
```diff
@@ -0,0 +1,321 @@
...
+    return items;
+}
+
+function scrape(doc, url) {
+    var newItem = new Zotero.Item('journalArticle');
+    var trs = doc.getElementsByTagName("tr"),
+        headers,
+        contents,
+        items = {};
```
aurimasv: nit: `items` is not a very good name for this. `data` or `fields` would be more appropriate

DABI.js
```diff
@@ -0,0 +1,321 @@
...
+    if (items["Anfangsseite"]) {
+        newItem.pages = items["Anfangsseite"] + (items["Endseite"] ? "-" + items["Endseite"] : "");
+    }
+
+    //Saving the tags to Zotero
+    if (items["Schlagwörter"]) {
+        var tags = items["Schlagwörter"].split("; ");
+        for (var i = 0; i < tags.length; i++) {
+            newItem.tags.push(tags[i]);
+        }
```
aurimasv: `newItem.tags = items["Schlagwörter"].split("; ");` is fine.

DABI.js
```diff
@@ -0,0 +1,321 @@
...
+        newItem.publicationTitle = items["Zeitschrift"].replace(/ : /g, ": ");
+    }
+
+    //Associating and saving the well formatted data to Zotero
+    var fieldMap = {
+        "date": items["Jahr"],
+        "issue": items["Heft"],
+        "volume": items["Band"],
+        "abstractNote": items["Abstract"]
+    };
```
aurimasv: why not just assign these directly? There doesn't seem to be any benefit in doing this twice.

DABI.js
```diff
@@ -0,0 +1,321 @@
...
+
+    for (var key in fieldMap) {
+        if (fieldMap.hasOwnProperty(key)) {
+            newItem[key] = fieldMap[key];
+        }
+    };
+
+    newItem.libraryCatalog = "DABI";
```
aurimasv: You can leave this out. It will get filled in automatically.

DABI.js
```diff
@@ -0,0 +1,321 @@
...
+            "creators": [{
+                "firstName": "Matti",
+                "lastName": "Stöhr",
+                "creatorType": "author"
+            }],
+            "date": "2012",
+            "abstractNote": "Zum Programm der AGMB-Tagung 2012 in Aachen gehörte u.a. der zweistündige Fortbildungsworkshop \"Literaturverwaltung im Fokus - Softwaretypen, bibliothekarische Services und mehr\". Im Beitrag werden weniger die referierten Workshopinhalte beschrieben, als vielmehr die Perspektive der Teilnehmerinnen und Teilnehmer anhand einer eMail-basierten Umfrage vorgestellt. Die Kernfrage lautet hierbei: War der Workshop für sie gewinnbringend?",
+            "issue": "3",
+            "libraryCatalog": "DABI",
+            "pages": "0-0",
```
aurimasv: Check for this case and make sure it doesn't happen. (1) if it's just `0`, we should leave pages blank. (2) If start and end is the same, it should just be start.

aurimasv:
Thanks for working on this! Looks like we're getting close.

DABI.js
```diff
@@ -0,0 +1,321 @@
...
+    "target": "^https?://dabi\\.ib\\.hu-berlin\\.de/",
+    "minVersion": "1.0",
+    "maxVersion": "",
+    "priority": 100,
+    "inRepository": true,
+    "translatorType": 4,
+    "browserSupport": "gcsibv",
+    "lastUpdated": "2016-01-08 11:43:21"
+}
+
```
zuphilip: Add here a license block, which should like https://github.com/zuphilip/translators/wiki/Common-code-blocks-for-translators#licence-block

zuphilip:
@jmiba I am happy to help you with any of these comments here. It would be nice if you could take the time to work this further out. A translator for the DABI would be good to have.

jmiba:
Thanks for the offer, @zuphilip, I haven't had the time yet to have a look into this but I will definitely work this out. Of course you can amend what you think fit if you can spare the time yourself.

zuphilip:
That is okay and there is no hurry. (Time is always short...)

zuphilip:
@jmiba What is the status here?

jmiba:
O @zuphilip, I know I've been lazy on that. I'll try to squeeze that in within the next couple of days.

zuphilip:
Okay, looking forward to hear back.

DABI.js
jmiba: Done

DABI.js
jmiba: Done

DABI.js
jmiba: Done

DABI.js
jmiba: Accepted

DABI.js
jmiba: Comment deleted

DABI.js
jmiba: Done

DABI.js
jmiba: I don't think that is the case. The DB does not host PDFs. Your example seems to have changed the ID (now http://dabi.ib.hu-berlin.de/cgi-bin/dabi/vollanzeige.pl?artikel_id=21283&modus=html), the full text resource is (and was?) elsewhere. The fact that ID changes are possible lets me argue that referencing the actual PDF when availble is most reasonable.

I'm going to change the handling for metadata-only entries. These will not include URLs any longer.

DABI.js
jmiba: Procedure changed

DABI.js
jmiba: dito

DABI.js
jmiba: changed

DABI.js
jmiba: Is there an automatic handling within Zotero of short titles?

DABI.js
jmiba: Done

DABI.js
jmiba: replaced

DABI.js
jmiba: done

DABI.js
jmiba: done

DABI.js
jmiba: deleted

DABI.js
jmiba: doesn't happen anymore

DABI.js
jmiba: done

DABI.js
jmiba: Used xpath

DABI.js
jmiba: Hope, I did that right.

DABI.js
jmiba: Done

jmiba:
@aurimasv, @zuphilip: Have made changes as requested, ready for review

zuphilip:
Okay, thank you. I changed some small things, cleaned the code a bit and proposed a PR against your branch https://github.com/jmiba/translators/pull/1 . If you accept this, then the PR here will update automatically.

One thing I also saw is the complicated code for the short title:
```js
		var short = newItem.title.replace(/^\W?(?:Die |Der |Das |\.{3}\s?)?/, '');
		short = short.replace(/\W?[,:?!."']/, '').split(' ').slice(0, 6).join(' ');
		newItem.shortTitle = short.substring(0, 1).toUpperCase() + short.slice(1);
```
I don't care much about short titles similar to what @aurimasv stated above. If it is important for your styles, then this is okay for me, because I don't see any danger that it will break the translator overall. @adam3smith What is your view on this?

DABI.js
```diff
@@ -0,0 +1,482 @@
...
+			{
+				"itemType": "journalArticle",
+				"title": "Anpassung der Personalstruktur der Fachhochschulbibliotheken in Nordrhein-Westfalen an die Erfordernisse der neunziger Jahre",
+				"creators": [],
+				"date": "1992",
+				"issue": "1",
+				"libraryCatalog": "DABI",
+				"pages": "364-372",
+				"publicationTitle": "Mitteilungsblatt des Verbandes der Bibliotheken des Landes Nordrhein-Westfalen",
+				"shortTitle": "Anpassung der Personalstruktur der Fachhochschulbibliotheken in",
```
adam3smith: This suggests for me the short-title code isn't good. I'd be inclined to let Zotero handle them automatically.

DABI.js
```diff
@@ -0,0 +1,482 @@
...
+						"lastName": "Beckstein",
+						"creatorType": "author"
+					}
+				],
+				"date": "2009",
+				"issue": "1",
+				"libraryCatalog": "DABI",
+				"pages": "46",
+				"publicationTitle": "Bibliotheksforum Bayern",
+				"shortTitle": "Bibliophile Flaggschiff Bayerns",
```
adam3smith: I also find this quite awkward. It might serve individual preferences (?), but for the most common usage of short titles, i.e. as short citation markers, I think this works much better with article.

DABI.js
zuphilip: From the code I see, that the short title consists of at most 6 words.

adam3smith: Yeah, I don't think that's acceptable. It'll too often produce nonsense such as this.

jmiba: How does Zotero handle short titles automatically? I'd be glad to drop this piece of code since I'm not happy with it either. Should I just do and accept the PR proposed by @zuphilip?

DABI.js
jmiba: Dito. And: You are right.

DABI.js
adam3smith: Zotero just splits before colons, exclamation marks, and question marks, which I think works well here (doesn't always in Germany, where we also see subtitles set off by periods). 

> Should I just do and accept the PR proposed by @zuphilip?

yes, please

zuphilip: The PR fixes some other things and therefore it is good to accept it anyways. Then, afterwards, we need to delete the lines dealing with shortTitle, and update the tests cases.

jmiba: How do I do it? Just say yes or click on some button (e.g. "Close pull request"?). In case you need me to say something so, yes.

zuphilip: Click on the green button "Create a merge commit" or "Squash and merge" here: https://github.com/jmiba/translators/pull/1

jmiba: Thank you for your help correcting and cleaning up the code. Made me learn something.

zuphilip: Okay, that worked. Now maybe `git pull <your-fork>...` and then delete the lines for the short title and update the test cases (use Scaffold). Can you do that?

jmiba: OK, done (I think).

zuphilip:
@adam3smith Can you do the final review?

zuphilip:
Ping @adam3smith

AbeJellinek:
Sorry, this should have been merged long ago! Looks good to me. Thank you for all your work getting this site supported.

jmiba:
Thanks a lot.

PR #992, opened 2016-01-04
Create Notre Dame Philosophical Reviews.js

retorquere:


Notre Dame Philosophical Reviews.js
```diff
@@ -0,0 +1,129 @@
...
+  item = new Zotero.Item('journalArticle');
+  item.publication = 'Notre Dame Philosophical Reviews';
+  item.ISSN = '1538 - 1617';
+  item.url = doc.location.href;
+  item.date = ((ref = doc.getElementsByClassName('pubdate')[0]) != null ? ref.getAttribute('datetime') : void 0) || '';
+  item.shortTitle = ((ref1 = doc.getElementsByClassName('entry-title')[0]) != null ? ref1.textContent : void 0) || '';
+  if (item.shortTitle) {
+    item.title = "Book review: " + item.shortTitle;
+  }
+  author = (ref2 = doc.getElementsByClassName('author')[0]) != null ? ref2.textContent : void 0;
```
adam3smith: add the author as `reviewedAuthor` rather than to the title (also, going by the tests somethign isn't working right).

Notre Dame Philosophical Reviews.js
```diff
@@ -0,0 +1,129 @@
...
+  reviewer = doc.getElementsByClassName('reviewers')[0];
+  if (reviewer) {
+    reviewedby = doc.getElementsByClassName('reviewed-by')[0];
+    if (reviewedby) {
+      reviewer.removeChild(reviewedby);
+    }
+    reviewer = reviewer.textContent.replace(/,.*/, '').trim();
+    item.creators = [Zotero.Utilities.cleanAuthor(reviewer, 'author')];
+  }
+  item.abstract = abstract = ((ref3 = doc.getElementsByClassName('entry-content')[0]) != null ? ref3.textContent : void 0) || '';
```
adam3smith: `item.abstractNote`

Notre Dame Philosophical Reviews.js
```diff
@@ -0,0 +1,129 @@
...
+    item.title += " (" + author + ")";
+  }
+  reviewer = doc.getElementsByClassName('reviewers')[0];
+  if (reviewer) {
+    reviewedby = doc.getElementsByClassName('reviewed-by')[0];
+    if (reviewedby) {
+      reviewer.removeChild(reviewedby);
+    }
+    reviewer = reviewer.textContent.replace(/,.*/, '').trim();
+    item.creators = [Zotero.Utilities.cleanAuthor(reviewer, 'author')];
```
adam3smith: since you'll have the reviewedAuthor now, better to push to the array rather than overwriting it.

Notre Dame Philosophical Reviews.js
```diff
@@ -0,0 +1,129 @@
...
+};
+
+doWeb = function(doc, url) {
+  var abstract, author, bibliography, item, ref, ref1, ref2, ref3, ref4, reviewedby, reviewer;
+  item = new Zotero.Item('journalArticle');
+  item.publication = 'Notre Dame Philosophical Reviews';
+  item.ISSN = '1538 - 1617';
+  item.url = doc.location.href;
+  item.date = ((ref = doc.getElementsByClassName('pubdate')[0]) != null ? ref.getAttribute('datetime') : void 0) || '';
+  item.shortTitle = ((ref1 = doc.getElementsByClassName('entry-title')[0]) != null ? ref1.textContent : void 0) || '';
```
adam3smith: I'd rather not use the full book title as the short title. That'll not work well citation-wise. I'd suggest just prepending "Review of " rather than Book Review (that's what we're doing on JSTOR) and use `item.title`. Then have Zotero create the short title automatically.

Notre Dame Philosophical Reviews.js
```diff
@@ -0,0 +1,129 @@
...
+
+detectWeb = function(doc, url) {
+  return 'journalArticle';
+};
+
+doWeb = function(doc, url) {
+  var abstract, author, bibliography, item, ref, ref1, ref2, ref3, ref4, reviewedby, reviewer;
+  item = new Zotero.Item('journalArticle');
+  item.publication = 'Notre Dame Philosophical Reviews';
+  item.ISSN = '1538 - 1617';
```
adam3smith: no spaces around hyphen for ISSN.

Notre Dame Philosophical Reviews.js
```diff
@@ -0,0 +1,129 @@
...
+        "creators": [
+          {
+            "firstName": "Leonard D.",
+            "lastName": "Katz",
+            "creatorType": "author"
+          }
+        ],
+        "date": "20050302T020002-05:00",
+        "ISSN": "1538 - 1617",
+        "extra": "Feldman, Fred, Pleasure and the Good Life: Concerning the Nature, Varieties, and Plausibility of Hedonism, Oxford University Press, 2004, 221pp, $35.00 (hbk), ISBN  019926516X.",
```
adam3smith: I'm not too fond of this in Extra -- if you want to keep it, let's push it to a note.

Notre Dame Philosophical Reviews.js
adam3smith: If you want to, you could put italics html tags around the book title -- makes sense to me. We should probably do that in JSTOR, too.

Notre Dame Philosophical Reviews.js
```diff
@@ -0,0 +1,129 @@
...
+        "itemType": "journalArticle",
+        "title": "Book review: Pleasure and the Good Life: Concerning the Nature, Varieties, and Plausibility of Hedonism",
+        "creators": [
+          {
+            "firstName": "Leonard D.",
+            "lastName": "Katz",
+            "creatorType": "author"
+          }
+        ],
+        "date": "20050302T020002-05:00",
```
adam3smith: since Zotero does also save&display the date as entered, could you prettify this a bit, i.e. put hyphens between yyyy-mm-dd and strip out the timestamp

adam3smith:
See comments in line. Beyond those, it'd be great to also add multiples, e.g. from the recent reviews screen and search results. Shouldn't be too hard.

retorquere:
OK, I've addressed all those. I've added a.o. `item.title = "Review of <i>" + title.textContent + "</i>";`, but in the newly generated testcases, the `</i>` is missing.

How do I go about updating this pull request again? I tke it you;d rather retain your comments, so closing this one and opening a new one is not your preference.

Are there docs on multiples? Never done those.

retorquere:
Ah no hold on, the training `</i> is missing in the shortTitle Zotero generates, not the title.

adam3smith:
you can update the PR by pushing the changes to your patch-2 branch or, when using the web interface, just clicking the pencil icon here: https://github.com/zotero/translators/pull/992/files

multiples: https://www.zotero.org/support/dev/translators/coding#saving_multiple_items or check out any translator -- the basics for those are pretty much the same for any site.

ugh on the trailing html marker, though I see how that happens (Zotero just "dumbly" chops of the title after the colon for the short title). You could see when the shortTitle gets created and if that happens before `item.complete()` you can fix it manually with a replace function.

adam3smith:
great on everything except the multiples and the missing `</i>` in the short title.

retorquere:
shortTitle is set in itemDone, so that's too late. The shortTitle algorithm is trivial though (just 17 lines) so I could just replicate that.

In the case of multiples, I should return "multiple" in detectWeb, yes? Bu then what? Do I kick off a scrape request (what is now doWeb) for each of them, and then present that array in Zotero.selectItems? Is there ready-made code I could look at?

adam3smith:
sure. There are lots of examples, check out e.g.
https://github.com/zotero/translators/blob/master/Sueddeutsche.de.js#L117
you can essentially use the same code with different xpath and move the item creation/scraping to a separate function (which we conventionally call scrape such as here). 
You'd then only have to adjust detectWeb to return "multiple" on relevant pages (you can check this out in the same translator).

adam3smith:
and sounds good on replicating the shortTitle creation.

retorquere:
OK, I think that should do it.

Notre Dame Philosophical Reviews.js
```diff
@@ -0,0 +1,207 @@
...
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+var detectWeb, doWeb, scrape, testCases;
+
+detectWeb = function(doc, url) {
+  if (url.match(/^https:\/\/ndpr\.nd\.edu\/news\/[0-9]+/)) {
```
adam3smith: for literal matches like the two below, we prefer `url.indexOf("string")!=-1`; where you do need regex, such as here, use `url.search(/regex/)!=-1` both for efficiency. Since detectWeb runs a fair amount, we care about even pretty small efficiency gains.

Notre Dame Philosophical Reviews.js
```diff
@@ -0,0 +1,207 @@
...
+        "libraryCatalog": "Notre Dame Philosophical Reviews",
+        "shortTitle": "Review of <i>Pleasure and the Good Life",
```
adam3smith: this still doesn't look right.

Notre Dame Philosophical Reviews.js
```diff
@@ -0,0 +1,207 @@
...
+    item.creators.push(Zotero.Utilities.cleanAuthor(reviewer, 'author'));
+  }
+  abstractNote = doc.getElementsByClassName('entry-content')[0];
+  if (abstractNote) {
+    item.abstractNote = abstractNote.textContent;
+  }
+  bibliography = doc.getElementsByClassName('bibliography')[0];
+  bibliography = ((bibliography && bibliography.textContent) || '').trim();
+  if (bibliography) {
+    item.notes = [bibliography];
```
adam3smith: notes are arrays of objects, not arrays of strings. Use
`item.notes.push({note:bibliography});`

Notre Dame Philosophical Reviews.js
```diff
@@ -0,0 +1,207 @@
+{
+  "translatorID": "a823550b-6475-4b20-8539-a3c416906228",
+  "label": "Notre Dame Philosophical Reviews",
+  "creator": "Emiliano Heyns",
+  "target": "^https://ndpr\\.nd\\.edu",
```
adam3smith: put a slash at the end of this -- I don't understand this 100%, but in case people use this with a proxy it'll make sure the regex still works.

retorquere:
done and done

retorquere:
BTW the array-of-strings did import as notes when I tried.

retorquere:
Do you want me to change the copyright notice?

adam3smith:
copyright is fine as is; we normally have the creator as copyright holder. As long as the license is AGLP, everyone is happy.

adam3smith:
glad to hear that array of strings import regularly now. We had a case in the past where it imported as an individual note for each character, which was a bit of a disaster. But better to do it right, though ;)

Everything looking good now, thanks!

retorquere:
Is putting the review in the abstractNote the right thing to do? It's the full review.

retorquere:
Ah, merged, super!

adam3smith:
hmmm -- I didn't catch that. How long are these typically? We don't want to overpopulate the abstract field, obviously. Typically I'd just put the first paragraph in where there's no actual abstract.

retorquere:
They can be pretty long -- 2000 - 3000 words are not uncommon. They don't come with an abstract, although I could perhaps cut off after the first double-newline. Which is not always reliable, as they are HTML, and I just grab the textContent. I could put them in a note I suppose. I couldn't find out how to add them as an HTML attachment.

retorquere:
No, I've looked at a few, and extracting an sensible abstract isn't going to work. Let me know what you think is most suitable, and I'll open a new pull request.

adam3smith:
ah, OK, let's fix that then. You can just do
`item.attachments.push({document: doc, title: "NDPR Full Text Snapshot"})`
(or specify a url instead of a document and then include a mimeType as in https://github.com/zotero/translators/blob/master/Sueddeutsche.de.js#L109 -- no need to specify snapshot: true, that's the default)
And then I'd just remove the abstract entirely.

adam3smith:
(at all: sorry, for the commit pile in the history, btw. I forgot to squash this)

retorquere:
item.attachments is always already an array?

retorquere:
Is there a way for me to put a cleaned-up version in the snapshot? I know exactly where the review is, I can cut away all the navigation.

adam3smith:
yes. Same for creators, tags, notes, and seeAlso (aka Related)

adam3smith:
and no, I don't think there's a good way to clean the Snapshot up. I wouldn't worry about it, there doesn't seem to be much trash on that page.

retorquere:
What specifically "NDPR Full Text Snapshot"? I can put in the review title.

adam3smith:
it's the label that appears in the center pane -- we (almost) always use a generic label since you already have the title in the item it's attached to and the label tells you the provenance of the full text. (I realize the Sueddeutsche translator uses the title; that predates more stringent review of and clearer standards for translators)

retorquere:
I've created a new pull request at https://github.com/zotero/translators/pull/993.

I see that snapshotting is turned off when I disable "Automatically take snapshots from webpages". Damn. In any case, this works. The snapshot isn't very pretty -- a lot of the navigation text and pictures are missing.

PR #998, opened 2016-01-10
add more OG types to EM/RDF

adam3smith:
a couple of commas on test changes inline to follow.

### DIFF
--- a/Embedded Metadata.js
+++ b/Embedded Metadata.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2015-09-29 18:37:33"
+	"lastUpdated": "2016-01-17 17:54:53"
 }
 
 /*
@@ -94,6 +94,8 @@ var _prefixes = {
 	og:"http://ogp.me/ns#",				// Used for Facebook's OpenGraph Protocol
 	article:"http://ogp.me/ns/article#",
 	book:"http://ogp.me/ns/book#",
+	music:"http://ogp.me/ns/music#",
+	video:"http://ogp.me/ns/video#",
 	rdf:"http://www.w3.org/1999/02/22-rdf-syntax-ns#"
 };
 
@@ -254,7 +256,7 @@ function init(doc, url, callback, forceLoadRDF) {
 
 			if(_prefixes[prefix]) {
 				var prop = tag.substr(delimIndex+1, 1).toLowerCase()+tag.substr(delimIndex+2);
-				
+
 				//bib and bibo types are special, they use rdf:type to define type
 				var specialNS = [_prefixes['bib'], _prefixes['bibo']];
 				if(prop == 'type' && specialNS.indexOf(_prefixes[prefix]) != -1) {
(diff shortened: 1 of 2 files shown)


Embedded Metadata.js
```diff
@@ -1115,16 +1122,12 @@ var testCases = [
 						"firstName": "Gregg",
 						"lastName": "Barrios",
 						"creatorType": "author"
-					},
-					{
-						"firstName": "LA Review of",
-						"lastName": "Books",
-						"creatorType": "author"
```
adam3smith: this comes as benefit of preferring name="author" metatag

Embedded Metadata.js
```diff
@@ -865,7 +877,7 @@ var testCases = [
 						"creatorType": "author"
 					}
 				],
-				"date": "2011/10/17",
+				"date": "2011",
```
adam3smith: this is a change in site data; comes straight from citation_data, so unrelated to these changes.

Embedded Metadata.js
```diff
@@ -1145,12 +1148,12 @@ var testCases = [
 				"title": "Northwestern Can't Quit ASA Over Boycott Because it is Not a Member",
 				"creators": [
 					{
-						"firstName": "Eugene",
-						"lastName": "Kontorovich",
+						"firstName": "",
+						"lastName": "https://www.facebook.com/EugeneKontorovich?ref=hl",
```
adam3smith: this is the only disadvantage we're getting from this. I'm unsure what to do about it--it comes from `property="article:author"`. I think that's still preferable to include, but am open for suggestions (e.g. we could test that for twitter and facebook URLs and reject it if those are in the string).

adam3smith:
@zuphilip -- if you could take a quick look over this if you seen anything problematic or have other/additional ideas, that'd be great.

aurimasv:
FWIW, LGTM

Embedded Metadata.js
```diff
@@ -571,8 +573,18 @@ function addLowQualityMetadata(doc, newItem) {
 		}
 	}
 	
-	if(!newItem.creators.length) getAuthorFromByline(doc, newItem);
-	
+	if(!newItem.creators.length) {
+		//the authors in the standard W3 author tag are safer than byline guessing
+		var w3authors = ZU.xpath(doc, '//meta[@name="author"]');
+		if (w3authors.length>0){
+			for (var i = 0; i<w3authors.length; i++){
```
zuphilip: I think the `if` is not needed here, because if the length is null you still can do the (empty) for-loop.

zuphilip: Hm.. well then the `else` part is not working and you have to test there for something... Never mind

Embedded Metadata.js
zuphilip: I don't understand why this changed here. When is this meta tag extracted and why did it change with your improvement above to extract the `<meta name="author">` tags?

BTW shouldn't in that page `article` then be defined as some namespace? Maybe it is just not the best test example here...

RDF.js
```diff
@@ -749,14 +749,20 @@ function importItem(newItem, node) {
 		var creatorType = possibleCreatorTypes[i];
 		if(creatorType == "author") {
 			creators = getFirstResults(node, [n.bib+"authors", n.dc+"creator", n.dc1_0+"creator",
-				n.dcterms+"creator", n.eprints+"creators_name",
+				n.dcterms+"creator", n.eprints+"creators_name", n.article+"author", n.video+"director", n.music+"musician",
```
zuphilip: Okay, I guess the `n.article+"author"` indicates the Facebook author from above.

I read at the <a href="http://ogp.me/#no_vertical">specification</a> that this is "profile array" where a profile consists of first_name, last_name, user_name, gender. However, I haven't seen such data in any of the examples. And <a href="https://developers.facebook.com/docs/reference/opengraph/object-type/article">here</a> it is written that this is "An array of Facebook profile URLs or IDs of the authors for this article" and that is AFAIS the implementation WordPress supports.

Therefore, I don't see much value in these Facebook links and would suggest to drop `n.article+"author"` completely. I guess it might be similar for the other two properties "director" and "musician".

zuphilip:
Okay, just my last comment is important. Everything else looks fine for me.

RDF.js
adam3smith: hmm -- I hadn't seen the bit about the Facebook profile URLs. The idea to include this originally came from here:
https://hbr.org/2015/08/how-to-do-walking-meetings-right
which just has regular names (we also pick the authors from the byline, but I figured that wouldn't always be the case.).

So given that, my suggestion would be to take them out here and put them in the EM translator and don't use the content where it contains facebook.com -- does that sound good?

zuphilip: Yes, that sound good. However, I would even take out every possible URL.

adam3smith:
OK made the change as discussed above, which works out nicely with all our tests. @zuphilip see if you think this is OK & I'll squash & commit

PR #1001, opened 2016-01-14
Fix ProQuest for bookmarklet

sircmyst:
The bookmarklet was failing in ProQuest search results list view on both the PQ and ebrary tabs. A condition meant to check for the ebrary results tab was returning true on the PQ results tab instead. This meant that the bookmarklet would reject the PQ results and try to call down the ebrary translator.

This update should fix getting the PQ results and also provide a workaround for ebrary results when using the bookmarklet. Since it can't use the ebrary translator without errors, we can instead scrape some basic metadata from the ebrary tab on the PQ search results page, which is better than nothing.


ProQuest.js
```diff
@@ -499,6 +498,33 @@ function getItemType(types) {
 	return guessType;
 }
 
+function scrapeEbraryResults(doc, refs) {
+	// Since we can't chase URLs, let's get what we can from the page
+	var ebraryResults = ZU.xpath(doc, '//div[@class="ebraryitem"]');
```
aurimasv: Build up this list in `getSearchResults` and pass a filtered list here based on user selection. E.g. https://github.com/zotero/translators/blob/e9423311b4a492dc7862c7d220f1ac5c23246b82/Cambridge%20Journals%20Online.js

ProQuest.js
```diff
@@ -499,6 +498,33 @@ function getItemType(types) {
...
+
+		item.url = refs[i];
+		item.title = ZU.xpathText(element, './a/@title');
+		item.date = metadata[0].slice(metadata[0].indexOf('(') + 1, metadata[0].indexOf(')'));
+		item.place = metadata[0].slice(0, metadata[0].indexOf('(')).trim();
+		item.publisher = metadata[1].trim();
+		
+		// Push the authors in reverse to restore the original order
+		for(var j = metadata.length - 1; j >= 2; j--)
+			item.creators.push(ZU.cleanAuthor(metadata[j], "author", true));
```
aurimasv: nit: use braces for the `for` block

ProQuest.js
```diff
@@ -499,6 +498,33 @@ function getItemType(types) {
...
 
+function scrapeEbraryResults(doc, refs) {
+	// Since we can't chase URLs, let's get what we can from the page
+	var ebraryResults = ZU.xpath(doc, '//div[@class="ebraryitem"]');
+	
+	for(var i = 0; i < refs.length; i++) {
+		var item = new Zotero.Item("book");
+		var relativeUrl = refs[i].slice(refs[i].indexOf("results.ebraryresults:"));
+		var element = ZU.xpath(ebraryResults, '//div[contains(@id, "ebrary_copy")][a[contains(@href, "' + relativeUrl + '")]]')[0];
+		var metadata = ZU.xpathText(element, './text()').trim().split(';').reverse();
```
aurimasv: I'd just use a regexp to parse that.

ProQuest.js
sircmyst: Refactored to do this. Funny enough, as I was doing this, I noticed a hidden span I hadn't noticed before that had a little extra data that wasn't visible on the page (ISBN, language, page count, LoC call number).

ProQuest.js
sircmyst: Done

ProQuest.js
sircmyst: I was able to simplify this down due to the refactoring above.

ProQuest.js
```diff
@@ -194,14 +200,26 @@ function doWeb(doc, url, noFollow) {
...
-					translator.translate();
-				});
+			if (isEbrary) {
+				if(Zotero.isBookmarklet) {
+					// The bookmarklet can't use the ebrary translator
+					
+					var urlKeys = Object.keys(items);
+					var refs = [];
+					
+					for(var i = 0; i < urlKeys.length; i++) {
```
aurimasv: `for (var i in items)` for iterating objects (not arrays), no need to get the keys beforehand.

ProQuest.js
```diff
@@ -499,6 +517,61 @@ function getItemType(types) {
 	return guessType;
 }
 
+function scrapeEbraryResults(refs) {
+	// Since we can't chase URLs, let's get what we can from the page
+	
+	for(var i = 0; i < refs.length; i++) {
+		var ref = refs[i];
+		var hiddenData = ZU.xpathText(ref.html, './span');
+		var visibleData = ZU.xpathText(ref.html.getElementsByClassName('results_list_copy'), './text()').trim().split(';').reverse();
```
aurimasv: `Array.prototype.map(ref.html.getElementsByClassName('results_list_copy'), function(el) { return node.textContent; })`

Why are you reversing? Variable number of authors? Please comment on the text structure you are parsing.

ProQuest.js
```diff
@@ -499,6 +517,61 @@ function getItemType(types) {
...
+		
+		var item = new Zotero.Item("book");
+
+		var date = /\(([\w\d\s]+)\)/.exec(visibleData[0]);
```
aurimasv: \d is already part of \w, so just \w\s

ProQuest.js
```diff
@@ -499,6 +517,61 @@ function getItemType(types) {
...
+		var place = /([\w,\s]+)\(/.exec(visibleData[0]);
```
aurimasv: this will match the same text as above, since comma is not required, though if the place precedes date, it would match that first. Please comment on what you're parsing

ProQuest.js
```diff
@@ -499,6 +517,61 @@ function getItemType(types) {
...
+		var isbn = /isbn,\svalue\s=\s\'(\d+)\'\n/i.exec(hiddenData);
```
aurimasv: ISBN-10 can contain X as the check digit. the `\n` seems unnecessary.

ProQuest.js
```diff
@@ -499,6 +517,61 @@ function getItemType(types) {
...
+		var language = /language_code,\svalue\s=\s\'([A-Za-z]+)\'\n/i.exec(hiddenData);
+		var numPages = /page_count,\svalue\s=\s\'(\d+)\'\n/i.exec(hiddenData);
+		var locNum = /lccn,\svalue\s=\s\'([\d\s\-\.\w]+)\'\n/i.exec(hiddenData);
```
aurimasv: \d is in \w, `.` does not need escaping inside `[]`. Tip: you don't need to escape `-` inside `[]` if you put it first.

ProQuest.js
sircmyst: Yes, there's a variable number of authors. This particular text string looks like this:
author, first; author, second; author, third; publisher name; publisher location (date)

Unfortunately, this string is not usefully described or broken up by span elements in markup. It was more efficient to split the string on ; and parse it in reverse, since there's no way to predict how many authors a book will have. I can include a comment informing the structure of the string we're looking at on the page.

The data in this string also exists within the same element that hiddenData comes from, so we could extract it that way instead. However, my dev team talked about it during our last scrum, and apparently the PQ platform is still updating their UI. There was some worry that the invisible span with the metadata may not stick around.

aurimasv: Ok. That's what I figured. Please add a comment to the code

sircmyst: No problem.

Also, the Array.prototype.map() call described above doesn't work. I've been fiddling with it in Scaffold and it keeps producing an error. Turns out getElementsByClassName() is returning an object, not an array or a string. Getting textContent from the returned object will also grab extra text we don't want because there's an <a> node at the same level as the text node, though we can split that on \n to separate the contents of the <a> text node out.

ProQuest.js
sircmyst: This regex and the previous are looking at a string like:
Los Angeles, CA, USA (June 2006)

This regex is meant to match everything before the ( character, which will include letters, spaces, and commas.
The previous regex will match what's between ( and ). This will not have commas.

ProQuest.js
sircmyst: ProQuest seems to be consistently returning ISBN-13 without separators, but this is a good thing to include just in case.

ProQuest.js
sircmyst: Pushed another revision with changes for this and the other issues.

ProQuest.js
aurimasv: Sorry, should be `Array.prototype.map.call(...`. Because the returned object is not a real array, you have to invoke the map function this way, otherwise, you could just call `.map(...` on the array.

sircmyst: Thanks, got this approach to work.

sircmyst:
Hi there, sorry to bother, just wanted to poke and see how things look? Do you feel anymore changes should be made to this pull request?

aurimasv:
Looks ok. Could you squash?

sircmyst:
Squashed and done!
(There's a merge from master still hanging out there, I was seeing a ton of other commits that looked like from other work and I was worried that I might do some damage if I messed with those. But all my commits are squashed down to just one.)

adam3smith:
oh sorry, this never got merged. Thanks!

PR #1004, opened 2016-01-26
First fix to BnF catalog change

symac:
As discussed with @adam3smith on twitter, I have updated the BnF translator following the change to their catalog they did today. There are still enhancement that should be done especially regarding the document type and I will try to to it later (as well as adding tests) but as a quick fix I think this version is fine.

### DIFF
--- a/Bibliotheque nationale de France.js
+++ b/Bibliotheque nationale de France.js
@@ -1,7 +1,7 @@
 {
 	"translatorID": "47533cd7-ccaa-47a7-81bb-71c45e68a74d",
 	"label": "Bibliothèque nationale de France",
-	"creator": "Florian Ziche",
+	"creator": "Florian Ziche, Sylvain Machefert",
 	"target": "^https?://[^/]*catalogue\\.bnf\\.fr",
 	"minVersion": "3.0",
 	"maxVersion": "",
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "g",
-	"lastUpdated": "2013-02-28 14:51:02"
+	"lastUpdated": "2016-01-26 22:29:34"
 }
 
 /*
@@ -476,15 +476,15 @@ var BnfClass = function() {
 
 	/* Get the UNIMARC URL for a given single result page. */
 	this.reformURL = function(url) {
-		return url.replace(/&FormatAffichage=[^&]*/, "")
-			.replace(/&idNoeud=[^&]*/, "") + "&FormatAffichage=4";
+		url = url.replace(".public", ""); // Not sure that would happen as there is no reason user switching to unimarc view before going back to normal
+		return url + ".unimarc";
 	};
 
 	
 	/* Get the results table from a list page, if any. Looks for //table[@class="ListeNotice"]. */
 	this.getResultsTable = function(doc) {
 		try {
-			var xPath = '//table[@class="ListeNotice"]';
+			var xPath = '//div[@class="liste-notices"]';
 			var xPathObject = doc.evaluate(xPath, doc, null, XPathResult.ANY_TYPE, null).iterateNext();
 			return xPathObject;
 		} catch(x) {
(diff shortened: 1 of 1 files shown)


adam3smith:
I'm going to merge this right away given the high-level of demand/frustration on this. 
I'll have some comments inline when I get the chance and @aurimasv or @zuphilip also may have thoughts for further improvements. Quick look to me suggested that also now test cases might be possible, in which case we'd love those.

adam3smith:
and, needless to say, merci!

symac:
@adam3smith no problem, I am now waiting for your comments as a quick fix by a casual JS coder will certainly look horrible to you :)

zuphilip:
Merci beaucoup! Actually, I just have seen one thing: I wouldn't use `doc.evaluate` but the Zotero functions `ZU.xpath`, `ZU.xpathText`. Moreover, I am wondering whether we need that much postprocessing after calling the MARC translator, which should also handle UNIMARC.

Bibliotheque nationale de France.js
```diff
@@ -528,34 +528,29 @@ var BnfClass = function() {
...
-		var baseUri = /^(https?:\/\/[^\/]+)/.exec(doc.location.href)[1];
-		var cellPath = '//td[@class="mn_partienoticesynthetique"]';
-		var spanPath = './/span';
+		var cellPath = '//div[@class="liste-notices"]/div[@class="notice-item"]';
 		var cells = doc.evaluate(cellPath, doc, null, XPathResult.ANY_TYPE, null);
-		var cell = undefined;
-		var regexLink = /\s*window.location='([^']+)'\s*/;
-		
-		//Cell loop
 		while(cell = cells.iterateNext()) {
```
aurimasv: declare `cell` as function-local (`var cell;`). Right now it's in the global scope.

Bibliotheque nationale de France.js
```diff
@@ -528,34 +528,29 @@ var BnfClass = function() {
...
-			var spans = doc.evaluate(spanPath, cell, null, XPathResult.ANY_TYPE, null);
-			//Span loop
-			while(span = spans.iterateNext()) {
-				if(title.length > 0) {
-					title += " – ";
+			var link = doc.evaluate('./div[@class="notice-contenu"]/a', cell, null, XPathResult.ANY_TYPE, null).iterateNext();
+			
+			var title = doc.evaluate('./h2', link, null, XPathResult.ANY_TYPE, null).iterateNext(); 
+			if (title) {
+				title = ZU.trim(title.textContent);
```
aurimasv: `ZU.trim` is deprecated. Use `.trim()`

Bibliotheque nationale de France.js
```diff
@@ -528,34 +528,29 @@ var BnfClass = function() {
...
+			} else 	{
+				// 2016-01-26 : Sometimes there is no H2, we then need to get everything, example when searching for : 
+				// "Se souvenir de Tonnay-Charente"
+				title = ZU.trim(link.textContent);
 			}
-				title += Zotero.Utilities.trim(span.textContent);
+			
+			var documentYear = doc.evaluate('./div[@class="notice-ordre"]', cell, null, XPathResult.ANY_TYPE, null).iterateNext();
+			if ( documentYear ) {
+				title += " / " + documentYear.textContent;
```
aurimasv: Is there a good reason to append the year? We generally do not do this.

Bibliotheque nationale de France.js
```diff
@@ -1,7 +1,7 @@
 {
 	"translatorID": "47533cd7-ccaa-47a7-81bb-71c45e68a74d",
 	"label": "Bibliothèque nationale de France",
-	"creator": "Florian Ziche",
+	"creator": "Florian Ziche, Sylvain Machefert",
 	"target": "^https?://[^/]*catalogue\\.bnf\\.fr",
```
aurimasv: would be good idea to take the opportunity to adjust the regexp. `"^https?://([^/]+\\.)?catalogue\\.bnf\\.fr/"`

aurimasv:
Just a few nits, but I agree with @zuphilip's comments as well.

Bibliotheque nationale de France.js
symac: No good reason no, had just in mind it was done like that usually, never mind, will remove it.

PR #1005, opened 2016-01-28
Create retsinformation.js

roaldfrosig:
Zotero translator for the Danish legal site www.retsinformation.dk


retsinformation.js
```diff
@@ -0,0 +1,182 @@
+{
+	"translatorID": "feef66bf-4b52-498f-a586-8e9a99dc07a0",
+	"label": "Retsinformation",
```
aurimasv: Please make sure this matches the file name (i.e. capitalize the file name)

retsinformation.js
```diff
@@ -0,0 +1,182 @@
...
+}
+
+function detectWeb(doc, url){
+	if (ZU.xpathText(doc, '//div[@class="wrapper2"]/table[@id="ctl00_MainContent_ResultGrid1"]')) {
+		return "multiple";
+	} else if (url.indexOf("R0710") != -1) {
+		threeLetters = getKortNavn(doc, url)[0].substr(0,3);
+		if (/(DOM|AFG|KEN|UDT)/.test(threeLetters)) {
+			return "case";
+		};
```
aurimasv: nit: no need to put `;` after `}`

retsinformation.js
```diff
@@ -0,0 +1,182 @@
...
+		if (/(EDP|ISP|FOU)/.test(threeLetters)) {
```
aurimasv: fyi, you don't need the `( )` here. No need to fix this, just saying.

retsinformation.js
```diff
@@ -0,0 +1,182 @@
...
+			return "report";
+		};
+		if (/\d{3}/.test(threeLetters)) {
+			return "bill";
+		} else return "statute";
```
aurimasv: since you're using if blocks above, use a block (`{ }`) for else as well.

retsinformation.js
```diff
@@ -0,0 +1,182 @@
...
+	} else return false;
+}
+
+function doWeb(doc, url) {
+	if (detectWeb(doc, url) == "multiple") {
+		var items = new Object();
+		Z.debug("L26"+url);
+		Z.debug("L27"+url.match("R0700.aspx\\?r"));
+		if (url.match(/(R0900|R0920|R0930)/)) {
```
aurimasv: `.test`, as above

retsinformation.js
```diff
@@ -0,0 +1,182 @@
...
+			var titles = ZU.xpath(doc, '//table[@id="ctl00_MainContent_ResultGrid1"]/tbody/tr[@class!="th"]/td[1]/a[1]');
+		} else if (url.match(/(R0210|R0310)/)) {
+			var titles = ZU.xpath(doc, '//table[@id="ctl00_MainContent_ResultGrid1"]/tbody/tr[@class!="th"]/td[2]/a[1]');
+		} else if (url.match("R0700.aspx\\?res")) {
+			Z.debug("L132"+url);
+			var titles = ZU.xpath(doc, '//table[@id="ctl00_MainContent_ResultGrid1"]/tbody/tr[@class!="th"]/td[3]/a[1]');
+		} else if (url.match(/(R0220|R0415|R0700)/)) {
+			var titles = ZU.xpath(doc, '//table[@id="ctl00_MainContent_ResultGrid1"]/tbody/tr[@class!="th"]/td[4]/a[1]');
+		}
```
aurimasv: Take all of the above (all the multiples logic) and put it in a function, then call it here and from detectWeb. See https://github.com/zotero/translators/blob/master/AIP.js for a simple example.

retsinformation.js
```diff
@@ -0,0 +1,182 @@
...
+		scrape(doc, url);
+	};
+}
+
+function clean(text) {
+	text = text.replace(/\n/g," ");
+	text = text.replace(/ +/g," ");
+	text = text.trim();
+	return text;
+}
```
aurimasv: `ZU.trimInternal` does the same thing (except multiple newlines = single space, which seems more correct anyway)

retsinformation.js
```diff
@@ -0,0 +1,182 @@
...
+
+function scrape(doc, url) {
+	var type = detectWeb(doc, url);
```
aurimasv: Instead of calling `detectWeb`, move the relevant logic out of `detectWeb` into something like `getItemType` and call it from here and `detectWeb`. `detectWeb` can return "multiple", which would not be right here.

retsinformation.js
```diff
@@ -0,0 +1,182 @@
...
+	Z.debug("L91"+type);
+	var newItem = new Zotero.Item(type);
+	newItem.title = getTitle(doc, url);
+	newItem.url = url;
+	if (type=="statute") {
+		newItem.nameOfAct = newItem.title
```
aurimasv: that's an alternative name for `title`. You don't need to fill those in if you fill in `title` above.

aurimasv: Actually..... this and a bunch of fields below are not real Zotero fields. See http://aurimasv.github.io/z2csl/typeMap.xml for available fields

retsinformation.js
```diff
@@ -0,0 +1,182 @@
...
+		newItem.codeNumber = getKortNavn(doc, url)[0];
```
aurimasv: Same as with title, if you can just set `.number`, seems like you won't have to repeat the same assignment for all different types. See http://aurimasv.github.io/z2csl/typeMap.xml for what maps to what base fields.

retsinformation.js
```diff
@@ -0,0 +1,182 @@
...
+		newItem.publicLawNumber = newItem.codeNumber;
```
aurimasv: I don't think this should be the case. Only one of those is probably right, but idk which one you should be setting. @adam3smith or @fbennett can probably answer that.

retsinformation.js
```diff
@@ -0,0 +1,182 @@
...
+		newItem.dateEnacted = getKortNavn(doc, url)[1];
+		newItem.shortTitle = getRessort(doc, url)[0];
+		if (/L(BK|OV)/.test(newItem.publicLawNumber.substr(0,3))) {
+			Zotero.debug("LOV eller LBK");
```
aurimasv: please print log messages in English

retsinformation.js
```diff
@@ -0,0 +1,182 @@
...
+		} else if (threeLetters=="FOU") {
+			newItem.seriesTitle = "Folketingets Ombudsmands udtalelser";
+		};
+		newItem.reportNumber = getKortNavn(doc, url)[0];
+	} else if (type=="bill") {
+		newItem = addAuthor (newItem, getRessort(doc, url)[1]);
+		newItem.billNumber = getKortNavn(doc, url)[0];
+		newItem.date = getRessort(doc, url)[2];
+	} else {
+		return;
```
aurimasv: you should never be doing that or have to do that. This should throw.

retsinformation.js
```diff
@@ -0,0 +1,182 @@
...
+	};
+	newItem.complete();
+	return;
```
aurimasv: no need to `return`

retsinformation.js
```diff
@@ -0,0 +1,182 @@
...
+		title = ZU.xpathText(doc, myXPath);
+	};
+	if (!title) {
+		myXPath = '//div[@class="wrapper2"]/div/div[@id="INDHOLD"]/h1[@class="TITLE"]';
+		title = ZU.xpathText(doc, myXPath);
+	};
+	if (!title) {
+		myXPath = '//div[@class="wrapper2"]/div/div[@id="INDHOLD"]/center/h1[@class="TITLE"]';
+		title = ZU.xpathText(doc, myXPath);
+	};
```
aurimasv: you can chain all of these into ` var title = ZU.xpath(doc, ".....") || ZU.xpath(doc, ".....") || ZU.xpath(doc, ".....")`. Or, if you don't actually want to prefer one xpath over the other, you could craft a single xpath that matches all of these cases.

retsinformation.js
```diff
@@ -0,0 +1,182 @@
...
+	} else {
+		return;
+	};
+	newItem.complete();
+	return;
+}
+
+function getTitle (doc, url) {
+	var myXPath = '//div[@class="wrapper2"]/div/p[@class="Titel2"]';
+	title = ZU.xpathText(doc, myXPath);
```
aurimasv: `var title`

retsinformation.js
```diff
@@ -0,0 +1,182 @@
...
+		title = ZU.xpathText(doc, myXPath);
+	};
+	if (!title) {
+		var myXPath = '/html/head/title';
+		title = ZU.xpathText(doc, myXPath);
+		title = title.substring(0,title.lastIndexOf("-"));
+		if (getRessort(doc, url)[0]) {
+			title = title.substring(getRessort(doc, url)[0].length+4,title.length);
+		};
+	};
```
aurimasv: Do you actually have a case for this? I'd rather not resort to something like that. If we can't find the title the translator should throw an error (it does if no title is specified), which reports back to zotero.org and gives us a heads-up that there is an issue.

retsinformation.js
```diff
@@ -0,0 +1,182 @@
...
+	Z.debug("L111"+title);
+	return title;
+}
+
+function getKortNavn (doc, url) {
+	var myXPath = '//span[@class="kortNavn"]';
+	if (!ZU.xpathText(doc, myXPath)) {
+		myXPath = '//span[@class="kortNavn historisk"]';
+	};
+	infoArray = ZU.xpathText(doc, myXPath);
```
aurimasv: `var infoArray`

retsinformation.js
```diff
@@ -0,0 +1,182 @@
...
+	infoArray = infoArray.replace(/ (Gældende|Historisk)/,"");
+	infoArray = infoArray.split(" af ");
+	return infoArray;
```
aurimasv: it would be more readable (above) if the values were assigned to an object with appropriate properties (e.g. `.date`) and the object returned.

retsinformation.js
```diff
@@ -0,0 +1,182 @@
...
+}
+
+function getRessort (doc, url) {
+	var myXPath = '//div[@class="ressort"]';
+	fodder = ZU.xpathText(doc, myXPath);
```
aurimasv: `var fodder`

retsinformation.js
```diff
@@ -0,0 +1,182 @@
...
+	fodder = fodder.trim();
+	if (fodder.slice(0,1) == "(") {
```
aurimasv: `.charAt(0)`

retsinformation.js
```diff
@@ -0,0 +1,182 @@
...
+		infoArray[0] = fodder.substr(1,fodder.indexOf(")")-1);
```
aurimasv: `.substring` is more appropriate here.

aurimasv: `infoArray` is undefined here

retsinformation.js
```diff
@@ -0,0 +1,182 @@
...
+	} else {
+		infoArray[0] = "";
+	};
+	pubDate = fodder.search(/\d{2}\-\d{2}\-\d{4}/);
```
aurimasv: `var pubDate`

retsinformation.js
```diff
@@ -0,0 +1,182 @@
...
+	return infoArray;
+}
+
+function addAuthor (item, author) {
+	item.creators[0] = {};
+	item.creators[0].creatorType = "author";
+	item.creators[0].lastName = author;
+	item.creators[0].fieldMode = 1;
+	return item;
+}
```
aurimasv: Shouldn't be returning, since it's modifying `item` in place.

Also, I think this can be done once above in `scrape`.

retsinformation.js
```diff
@@ -0,0 +1,182 @@
...
+	"target": "^https?://(?:www\\.)?retsinformation\\.dk/",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "g",
+	"lastUpdated": "2016-01-27 14:38:37"
+}
+
```
zuphilip: You should add the AGPL licence explicitely here, see for example my block for that here: https://github.com/zuphilip/translators/wiki/Common-code-blocks-for-translators#licence-block

retsinformation.js
roaldfrosig: Done.

retsinformation.js
roaldfrosig: Done.

retsinformation.js
roaldfrosig: I'm not sure which ( ) I can remove?

retsinformation.js
roaldfrosig: Done.

retsinformation.js
roaldfrosig: Done.

retsinformation.js
roaldfrosig: Done.

retsinformation.js
roaldfrosig: Done.

retsinformation.js
roaldfrosig: Done.

retsinformation.js
roaldfrosig: Done.

retsinformation.js
roaldfrosig: The document you link to says the Zotero field is called codeNumber, so I haven't changed it.

retsinformation.js
roaldfrosig: I know it's not an ideal solution. I believe that the publicLawNumber field is the most correct one, but I have made our CSL-files under the assumption that we used the codeNumber field, and I've told our students to manually assign the law-number into the codeNumber field. That why I assign the number to both fields.

retsinformation.js
roaldfrosig: All log messages deleted.

retsinformation.js
roaldfrosig: Removed.

retsinformation.js
roaldfrosig: Removed.

retsinformation.js
roaldfrosig: I'm not that handy with XPaths, but I've done something like what you suggest.

retsinformation.js
roaldfrosig: Done.

retsinformation.js
roaldfrosig: No, I don't have a case for this. But as the site is rather idiosynchratically built, I think there's a good chance that there will be cases where there's an html-structure different from the 5 I've found so far.

retsinformation.js
roaldfrosig: Changed.

retsinformation.js
roaldfrosig: Done.

retsinformation.js
roaldfrosig: Done.

retsinformation.js
roaldfrosig: Done.

retsinformation.js
roaldfrosig: Done. And done.

retsinformation.js
roaldfrosig: Done.

roaldfrosig: Done.

retsinformation.js
roaldfrosig: I tried to not return item - it didn't work.

I made this function to avoid writing the code multiple times. I can't see how to accomplish the same inside the scrape function.

retsinformation.js
roaldfrosig: Done.

Retsinformation.js
```diff
@@ -0,0 +1,191 @@
...
+    GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
+    along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+    ***** END LICENSE BLOCK *****
+*/
+
+function detectWeb(doc, url){
+	if (ZU.xpathText(doc, '//div[@class="wrapper2"]/table[@id="ctl00_MainContent_ResultGrid1"]') && getSelectItems(doc, url).length) {
```
aurimasv: Please split this over multiple lines.
```
if (ZU.xpathText(doc, '//div[@class="wrapper2"]/table[@id="ctl00_MainContent_ResultGrid1"]')
  && getSelectItems(doc, url).length
) {

File: Retsinformation.js [hunk 27]
```diff
@@ -0,0 +1,191 @@
...
+		return "multiple";
+	} else if (url.indexOf("R0710") != -1) {
+		type = getType(doc, url);
+		return type;
+	} else return false;
```
aurimasv: don't need this

Retsinformation.js
```diff
@@ -0,0 +1,191 @@
...
+	if (/(R0900|R0920|R0930)/.test(url)) {
+		var titles = ZU.xpath(doc, '//table[@id="ctl00_MainContent_ResultGrid1"]/tbody/tr[@class!="th"]/td[1]/a[1]');
+	} else if (/(R0210|R0310)/.test(url)) {
+		var titles = ZU.xpath(doc, '//table[@id="ctl00_MainContent_ResultGrid1"]/tbody/tr[@class!="th"]/td[2]/a[1]');
+	} else if (/R0700.aspx\\?res/.test(url)) {
+		var titles = ZU.xpath(doc, '//table[@id="ctl00_MainContent_ResultGrid1"]/tbody/tr[@class!="th"]/td[3]/a[1]');
+	} else if (/(R0220|R0415|R0700)/.test(url)) {
+		var titles = ZU.xpath(doc, '//table[@id="ctl00_MainContent_ResultGrid1"]/tbody/tr[@class!="th"]/td[4]/a[1]');
+	}
+	return titles;
```
aurimasv: What if none of the `if` statements above match?
just a suggestion, but you can replace all instances of `var titles = ` with `return `. Then return `[]` at the end if nothing matches.

Retsinformation.js
```diff
@@ -0,0 +1,191 @@
...
+function scrape(doc, url) {
+	var type = getType(doc, url);
+	var newItem = new Zotero.Item(type);
+	newItem.title = getTitle(doc, url);
+	newItem.url = url;
+	if (type=="statute") {
+//		newItem.nameOfAct = newItem.title
+		newItem.codeNumber = getKortNavn(doc, url).id;
+		newItem.publicLawNumber = newItem.codeNumber;
+		newItem.dateEnacted = getKortNavn(doc, url).date;
```
aurimasv: call `getKortNavn` once, use the returned result multiple times.

Retsinformation.js
```diff
@@ -0,0 +1,191 @@
...
+
+function scrape(doc, url) {
+	var type = getType(doc, url);
+	var newItem = new Zotero.Item(type);
+	newItem.title = getTitle(doc, url);
+	newItem.url = url;
+	if (type=="statute") {
+//		newItem.nameOfAct = newItem.title
+		newItem.codeNumber = getKortNavn(doc, url).id;
+		newItem.publicLawNumber = newItem.codeNumber;
```
aurimasv: pick one of these fields, not both.

Retsinformation.js
```diff
@@ -0,0 +1,191 @@
...
+		newItem.dateEnacted = getKortNavn(doc, url).date;
+		newItem.shortTitle = getRessort(doc, url).shortTitle;
+		if (/L(BK|OV)/.test(newItem.publicLawNumber.substr(0,3))) {
+		} else {
```
aurimasv: do you mean `if (!/..../.test(....)) {`?

Retsinformation.js
```diff
@@ -0,0 +1,191 @@
...
+			newItem = addAuthor (newItem, getRessort(doc, url).ressort);
+		};
```
aurimasv: nit: drop the semicolon

Retsinformation.js
```diff
@@ -0,0 +1,191 @@
...
+	newItem.url = url;
+	if (type=="statute") {
+//		newItem.nameOfAct = newItem.title
+		newItem.codeNumber = getKortNavn(doc, url).id;
+		newItem.publicLawNumber = newItem.codeNumber;
+		newItem.dateEnacted = getKortNavn(doc, url).date;
+		newItem.shortTitle = getRessort(doc, url).shortTitle;
+		if (/L(BK|OV)/.test(newItem.publicLawNumber.substr(0,3))) {
+		} else {
+			newItem = addAuthor (newItem, getRessort(doc, url).ressort);
```
aurimasv: you certainly do not need to re-assign `newItem`

Retsinformation.js
```diff
@@ -0,0 +1,191 @@
...
+		};
+	} else if (type=="case") {
+		newItem.caseName = newItem.title
```
aurimasv: drop this line.

Retsinformation.js
aurimasv: Same for `getRessort`

Retsinformation.js
```diff
@@ -0,0 +1,191 @@
...
+		newItem = addAuthor (newItem, getRessort(doc, url).ressort);		
+		newItem.docketNumber = getKortNavn(doc, url).id;
+		newItem.dateDecided = getKortNavn(doc, url).date;
+		newItem.shortTitle = getRessort(doc, url).shortTitle;
+	} else if (type=="report") {
+		newItem = addAuthor (newItem, getRessort(doc, url).ressort);
+		newItem.date = getRessort(doc, url).pubDate;
+		newItem.institution = "Folketinget";
+		threeLetters = getKortNavn(doc, url).threeLetters;
```
aurimasv: `var threeLetters`

Retsinformation.js
```diff
@@ -0,0 +1,191 @@
...
+	return titles;
+}
+
+function scrape(doc, url) {
+	var type = getType(doc, url);
+	var newItem = new Zotero.Item(type);
+	newItem.title = getTitle(doc, url);
+	newItem.url = url;
+	if (type=="statute") {
+//		newItem.nameOfAct = newItem.title
```
aurimasv: remove

Retsinformation.js
```diff
@@ -0,0 +1,191 @@
...
+		newItem = addAuthor (newItem, getRessort(doc, url).ressort);
+		newItem.date = getRessort(doc, url).pubDate;
+		newItem.institution = "Folketinget";
+		threeLetters = getKortNavn(doc, url).threeLetters;
+		if (threeLetters=="EDP") {
+			newItem.seriesTitle = "Folketingets Ombudsmands driftundersøgelser";
+		} else if (threeLetters=="ISP") {
+			newItem.seriesTitle = "Folketingets Ombudsmands Inspektioner";
+		} else if (threeLetters=="FOU") {
+			newItem.seriesTitle = "Folketingets Ombudsmands udtalelser";
```
aurimasv: Title case? Here and above.

Retsinformation.js
```diff
@@ -0,0 +1,191 @@
...
+	} else if (type=="bill") {
+		newItem = addAuthor (newItem, getRessort(doc, url).ressort);
+		newItem.billNumber = getKortNavn(doc, url).id;
+		newItem.date = getRessort(doc, url).pubDate;
+	}
+	newItem.complete();
+}
+
+function getType (doc, url) {
+	threeLetters = getKortNavn(doc, url).threeLetters;
```
aurimasv: `var threeLetters`

Retsinformation.js
```diff
@@ -0,0 +1,191 @@
...
+
+function getTitle (doc, url) {
+	var baseXPath = '//div[@class="wrapper2"]/div/';
+	var xPath1 = baseXPath + 'p[@class="Titel2"]';
+	baseXPath = baseXPath + 'div[@id="INDHOLD"]/';
+	var xPath2 = baseXPath + 'p[@class="Titel"]';
+	var xPath3 = baseXPath + 'font/p[@align="CENTER"]';
+	var xPath4 = baseXPath + 'h1[@class="TITLE"]';
+	var xPath5 = baseXPath + 'center/h1[@class="TITLE"]';
+	var title = ZU.xpathText(doc, xPath1) || ZU.xpathText(doc, xPath2) || ZU.xpathText(doc, xPath3) || ZU.xpathText(doc, xPath4) || ZU.xpathText(doc, xPath5);
```
aurimasv: split over multiple lines.
```
var title = ..... || .... || ....
  || .... || ....;
```

Retsinformation.js
```diff
@@ -0,0 +1,191 @@
...
+	if (!title) {
+		var myXPath = '/html/head/title';
```
aurimasv: `doc.title`

Retsinformation.js
```diff
@@ -0,0 +1,191 @@
...
+		title = ZU.xpathText(doc, myXPath);
+		title = title.substring(0,title.lastIndexOf("-"));
+		if (getRessort(doc, url)[0]) {
+			title = title.substring(getRessort(doc, url)[0].length+4,title.length);
```
aurimasv: Comment on what you're doing here. What's the format of the title?

Retsinformation.js
```diff
@@ -0,0 +1,191 @@
...
+		};
+	}; 
+	title = ZU.trimInternal(title);
+	return title;
```
aurimasv: nit: `return ZU.trimInternal(title)`

Retsinformation.js
```diff
@@ -0,0 +1,191 @@
...
+	return item;
+}
+
+function getRessort (doc, url) {
+	var myXPath = '//div[@class="ressort"]';
+	var fodder = ZU.xpathText(doc, myXPath);
+	fodder = fodder.trim();
+	var item = new Object();
+	if (fodder.charAt(0) == "(") {
+		item.shortTitle = fodder.substring(1,fodder.lastIndexOf(")")-1);
```
aurimasv: I don't think you want the `-1`. `.substring` does not include the last index.

Retsinformation.js
```diff
@@ -0,0 +1,191 @@
...
+	}
+	var datePosition = fodder.search(/\d{2}\-\d{2}\-\d{4}/);
+	item.ressort = fodder.substr(datePosition+10);
+	item.pubDate = fodder.substr(datePosition,10);
```
aurimasv: What format are you parsing here exactly? this doesn't seem right. Also, in cases like this, match with regexp and use the parts that match
```
var m = fodder.match(/(\d{2}-\d{2}-\d{4}) (something here)/;
item.ressort = m[1];
item.pubDate = m[2];
```
(I'm also interested in the order of day and month in that date)

aurimasv:
Please add tests using Scaffold. Let us know if you need some guidance on how to do that.

Retsinformation.js
roaldfrosig: Done.

Retsinformation.js
roaldfrosig: Removed.

Retsinformation.js
roaldfrosig: Changed structure, ending with an 'else' statement.

Retsinformation.js
roaldfrosig: Done. Thanks, that's much better!

Retsinformation.js
roaldfrosig: Done.

Retsinformation.js
roaldfrosig: Yes. I've changed it now.

Retsinformation.js
roaldfrosig: Done.

Retsinformation.js
roaldfrosig: Done.

Retsinformation.js
roaldfrosig: Done.

Retsinformation.js
roaldfrosig: I've skipped this variable altogether.

Retsinformation.js
roaldfrosig: Done.

Retsinformation.js
roaldfrosig: Title capitalisation is odd, but it's taken directly from the webpage, e.g.: https://www.retsinformation.dk/Forms/R0415.aspx?res=104&nres=7

Retsinformation.js
roaldfrosig: Done.

Retsinformation.js
roaldfrosig: Done.

Retsinformation.js
roaldfrosig: Done.

Retsinformation.js
roaldfrosig: Done. - And corrected stupid oversight.

Retsinformation.js
roaldfrosig: Done.

Retsinformation.js
roaldfrosig: You are right of course.

Retsinformation.js
roaldfrosig: I've added a note to the code here.
I could get the date using 'match' instead of 'search', but I can't see how I can extract the 'ressort' part of the string - it's the last part of the string after the date.

roaldfrosig:
I've updated the.js file again. Once more I'd like to thank you for your extensive and detailed comments and corrections aurimasv.
I can't figure out how to run tests using scaffold. Are there any online guidelines for that?

zuphilip:
To add test cases in scaffold:
 1. Open an example url in firefox
 2. Open your translator in Scaffold
 3. Go to `Testing` and click `New Web`
 4. Repeat for other urls and save

aurimasv:
Closed by 7811d7e14935f2fc78b3c5a55e0eacd57600b235

Thank you for your work. I squashed your commits and merged them in. I also cleaned up the code a bit to be more consistent with our practices. Please take a look at 7811d7e14935f2fc78b3c5a55e0eacd57600b235 if you are interested. All the tests pass (except for re-formatted date), so I think the logic remains the same.

PR #1006, opened 2016-01-29
HCSP website could be accessed either with or without "www"

jomuller:
The pages on the HCSP's website could be accessed either with or without "www". Examples:

- http://www.hcsp.fr/Explore.cgi/avisrapportsdomaine?clefr=533
- http://hcsp.fr/Explore.cgi/avisrapportsdomaine?clefr=533

leads exactly to the same page but with two different URLs. I changed the regex for the target to be able to detect both.

### DIFF
--- a/HCSP.js
+++ b/HCSP.js
@@ -2,14 +2,14 @@
 	"translatorID": "3e375db7-5f6e-4861-ab8b-3ddb0f4accc9",
 	"label": "HCSP",
 	"creator": "Joris Muller",
-	"target": "^https?://www\\.hcsp\\.fr/explore\\.cgi/",
+	"target": "^https?://(www\\.)?hcsp\\.fr/explore\\.cgi/",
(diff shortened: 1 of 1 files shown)


adam3smith:
thanks -- you need to update the last updated date, still (otherwise translators don't get updated for users). The rebase&fixup into a single commit: https://help.github.com/articles/about-git-rebase/

jomuller:
I didn't practice git since 6 monthes, then I totally destroyed my local repo trying to rebase (very dangerous command by the way). Then I pulled everything from zotero repo, redone the changes.

zuphilip:
Yes, everything looks fine now.

(BTW I often copy the whole git directory _before_ doing some git things which I am not so familiar with. If something goes wrong I still have my second copy:)

adam3smith:
thanks -- sorry for the trouble

PR #1007, opened 2016-01-30
[IEEE Xplore] Fix PDF download for no-frame (mobile) pages

joshwells:
Discovered Problem:  IEEE Xplore pages do not use frames when the user agent is a mobile browser (test case is iOS).  As a result, the frame associated with the PDF download link is not present for mobile devices, and PDF download does not happen on iOS devices because the script looks for that frame.  The problem can be observed in Firefox by changing the user agent to "Mozilla/5.0 (iPad; CPU OS 8_4_1 like Mac OS X) AppleWebKit/600.1.4 (KHTML, like Gecko) Version/8.0 Mobile/12H321 Safari/600.1.4".

Proposed Solution:  I have expanded the REGEX expression to find the PDF path for pages containing frames and pages not containing frames.  A test case was also updated to show apparent, correct results.

Assuming this fix is approved, it will need to be pushed to the bookmarklets to be useful.

### DIFF
--- a/IEEE Xplore.js
+++ b/IEEE Xplore.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2016-01-30 18:30:13"
+	"lastUpdated": "2016-01-30 20:32:15"
 }
 
 function detectWeb(doc, url) {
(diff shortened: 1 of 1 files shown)


IEEE Xplore.js
```diff
@@ -356,6 +361,10 @@ var testCases = [
 				"volume": "52",
 				"attachments": [
 					{
+						"title": "IEEE Xplore Full Text PDF",
+						"mimeType": "application/pdf"
+					},
+					{
```
aurimasv: How did this end up in the tests? Was the test run with user agent set to mobile? Was this not working for desktops also?

If the former, then you should undo this change, because the tests will fail on Zotero test servers, which do not run as mobile. If the latter, then... are PDFs no longer displayed in frames anywhere? (doesn't seem to be consistent with PR description)

IEEE Xplore.js
```diff
@@ -141,12 +141,17 @@ function scrape (doc, url) {
 			}
 			if (pdf) {
 				ZU.doGet(pdf, function (src) {
-					var m = /<frame src="(.*\.pdf.*)"/.exec(src);
-					if (m) item.attachments = [{
+					var m = /<frame src="(.*\.pdf.*)"|<meta HTTP-EQUIV="REFRESH" content="0; url=(.*\.pdf.*)"/.exec(src);
+					if (m[1]) item.attachments = [{
```
aurimasv: To not duplicate code
```
var pdfUrl = m && (m[1] || m[2]);
if (pdfUrl) {
  item.attachments.push({
     url: pdfUrl,
     title: "IEEE Xplore Full Text PDF",
     mimeType: "application/pdf"
  });
}
item.attachments.push({
  url: url,
  title: "IEEE Xplore Abstract Record",
  mimeType: "text/html"
});
```

joshwells:
Just checking to see if there is anything else I can do to help this request along.  Has anyone had a chance to review it?  I'm antsy to get this working for my mobile devices.

aurimasv:
Closed by f2db2f5afe58745456f5ca188d37cc4196c3956e

I squashed the merge commit to keep the log clean. I also added a test for an open access PDF and did a little code reformatting. Unfortunately, we cannot currently test with different user-agent, so the change in this PR is not tested, but it seems to be working using ModifyHeaders.

Thanks for working on this! Sorry it took a while to merge.

PR #1009, opened 2016-01-31
Various fixes to CAIRN.info translator

nchachereau:
This PR proposes various improvements to the translator for Cairn.info, an important portal for journals in French.

The most important problem, the one that prompted me to look into the translator, was PDF importation from a page with multiple results: because a closure was created, the importation operated on the same page and thus attached the same PDF to all items.

Other improvements:
- tag importation did not work
- the metadata on volume and issue published by Cairn.info is often wrong, the translator fixes it
- importing multiple items did not work on some pages
- using Cairn.info (the name the portal uses) instead of the domain www.cairn.info for the Library Catalog field


CAIRN Info.js
```diff
@@ -67,44 +67,63 @@ function doWeb(doc,url)
...
+	translator.setTranslator("951c027d-74ac-47d4-a107-9c3069ab7b48");
+	translator.setDocument(doc);
+	translator.setHandler("itemDone", function(obj, item) {
+		// Cairn.info uses non-standard keywords:
+		// we import them here, as the Embedded Metadata translator
+		// cannot catch them.
+		item.tags = [];
+		var keywords = ZU.xpathText(doc, '//meta[@name="article-mot_cle"]/@content');
+		if (keywords) keywords = keywords.split(/\s*[,;]\s*/);
+		for (i in keywords) {
```
aurimasv: use `for (var i=0; i<keywords.length; i++)` to iterate arrays

CAIRN Info.js
```diff
@@ -67,44 +67,63 @@ function doWeb(doc,url)
...
+			if (keywords[i].search(/[^\s]/) != -1) {
```
aurimasv: `\S` == `[^\s]`, but `if (keywords[i].trim())` should do the same and, I think, is a bit clearer.

CAIRN Info.js
```diff
@@ -67,44 +67,63 @@ function doWeb(doc,url)
...
+				item.tags.push(keywords[i])
+			}
+		}
+
+		// The default value for PDF download is on an HTML page that
+		// calls the actual download. We need to correct the attachment
+		// URL after the import translator has run.
+		for (var i=item.attachments.length - 1; i>=0; i--) {
```
aurimasv: why backwards?

CAIRN Info.js
```diff
@@ -67,44 +67,63 @@ function doWeb(doc,url)
...
+				item.attachments[i].url += "&download=1";
+			}
+		}
+
+		// Correct volume and issue information
+		if (item.volume.search(/^n°/i) != -1) {
+			item.issue = item.volume.split(/n°/i)[1].trim();
+			item.volume = '';
+		} else if (item.volume.search(/^Vol./i) != -1) {
+			item.volume = item.volume.split('Vol. ')[1];
```
aurimasv: omit the sapce in 'Vol. ' and them `.trim()` after. (it's testing for 'Vol.' without the space, but then assumes that it has a space for the split)

CAIRN Info.js
```diff
@@ -67,44 +67,63 @@ function doWeb(doc,url)
...
+		}
+		if (item.volume.search(/^\d+-\d+$/) != -1) {
+			var volume = item.volume.split('-')
+			item.volume = volume[0];
+			item.issue = volume[1];
+		}
+
+		// Other fixes
+		item.libraryCatalog = 'Cairn.info';
```
aurimasv: Instead, I would `delete item.libraryCatalog` and let Zotero fill this in automatically.

CAIRN Info.js
```diff
@@ -116,6 +135,143 @@ var testCases = [
...
+					}
+				],
+				"date": "2014-05-13",
+				"ISSN": "0048-8046",
+				"issue": "1",
+				"language": "fr",
+				"libraryCatalog": "Cairn.info",
+				"pages": "65-92",
+				"publicationTitle": "Revue européenne des sciences sociales",
+				"url": "http://www.cairn.info/resume.php?ID_ARTICLE=RESS_521_0065",
```
aurimasv: @adam3smith is this ok?

CAIRN Info.js
```diff
@@ -116,6 +135,143 @@ var testCases = [
...
+					},
+					{
+						"title": "Snapshot"
+					}
+				],
+				"tags": [],
+				"notes": [],
+				"seeAlso": []
+			}
+		]
```
aurimasv: So all of these new test cases test something different? Could you comment (here on github) on what they test?

aurimasv:
Thank you for working on the translator. I'm aware that most of my comments are not related to your changes, but if you could address them, that would be great. Mostly this looks good to me, just the nits I commented on.

CAIRN Info.js
adam3smith: yes -- it's a stable link to a full-text source. Makes sense to include as url.

CAIRN Info.js
```diff
@@ -67,44 +67,63 @@ function doWeb(doc,url)
...
-		translator.getTranslatorObject(function (obj) {
-				obj.doWeb(doc, url);
-				});
+		scrape(doc);
 	}
 }
+
+
+function scrape(doc)
+{
```
adam3smith: let's place the curly brackets all ofter then function, i.e. `function scrape(doc) {` See also line 49 abve (I realize that probably wasn't you, but might as well fix it now while we're at it).

CAIRN Info.js
```diff
@@ -116,6 +135,143 @@ var testCases = [
...
+				],
+				"date": "2014-05-13",
+				"ISSN": "0048-8046",
+				"issue": "1",
+				"language": "fr",
+				"libraryCatalog": "Cairn.info",
+				"pages": "65-92",
+				"publicationTitle": "Revue européenne des sciences sociales",
+				"url": "http://www.cairn.info/resume.php?ID_ARTICLE=RESS_521_0065",
+				"volume": "52",
```
nchachereau: This first test handles transforming volume 52-1 into volume 52 and issue 1.

CAIRN Info.js
```diff
@@ -116,6 +135,143 @@ var testCases = [
...
+				"creators": [
+					{
+						"firstName": "Éric",
+						"lastName": "Derennes",
+						"creatorType": "author"
+					}
+				],
+				"date": "2012-04-17",
+				"ISSN": "0035-3264",
+				"issue": "661",
```
nchachereau: This second test handles transforming volume "n° 661" into issue 661.

CAIRN Info.js
```diff
@@ -116,6 +135,143 @@ var testCases = [
...
+				"date": "2011-10-01",
+				"ISSN": "0035-2969",
+				"issue": "3",
+				"language": "fr",
+				"libraryCatalog": "Cairn.info",
+				"pages": "537-557",
+				"publicationTitle": "Revue française de sociologie",
+				"shortTitle": "Transformation de l'État ou changement de régime ?",
+				"url": "http://www.cairn.info/resume.php?ID_ARTICLE=RFS_523_0537",
+				"volume": "52",
```
nchachereau: This last test deals with "Vol. 52".

nchachereau:
Thank you for your feedback! I addressed your comments in a few new commits. Concerning the "Library Catalog" field, I changed the translator's label, and this is enough for the tests to pass. However, should I also rename the file to Cairn.info.js?

aurimasv:
>should I also rename the file to Cairn.info.js?

That would be preferred, but is not strictly necessary.

Thanks for working on this! Looks good to me. Could you please [squash the commits and force-push to the same branch](https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit)?

nchachereau:
Sure! Any advice on the commit message, as there are several improvements involved?

I will also rename the file, if it is preferred.

adam3smith:
you can edit the commit message during squash. Briefly summarize the main changes in the message (don't keep all individual commit messages, which I believe is the git default), but no need to write novels ;)

nchachereau:
I squashed the commits, is everything this way?

aurimasv:
Thanks!

Issue #771, opened 2014-09-01
Error: BibTeX keys should not begin with underscore


zuphilip:
I encountered the same problem as described in <a href="https://forums.zotero.org/discussion/971/troubles-when-export-to-bibtex/">this old discussion thread</a>: A BibTeX key should not begin with an underscore `_` but this will actually happen if an entry has no author (e.g. a webpage).

I suggest to add after <a href="https://github.com/zotero/translators/blob/master/BibTeX.js#L1088">line 1088 of the BibTeX translator</a> as following:
`
basekey = basekey.replace(/^_/, "");
`
or maybe the underscore should be escaped as well.

PR #1013, opened 2016-02-04
Remove colon from bibtex cite key

matthieu-foucault:
Having a colon in bibtex cite key will cause issues when using the babel latex package with the french language and the pagebackref option from hyperref.
Babel will insert a space before the colon at some point, which will confuse hyperref.

### DIFF
--- a/BibTeX.js
+++ b/BibTeX.js
@@ -1066,7 +1066,7 @@ var numberRe = /^[0-9]+/;
 // Also remove markup
 var citeKeyTitleBannedRe = /\b(a|an|the|some|from|on|in|to|of|do|with|der|die|das|ein|eine|einer|eines|einem|einen|un|une|la|le|l\'|el|las|los|al|uno|una|unos|unas|de|des|del|d\')(\s+|\b)|(<\/?(i|b|sup|sub|sc|span style=\"small-caps\"|span)>)/g;
 var citeKeyConversionsRe = /%([a-zA-Z])/;
-var citeKeyCleanRe = /[^a-z0-9\!\$\&\*\+\-\.\/\:\;\<\>\?\[\]\^\_\`\|]+/g;
+var citeKeyCleanRe = /[^a-z0-9\!\$\&\*\+\-\.\/\;\<\>\?\[\]\^\_\`\|]+/g;
 
 var citeKeyConversions = {
 	"a":function (flags, item) {


aurimasv:
Some relevant discussions on why changing the citekey format is problematic: #771, #570.

Based on my quick search, this seems like an issue with babel, rather than Zotero, but it's probably easier for us to avoid breaking babel than for babel to change what they do. I also could not find any workarounds. If this issue does warrant changing the citekey cleanup, we would probably want to change it significantly (based on the linked discussions). We can offer a way for unhappy users to revert back to the old style via a hidden pref though. I'll wait for @adam3smith, @zuphilip, and @karnesky to pitch in before we proceed.

karnesky:
This is an edge-case where people can quibble.

French babble is very useful and is semi-popular, but colons are fairly common in BibTeX citekeys "in the wild". Many of the (few) threads I've found that describe this issue encourage using `\shorthandoff{:}` to change the behavior of french babble instead of changing the bib file. That is my inclination here as well.  

Better BibTeX had users complain when they (inadvertently) stripped colons!  

I'm not strictly opposed to changing how we auto-generate keys (and, in the other threads, you can see I lobby for the change to make "prettier" keys even if the ones that had been generated are valid technically). I don't feel strongly about what we do for auto-generated keys here. But my gut says we may start to disappoint more users than we'd help if we changed how colons are handled.

If we do make some change for this, I'd agree with the suggestion that it should be in a separate function. But the biggest benefit I see is not to allow users to revert via a hidden pref, but for the longer-term. When we eventually get user-modifiable citekeys, it might be helpful to differentiate the characters that are forbidden generally from those that are avoided due to an odd package here-or-there.

aurimasv:
>But the biggest benefit I see is not to allow users to revert via a hidden pref, but for the longer-term. When we eventually get user-modifiable citekeys, it might be helpful to differentiate the characters that are forbidden generally from those that are avoided due to an odd package here-or-there.

Sorry, I'm not sure I understood. Are you against users being able to preserve current behavior via hidden pref? The way I envision this is putting the "allowed citekey characters" in a hidden pref. If someone is modifying this to include additional characters, I would rather let them include characters that are "forbidden" and see how things break on latex end, rather than have odd end-user behavior where some of the "allowed" characters in the hidden pref are actually not allowed anyway.

karnesky:
> Are you against users being able to preserve current behavior via hidden pref? 
No, that's fine.

> The way I envision this is putting the "allowed citekey characters" in a hidden pref.
That's clearer (I thought you wanted a switch between old and new behavior).

I would actually be fine if auto-generated citekeys consisted of only latin-alpha and numeric characters and perhaps '-'. That's a pretty radical change & would impact lots of users. Your proposed implementation sounds like a good way to make something like that more palatable.

I'm not sure if implementing it because of the colon "issue" makes much sense, but could see it being useful if enough of these small issues kept popping up or we can wait to push such a change when we're making other radical changes (such as allowing user-generated keys).

What I meant to say was: If we had user-modifiable citekeys, it might be nice for Zotero to be able to warn or prevent the use of some forbidden characters. This would essentially be that small set of symbols that citeKeyCleanRe strips out _now_ and would _not_ include the colon.

aurimasv:
>I'm not sure if implementing it because of the colon "issue" makes much sense, but could see it being useful if enough of these small issues kept popping up

We've been talking about this for a while, so I feel like at some point we should just do it. None of us disagree that using weird characters in citekey is not great, we just don't want people to get mad. I'm saying we can offer them a way back if they want, but going forward we encourage exporting cleaner citekeys.

>or we can wait to push such a change when we're making other radical changes (such as allowing user-generated keys).

I can toss that into the same set of changes. We would put the citekey format in a hidden pref as well and set the default to the format we use now. We should discuss what kind of customization we want in a different issue. Would someone mind leading that? Maybe there's some existing syntax we can adopt (e.g. zotfile wildcards?)

>it might be nice for Zotero to be able to warn or prevent the use of some forbidden characters

We can't warn (unless we break the export entirely), but we could strip out the forbidden characters automatically. I feel like that would be more confusing though for the user. There might be some cases where "forbidden" characters don't actually break anything. I'm not set on this idea, but that's what I'm leaning towards.

karnesky:
I feel we're talking past each other.

What I refer to is a key that a user can generate manually that would require a new database field, rather than the ability to customize the auto-generation of the keys.  A colon in such a _manually-generated_ key is not a "weird" character.

> None of us disagree that using weird characters in citekey is not great,

Yes.  Weird characters in auto-generated keys are not great. Let's fix this with a white list, as you propose.

aurimasv:
>What I refer to is a key that a user can generate manually that would require a new database field, rather than the ability to customize the auto-generation of the keys.

Oh, yes, I misunderstood.

zuphilip:
I just briefly read the discussions here. I would appreciate if we could generate nice bibtex keys without any special character. :+1: The hidden preference option sounds like a way to also support users which prefer/used for some reasons special characters in their bibtex keys. Thus, I support your suggestions and looking forward for an implementation. :+1:

adam3smith:
@zuphilip since aurimasv isn't around anymore, we should probably work on this. I'm going to keep the discussion here even though it's technically better in a separate ticket, but I want to preserve the thread above.

1. Whitelist: @karnesky suggests [A-Za-z0-9\-] above & I think that makes sense. Anything else we should add? We're stripping accents&umlauts?
2. Hidden Preference: how about BibTeX.export.simpleCitekey where true is citekeys restricted to the whitelist and false is the current behavior
3. Hidden Preference default: this one is tricky -- we'll break things, massively so, if we set the default to true and e.g. in the translation server, people don't have an option (this affects Overleaf, in particular). On the other hand, the effect of this is rather limited if we leave the default to false. Thoughts?

karnesky:
1: Yes, strip accents & umlauts.  Preferred behavior may be to mimic [JabRef's UnicodeToReadableCharMap](https://github.com/JabRef/jabref/blob/ebbeb1d249edf7134abf92d24c2031ae54d19881/src/main/java/org/jabref/model/strings/UnicodeToReadableCharMap.java), though this is extra work.

2: I think this is fine.

3: I have a slight preference for setting it to 'true', as it'll work better in the legacy workflows that still have 7-bit tools & because it is easy for people to revert to old behavior.  But this preference is very weak.

zuphilip:
_To 1:_ We are already stripping accents and umlauts, see https://github.com/zotero/translators/blob/master/BibTeX.js#L1187 . I cannot think of any case when it is important to have any of these special characters in the bibtex-key. We have the `-` in double names or compositions of word. There might be some cases when these symbols are already in bibtex keys, e.g. when the title contains mathematical symbols, but these cases should work without special characters as well. Thus, I would be fine to switch to this very simple restricted set of allowed symbols for the citekey.

_To 2:_ The hidden preferences could be too hidden to have any effect. Could we also do this as an export option (`displayOptions`)?

_To 3:_ Personally, I would prefer to use the new option as default. I doubt that this will break a lot in an individual document and some missing references should then be easy to fix. However, I see that these changes will go in the background for most users and therefore, one might simply not expect that something which worked before, does not work anymore. I don't know the details for overleaf integration... BTW [we already changed the bibtex keys once this year](https://github.com/zotero/translators/pull/1243), and I am not aware that a lot of people complained about this. Are you?

joelostblom:
If anyone else is having issues with this, a workaround is to run your exported bibtex file through the linux command line tool `tr` and replace all the occurrence of `:_` with just `_` via the command `tr ':_' '_' < old-bib.bib > new-bib.bib`. Modify what to replace according to the pattern in your bibtex keys, in my case it is usually `software_title:_year`.

PR #1018, opened 2016-02-19
New translator: Scholars Portal Journals

bkawula:


Scholars Portal Journals.js
```diff
@@ -0,0 +1,172 @@
+{
+	"translatorID": "5ac0fd37-5578-4f82-8340-0e135b6336ee",
+	"label": "Scholars Portal Journals",
+	"creator": "Bartek Kawula",
+	"target": "https?://journals[1-2]\\.scholarsportal\\.info/",
```
aurimasv: `"^https?://journals\d\\.scholarsportal\\.info/"` (it's supporting 1-4 atm, but there's no harm in making this more future-proof)

Scholars Portal Journals.js
```diff
@@ -0,0 +1,172 @@
...
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2016-02-19 04:18:37"
+}
+
+function detectWeb(doc, url) { 
+	// see if saved list is toggled open
+	if (doc.getElementsByClassName('inner-wrap-open') [0]) {
```
aurimasv: This will need to be a bit more involved. Since the page is not being loaded, detectWeb is not triggered when the list is shown/hidden. You will want to use `Zotero.monitorDOMChanges`. For an example, you can take a look at a couple translators [here](https://github.com/zotero/translators/search?utf8=%E2%9C%93&q=monitorDomChanges) (some are more complicated than others). In this case, I think you would want something like this at the beginning of detectWeb:
```
var myArticles = doc.getElementById('my_articles');
if (myArticles) {
  Zotero.monitorDOMChanges(myArticles.parentElement, {attributes: true, attributeFilter: ["style"]});
}
```

Scholars Portal Journals.js
```diff
@@ -0,0 +1,172 @@
...
+		return 'multiple'
+	} else {
+		if(url.indexOf("/search?q") != -1 || url.indexOf("/browse/") != -1){
```
aurimasv: This should just be an `else if`

Scholars Portal Journals.js
```diff
@@ -0,0 +1,172 @@
...
+			return 'multiple'
```
aurimasv: In both cases where you are returning "multiple" you should use `getItems` to check that items are available for scraping.

Scholars Portal Journals.js
```diff
@@ -0,0 +1,172 @@
...
+		}
+		else {
+			return 'journalArticle';
```
aurimasv: that's not strict enough. It would display Zotero icon on the landing page. You need to perform some checks to be 99.9% sure that there is something to be scraped from the page.

Scholars Portal Journals.js
```diff
@@ -0,0 +1,172 @@
...
+	if(type == "multiple") {
+		var list = getItems(doc, url);
+		Zotero.selectItems(list, function(selectedItems) {
+			if(!selectedItems) return true;
+			var uri = [];
+			for(var i in selectedItems) {
+				uri.push(list[i].uri)
+			}
+			for(var i=0; i<uri.length; i++) {
+				var url = uri[i].replace(/(\/details)/g,"");
```
aurimasv: no parentheses in the regexp. Would also help to explain the reasoning here.

Scholars Portal Journals.js
```diff
@@ -0,0 +1,172 @@
...
+				scrape(doc, url)
+			}
```
aurimasv: merge the two loops, but you should collect all the URLs together first, then call `ZU.doGet` on all and `scrape` the resulting text. This performs sequential HTTP requests, rather than spamming the servers.

Scholars Portal Journals.js
```diff
@@ -0,0 +1,172 @@
...
+		var a = url.indexOf("details");
+		var b = url.indexOf("xml");
+		url = url.substring(a+7,b+3);
+		scrape(doc, url)
+	}
+}
+
+function getItems(doc, url) {
+	if (doc.getElementsByClassName('inner-wrap-open') [0]) {
+		var titles = ZU.xpath(doc.getElementById('my-articles-list'), '//div[@class = "title"]/h3/a');
```
aurimasv: if you're anchoring the `ZU.xpath` at an element, your xpath should start with a `.`

Scholars Portal Journals.js
```diff
@@ -0,0 +1,172 @@
...
+		var items = {};
+		for (var i=0; i<titles.length; i++) { 
+			var title = ZU.trimInternal(titles[i].textContent);
+			var uri = ZU.trimInternal(titles[i].pathname);
```
aurimasv: `.pathname` is not supported in most browsers. Use `.href`. `ZU.trimInternal` here is not necessary.

Scholars Portal Journals.js
```diff
@@ -0,0 +1,172 @@
...
+			items[i] = {'title':title,'uri':uri};
```
aurimasv: The general approach in Zotero is to use the unique id (`uri` in this case) as the key for the map and title as the value. So `items[uri] = title`. You can then take `selectedItems` above and use the keys directly.

Scholars Portal Journals.js
```diff
@@ -0,0 +1,172 @@
...
+		}
+		return items
+	} else {
+		if(url.indexOf("/browse") != -1){
+			var titles = ZU.xpath(doc, '//div/h4/a');
+		}
+		else {
+			var titles = ZU.xpath(doc.getElementById('result-list'), '//div[@class = "details"]/h3/a');
```
aurimasv: xpath should start with `.`

Scholars Portal Journals.js
```diff
@@ -0,0 +1,172 @@
...
+		}
+		var items = {};
+		for (var i=0; i<titles.length; i++) { 
+			var title = titles[i].textContent;
```
aurimasv: `ZU.trimInternal` just in case

Scholars Portal Journals.js
```diff
@@ -0,0 +1,172 @@
...
+			var uri = titles[i].pathname;
```
aurimasv: `.href`

Scholars Portal Journals.js
```diff
@@ -0,0 +1,172 @@
...
+			items[i] = {'title':title,'uri':uri};
+		}
+		return items
+	}
+}
+
+function extras(doc, item) { 
+	var currentdate = new Date(); 
+	item.accessed = currentdate.toLocaleString();
```
aurimasv: You do not need this. It will be filled in automatically.

Scholars Portal Journals.js
```diff
@@ -0,0 +1,172 @@
...
+	return item
+}
+
+function scrape(doc, uri) { 
+		var risURL = 'http://'+doc.domain+'/ris?uri='+uri;
```
aurimasv: use relative URLs `/ris?uri=...` and `encodeURIComponent()` the `uri`

Also, too many indents.

Scholars Portal Journals.js
```diff
@@ -0,0 +1,172 @@
...
+		var pdfURL = "/pdf" + uri;
+		ZU.doGet(risURL, function(text) {
+			var translator = Zotero.loadTranslator("import");
+			translator.setTranslator("32d59d2d-b65a-4da4-b0a3-bdd3cfb979e7");
```
aurimasv: make a note of which translator you're loading here

Scholars Portal Journals.js
```diff
@@ -0,0 +1,172 @@
...
+			translator.setString(text);
+			translator.setHandler("itemDone", function(obj, item) {
+				item = extras(doc, item);
+				item.url = "http://resolver.scholarsportal.info/resolve" + uri;
```
aurimasv: Do not link to resources where we did not import the metadata from, especially if the resulting URL may not resolve at all. This should point to the location where the full-text content was obtained. You can link it to the scholasportal page, since it seems that PDFs are hosted here (though only if the PDF is actually available)

Scholars Portal Journals.js
```diff
@@ -0,0 +1,172 @@
...
+				item.attachments = [];
+				item.attachments.push({
```
aurimasv:
```
item.attachments = [{
  ...
}];
```

Scholars Portal Journals.js
```diff
@@ -0,0 +1,172 @@
...
+	"target": "https?://journals[1-2]\\.scholarsportal\\.info/",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2016-02-19 04:18:37"
+}
+
```
zuphilip: Add the licence block here. You can adapt my template for that https://github.com/zuphilip/translators/wiki/Common-code-blocks-for-translators#licence-block

bkawula:
I just committed my updates. Thank you aurimasv and zuphilip for your suggestions!

bkawula:
So uh, can I get a yay or nay on my latest commit?

Scholars Portal Journals.js
```diff
@@ -0,0 +1,197 @@
...
+function scrape(text, doc) { 
+	// loading RIS transformer. 
+	var translator = Zotero.loadTranslator("import");
+	translator.setTranslator("32d59d2d-b65a-4da4-b0a3-bdd3cfb979e7");
+	translator.setString(text);
+	translator.setHandler("itemDone", function(obj, item) {
+		var uri = getURI(item.attachments[0].path);
+		var pdfURL = "/pdf" + uri;
+		item.url = "http://journals.scholarsportal.info/details" + uri;
+		item.attachments = [{}];
```
adam3smith: you're adding an empty attachment here. You can either remove the curly brackets (i.e. the empty object) here, or, perhaps better, delete this line and, instead of pushing to the attachments array do

```
item.attachments = [{
		url: pdfURL,
		title: "Scholars Portal Full Text PDF",
		mimeType: "application/pdf"
	}]
```

Scholars Portal Journals.js
```diff
@@ -0,0 +1,197 @@
...
+
+function doWeb(doc, url) {
+	var type = detectWeb(doc, url);
+	if(type == "multiple") {
+		var list = getItems(doc, url);
+		Zotero.selectItems(list, function(selectedItems) {
+			if(!selectedItems) return true;
+			var articles = [];
+			for(var i in selectedItems) {
+				var article = 'http://'+doc.domain+'/ris?uri='+i
```
adam3smith: I wasn't part of the original discussion, but is there no better way to do this? Zotero will recognize relative URLs, so `'/ris?uri=' + i` should really be enough and e.g. prevent problems with https vs. http

adam3smith:
Sorry about that -- see my two comments in line. Otherwise looks good to me. @aurimasv  @zuphilip  -- do you have anything else?

zuphilip:
I haven't look close here and I will not have time to do that soon. The license is there now, which I asked in the first round. That is good.

bkawula:
Alright. Let me know if there's anything else.

adam3smith:
looks good to me. Will merge this weekend unless there are any objections.

bkawula:
nudge nudge ;)

adam3smith:
sorry about that. Yes, good to go, could you [rebase and squash](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) this into a single commit with a single commit message (I'd just use fixup rather than squash).

bkawula:
Closing this request. Consolidated pull request is [here](https://github.com/zotero/translators/pull/1038/commits/e010c225ed21421fee0952c7028733917fd71958).

PR #1026, opened 2016-03-20
Change The Hindu scraper to find author names correctly

piyushs:
The Hindu has slightly modified the format of its webpages, and the old
translator was not able to scrape author names correctly any more.  This
commit fixes this bug, and makes sure that all the earlier translator
tests pass.

### DIFF
--- a/The Hindu.js
+++ b/The Hindu.js
@@ -9,7 +9,7 @@
     "inRepository": true,
     "translatorType": 4,
     "browserSupport": "gcv",
-    "lastUpdated": "2013-11-19 08:23:18"
+    "lastUpdated": "2016-03-22 15:08:45"
 }
 
 /*****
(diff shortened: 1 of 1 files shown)


The Hindu.js
```diff
@@ -9,7 +9,7 @@
     "inRepository": true,
     "translatorType": 4,
     "browserSupport": "gcv",
-    "lastUpdated": "2013-11-19 08:23:18"
+	"lastUpdated": "2016-03-20 03:42:31"
```
aurimasv: please fix indent

piyushs: Fixed now

zuphilip:
@adam3smith This looks good for me. Maybe we should run the test cases again. But it can also merged as it is, I guess.

zuphilip:
Ping @adam3smith 
(This fixes the author scraping and the test cases seems to be unchanged.)

adam3smith:
thanks for the ping

Issue #1025, opened 2016-03-19
TEI Output Bug: Only the last item in a group of entries is output


acmuller:
TEI export from local Zotero in Firefox only outputs the last item in a list of bibliographies.

adam3smith:
we're aware: https://groups.google.com/d/msg/zotero-dev/FWCP_R5S_hg/WQAVMPaTEQAJ it's a bit onf an odd issue and I'm waiting for a reply before I'll try to get it fixed (you should be able to fix it locally by reverting my changes here: 
https://github.com/zotero/translators/commit/23f345b70fe721e0050d0a624aff24b3898354b3
(I won't do this for the translator as a whole as that'd break API-based translation and because it _should_ work the way it is, so it's a Zotero fix to make, as far as I can see).

dstillman:
No, it's the translator. It's still using `item.itemID`, which doesn't exist in non-legacy mode. Since the translator [uses that as a key](https://github.com/zotero/translators/blob/23f345b70fe721e0050d0a624aff24b3898354b3/TEI.js#L586), but they're all `undefined`, the resulting object ends up with only one property, `"undefined"`. I'm guessing this can be fixed by just using a counter instead.

(In non-legacy mode there's no unique id. Basically, since this is supposed to be the data payload of API JSON, the only real candidate for a unique id would have been the optional `key`, and that doesn't uniquely identify the item without a user/group anyway. If we wanted the possibility of round-tripped data via export/import we'd probably have to stuff a URI in, even if that's not valid API JSON.)

(/cc @aurimasv, who worked on this)

aurimasv:
We do provide uri in [the translator export format](https://github.com/zotero/zotero/blob/4.0/test/tests/data/translatorExport.js), so that could be used as a mostly unique key, granted that it's not guaranteed to be globally unique, or persistent, for un-synced items.

adam3smith:
@acmuller  could you download and test the translator here?
https://raw.githubusercontent.com/adam3smith/translators/tei/TEI.js

acmuller:
On 3/22/2016 2:21 PM, Sebastian Karcher wrote:

> @acmuller <https://github.com/acmuller> could you download and test the
> translator here?
> https://raw.githubusercontent.com/adam3smith/translators/tei/TEI.js

It works!

Thank you for taking care of this.

Charles

---------------------------
A. Charles Muller

Graduate School of Humanities and Sociology
Faculty of Letters
University of Tokyo
[...]

PR #1029, opened 2016-03-22
* TEI.js: switch to item.uri

adam3smith:


adam3smith:
closes https://github.com/zotero/translators/issues/1025

PR #1030, opened 2016-03-23
MIDAS Journal Translator

str4w:
This is a translator for the journals hosted by Kitware - Insight Journal, VTK Journal, etc.

### DIFF
--- a/MIDAS Journals.js
+++ b/MIDAS Journals.js
@@ -1,0 +1,122 @@
+{
+	"translatorID": "86c35e86-3f97-4e80-9356-8209c97737c2",
+	"label": "MIDAS Journals",
+	"creator": "Rupert Brooks",
+	"target": "(insight-journal|midasjournal|vtkjournal).org/browse/publication",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsib",
+	"lastUpdated": "2015-11-10 14:10:06"
+}
+
+/*
+   Midas Journal Translator
+   (Includes ITKJournal,InsightJournal,VTKJournal)
+   Copyright (C) 2016 Rupert Brooks
+   This program is free software: you can redistribute it and/or modify
+   it under the terms of the GNU Affero General Public License as published by
+   the Free Software Foundation, either version 3 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU Affero General Public License for more details.
+
+   You should have received a copy of the GNU Affero General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+function detectWeb(doc, url) {
+	if (url.match("browse/publication")) return "journalArticle";
+}
+
+function scrape(doc, url) {
+   var namespace = doc.documentElement.namespaceURI; 
+   var nsResolver = namespace ? function(prefix) {
+	  if (prefix == 'x') return namespace; else return null; 
+	   } : null; 
+   var titleXPath = doc.evaluate('//*[@id="publication"]/div[@class="title"]', doc, nsResolver, XPathResult.ANY_TYPE, null); 
+   var authorXPath = doc.evaluate('//*[@id="publication"]/div[@class="authors"]', doc, nsResolver, XPathResult.ANY_TYPE, null); 
+   var urlXPath = doc.evaluate('//*[@id="publication"]/table/tbody/tr/td/a', doc, nsResolver, XPathResult.ANY_TYPE, null); 
+   var journalXPath = doc.evaluate('//*[@id="publication"]/div[@class="journal"]/a[1]', doc, nsResolver, XPathResult.ANY_TYPE, null); 
+   var issueXPath = doc.evaluate('//*[@id="publication"]/div[@class="journal"]/a[2]', doc, nsResolver, XPathResult.ANY_TYPE, null); 
+   var submittedXPath = doc.evaluate('//*[@id="publication"]/div[@class="submittedby"]', doc, nsResolver, XPathResult.ANY_TYPE, null); 
+   var abstractXPath = doc.evaluate('//*[@id="publication"]/div[@class="abstract"]', doc, nsResolver, XPathResult.ANY_TYPE, null); 
+   var downloadAllXPath = doc.evaluate('//a[contains(text(),"Download All")]/@href', doc, nsResolver, XPathResult.ANY_TYPE, null); 
+   var pdfXPath1 = doc.evaluate('//a[contains(text(),"Download Paper")]/@href', doc, nsResolver, XPathResult.ANY_TYPE, null); 
+   //var pdfXPath2 = doc.evaluate('//a[contains(text(),".pdf")]/@href', doc, nsResolver, XPathResult.ANY_TYPE, null); 
+ 
+ 
+   pdfhref=pdfXPath1.iterateNext();
+   //if(!pdfhref) {
+   //	 pdfhref=pdfXPath2.iterateNext();
+   //}
+   if(pdfhref) {
+	  tmp=pdfhref.baseURI.split('/');
+	  pdflink=tmp[0]+'//' +tmp[2]+pdfhref.value;
+	  //Zotero.debug(pdflink);
+   }
+   datematch=new RegExp("on\\w+([0-9]+)\\-([0-9]+)\\-([0-9]+)");
+   datematch=new RegExp("on +([0-9]+)\\-([0-9]+)\\-([0-9]+)");
+   submittedString=submittedXPath.iterateNext().textContent;
+   dateparse=datematch.exec(submittedString);
+  
+   authors=authorXPath.iterateNext().textContent.split(/, */);
+   var newItem = new Zotero.Item("journalArticle");
+   newItem.title = titleXPath.iterateNext().textContent;
+   newItem.url = urlXPath.iterateNext().textContent; 
+   newItem.publicationTitle = journalXPath.iterateNext().textContent;
+   newItem.seriesTitle = issueXPath.iterateNext().textContent;
+   newItem.abstractNote = abstractXPath.iterateNext().textContent;
+   newItem.date = dateparse[3]+"-"+dateparse[1]+"-"+ dateparse[2];
+   splitDownloadPath = downloadAllXPath.iterateNext().value.split('/');
+   version=splitDownloadPath[splitDownloadPath.length-1]
+   newItem.extra="Revision: " + version
+   newItem.issue=splitDownloadPath[splitDownloadPath.length-2]
+
+   var a=0;
+   while (a<authors.length) {
+   	   splitname=authors[a].split(' ');
+   	   a=a+1;
+   	   //Zotero.debug("splitname");
+   	   //Zotero.debug(splitname);
+   	   var z=1;
+   	   name="";
+   	   while (z<splitname.length) {
+   	   	  name=name+splitname[z]+" ";
+   	   	  z=z+1;
+   	   }
+   	   name=name+splitname[0];
+   	   newItem.creators.push(Zotero.Utilities.cleanAuthor(name,"author"));
+   }
+   //author
+   //date
+   //pdf	
+   if(pdfhref) {
+   		newItem.attachments.push({
+		   title: newItem.publicationTitle+" "+newItem.issue + " v" + version + " PDF",
+		   url: pdflink,
+		   mimeType: "application/pdf"
+		});
+   }
+	newItem.attachments.push({
+		title: newItem.publicationTitle+" "+newItem.issue + " v" + version +" Snapshot",
+		url: url,
+		mimeType:"text/html"
+	});
+
+   newItem.complete();
+}
+
+function doWeb(doc, url) {
+	if(detectWeb(doc,url)=="journalArticle") {
+	   var articles = new Array();
+	   articles = [url];
+	   Zotero.Utilities.processDocuments(articles, scrape, function(){Zotero.done();});
+	   Zotero.wait();
+	}
+}
--- a/MIDAS Journals.js
+++ b/MIDAS Journals.js
@@ -2,14 +2,14 @@
 	"translatorID": "86c35e86-3f97-4e80-9356-8209c97737c2",
 	"label": "MIDAS Journals",
 	"creator": "Rupert Brooks",
-	"target": "(insight-journal|midasjournal|vtkjournal)\\.org/browse/publication",
+	"target": "^https?://(www\\.)?(insight-journal|midasjournal|vtkjournal)\\.org/",
 	"minVersion": "3.0",
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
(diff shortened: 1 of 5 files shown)


MIDAS Journals.js
```diff
@@ -0,0 +1,122 @@
...
+   newItem.title = titleXPath.iterateNext().textContent;
+   newItem.url = urlXPath.iterateNext().textContent; 
+   newItem.publicationTitle = journalXPath.iterateNext().textContent;
+   newItem.seriesTitle = issueXPath.iterateNext().textContent;
+   newItem.abstractNote = abstractXPath.iterateNext().textContent;
+   newItem.date = dateparse[3]+"-"+dateparse[1]+"-"+ dateparse[2];
+   splitDownloadPath = downloadAllXPath.iterateNext().value.split('/');
+   version=splitDownloadPath[splitDownloadPath.length-1]
+   newItem.extra="Revision: " + version
+   newItem.issue=splitDownloadPath[splitDownloadPath.length-2]
```
zuphilip: All three lines above are missing a semicolon.

MIDAS Journals.js
```diff
@@ -0,0 +1,122 @@
...
+
+   var a=0;
+   while (a<authors.length) {
+   	   splitname=authors[a].split(' ');
+   	   a=a+1;
+   	   //Zotero.debug("splitname");
+   	   //Zotero.debug(splitname);
+   	   var z=1;
+   	   name="";
```
zuphilip: `var` is missing

MIDAS Journals.js
```diff
@@ -0,0 +1,122 @@
+{
+	"translatorID": "86c35e86-3f97-4e80-9356-8209c97737c2",
+	"label": "MIDAS Journals",
+	"creator": "Rupert Brooks",
+	"target": "(insight-journal|midasjournal|vtkjournal).org/browse/publication",
```
zuphilip: What is in front this? Normally we have something like `^https?://(www)?`. Can we do this here also?

Moreover, escape the point in the regexp.

zuphilip:
Can you give some specific examples for your translator?

MIDAS Journals.js
```diff
@@ -0,0 +1,122 @@
...
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU Affero General Public License for more details.
+
+   You should have received a copy of the GNU Affero General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+function detectWeb(doc, url) {
+	if (url.match("browse/publication")) return "journalArticle";
```
aurimasv: this is already guaranteed by the target regexp above. Since the only required field for translation is `item.title`, you probably want to check here that you can get the title.

MIDAS Journals.js
```diff
@@ -0,0 +1,122 @@
...
+}
+
+function scrape(doc, url) {
+   var namespace = doc.documentElement.namespaceURI; 
+   var nsResolver = namespace ? function(prefix) {
+	  if (prefix == 'x') return namespace; else return null; 
+	   } : null;
```
aurimasv: Drop all of this and, below, use `ZU.xpath(doc, "your xpath here")` for an element array or `ZU.xpathText(doc, "your xpath here")` for the `.textContent` of the elements.

MIDAS Journals.js
```diff
@@ -0,0 +1,122 @@
...
+	   } : null; 
+   var titleXPath = doc.evaluate('//*[@id="publication"]/div[@class="title"]', doc, nsResolver, XPathResult.ANY_TYPE, null); 
+   var authorXPath = doc.evaluate('//*[@id="publication"]/div[@class="authors"]', doc, nsResolver, XPathResult.ANY_TYPE, null); 
+   var urlXPath = doc.evaluate('//*[@id="publication"]/table/tbody/tr/td/a', doc, nsResolver, XPathResult.ANY_TYPE, null); 
+   var journalXPath = doc.evaluate('//*[@id="publication"]/div[@class="journal"]/a[1]', doc, nsResolver, XPathResult.ANY_TYPE, null); 
+   var issueXPath = doc.evaluate('//*[@id="publication"]/div[@class="journal"]/a[2]', doc, nsResolver, XPathResult.ANY_TYPE, null); 
+   var submittedXPath = doc.evaluate('//*[@id="publication"]/div[@class="submittedby"]', doc, nsResolver, XPathResult.ANY_TYPE, null); 
+   var abstractXPath = doc.evaluate('//*[@id="publication"]/div[@class="abstract"]', doc, nsResolver, XPathResult.ANY_TYPE, null); 
+   var downloadAllXPath = doc.evaluate('//a[contains(text(),"Download All")]/@href', doc, nsResolver, XPathResult.ANY_TYPE, null); 
+   var pdfXPath1 = doc.evaluate('//a[contains(text(),"Download Paper")]/@href', doc, nsResolver, XPathResult.ANY_TYPE, null);
```
aurimasv: Move all of these down to where they are actually assigned to the Zotero item.

MIDAS Journals.js
```diff
@@ -0,0 +1,122 @@
...
+   var journalXPath = doc.evaluate('//*[@id="publication"]/div[@class="journal"]/a[1]', doc, nsResolver, XPathResult.ANY_TYPE, null); 
+   var issueXPath = doc.evaluate('//*[@id="publication"]/div[@class="journal"]/a[2]', doc, nsResolver, XPathResult.ANY_TYPE, null); 
+   var submittedXPath = doc.evaluate('//*[@id="publication"]/div[@class="submittedby"]', doc, nsResolver, XPathResult.ANY_TYPE, null); 
+   var abstractXPath = doc.evaluate('//*[@id="publication"]/div[@class="abstract"]', doc, nsResolver, XPathResult.ANY_TYPE, null); 
+   var downloadAllXPath = doc.evaluate('//a[contains(text(),"Download All")]/@href', doc, nsResolver, XPathResult.ANY_TYPE, null); 
+   var pdfXPath1 = doc.evaluate('//a[contains(text(),"Download Paper")]/@href', doc, nsResolver, XPathResult.ANY_TYPE, null); 
+   //var pdfXPath2 = doc.evaluate('//a[contains(text(),".pdf")]/@href', doc, nsResolver, XPathResult.ANY_TYPE, null); 
+ 
+ 
+   pdfhref=pdfXPath1.iterateNext();
```
aurimasv: missing `var `. Apply throughout, please.

MIDAS Journals.js
```diff
@@ -0,0 +1,122 @@
...
+   //if(!pdfhref) {
+   //	 pdfhref=pdfXPath2.iterateNext();
+   //}
+   if(pdfhref) {
+	  tmp=pdfhref.baseURI.split('/');
+	  pdflink=tmp[0]+'//' +tmp[2]+pdfhref.value;
+	  //Zotero.debug(pdflink);
+   }
+   datematch=new RegExp("on\\w+([0-9]+)\\-([0-9]+)\\-([0-9]+)");
```
aurimasv: Instead of `new RegExp(...)` use `/.../`

MIDAS Journals.js
```diff
@@ -0,0 +1,122 @@
...
+   datematch=new RegExp("on +([0-9]+)\\-([0-9]+)\\-([0-9]+)");
```
aurimasv: add spaces before and after `=`. Apply throughout please.

MIDAS Journals.js
```diff
@@ -0,0 +1,122 @@
...
+   newItem.seriesTitle = issueXPath.iterateNext().textContent;
+   newItem.abstractNote = abstractXPath.iterateNext().textContent;
+   newItem.date = dateparse[3]+"-"+dateparse[1]+"-"+ dateparse[2];
+   splitDownloadPath = downloadAllXPath.iterateNext().value.split('/');
+   version=splitDownloadPath[splitDownloadPath.length-1]
+   newItem.extra="Revision: " + version
+   newItem.issue=splitDownloadPath[splitDownloadPath.length-2]
+
+   var a=0;
+   while (a<authors.length) {
```
aurimasv: This (and below) should be a `for` loop. `for (var i=0; i<authors.length; i++) { ...`

MIDAS Journals.js
```diff
@@ -0,0 +1,122 @@
...
+   	   a=a+1;
+   	   //Zotero.debug("splitname");
+   	   //Zotero.debug(splitname);
+   	   var z=1;
+   	   name="";
+   	   while (z<splitname.length) {
+   	   	  name=name+splitname[z]+" ";
+   	   	  z=z+1;
+   	   }
+   	   name=name+splitname[0];
```
aurimasv: Please leave a comment in the code of what kind of string you are processing in this whole block. I can see that you're reshuffling something like `Prastawa M.` into `M. Prastawa` and then calling `cleanAuthor` to parse it. It would be much simpler and cleaner to
```
var m = "Prastawa M.".match(/(\S+)\s+(.+)/);
item.creators.push({
    firstName: m[2],
    lastName: m[1],
    creatorType: 'author'
})`
```

MIDAS Journals.js
```diff
@@ -0,0 +1,122 @@
...
+   	   newItem.creators.push(Zotero.Utilities.cleanAuthor(name,"author"));
+   }
+   //author
+   //date
+   //pdf	
+   if(pdfhref) {
+   		newItem.attachments.push({
+		   title: newItem.publicationTitle+" "+newItem.issue + " v" + version + " PDF",
```
aurimasv: Simply "Full Text PDF" is sufficient. The specifics are available in the metadata

MIDAS Journals.js
```diff
@@ -0,0 +1,122 @@
...
+		   url: pdflink,
+		   mimeType: "application/pdf"
+		});
+   }
+	newItem.attachments.push({
+		title: newItem.publicationTitle+" "+newItem.issue + " v" + version +" Snapshot",
+		url: url,
+		mimeType:"text/html"
+	});
```
aurimasv: There is generally no reason to attach a snapshot if we're attaching a full text pdf

MIDAS Journals.js
```diff
@@ -0,0 +1,122 @@
...
+
+   newItem.complete();
+}
+
+function doWeb(doc, url) {
+	if(detectWeb(doc,url)=="journalArticle") {
```
aurimasv: Given you detectWeb, this is guaranteed.

MIDAS Journals.js
```diff
@@ -0,0 +1,122 @@
...
+	   var articles = new Array();
+	   articles = [url];
+	   Zotero.Utilities.processDocuments(articles, scrape, function(){Zotero.done();});
```
aurimasv: I'm not sure what you're trying to do here. The page is already loaded. Why re-load it?

MIDAS Journals.js
```diff
@@ -0,0 +1,122 @@
...
+	   Zotero.wait();
```
aurimasv: This is no longer necessary. It was needed a long time ago.

aurimasv:
As @zuphilip says, please provide some examples and, preferably, add test pages using Scaffold.

zuphilip:
@str4w What is the status here?

str4w:
Oops.  Honestly, I assessed the number of comments at the time as "Too large to complete this week", and then I totally forgot about it after that.  It is back on my radar, I will try to look at it in the near future.

adam3smith:
looking at this quickly, wouldn't it make sense to just import BibTeX or Medline (for which we already have import translators) and then add journal title and ISSN quickly based on the URL?

zuphilip:
BibTeX does not look that good, e.g. for http://www.vtkjournal.org/browse/publication/880 we have
```
  Author	= "I. Herrera and C. Buchart and D. Borro",
```
Medline format could work also as there are also the full first names there, but publication title seems missing and maybe more. That is the reason I asked for examples...

adam3smith:
right, publication titles seem to be missing everywhere -- that's why I thought we'd add them based on URL.

str4w:
Sorry this dragged so long.  This should be improved from the last iteration. Questions - 

Looking at some of the other examples, sometimes they use
Zotero.<method>
Z.<method>
ZU.<method>
what is the correct way, what are these globals?

More importantly, this would be cleaner and the data would be more complete if it could pull the exported XML (sadly, as the previous comments point out, that data is incomplete and we can't just use that for everything).  To get that XML, need to make an HTTP post and parse the output - how does one do that - can the Zotero.doGet method do it?

MIDAS Journals.js
str4w: would be even better to take prestructured data from the XML, see below comment

MIDAS Journals.js
str4w: I see your point, but most of the other translators I looked at (at least ArXiv, and ACM) do this.  It seems to be the standard practice.

zuphilip:
@str4w I worked directly on top of your work here and created a PR into your fork: https://github.com/str4w/translators/pull/1 . If you merge this PR then it will be automatically update here.

There are some special functions which you can use in coding Zotero translators and they either start with `Zotero.` (or short `Z.`) or `Zotero.Utilities.` (short `ZU.`). See also https://www.zotero.org/support/dev/translators/functions

We need a post call here and therfore I used `Zotero.doPost`. It is important then to continue inside this function and not afterwards. See my changes.

str4w:
Thanks for the explanations and improvements.  I merged with my fork so it appears here.

zuphilip:
Okay, great. Let me know whether there is anything left from your point of view.

str4w:
I'm running a few tests to be sure.  A couple things - I see that you have added multiples.  However, I tried this link
http://www.vtkjournal.org/browse/journal/53
and detectWeb did not detect it.

I tried this link http://www.insight-journal.org/browse/publication/921 and the doWeb returns the tags below.  I don't understand the + and -.  Is it normal? I am not actually using the autogenerated tags in Zotero, so this doesn't trouble me directly, just wondering.

             "tags": [
               {
         -       "tag": "Reproducibility"
         +       "tag": "IPython"
               }
               {
         -       "tag": "Tardigrades"
         +       "tag": "Reproducibility"
               }
               {
                 "tag": "Scipy"
               }
               {
         -       "tag": "IPython"
         +       "tag": "Tardigrades"
               }
               {
         -       "tag": "notebook"
         +       "tag": "dexy"
               }
               {
         -       "tag": "dexy"
         +       "tag": "docker"
               }
               {
                 "tag": "github"
               }
               {
         -       "tag": "docker"
         +       "tag": "notebook"
               }
             ]

zuphilip:
Okay, I added another path for this multiple with some filtering on the expected urls. Moroever, some clean-ups. Please merge again: https://github.com/str4w/translators/pull/2

Yes, what you see with the tags is normal. The differences which are indicated with the -/+ are just the ordering, which is performed alphabetically for the test cases.

str4w:
I checked a bunch of articles, ran the tests, looks good.   I added the ISSN tag, since it is available in many cases, why not have it?

Are you sure you want the article number to go into the pages field?  If that is consistent with what other translators do then it would make sense, it just doesn't fit the literal definition of that field as I understand it.

Assuming the ISSN tag is ok with you and the page field for the article number was intentional, looks good to me, go for it.

zuphilip:
> Are you sure you want the article number to go into the pages field? If that is consistent with what other translators do then it would make sense, it just doesn't fit the literal definition of that field as I understand it.

@adam3smith Do I remember correctly? Can you confirm (or correct) that saving an article id is best done in the `pages` field?

(I will merge it as soon as we clarified this.)

zuphilip:
Merged! :tada: Thank you @str4w for the work! (You should reset your master branch now before any new work.)

PR #1034, opened 2016-03-25
Update Wikisource.js

guitarzxt:
This takes the Dictionary of National Biography and supplements, and the Encyclopedia Britannica 1911 and 1922 supplement.

This takes the original scraper as it was last year, and adds specific information for a few of the sub sites for wikisource. All others should remain manuscripts as is currently the case.

### DIFF
--- a/Wikisource.js
+++ b/Wikisource.js
@@ -9,16 +9,16 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2015-08-05 23:08:53"
+	"lastUpdated": "2015-08-15 23:08:53"
 }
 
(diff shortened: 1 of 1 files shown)


Wikisource.js
```diff
@@ -9,16 +9,16 @@
...
...er(b);return this};this.sub=function(b){var c=function(f,e){var d=e.evaluate(b,f,null,XPathResult.ANY_TYPE,null);if(d){return d.iterateNext()}else{return undefined}};this.addFilter(c);return this};this.evaluate=function(f){var e=f.evaluate(this._xpath,f,null,XPathResult.ANY_TYPE,null);var d=e.resultType;var c=new Array();if(d==XPathResult.STRING_TYPE){c.push(e.stringValue)}else{if(d==XPathResult.ORDERED_NODE_ITERATOR_TYPE||d==XPathResult.UNORDERED_NODE_ITERATOR_TYPE){var b;while((b=e.iterateNext())){c.push(b)}}}c=this._applyFilters(c,f);if(c.length==0){return false}else{return c}}};FW._Xpath.prototype=new FW._StringMagic();FW.detectWeb=function(e,b){for(var c in FW._scrapers){var d=FW._scrapers[c];var f=d.evaluateThing(d.itemType,e,b);var a=d.evaluateThing(d.detect,e,b);if(a.length>0&&a[0]){return f}}return undefined};FW.getScraper=function(b,a){var c=FW.detectWeb(b,a);return FW._scrapers.filter(function(d){return(d.evaluateThing(d.itemType,b,a)==c)&&(d.evaluateThing(d.detect,b,a))})[0]};FW.doWeb=function(c,a){var b=FW.getScraper(c,a);b.makeItems(c,a,[],function(f,e,g,d){e.callHook("scraperDone",f,g,d);if(!f.title){f.title=""}f.complete()},function(){Zotero.done()});Zotero.wait()};
```
aurimasv: This has to be kept for the framework translator to work

Wikisource.js
```diff
@@ -31,28 +31,203 @@ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
 GNU General Public License for more details.
 
 You should have received a copy of the GNU General Public License
-along with this program. If not, see <http://www.gnu.org/licenses/>.
+along with this program. If not, see <http://www.gnu.org/licenses/>.-
```
aurimasv: revert this please

aurimasv:
Generally speaking, I'm not sure that it is feasible for Zotero to support specific sections of wikisource.org. IMO, any additions to this translator should be generic.

@adam3smith, @zuphilip, any thoughts on this? Seems like a slippery slope.

zuphilip:
Yeah, I agree here with @aurimasv . It is not the job of Zotero to add the book editors of specific dictionaries/encyclopadia (this could blow up the translator almost infinitely). This information (if important) should somewhere be on the wikisource website. From there we could try to extract the information again in a generic way.

The bibliographic data from wikisource can be found by clicking on "Cite this page", e.g. https://en.wikisource.org/w/index.php?title=Special:CiteThisPage&page=Gell%2C_John_%281593-1671%29_%28DNB00%29&id=3406712 . Actually, the BibTeX does look quite poor and the useful bibliographic information from there is not much more than the title, date and the url.

Actually, in https://en.wikisource.org/w/index.php?title=Special:CiteThisPage&page=Alice%27s_Adventures_in_Wonderland_%281866%29%2FChapter_2&id=5644792 I would the author `Lewis Carroll` to show up in the bibliographic data somewhere. The information is saved by using the template https://en.wikisource.org/wiki/Template:Header#Usage_examples but maybe you adjust the generation of the "Cite this Page". Also in your dictionaries the data about the book editors and dictionary title should be part of the bibliographic data provided by wikisource.

Wikisource.js
guitarzxt: I don't think I removed it, so your assumption that Scaffold removed it is probable.

adam3smith: it looks like you might be using copy&paste from scaffold rather than copying the file? (The test is also missing in your version).

guitarzxt:
Here's the deal: I am working on a digital humanities project where I and others are going through and writing biographies for at least 500 people, and maybe ultimately as many as 1100. Our primary sources are the Dictionary of National Biography and the 1911 Encyclopedia Britannica and supplements. I would love nothing more than to have Zotero be able to detect which sources on wikisource are Encyclopedias and which are other things and assign the correct type, then scrape all the information needed, by having a general scraper that works for all of wikisource. Unfortunately I think my options are to either use the current scraper and switch it to an encyclopedia and then manually add in the editor, volume, publication date, etc for each source for each author, or use this scraper, which has been working great for about the last 7 months. 

The problem with Wikisource is that often the various things within wikisource are slightly different. If I can remember correctly, many of the pieces of information are put in different places which cannot be easily differentiated with a general scraper.

Wikisource.js
guitarzxt: It tells me uploads are disabled

guitarzxt: Okay, I uploaded it now and made another pull request. That part of line 15 was removed in the original.

adam3smith:
I'm less certain about this than @aurimasv and @zuphilip . I think it's definitely true that
1. Ideally wikisource would be better organized and we would only need one single translator variant and
2. We can't write special code for everyting in wikisource
but I'm not convinced that should mean we shouldn't cover any special cases. Both the DNB and the 1911 Encyclopeda are major works, not just in their importance but also in sheer volume. Would it really be such a problem to include them?

zuphilip:
Well, I am especially concerned about lines like this one: ```item.date = ~~((volNum-1)/4)+1885;```. We shouldn't hardcode such information in the translators. But it would be no problem is to parse different layout of wikisource differently or do some other look up. I guess that `encyclopediaArticles` and `bookChapters` have a different layout of bibliographic information in wikisource.

Looking again at the example https://en.wikisource.org/wiki/Gell,_John_%281593-1671%29_%28DNB00%29 we can also click on "Source" which will bring us to https://en.wikisource.org/wiki/Index:Dictionary_of_National_Biography_volume_21.djvu . This side seems to contain all the missing information from the container. It should be possible to combine the information of these two sites in order to create optimal result. But I guess it then makes sense to change this also into a real translator without the framework.

It does not help to open a new PR #1048 before we discussed this further.

adam3smith:
> But I guess it then makes sense to change this also into a real translator without the framework.

yes to this. As this gets more complex, we'll want to do it properly. 

Not sure (and I actually mean "not sure") I agree about not hardcoding parts of translators. We do code, e.g., ISSN and title for newspapers.

zuphilip:
<blockquote>Not sure [...] I agree about not hardcoding parts of translators. We do code, e.g., ISSN and title for newspapers. </blockquote>

Yeah, but isn't this more for a translator for exactly one journal or exactly one newspaper? I feel that it is different with wikisource, where we could have thousands of different values for the editors or ISSN.

However, I hope that we don't have to hardcode anything, if we cleverly combine all the information which is on the page and maybe on the page linked under "Source" (see my previous comment). Someone has just to start with a complete rewrite of the translator...

zuphilip:
Okay, I began to completely rewrite Wikisource, but this needs more polishing: https://github.com/zuphilip/translators/commit/5af04b7a418c7f1a57dfc78c9060554235fa56c1 / CC @adam3smith

guitarzxt:
I much appreciate it sirs. I would offer to help but my contribution in the translator submitted above was my first time using javascript, and the first programming I have done since I switched from a Computer Science major to a Philosophy major in 2003.

adam3smith:
superseded by #1050

PR #1041, opened 2016-04-03
Add various new fields to Primo

adam3smith:
- improve call numbers
- add full text URLs
- add links to finding aids
- some minor parsing improvements for publishers and contributors
- update tests

@zuphilip -- the folks at Harvard University Library requested these changes (the item type and field ones, not the parsing) and told me they made sense across Primo instances. If you could have a quick look that all this looks right to you? (note that there is some Harvard specific code)

@cbaksik -- FYI

### DIFF
--- a/Primo.js
+++ b/Primo.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsb",
-	"lastUpdated": "2016-04-03 22:15:00"
+	"lastUpdated": "2016-04-05 04:56:07"
 }
 
 /*
@@ -80,7 +80,7 @@ function doWeb(doc, url) {
 				urls.push({url: i, id: itemIDs[i].id, docID: itemIDs[i].docID});
 			}
 			fetchPNX(urls);
-		})
+		});
 	} else {
 		fetchPNX([{url: url, id: 0, docID: getDocID(url)}]);
 	}
@@ -280,7 +280,7 @@ function importPNX(text) {
 	item.title = ZU.xpathText(doc, '//display/title');
 	if(item.title) {
 		item.title = ZU.unescapeHTML(item.title);
-	    item.title = item.title.replace(/\s*:/, ":");
+		item.title = item.title.replace(/\s*:/, ":");
 	}
 	var creators = ZU.xpath(doc, '//display/creator');
 	var contributors = ZU.xpath(doc, '//display/contributor');
@@ -300,7 +300,7 @@ function importPNX(text) {
 			var splitAu = author.split(',');
 			if (splitAu.length > 2) continue;
 			var name = splitAu[1].trim().toLowerCase() + ' '
-				+ splitAu[0].trim().toLowerCase()
+				+ splitAu[0].trim().toLowerCase();
 			splitGuidance[name] = author;
 		}
 	}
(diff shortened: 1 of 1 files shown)


Primo.js
```diff
@@ -369,9 +384,51 @@ function importPNX(text) {
...
+	}
+	if (callArray.length) {
+		item.callNumber = callArray.join(", ");
+	}
+	else {
+		ZU.xpathText(doc, '//enrichment/classificationlcc');
+	}
 	
+	//Harvard specific code, requested by Harvard Library:
+	var library = ZU.xpathText(doc, '//browse/institution')
```
zuphilip: Missing semicolon (also in lines 83 and 303)

Primo.js
```diff
@@ -265,8 +278,10 @@ function importPNX(text) {
 	}
 	
 	item.title = ZU.xpathText(doc, '//display/title');
-	if(item.title) item.title = ZU.unescapeHTML(item.title);
-	
+	if(item.title) {
+		item.title = ZU.unescapeHTML(item.title);
+	    item.title = item.title.replace(/\s*:/, ":");
```
zuphilip: Is this for separating a subtitle? And fix indenting here.

Primo.js
```diff
@@ -369,9 +384,51 @@ function importPNX(text) {
...
+	for (var i = 0; i<callNumber.length; i++) {
+		if (callNumber[i].textContent.search(/\$\$D.+\$/) != -1) {
+			callArray.push(callNumber[i].textContent.match(/\$\$D(.+?)\$/)[1]);
+		}
+	}
+	if (!callArray.length) {
+		callNumber = ZU.xpath(doc, '//display/availlibrary');
+		for (var i = 0; i<callNumber.length; i++) {
+			if (callNumber[i].textContent.search(/\$\$2.+\$/) != -1) {
+				callArray.push(callNumber[i].textContent.match(/\$\$2\(?(.+?)(?:\s*\))?\$/)[1]);
```
zuphilip: Here is an example for us:
```
<availlibrary>
   $$IMAN$$L$$1geschlossenes Magazin$$2(ZL 474 )$$Savailable$$X49MAN_INST$$YWEST$$ZT-MAG$$P1
</availlibrary>
<availlibrary>
   $$IMAN$$L$$1geschlossenes Magazin$$2(ZL 474 )$$Savailable$$X49MAN_INST$$YWEST$$ZT-MAG$$P2
</availlibrary>
```
So, yes, I guess that should work.

zuphilip:
I cannot test and look much further at the moment, but the changes in general look good for me to. Actually, they will also be an improvement for us (call number and links). Thus, thank you very much for the work!

Primo.js
cbaksik: It's for removing a single space in front of a semicolon. In library cataloging, titles and subtitles are separate by a space-semicolon-space, but in citation styles, the space in front of the semicolon needs to be removed, and this change will prevent users from having to do it manually in Zotero

Primo.js
adam3smith: this looks to me like I should be removing duplicates in the array before converting it to a string. Will do that.

Primo.js
```diff
@@ -504,11 +562,12 @@ var testCases = [
...
 				],
 				"date": "1973",
-				"ISBN": "0600393046",
+				"ISBN": "9780600393047",
+				"callNumber": "9B6655",
 				"language": "eng",
 				"libraryCatalog": "Primo",
 				"numPages": "252",
-				"publisher": "Hamlyn",
+				"publisher": "Hamlyn,",
```
adam3smith: note to self: check where that comma is coming from and remove.

Primo.js
adam3smith: we already do this for import via MARC and from WorldCat, so this should be entirely uncontroversial.

adam3smith:
@cbaksik - this is now live with Zotero (translators update every 24hs or use the "Update Now" button in the general tab of the preferences) & ready for testing. Feedback here or via e-mail.

Primo.js
zuphilip: Yeah, I was just wondering... BTW I think the space-semicolon-space comes from the [ISBD](https://de.wikipedia.org/wiki/International_Standard_Bibliographic_Description) format.

PR #1042, opened 2016-04-04
Add new Trove translator

wragge:
The 'National Library of Australia - Australian Newspapers' translator was broken by a recent update to the site. I took the opportunity to create a new translator for (almost) the whole of Trove, of which the newspapers are now just one part. If this translator is ok, then the old 'Australian Newspapers' translator can be removed from the repository.

### DIFF
--- a/National Archives of Australia.js
+++ b/National Archives of Australia.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2015-06-25 06:51:40"
+	"lastUpdated": "2016-04-05 11:11:08"
 }
 
 /*
@@ -207,14 +207,10 @@ function scrapeSeries(doc, url) {
 	item.date = meta['contents dates'];
 	item.medium = meta['predominant physical format'];
 	item.abstractNote = meta['series note'];
+	item.archiveLocation = meta['series number'];
 	
 	var seriesNumber = encodeURIComponent(meta['series number']);
-	item.attachments.push({
-		title: "National Archives of Australia Record",
-		url: 'http://www.naa.gov.au/cgi-bin/Search?O=S&Number=' + seriesNumber,
-		mimeType: 'text/html',
-		snapshot: false
-	})
+	item.url = 'http://www.naa.gov.au/cgi-bin/Search?O=S&Number=' + seriesNumber;
 	
 	// Agencies recording into this series
 	var agencies = ZU.xpath(doc, '//div[@id="provenanceRecording"]//div[@class="linkagesInfo"]');
--- a/National Archives of Australia.js
+++ b/National Archives of Australia.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2016-04-05 11:11:08"
+	"lastUpdated": "2016-05-14 04:30:12"
 }
 
 /*
@@ -502,14 +502,10 @@ var testCases = [
 				"date": "14 Aug 1944 - 31 Oct 1944",
 				"abstractNote": "This series consists of one volume bound in black with the title 'A report on war crimes by individual members of the armed forces of the enemy against Australians by Sir William Webb Kt' embossed on the front cover in gold.\n\nBackground\n\nThe United Nations War Crimes Commission had two stated objectives (1) to hear evidence of war crimes brought to it by member governments and to list the perpetrator for arrest and (2) to make recommendations to member governments on how war criminals could be brought to trial. It held its first meeting on 20 October 1943 and in reporting to Dr Evatt the Secretary of the Department of External Affairs recommended that a Commission be given to Sir William Webb to investigate war crimes against Australians and to bring to the government such cases as could be forwarded to the UNWCC. \n\nOn 9 February Dr Evatt approached Sir William with an invitation and this was accepted on 24 February. The new commission was issued on 8 June 1944 with prime responsibility for administrative matters held by the Department of External Affairs though the report was to be submitted also to the Attorney Generals Department.\n\nThe hearings commenced on 14 August and concluded on 20 October 1944. The report was tendered to the Minister on 31 October 1944.",
 				"archive": "National Archives of Australia",
+				"archiveLocation": "A10950",
 				"libraryCatalog": "National Archives of Australia",
-				"attachments": [
-					{
-						"title": "National Archives of Australia Record",
-						"mimeType": "text/html",
-						"snapshot": false
-					}
-				],
+				"url": "http://www.naa.gov.au/cgi-bin/Search?O=S&Number=A10950",
+				"attachments": [],
 				"tags": [],
 				"notes": [],
 				"seeAlso": []
--- a/Trove.js
+++ b/Trove.js
@@ -1,0 +1,326 @@
+{
+	"translatorID": "8efcb7cb-4180-4555-969a-08e8b34066c4",
+	"label": "Trove",
+	"creator": "Tim Sherratt",
+	"target": "^https?://trove\\.nla\\.gov\\.au/(?:newspaper|gazette|work|book|article|picture|music|map|collection)/",
+	"minVersion": "2.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsbv",
+	"lastUpdated": "2016-04-04 09:40:10"
+}
+
+/*
+   Trove Translator
+   Copyright (C) 2016 Tim Sherratt (tim@discontents.com.au, @wragge)
+
+   This program is free software: you can redistribute it and/or modify
+   it under the terms of the GNU Affero General Public License as published by
+   the Free Software Foundation, either version 3 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU Affero General Public License for more details.
+
+   You should have received a copy of the GNU Affero General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+function detectWeb(doc, url) {
+	if (url.match(/\/result\?/) || url.match(/\/newspaper\/page/)) {
+		return "multiple";
+	} else if (url.match(/\/newspaper\/article\//i)) {
+		return "newspaperArticle";
+	} else if (url.match(/\/work\//i)) {
+		return "book";
+	}
+}
+
+function getSearchResults(doc, url) {
+	var items = {};
+	var results;
+	if (url.match(/\/result\?/)) {
+		results = ZU.xpath(doc, "//div[@id='mainresults']//li/dl/dt/a");
+	} else if (url.match(/\/newspaper\/page/)) {
+		results = ZU.xpath(doc, "//ol[@class='list-unstyled articles']/li/h4/a");
+	}
+	for (var i=0; i<results.length; i++) {
+		var url = results[i].href;
+		var title = ZU.trimInternal(results[i].textContent);
+		items[url] = title;
+	}
+	return items;
+}
+
+function doWeb(doc, url) {
+	if (detectWeb(doc, url) == "multiple") {
+		Zotero.selectItems(getSearchResults(doc, url), function (items) {
+			if (!items) {
+				return true;
+			}
+			var urls = [];
+			for (var i in items) {
+				urls.push(i);
+			}
+			ZU.processDocuments(urls, scrape);
+		});
+	} else {
+		scrape(doc, url);
+	}
+
+}
+
+function scrape(doc, url) {
+	if (url.match(/\/newspaper\/article\//i)) {
+		scrapeNewspaper(doc, url);
+	} else {
+		scrapeWork(doc, url);
+	}
+}
+
+function scrapeNewspaper(doc, url) {
+	// There's a BibTex citation embedded in the page, seems like the easiest way to get the details.
+
+	var bibtex = ZU.xpathText(doc, "//textarea[@id='bibtex-citation']");
+
+	var translator = Zotero.loadTranslator("import");
+	translator.setTranslator("9cb70025-a888-4a29-a210-93ec52da40d4");
+	translator.setString(bibtex);
+
+	// Clean up the BibTex results and add some extra stuff.
+
+	translator.setHandler("itemDone", function (obj, item) {
+		var articleID = item.url.match(/nla\.news-article\d+/)[0];
+		item.itemType = 'newspaperArticle';
+		item.abstractNote = ZU.xpathText(doc, "//meta[@property='og:description']/@content");
+		delete item.type;
+		delete item.itemID;
+
+		// Attaching a PDF is a 2 stage process.
+		// First you need to tell the service to start generating the PDF.
+		// You get back a hash id to include in the PDF url.
+
+		var renditionURL = "http://trove.nla.gov.au/newspaper/rendition/"
+		var prepURL = renditionURL + articleID + '/level/3/prep'
+
+		ZU.HTTP.doGet(prepURL, function(hashID) {
+
+			// It takes some time to generate the pdfs after the prep call.
+			// You can ping the service to see if the PDF is ready,
+			// but when I tried this I was denied permission to the statusText of the response object 
+			// which is necessary because statusText == 'Locked' if the PDF isn't ready.
+			// So I took out the ping call completely and just put in a 5 second pause which seems to work ok.
+			// Note that the service itself uses a 5 second delay if the ping fails.
+
+			var timeout = Date.now() + 5000;
+			while (Date.now() < timeout){}
+			item.attachments.push({url: renditionURL + articleID + '.3.pdf?followup=' + hashID, title: item.publicationTitle + ', ' + item.date + ', p. ' + item.pages, mimeType:'application/pdf'});
+
+			// Get the OCRd text and save in a note.
+			var textURL = renditionURL + articleID + '.txt';
+			ZU.HTTP.doGet(textURL, function(text) {
+				item.notes.push({note: text.trim()});
+				item.complete();
+			});
+
+		});
+	});
+	translator.translate();	
+}
+
+var troveTypes = {
+	"Book": "book",
+	"Article Article/Book chapter": "bookSection",
+	"Thesis": "thesis",
+	"Archived website": "webpage",
+	"Conference Proceedings": "book",
+	"Audio book": "book",
+	"Article": "journalArticle",
+	"Article Article/Journal or magazine article": "journalArticle",
+	"Article Article/Conference paper": "conferencePaper",
+	"Article Article/Report": "report",
+	"Photograph": "artwork",
+	"Poster, chart, other": "artwork",
+	"Art work": "artwork",
+	"Object": "artwork",
+	"Microform Photograph": "artwork",
+	"Microform Object": "artwork",
+	"Sound": "audioRecording",
+	"Video": "videoRecording",
+	"Printed music": "book",
+	"Map": "map",
+	"Unpublished": "manuscript",
+	"Published": "document"
+}
+	
+
+function checkType(string) {
+	types = string.split("; ");
+	newString = types.join(" ");
+	if (newString in troveTypes) {
+		return troveTypes[newString];
+	} else {
+		while (types.length > 0) {
+			types.pop();
+			newString = types.join(" ");
+			if (newString in troveTypes) {
+				return troveTypes[newString];
+			}
+
+		}
+	}
+	return "book";
+}
+
+function scrapeWork(doc, url) {
+	var thumbnailURL;
+
+	// Remove all params from url
+	var workURL = url.match(/http.*\/work\/\d+/)[0];
+	var bibtexURL = workURL + '?citationFormat=BibTeX';
+
+	// Need to get version identifier for the BibText url
+	var versionID = doc.body.innerHTML.match(/displayCiteDialog\(\'(.+?)\'/);
+	if (versionID !== null) {
+		bibtexURL += '&selectedversion=' + versionID[1];
+		thumbnailURL = ZU.xpathText(doc, "//a/img[@class='mosaic ui-shdw']/@src");
+	} else {
+		// It's a work -- so thumbnails are different
+		thumbnailURL = ZU.xpathText(doc, "//li[@class='imgfirst']//img/@src");
+	}
+
+	// Get the BibTex and feed it to the translator.
+	ZU.HTTP.doGet(bibtexURL, function(bibtex) {
+		var translator = Zotero.loadTranslator("import");
+		translator.setTranslator("9cb70025-a888-4a29-a210-93ec52da40d4");
+		translator.setString(bibtex);
+		translator.setHandler("itemDone", function (obj, item) {
+			item.itemType = checkType(item.type);
+
+			// This gives a better version-aware url.
+			item.url = ZU.xpathText(doc, "//meta[@property='og:url']/@content");
+			item.abstractNote = ZU.xpathText(doc, "//meta[@property='og:description']/@content");
+			delete item.itemID;
+			delete item.type;
+			if (thumbnailURL !== null) {
+				item.attachments.push({url: thumbnailURL, title: item.title, mimeType:'image/jpeg'});
+			}
+			item.complete();
+		});
+		translator.translate();
+	});
+}
+
+/** BEGIN TEST CASES **/
+var testCases = [
+	{
+		"type": "web",
+		"url": "http://trove.nla.gov.au/version/208674371",
+		"items": [
+			{
+				"itemType": "artwork",
+				"title": "THIRROUL - Hotels - Rex Hotel",
+				"creators": [
+					{
+						"firstName": "1910-1981, William A. (William Alan)",
+						"lastName": "Bayley",
+						"creatorType": "author"
+					}
+				],
+				"date": "1960",
+				"abstractNote": "Thirroul Hotels REX HOTEL ca. 1956 1950-1960 Illawarra Region",
+				"libraryCatalog": "Trove",
+				"url": "http://trove.nla.gov.au/version/208674371",
+				"attachments": [
+					{
+						"title": "THIRROUL - Hotels - Rex Hotel",
+						"mimeType": "image/jpeg"
+					}
+				],
+				"tags": [],
+				"notes": [],
+				"seeAlso": []
+			}
+		]
+	},
+	{
+		"type": "web",
+		"url": "http://trove.nla.gov.au/version/11567057",
+		"items": [
+			{
+				"itemType": "book",
+				"title": "Experiences of a meteorologist in South Australia",
+				"creators": [
+					{
+						"firstName": "Clement L. (Clement Lindley)",
+						"lastName": "Wragge",
+						"creatorType": "author"
+					}
+				],
+				"date": "1980",
+				"ISBN": "9780908065073",
+				"abstractNote": "In 14 libraries. 24 p. : ill. ; 22 cm. Wragge, Clement L. (Clement Lindley), 1852-1922. South Australia. Climate, 1883-1884. Meteorologists -- South Australia -- Biography. South Australia -- Climate -- History.",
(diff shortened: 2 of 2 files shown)


Trove.js
```diff
@@ -0,0 +1,326 @@
...
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU Affero General Public License for more details.
+
+   You should have received a copy of the GNU Affero General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+function detectWeb(doc, url) {
+	if (url.match(/\/result\?/) || url.match(/\/newspaper\/page/)) {
```
adam3smith: here and below, we prefer using .indexOf("string") != -1 to test for literal strings and .search(/regex/) != -1 where you need rege capabilities. It's a lot more efficient.

Trove.js
```diff
@@ -0,0 +1,326 @@
...
+		item.itemType = 'newspaperArticle';
+		item.abstractNote = ZU.xpathText(doc, "//meta[@property='og:description']/@content");
+		delete item.type;
+		delete item.itemID;
+
+		// Attaching a PDF is a 2 stage process.
+		// First you need to tell the service to start generating the PDF.
+		// You get back a hash id to include in the PDF url.
+
+		var renditionURL = "http://trove.nla.gov.au/newspaper/rendition/"
```
adam3smith: using a relative link should be more stable, e.g. if the site supports https (i.e. `"/newspaper/rendition"`)

Trove.js
```diff
@@ -0,0 +1,326 @@
...
+			// It takes some time to generate the pdfs after the prep call.
+			// You can ping the service to see if the PDF is ready,
+			// but when I tried this I was denied permission to the statusText of the response object 
+			// which is necessary because statusText == 'Locked' if the PDF isn't ready.
+			// So I took out the ping call completely and just put in a 5 second pause which seems to work ok.
+			// Note that the service itself uses a 5 second delay if the ping fails.
+
+			var timeout = Date.now() + 5000;
+			while (Date.now() < timeout){}
+			item.attachments.push({url: renditionURL + articleID + '.3.pdf?followup=' + hashID, title: item.publicationTitle + ', ' + item.date + ', p. ' + item.pages, mimeType:'application/pdf'});
```
adam3smith: we perfer a generic label like `Trove Full Text PDF` for the title here (the filename is generated automatically by Zotero)

Trove.js
```diff
@@ -0,0 +1,326 @@
...
+		}
+	}
+	return "book";
+}
+
+function scrapeWork(doc, url) {
+	var thumbnailURL;
+
+	// Remove all params from url
+	var workURL = url.match(/http.*\/work\/\d+/)[0];
```
adam3smith: we're positive this always matches? Otherwise the [0] throws an error here which should be avoided.

Trove.js
```diff
@@ -0,0 +1,326 @@
...
+	{
+		"type": "web",
+		"url": "http://trove.nla.gov.au/version/208674371",
+		"items": [
+			{
+				"itemType": "artwork",
+				"title": "THIRROUL - Hotels - Rex Hotel",
+				"creators": [
+					{
+						"firstName": "1910-1981, William A. (William Alan)",
```
adam3smith: we should be able to do better than this, no? Can we clean up names some?

Trove.js
```diff
@@ -0,0 +1,326 @@
...
+					{
+						"firstName": "Clement L. (Clement Lindley)",
+						"lastName": "Wragge",
+						"creatorType": "author"
+					}
+				],
+				"date": "1980",
+				"ISBN": "9780908065073",
+				"abstractNote": "In 14 libraries. 24 p. : ill. ; 22 cm. Wragge, Clement L. (Clement Lindley), 1852-1922. South Australia. Climate, 1883-1884. Meteorologists -- South Australia -- Biography. South Australia -- Climate -- History.",
+				"extra": "Reprinted from Good words for 1887/ edited by Donald Macleod, published: London: Isbister and Co",
```
adam3smith: which field in Bibtex is this? I'd generally prefer this in a note, but that might require fixing bibtex import rather than here.

adam3smith:
See comments inline; looks good to me otherwise.

Trove.js
wragge: It's in the 'note' field. I haven't done anything about this.

Trove.js
wragge: Done.

Trove.js
wragge: Done.

Trove.js
wragge: Done -- named 'Trove Newspaper PDF' (not strictly full text as it's actually the images).

Trove.js
wragge: Improved -- uses indexOf and slice.

Trove.js
wragge: I think it's fairly rare for the dates to creep in to the BibTex, but I've added a function to clean dates out of first names.

wragge:
Thanks for the comments. I think I've addressed them all.

Trove.js
```diff
@@ -0,0 +1,340 @@
...
+		return "newspaperArticle";
+	} else if (url.indexOf('/work/') != -1) {
+		return "document";
+	}
+}
+
+function getSearchResults(doc, url) {
+	var items = {};
+	var results;
+	if (url.match(/\/result\?/)) {
```
adam3smith: some more `match` to be replaced here

Trove.js
```diff
@@ -0,0 +1,340 @@
...
+		results = ZU.xpath(doc, "//div[@id='mainresults']//li/dl/dt/a");
+	} else if (url.match(/\/newspaper\/page/)) {
```
adam3smith: here

Trove.js
```diff
@@ -0,0 +1,340 @@
...
+			ZU.processDocuments(urls, scrape);
+		});
+	} else {
+		scrape(doc, url);
+	}
+
+}
+
+function scrape(doc, url) {
+	if (url.match(/\/newspaper\/article\//i)) {
```
adam3smith: and here

Trove.js
```diff
@@ -0,0 +1,340 @@
...
+			item.itemType = checkType(item.type);
+			item.creators = cleanCreators(item.creators);
+
+			// This gives a better version-aware url.
+			item.url = ZU.xpathText(doc, "//meta[@property='og:url']/@content");
+			item.abstractNote = ZU.xpathText(doc, "//meta[@property='og:description']/@content");
+			delete item.itemID;
+			delete item.type;
+			if (thumbnailURL !== null) {
+				item.attachments.push({url: thumbnailURL, title: item.title, mimeType:'image/jpeg'});
```
adam3smith: and if you could use a generic title here, too, as above. I overlooked this on the first go.

Trove.js
```diff
@@ -0,0 +1,340 @@
...
+		ZU.HTTP.doGet(prepURL, function(hashID) {
+
+			// It takes some time to generate the pdfs after the prep call.
+			// You can ping the service to see if the PDF is ready,
+			// but when I tried this I was denied permission to the statusText of the response object 
+			// which is necessary because statusText == 'Locked' if the PDF isn't ready.
+			// So I took out the ping call completely and just put in a 5 second pause which seems to work ok.
+			// Note that the service itself uses a 5 second delay if the ping fails.
+
+			var timeout = Date.now() + 5000;
```
adam3smith: @aurimasv @zuphilip  -- that's the only potentially problematic bit I see otherwise. Looks good to me, any concerns from your side?

adam3smith:
@wragge  -- you may want to wait with fixing my other nits above until we hear back from a 2nd person about the timeout, up to you.

Trove.js
wragge: As I note in the comment, this is the only way I can see to reliably get the PDF. If I could get access to the response object I could do a quick check, but if that came back "Locked" I'd still need to do the timeout. This is exactly what the Trove UI itself does -- pings, if status is locked it includes a 5 second timeout (except that it can use setTimeout which seems to be blocked within Zotero).

wragge:
Argh sorry didn't mean for that last commit to be added to this pull request. It's only very minor though...

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+function getSearchResults(doc, url) {
+	var items = {};
+	var results;
+	if (url.indexOf('/result?') != -1) {
+		results = ZU.xpath(doc, "//div[@id='mainresults']//li/dl/dt/a");
+	} else if (url.indexOf('/newspaper/page') != -1) {
+		results = ZU.xpath(doc, "//ol[@class='list-unstyled articles']/li/h4/a");
+	}
+	for (var i=0; i<results.length; i++) {
+		var url = results[i].href;
```
zuphilip: The variable `url` is a parameter of the function and therefore the variable is already defined. I would prefer here to use another name for this new variable.

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+		item.itemType = 'newspaperArticle';
+		item.abstractNote = ZU.xpathText(doc, "//meta[@property='og:description']/@content");
+		delete item.type;
+		delete item.itemID;
+
+		// Attaching a PDF is a 2 stage process.
+		// First you need to tell the service to start generating the PDF.
+		// You get back a hash id to include in the PDF url.
+
+		var renditionURL = "/newspaper/rendition/"
```
zuphilip: Missing semicolon

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+		var prepURL = renditionURL + articleID + '/level/3/prep'
```
zuphilip: Missing semicolon

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+	"Object": "artwork",
+	"Microform Photograph": "artwork",
+	"Microform Object": "artwork",
+	"Sound": "audioRecording",
+	"Video": "videoRecording",
+	"Printed music": "book",
+	"Map": "map",
+	"Unpublished": "manuscript",
+	"Published": "document"
+}
```
zuphilip: Missing semicolon

Trove.js
adam3smith: ah yes, most definitely, good catch.

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+	var results;
+	if (url.indexOf('/result?') != -1) {
+		results = ZU.xpath(doc, "//div[@id='mainresults']//li/dl/dt/a");
+	} else if (url.indexOf('/newspaper/page') != -1) {
+		results = ZU.xpath(doc, "//ol[@class='list-unstyled articles']/li/h4/a");
+	}
+	for (var i=0; i<results.length; i++) {
+		var url = results[i].href;
+		var title = ZU.trimInternal(results[i].textContent);
+		items[url] = title;
```
adam3smith: and then here, too, of course.

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+		}
+	}
+	return "book";
+}
+
+function cleanCreators(creators) {
+	newCreators = [];
+	for (var i = 0; i < creators.length; i++) {
+		var creator = creators[i];
+		creator.firstName = creator.firstName.replace(/\(?\d{4}-\d{0,4}\)?,?/, "").trim()
```
zuphilip: Missing semicolon

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+		newCreators.push(creator);
+	}
+	return newCreators
```
zuphilip: Missing semicolon

Trove.js
zuphilip: Yes, this looks problematic and it could lead to larger waiting times.
E.g. the 20 items of [this search](http://trove.nla.gov.au/book/result?l-australian=y&q=wragge) would lead to 100 sec waiting time, which is more than 1,5 minutes!

adam3smith: but is this a trade-off worth making? Would it be that confusing if a download of 20 items with images takes a while? 
The way I understand, the alternative is removing that component altogether, right?
One issue I'm seeing is that this would halt the entire import process. I haven't thought this through, but could we put the attachment process into an asynchronous function, so the item saves and we're "just" waiting for the attachment?

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+			var timeout = Date.now() + 5000;
+			while (Date.now() < timeout){}
+			item.attachments.push({url: renditionURL + articleID + '.3.pdf?followup=' + hashID, title: 'Trove newspaper PDF', mimeType:'application/pdf'});
+
+			// Get the OCRd text and save in a note.
+			var textURL = renditionURL + articleID + '.txt';
+			ZU.HTTP.doGet(textURL, function(text) {
+				item.notes.push({note: text.trim()});
+				item.complete();
+			});
```
zuphilip: It looks that the `hashID` is not needed here, and therefore I guess you can put this lines also outside the outer `doGet`.

wragge: Where would the item.complete() go then?

Trove.js
wragge: Yeah but you wouldn't be waiting 100 sec without anything happening. You'd see an item was being processed every 5 sec or so. And in any case if you asked any user would they rather wait for 100 sec, or manually view, download, and add to Zotero every individual item I can tell you what the answer would be...

The automatic download of newspaper article PDFs from Trove has brought 'wows' in various workshop/demos I've given -- it's been a very popular feature and I would not want to have it removed.

zuphilip: Okay, I am convinced that the waiting time is worth the advantages.

(OT @wragge I like your presentation of your digitalized newspapers very much and maybe I can contact you on that separately.)

Trove.js
zuphilip: Ah, I missed that one. Maybe write a comment then.

National Archives of Australia.js
```diff
@@ -207,14 +207,10 @@ function scrapeSeries(doc, url) {
...
+	item.archiveLocation = meta['series number'];
 	
 	var seriesNumber = encodeURIComponent(meta['series number']);
-	item.attachments.push({
-		title: "National Archives of Australia Record",
-		url: 'http://www.naa.gov.au/cgi-bin/Search?O=S&Number=' + seriesNumber,
-		mimeType: 'text/html',
-		snapshot: false
-	})
+	item.url = 'http://www.naa.gov.au/cgi-bin/Search?O=S&Number=' + seriesNumber;
```
aurimasv: Update tests please.

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU Affero General Public License for more details.
+
+   You should have received a copy of the GNU Affero General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+function detectWeb(doc, url) {
+	if (url.indexOf('/result?') != -1 || url.indexOf('/newspaper/page') != -1) {
+		return "multiple";
```
aurimasv: make sure there are actual results to import (i.e. call `getSearchResults`)

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+	} else if (url.indexOf('/newspaper/article') != -1) {
+		return "newspaperArticle";
+	} else if (url.indexOf('/work/') != -1) {
+		return "document";
```
aurimasv: Since this is not actually the item type that will be imported, the test that you included at the bottom of the translator will fail. I would probably use the most likely item type here (seems to be artwork).

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+}
+
+function getSearchResults(doc, url) {
+	var items = {};
+	var results;
+	if (url.indexOf('/result?') != -1) {
+		results = ZU.xpath(doc, "//div[@id='mainresults']//li/dl/dt/a");
+	} else if (url.indexOf('/newspaper/page') != -1) {
+		results = ZU.xpath(doc, "//ol[@class='list-unstyled articles']/li/h4/a");
+	}
```
aurimasv: And if neither?

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+	for (var i=0; i<results.length; i++) {
+		var url = results[i].href;
+		var title = ZU.trimInternal(results[i].textContent);
+		items[url] = title;
+	}
+	return items;
```
aurimasv: generally, our pattern has been to return `false` if there were no results and the object otherwise. It's a bit annoying to check that there is nothing in the object otherwise.

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+
+			// It takes some time to generate the pdfs after the prep call.
+			// You can ping the service to see if the PDF is ready,
+			// but when I tried this I was denied permission to the statusText of the response object 
+			// which is necessary because statusText == 'Locked' if the PDF isn't ready.
+			// So I took out the ping call completely and just put in a 5 second pause which seems to work ok.
+			// Note that the service itself uses a 5 second delay if the ping fails.
+
+			var timeout = Date.now() + 5000;
+			while (Date.now() < timeout){}
```
aurimasv: This is not good. The way we process translator code may have changed recently (in prep for e10s, perhaps), but this would effectively lock up the browser for 5 seconds, no? I know we don't expose a number of global (window) functions in translators, though you may still be able to get to them via document object (`document.defaultView`). See if you can use that to get `setTimeout`. If no, and this does lock up the browser, then I would say this is not acceptable.

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+			item.attachments.push({url: renditionURL + articleID + '.3.pdf?followup=' + hashID, title: 'Trove newspaper PDF', mimeType:'application/pdf'});
```
aurimasv: use multiple lines

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+
+			// Get the OCRd text and save in a note.
```
aurimasv: I'm not sure that we should be getting the OCRd text and attaching the PDF. I think one or the other, but not both.

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+	"Art work": "artwork",
+	"Object": "artwork",
+	"Microform Photograph": "artwork",
+	"Microform Object": "artwork",
+	"Sound": "audioRecording",
+	"Video": "videoRecording",
+	"Printed music": "book",
+	"Map": "map",
+	"Unpublished": "manuscript",
+	"Published": "document"
```
aurimasv: not a great choice, since it has very limited fields and is not cited well. We generally default to journalArticle or book.

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+}
+	
+
+function checkType(string) {
```
aurimasv: Could you add a comment describing what this function does? What's the expected format of the input to this function?

Also, please use something more descriptive than "string"

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+	types = string.split("; ");
```
aurimasv: missing `var ` (here and below)

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+	newString = types.join(" ");
+	if (newString in troveTypes) {
```
aurimasv: `.hasOwnProperty` is actually what you want

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+		return troveTypes[newString];
+	} else {
+		while (types.length > 0) {
+			types.pop();
```
aurimasv: pop at the end of the while and skip the whole if statement above

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+				return troveTypes[newString];
+			}
+
+		}
+	}
+	return "book";
+}
+
+function cleanCreators(creators) {
+	newCreators = [];
```
aurimasv: `var`

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+	for (var i = 0; i < creators.length; i++) {
+		var creator = creators[i];
+		creator.firstName = creator.firstName.replace(/\(?\d{4}-\d{0,4}\)?,?/, "").trim()
+		newCreators.push(creator);
```
aurimasv: You're modifying the passed-in `creators` object anyway. Why return a new object in the end?

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+	}
+	return newCreators
+}
+
+function scrapeWork(doc, url) {
+	var thumbnailURL;
```
aurimasv: nit: move closer to where it's used

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+
+	// Remove all params from url
+	var workURL = url.slice(0, url.indexOf('?'));
```
aurimasv: I prefer `.replace(/[?#].*/, '')`, since the logic above does not account for `?` missing altogether. (`var a = 'abcd'; a.slice(0, a.indexOf('e')) == 'abc'`)

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+	var bibtexURL = workURL + '?citationFormat=BibTeX';
+
+	// Need to get version identifier for the BibText url
+	var versionID = doc.body.innerHTML.match(/displayCiteDialog\(\'(.+?)\'/);
+	if (versionID !== null) {
+		bibtexURL += '&selectedversion=' + versionID[1];
+		thumbnailURL = ZU.xpathText(doc, "//a/img[@class='mosaic ui-shdw']/@src");
+	} else {
+		// It's a work -- so thumbnails are different
+		thumbnailURL = ZU.xpathText(doc, "//li[@class='imgfirst']//img/@src");
```
aurimasv: What happens here with the bibtexURL? Is it ok for it not to have the ID? Please comment

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+		translator.setTranslator("9cb70025-a888-4a29-a210-93ec52da40d4");
+		translator.setString(bibtex);
+		translator.setHandler("itemDone", function (obj, item) {
+			item.itemType = checkType(item.type);
+			item.creators = cleanCreators(item.creators);
+
+			// This gives a better version-aware url.
+			item.url = ZU.xpathText(doc, "//meta[@property='og:url']/@content");
+			item.abstractNote = ZU.xpathText(doc, "//meta[@property='og:description']/@content");
+			delete item.itemID;
```
aurimasv: generally speaking, you don't need to worry about deleting these. If they don't belong to a type, they will be omitted automatically.

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+			delete item.type;
+			if (thumbnailURL !== null) {
+				item.attachments.push({url: thumbnailURL, title: 'Trove thumbnail image', mimeType:'image/jpeg'});
```
aurimasv: use multiple lines, but is this actually valuable? Could we not attach the full image? (I'm guessing this may be too complicated) Attaching the thumbnail may give people the wrong impression that they have a full copy of the image (since we don't attach thumbnails or incomplete articles anywhere else). It would probably be more useful to attach a link to the website hosting the full image.

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+	{
+		"type": "web",
+		"url": "http://trove.nla.gov.au/version/208674371",
+		"items": [
+			{
+				"itemType": "artwork",
+				"title": "THIRROUL - Hotels - Rex Hotel",
+				"creators": [
+					{
+						"firstName": "William A. (William Alan)",
```
aurimasv: Could we clean this up?

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+				"title": "Experiences of a meteorologist in South Australia",
+				"creators": [
+					{
+						"firstName": "Clement L. (Clement Lindley)",
+						"lastName": "Wragge",
+						"creatorType": "author"
+					}
+				],
+				"date": "1980",
+				"ISBN": "9780908065073",
```
aurimasv: Was this imported with Scaffold? I thought Scaffold would hyphenate these automatically, no?

Trove.js
```diff
@@ -0,0 +1,337 @@
...
+				"attachments": [
+					{
+						"title": "Trove newspaper PDF",
+						"mimeType": "application/pdf"
+					}
+				],
+				"tags": [],
+				"notes": [
+					{
+						"note": "'WRAGGE' - we have received a copy of the above, which is a journal devoted chiefly to the science of meteorology. It is owned and conducted by Mr. Clement Wragge."
```
aurimasv: Is this a useful note? Are notes in general from this website useful?

aurimasv:
Please take a look at some of the comments that got hidden by your recent commits. Looks like we overlapped.

Trove.js
aurimasv: Actually, this may not be possible. Since setTimeout is async, you would need to call incrementAsyncProcesses before and decrementAsyncProcesses at the end of the call. Otherwise Zotero doesn't know that it needs to wait for something else to finish. Those functions are not exposed to translators. So, if the while loop locks up the browser, which I think it does, this would not be acceptable.

Trove.js
zuphilip: It is the full text of the short newspaper article, see http://trove.nla.gov.au/newspaper/article/70068753 (the OCR version of the article is saved into a note)

Trove.js
wragge: For reasons explained above the PDF attachment needs to be there. Users will cope with a 5 sec delay.

Trove.js
wragge: Yes both. PDF contains the article images. OCR is searchable text.

Trove.js
wragge: Trove is an aggregator -- images come from hundreds of different sources. Thumbnail links are the only reliable ones. I believe users will find them valuable as a quick check on the identity of the image. But I will also attach a link to the providing organisations where they are provided (this will be a link to an HTML page not to an image).

Trove.js
wragge: Yes it's useful

Trove.js
adam3smith: @aurimasv is raising a different issue though: A 5 sec *delay* is one thing, and I'm fine with that. A 5 sec (let alone 1min +) *browser freeze* I don't think is acceptable behavior for Zotero to cause, no matter how useful the PDF attachment. 
I haven't actually tested this yet, so if the browser doesn't freeze, I completely agree it's worth it.

wragge: Not sure what is meant by freeze. If I select multiple newspaper articles using Zotero I can switch tabs while I'm waiting and do whatever else I want in the browser -- so it's not frozen?

Trove.js
zuphilip: Doesn't have the pdf the text as well? Zotero can index pdfs and search them as well.

Trove.js
dstillman: @wragge Are you testing this with Standalone or Firefox? I'm pretty sure this would lock up Firefox. With Standalone, it might run in the connector, but I think it would lock up the tab and peg the CPU in its helper process, even if you can still switch to another tab.

wragge:
I think I've addressed most of the issues raised.

To get around the PDF timeout issue, I've created a proxy server which returns the PDF url when it's ready.

While I was thinking about this, they changed the location of the BibTex files for newspaper articles, so I had to fix that as well.

adam3smith:
just wanted to say that this isn't being ignored; it just hit at a very busy time, I'll review asap, but worst case might take another two weeks. Apologies.

wragge:
Ok, no worries. Thanks for letting me know.

Trove.js
adam3smith: this still needs to get fixed: `var newCreators = [];`

Trove.js
adam3smith: still need an answer for this.

Trove.js
adam3smith: could you push this into a note and remove from extra? I'd rather not change bibtex import in general, but this is a poor fit for extra.

Trove.js
```diff
@@ -0,0 +1,334 @@
...
+					}
+				],
+				"seeAlso": []
+			}
+		]
+	},
+	{
+		"type": "web",
+		"url": "http://trove.nla.gov.au/newspaper/result?l-australian=y&q=wragge",
```
adam3smith: I get an error testing on this page. The Trove side throws
`main / UI Layout Initialization Warning
The layout-container "main layout / DIV#ui-layout-main" has no height.
Therefore the layout is 0-height and hence 'invisible'!`

Trove.js
```diff
@@ -0,0 +1,334 @@
...
+		"items": "multiple"
+	},
+	{
+		"type": "web",
+		"url": "http://trove.nla.gov.au/book/result?l-australian=y&q=wragge",
```
adam3smith: this one works OK .

zuphilip:
Superseeded by #1208.

PR #1045, opened 2016-04-07
Add journal title and abbreviation.

Jmuccigr:


adam3smith:
thanks. We'd want the tests updated, too -- see if that works with regular Scaffold. It might not, in which case I'll probably just do this for you.

Jmuccigr:
OK, let me try that.

Jmuccigr:
OK, more updates in there now along with the examples (which, if I understand it correctly, were already wrong). Again, I **think** I did it right.

Bryn Mawr Classical Review.js
```diff
@@ -159,18 +152,20 @@ var testCases = [
 				"seeAlso": [],
 				"attachments": [
 					{
-						"title": "Review of: The Classical Tradition",
+						"title": "Review of: Hamburg, Museum für Kunst und Gewerbe, Band 2: Unteritalisch rotfigure Keramik. Corpus vasorum antiquorum. Deutschland, Bd. 91; Hamburg Bd. 2"
```
aurimasv: this doesn't look right for http://bmcr.brynmawr.edu/2013/2013-01-44.html

Bryn Mawr Classical Review.js
```diff
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2015-06-02 10:52:41"
+	"lastUpdated": "2016-04-07 21:36:13"
 }
```
zuphilip: AFAIS the contributors to this translators are the usual persons and John. Therefore, we can also add the AGPL license block here. CC @adam3smith @aurimasv Right?

Bryn Mawr Classical Review.js
zuphilip: John, maybe you did a typo http://bmcr.brynmawr.edu/2013/2013-01-04.html

@aurimasv Normal testing in scaffold is currently broken https://github.com/zotero/scaffold/issues/26 but I have a PR https://github.com/zotero/scaffold/pull/27 for it.

Jmuccigr:
Yep, I didn't start with Scaffold, so I was doing some cutting and pasting and lost a few commas. They're in there now.

As for the AGPL license, that works for me.

Bryn Mawr Classical Review.js
```diff
@@ -76,9 +95,9 @@ function doWeb(doc, url) {
...
 		if (doc.evaluate('/html/body/center/table/tbody/tr/td/center/table/tbody/tr/td//font', doc, nsResolver, XPathResult.ANY_TYPE, null).iterateNext()) {
-			item.date = Zotero.Utilities.trimInternal(doc.evaluate('/html/body/center/table/tbody/tr/td/center/table/tbody/tr/td//font', doc, nsResolver, XPathResult.ANY_TYPE, null).iterateNext().textContent.replace("Bryn Mawr Classical Review ", "").replace(/\./g, "/"));
+			item.date = Zotero.Utilities.trimInternal(doc.evaluate('/html/body/center/table/tbody/tr/td/center/table/tbody/tr/td//font', doc, nsResolver, XPathResult.ANY_TYPE, null).iterateNext().textContent.replace("Bryn Mawr Classical Review ", "").replace(/\./g, "/")).substring(0,7);
 		} else {
-			item.date = Zotero.Utilities.trimInternal(doc.evaluate('/html/body/h3', doc, nsResolver, XPathResult.ANY_TYPE, null).iterateNext().textContent.replace("Bryn Mawr Classical Review ", "").replace(/\./g, "/"))
+			item.date = Zotero.Utilities.trimInternal(doc.evaluate('/html/body/h3', doc, nsResolver, XPathResult.ANY_TYPE, null).iterateNext().textContent.replace("Bryn Mawr Classical Review ", "").replace(/\./g, "/")).substring(0,7)
```
aurimasv: Why this change (here and above) to omit the day?

aurimasv: Oh, ok. I think I understand. The last two digits are not a date. Is that an article number? Could we put it into "pages" field? We generally use that for electronic article number.

Bryn Mawr Classical Review.js
```diff
@@ -134,23 +154,13 @@ var testCases = [
...
-						"lastName": "Grafton",
-						"creatorType": "reviewedAuthor"
-					},
-					{
-						"firstName": "Glenn W.",
-						"lastName": "Most",
-						"creatorType": "reviewedAuthor"
+ 						"firstName": "Keely Elizabeth",
+ 						"lastName": "Heuer",
+ 						"creatorType": "author"
```
aurimasv: This still looks wrong. "Reviewed by Christina S. Kraus"

Bryn Mawr Classical Review.js
```diff
@@ -134,23 +154,13 @@ var testCases = [
...
 					},
 					{
-						"firstName": "Salvatore",
-						"lastName": "Settis",
+						"firstName": "Rolf",
+						"lastName": "Hurschmann",
```
aurimasv: This is wrong too

zuphilip: One article is
http://bmcr.brynmawr.edu/2013/2013-01-04.html
and the other
http://bmcr.brynmawr.edu/2013/2013-01-44.html

@Jmuccigr Just make sure that url you provide and the data are corresponding to each other.

Bryn Mawr Classical Review.js
Jmuccigr: No, it's not a page number. It's just an ordinal. There are no page numbers  in this online journal.

Bryn Mawr Classical Review.js
Jmuccigr: Yes, the second test case is 04, not 44. It looks like the original test case had both 04 and 44.

The other test case is http://bmcr.brynmawr.edu/2010/2010-01-02.html, again inherited from the original.

To confuse things, Kraus was the author of both the current first example and the mistaken second example. (I've not dug further back into the history of the file to see what happened.)

aurimasv: >Yes, the second test case is 04, not 44. It looks like the original test case had both 04 and 44.

Where did it have both? All I see is 44 and all the metadata looks right (except for the date). Am I missing something?

Bryn Mawr Classical Review.js
aurimasv: I would be ok with the whole date/ordinal thing being stored as the article number in `pages` field. (I know it's not literally the page number, but for citation purposes it serves the same purpose)

Bryn Mawr Classical Review.js
Jmuccigr: Sorry, you're right. I was looking at the version on my computer. In any case, there are now two examples and both work, no?

Bryn Mawr Classical Review.js
Jmuccigr: I'd rather get some definitive word from BMCR on how they'd like to see these cited.

As I said, in earlier years the numbers didn't indicate the month, so it might be best to use the whole numerical bit as the issue, or at least handle the different numbering systems differently.

I don't know how it works elsewhere in Zotero, but putting issue info in as a page seems odd to me. You're going to get cites that says "p. 18" when there are no such pages and when the article appears to go on for multiple pages when printed.

Jmuccigr:
I'm still working with the editor on the various changes over time to the journal's numbering system, but I think this set of changes is at least an improvement over the current. Are there other issues to address right now?

Thanks.

Jmuccigr:
?

Jmuccigr:
Anything?

adam3smith:
sorry, you have to be more patient on translators. They're more critical code than citation styles (in that one broken translator can break Zotero for all users, not in that they're more important), so review needs to be thorough and can take a bit of time. Ping every couple of weeks so it doesn't get forgotten.

zuphilip:
It looks fine for me. Let me ping @adam3smith and see if he can do a final review.

Jmuccigr:
Good to know about the pace of things. Responses came pretty quick before, so I figured this had just gotten lost.

As I wrote above, some more needs to be done with this to account for the change in numbering, but this version is at least an improvement over the existing, which will give dates like April 46th for the latest review.

Bryn Mawr Classical Review.js
```diff
@@ -9,9 +9,26 @@
...
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+	GNU Affero General Public License for more details.
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+	***** END LICENSE BLOCK *****
+*/
+
 function detectWeb(doc, url) {
 	if (url.match(/by_reviewer/) || url.match(/by_author/) || url.match(/recent.html/) || url.match(/\/\d{4}\/$/)) {
```
adam3smith: I know this isn't your code, but while we're doing this, could we fix this properly:
`if (url.search(/by_reviewer|by_author|recent\.html|\/\d{4}\/$) != -1)`

Bryn Mawr Classical Review.js
```diff
@@ -59,6 +76,8 @@ function doWeb(doc, url) {
 	Zotero.Utilities.processDocuments(arts, function(doc) {
 		var item = new Zotero.Item("journalArticle");
 		var title = doc.evaluate('//h3/i', doc, nsResolver, XPathResult.ANY_TYPE, null).iterateNext().textContent;
```
adam3smith: another cleanup request:
replace `nsResolver` with `null` in all instances of `doc.evaluate` and delete line 41-44 above, i.e.

```
var ns = doc.documentElement.namespaceURI;
 	var nsResolver = ns ? function(prefix) {
 		if (prefix == 'x') return ns; else return null;
 	} : null;
```

Bryn Mawr Classical Review.js
```diff
@@ -59,6 +76,8 @@ function doWeb(doc, url) {
...
+		item.publicationTitle = "Bryn Mawr Classical Review";
+		item.journalAbbreviation = "BMCR";
```
adam3smith: along these lines, please add
`item.ISSN = "1055-7660"`

Bryn Mawr Classical Review.js
adam3smith: I'm OK leaving this as is for now (though I agree that very likely the article number should go into the page field), but couldn't we get the date from the URL rather than from this hideous xpath?

Bryn Mawr Classical Review.js
```diff
@@ -59,6 +76,8 @@ function doWeb(doc, url) {
 	Zotero.Utilities.processDocuments(arts, function(doc) {
```
adam3smith: Ideally, the whole translator should be slightly restructed with  `Zotero.selectItems` written as a callback function and the part below this as a separate `scrape` function. Will work better in all kinds of ways.
See https://github.com/zotero/translators/blob/master/Digital%20Medievalist.js#L25 for an example. Let me know if that's unclear.

Bryn Mawr Classical Review.js
Jmuccigr: Yes, I noticed that too, but was being minimalist in my changes. :-) Note that for either option, articles from the 90s won't work because they used just a 2-number year designator, like 98.

As I keep repeating, I'd like first to get to something that works for current articles, then fix it so that it works for all of them.

Bryn Mawr Classical Review.js
Jmuccigr: OK, did this one and updated the tests too.

I'll try to get to the other ones, though I'm not sure I can handle them all.

Bryn Mawr Classical Review.js
zuphilip: I would suggest to just use the functions `ZU.xpath` and `ZU.xpathText` for that.

zuphilip:
I will continue your work here and trying to address the remaining issues.

adam3smith:
superseded by https://github.com/zotero/translators/pull/1115

PR #1055, opened 2016-04-20
Display library catalog in Primo

adam3smith:
Using the domain with some exlibris stuff stripped out

@zuphilip -- does this makes sense to you? Seems like good idea to present better catalog name.

### DIFF
--- a/Primo.js
+++ b/Primo.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsb",
-	"lastUpdated": "2016-04-12 03:30:36"
+	"lastUpdated": "2016-04-20 21:14:30"
 }
 
 /*
@@ -178,7 +178,7 @@ function fetchPNX(itemData) {
 				PNXUrlGenerator.confirmed = true;
 			}
 			
-			importPNX(text);
+			importPNX(text, url);
 		},
 		function() {
 			if(!gotPNX && PNXUrlGenerator.nextFunction()) {
@@ -193,8 +193,9 @@ function fetchPNX(itemData) {
 }
 
 //import PNX record
-function importPNX(text) {
+function importPNX(text, url) {
 	//Note that if the session times out, PNX record will just contain a "null" entry
+	//Z.debug(url)
 	Z.debug(text);
 	//a lot of these apply only to prim records, mainly (but no exclusively) served by the jsp file
 	text = text.replace(/\<\/?xml-fragment[^\>]*\>/g, "")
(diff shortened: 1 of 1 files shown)


zuphilip:
I am not sure that every Primo instance is (already) hosted by Ex Libris. But certainly one can change the url which you see, e.g. http://primo.bib.uni-mannheim.de/primo_library/libweb/action/search.do?vid=MAN_UB or http://primo.uni-trier.de/primo_library/libweb/action/search.do?mode=Basic&vid=TRI_UB_I&vl%28freeText0%29=zumstein&fn=search&tab=default_tab&

However, it looks that the institution can also be derived from the PNX field `browse/institution` at least in some abbreviated form.

adam3smith:
right, right, this doesn't assume hosting by exlibris. It just takes the domain name and then strips the exlibris part (if it's there). So for you it'd be primo.bib.uni-mannheim.de

I considered going with browse or using sourceid, but then I'd have to write a mapping table for every involved library. Otherwise it'd give me, as you say, just an abbreviation. That's no good.

zuphilip:
> Otherwise it'd give me, as you say, just an abbreviation. That's no good. 

I can agree with that but I am not sure if an url part is anyhow better. For the examples mentioned in the translator it would give us:
```
purdue-primo-prod
primo.bib.uni-mannheim.de
limo.libis.be
virtuose.uqam.ca
searchit.princeton.edu
```
I cannot say that this is for me especially nice or semantically meaningful as a `item.libraryCatalog`. But, well, maybe it is the best we can get at the moment...

Primo.js
```diff
@@ -442,6 +447,7 @@ function importPNX(text) {
 			item.attachments.push({url: ZU.xpathText(doc, '//display/lds03'), title: "HOLLIS Permalink", snapshot: false});
```
zuphilip: Actually, it would be nice to scrape the permalink in general, which normally means make another call to some url e.g. `http://primo.bib.uni-mannheim.de/primo_library/libweb/action/permalink.do?docId=MAN_ALMA21146476460002561&vid=MAN_UB&fn=permalink` and then extract that link.

adam3smith: going to leave that for another time. Not crazy about the extra call.

adam3smith:
> For the examples mentioned in the translator it would give us: ...

This is in line with what we do for library catalogs, e.g.
https://github.com/zotero/translators/blob/master/Library%20Catalog%20%28InnoPAC%29.js#L143
https://github.com/zotero/translators/blob/master/Library%20Catalog%20%28Encore%29.js#L175
https://github.com/zotero/translators/blob/master/Library%20Catalog%20%28DRA%29.js#L73

and more. Not adding " Library Catalog" here because I know it's technically not a library catalog.

Primo.js
zuphilip: Yeah, the extra call might not be worth for this information. Maybe we can also construct the permalink from the id and url... Well, as you mentioned, maybe for another time.

PR #1056, opened 2016-04-22
Add translator for Library Genesis

revcherrycoke:
This adds a new translator for [Library Genesis](http://libgen.io).

Currently only single items accessible through the "LibGen (Sci-Tech)" search are supported (i.e. scientific literature).
It uses the provided BibTex file to obtain the bulk of the metadata, with some manual adjustments. It also adds the text from the first provided mirror as an attachment.


Library Genesis.js
```diff
@@ -0,0 +1,245 @@
...
+	return 'book';
+  }
+}
+
+function doWeb(doc, url) {
+  var md5Hash = MD5_REGEX.exec(url)[1];
+
+  // To save some work, we use the provided bibtex file to retrieve the
+  // metadata and use Zotero's built-in bibtex importer
+  var bibtexUrl = "http://libgen.io/book/bibtex.php?md5=" + md5Hash;
```
adam3smith: use a relative url, i.e. `'/book/bibtex.php?md5=' + md5Hash;` Robust to https and uses the same page you're on.

Library Genesis.js
```diff
@@ -0,0 +1,245 @@
...
+  ZU.processDocuments(bibtexUrl, function(bibtexDoc) {
+	var bibtexStr = bibtexDoc.getElementsByTagName("textarea")[0].value
+	var translator = Zotero.loadTranslator('import');
+	translator.setTranslator('9cb70025-a888-4a29-a210-93ec52da40d4');
+	translator.setString(bibtexStr);
+
+	translator.setHandler('itemDone', function(obj, item) {
+	  // The bibtex messes up multiple authors, so we set them ourselves
+	  var authorStr = ZU.xpathText(doc, "//tbody/tr[3]/td[2]/b");
```
adam3smith: why not parse this from the bibtex? That seems more robust and also means you don't handle two documents at the same time. I'm worried who well that'll work e.g. in the bookmarklet. Also makes it harder to work with multiples later on.

Library Genesis.js
```diff
@@ -0,0 +1,245 @@
...
+			if (item.ISBN.indexOf(splitChar) != -1) {
+		  	item.ISBN = item.ISBN.split(splitChar)[0];
+			}
+	  	});
+	  }
+
+	  // Add the full text attachment
+	  var extension = ZU.xpathText(doc, "//tbody/tr[11]/td[4]");
+	  var downloadUrl = (
+		"http://libgen.io/get/" + md5Hash + "/" + md5Hash + "." + extension);
```
adam3smith: use relative url

Library Genesis.js
```diff
@@ -0,0 +1,245 @@
+{
+	"translatorID": "11614156-f421-4e89-8ce0-a5e69ce3ebed",
+	"label": "Library Genesis",
+	"creator": "Reverend Wicks Cherrycoke",
+	"target": "libgen.io|lib.gen.rus.ec",
```
adam3smith: `^https?://(libgen\\.io|lib\\.gen\\.rus\\.ec)/`

Library Genesis.js
```diff
@@ -0,0 +1,245 @@
...
+    if (item.ISBN) {
+      [' ', ','].forEach(function(splitChar) {
+      if (item.ISBN.indexOf(splitChar) != -1) {
+        item.ISBN = item.ISBN.split(splitChar)[0];
+      }
+      });
+    }
+
+    // Add the full text attachment
+    var extension = ZU.xpathText(doc, "//tbody/tr[11]/td[4]");
```
adam3smith: Two things here:
1. On a general level, I'm a bit concerned about the automatic download here. Some of these files are huge, and I'm not sure how comfortable I am having Zotero auto-download files with sizes like 81MB. @zuphilip @aurimasv  -- thoughts on that one?
2. On a more practical level, the xpath seems very fragile. Wouldn't it be better to test for the `following-sibling` of the node that `contains(text(), "Extension:")` or something along those lines?

adam3smith:
Apart from that one issue noted above, something also broke the indenting in your last commit. We use one tab per level. I recommend using jsbeautify for that purpose, but I'm sure there are other ways, too.

revcherrycoke:
Sorry about the indentation, it should be back to hardtabs now.
I made the XPath for the file extension as generic as possible, thank you for your suggestions!
About the issue with the file attachments, do you think it would make sense to add a threshold (e.g. "do not auto-download if size > 10mb, attach link instead")?

adam3smith:
yeah, I think a threshold is the way to go, but it should be a Zotero wide setting, ideally adjustable by users (10MB seems like a good default to me, too). 
@dstillman -- I feel like we've had this conversation before. Do I remember correctly that that's basically what you had in mind?

Library Genesis.js
```diff
@@ -0,0 +1,246 @@
...
+	"target": "^https?://(libgen\\.io|lib\\.gen\\.rus\\.ec)/",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsi",
+	"lastUpdated": "2016-04-22 22:11:28"
+}
+
```
zuphilip: Add an AGPL license block here for the translator code.

Library Genesis.js
```diff
@@ -0,0 +1,268 @@
...
+	}
+}
+
+function doWeb(doc, url) {
+	var md5Hash = MD5_REGEX.exec(url)[1];
+
+	// To save some work, we use the provided bibtex file to retrieve the
+	// metadata and use Zotero's built-in bibtex importer
+	var bibtexUrl = "/book/bibtex.php?md5=" + md5Hash;
+	ZU.processDocuments(bibtexUrl, function(bibtexDoc) {
```
adam3smith: indenting is still not quite right here -- everything that's running inside this should be indented.

Library Genesis.js
```diff
@@ -0,0 +1,268 @@
...
+		if (item.ISBN) {
+			[' ', ','].forEach(function(splitChar) {
+			if (item.ISBN.indexOf(splitChar) != -1) {
+				item.ISBN = item.ISBN.split(splitChar)[0];
+			}
+			});
+		}
+
+		// Add the full text attachment
+		var extension = ZU.xpathText(
```
adam3smith: I'm sorry, but let's comment the entire attachment part out until we have a proper solution.

PR #1059, opened 2016-05-08
Create National Library of Norway.js

danmichaelo:
I’m relying mostly on the RIS translator.

To detect the document type without having to fetch the RIS file, I'm using an input field part of a feedback form (`//input[@id="mediaType"]`). This works, but seems a bit fragile, so I will try to get in contact with the library about a better solution. The best would be to have something like RDFa markup I guess.

### DIFF
--- a/National Library of Norway.js
+++ b/National Library of Norway.js
@@ -1,0 +1,205 @@
+{
+	"translatorID": "7bd82b70-1ed7-4aaf-bea2-bcbe6429a3ee",
+	"label": "National Library of Norway",
+	"creator": "Dan Michael O. Heggø",
+	"target": "^https?://(www\\.)?nb\\.no/",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2016-05-20 04:28:12"
+}
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	National Library of Norway Translator
+	Copyright © 2016 Dan Michael O. Heggø
+
+	This file is part of Zotero.
+
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+function detectWeb(doc, url) {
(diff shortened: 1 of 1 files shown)


National Library of Norway.js
```diff
@@ -0,0 +1,205 @@
...
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+function detectWeb(doc, url) {
+	if (url.match("/nbsok/search?")) {
```
adam3smith: use `indexOf("string") != -1` to test for the presence of literal strings (much more efficient)

National Library of Norway.js
```diff
@@ -0,0 +1,205 @@
...
+		return "multiple";
+	} else if (url.match("/nbsok/nb/")) {
```
adam3smith: same here

National Library of Norway.js
```diff
@@ -0,0 +1,205 @@
...
+				.replace(/ +/g, ' ')            // multiple spaces
+				;
+		});
+
+		if (item.numPages) {
+			item.numPages = extractNumPages(item.numPages);
+		}
+
+		// Add permalink
+		fields = ZU.getItemArray(doc, doc, 'http://urn.nb.no/URN:NBN');
```
adam3smith: `var fields`

adam3smith:
some nits inline, looks nice otherwise. Thanks!

National Library of Norway.js
```diff
@@ -0,0 +1,205 @@
...
+		Object.keys(fields).forEach(function(url) {
+			item.url = url;
+		});
```
zuphilip: This looks fishy: In each loop step you are replacing the previous `item.url` with a new `url`? Also I am not sure you really need to call the function [ZU.getItemArray](https://github.com/zotero/zotero/blob/master/chrome/content/zotero/xpcom/utilities_translate.js#L121) above. What would you like to do here?

danmichaelo: It was just a quick and dirty way to get the link that starts with `http://urn.nb.no/URN:NBN`. This rewrite is probably more efficient and transparent? (I know the link is somewhere inside `#preview_metadata`, so I can narrow in on that.)

```javascript
                var container = doc.getElementById('preview_metadata');
		if (container) {
			var links = container.getElementsByTagName('a');
			for(var i=0; i<links.length; i++) {
				if (links[i].href && links[i].href.indexOf('urn.nb.no/URN:NBN') != -1) {
					item.url = links[i].href;
					break;
				}
			}
		}
````

File: National Library of Norway.js [hunk 5]
```diff
@@ -0,0 +1,211 @@
...
+
+		if (item.numPages) {
+			item.numPages = extractNumPages(item.numPages);
+		}
+
+		// Add permalink
+		var container = doc.getElementById('preview_metadata');
+		if (container) {
+			var links = container.getElementsByTagName('a');
```
adam3smith: just use `ZU.xpathText(doc, './a[contains(@href, "urn.nb.no/URN:NBN")])`;
Should be basically as efficient and save you 5 lines.

danmichaelo: Perfect, thanks! Just had to do './/', since it's in a subdiv

adam3smith:
thanks!

PR #1064, opened 2016-05-18
two fixes to ePrint IACR

real-or-random:


ePrint IACR.js
```diff
@@ -56,8 +59,7 @@ function scrape(doc, url) {
 	
 	newItem.date = year;
 	newItem.reportNumber = no;
-	//we want to use this later & make sure we don't make http--> https requests or vice versa. 
-	newItem.url = url.match(/^https?:\/\/[^\/]+/)[0] + "/" + year + "/" + no;
+	newItem.url = "https://eprint.iacr.org/" + year + "/" + no;
```
adam3smith: I still get regular http, e.g. at http://eprint.iacr.org/2005/033 -- please revert this.

ePrint IACR.js
```diff
@@ -36,8 +36,10 @@ function scrape(doc, url) {
 		var no   = reportNo[2];
 	}
 	var title = doc.evaluate(titleXPath, doc, null, XPathResult.ANY_TYPE, null).iterateNext().textContent;
+	title = compressSpaces(title);
```
adam3smith: We have `ZU.trimInternal(string)` for this. Don't write a custom function

real-or-random:
Should be better now.

Sorry, I had forgot that I use the HTTPS Everywhere plugin, which redirects automatically.

adam3smith:
thanks!

PR #1066, opened 2016-05-26
Enable 5 translators for translation-server and add url to Google Books

mvolz:
Enable JSTOR, Primo, ScienceDirect, Wiley Online,
and Google Books translators for translation-server. 

Add regex for 12 character google books ID,
and add url to citation based on that ID.

Change-Id: I706250304fb8df4e53ee45c3e32ced7970e7f19e

### DIFF
--- a/Google Books.js
+++ b/Google Books.js
@@ -8,7 +8,7 @@
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
-	"browserSupport": "gcsb",
+	"browserSupport": "gcsbv",
 	"lastUpdated": "2015-09-24 15:24:15"
 }
 
@@ -116,17 +116,22 @@ function parseXML(text) {
 	}
 	
 	var ISBN;
-	const ISBN10Re = /(ISBN:)(\w{10})$/;
-	const ISBN13Re = /(ISBN:)(\w{13})$/;
+	const ISBN10Re = /(?:ISBN:)(\w{10})$/;
+	const ISBN13Re = /(?:ISBN:)(\w{13})$/;
+	const booksIDRe = /^(\w{12})$/;
 	var identifiers = ZU.xpath(xml, "dc:identifier", ns);
 	for (var i in identifiers) {
 		var ISBN10Match = ISBN10Re.exec(identifiers[i].textContent);
 		var ISBN13Match = ISBN13Re.exec(identifiers[i].textContent);
+		var booksIDMatch = booksIDRe.exec(identifiers[i].textContent);
 		if (ISBN10Match != null) {
-			ISBN = ISBN10Match[2];
+			ISBN = ISBN10Match[1];
 		}
 		if (ISBN13Match != null) {
-			ISBN = ISBN13Match[2];
+			ISBN = ISBN13Match[1];
+		}
+		if (booksIDMatch != null) {
+			newItem.url = 'https://books.google.com/books?id=' + booksIDMatch[1];
 		}
 	}
 	newItem.ISBN = ISBN;
--- a/JSTOR.js
+++ b/JSTOR.js
@@ -8,7 +8,7 @@
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
-	"browserSupport": "gcsib",
+	"browserSupport": "gcsibv",
 	"lastUpdated": "2016-05-13 22:51:52"
 }
 
--- a/Primo.js
+++ b/Primo.js
@@ -8,7 +8,7 @@
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
-	"browserSupport": "gcsb",
+	"browserSupport": "gcsbv",
 	"lastUpdated": "2016-04-22 09:32:07"
 }
 
--- a/ScienceDirect.js
+++ b/ScienceDirect.js
@@ -8,7 +8,7 @@
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
-	"browserSupport": "gcsib",
+	"browserSupport": "gcsibv",
 	"lastUpdated": "2015-12-08 05:47:08"
 }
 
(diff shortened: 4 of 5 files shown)


zuphilip:
That looks good. I am wondering whether we can change the `v` in larger scale? The search translator do not work at translation server, but are there any other restrictions?

(I added this PR to the WikiCite pad for documentation.)

mvolz:
Yes, I think there are probably lots we can add the v flag to. these are just the ones that we've already been using in production. When the automated tests were running they'd actually show which ones succeeded, partially failed, or failed on translation-server. Unfortunately have to check them all manually now! 

Some don't make sense to use because they require credentials, i.e. the Zotero one.

dstillman:
> Unfortunately have to check them all manually now! 

I'm not sure what the state of the automated translator tests is (@simonster would know), but there's a test mode you can run locally:

`./build/run-translation-server.sh -test /full/path/to/output/file.html`

I'm going to try tonight to update the translation server to work with the latest version of Zotero — it's currently running a version from 2013, and there are probably [various problems](https://groups.google.com/d/topic/zotero-dev/fJdp0NK_2ec/discussion) because of that.

Google Books.js
```diff
@@ -116,17 +116,22 @@ function parseXML(text) {
...
 		if (ISBN10Match != null) {
-			ISBN = ISBN10Match[2];
+			ISBN = ISBN10Match[1];
 		}
 		if (ISBN13Match != null) {
-			ISBN = ISBN13Match[2];
+			ISBN = ISBN13Match[1];
+		}
+		if (booksIDMatch != null) {
+			newItem.url = 'https://books.google.com/books?id=' + booksIDMatch[1];
```
zuphilip: If we want this change also rolled out in Zotero, then one needs to change the `lastUpdated` field. Moreover, one needs to update the test cases.

adam3smith: I don't think we'd want this. We only include the url where we have a full text at that URL; that's not the case for all books on google books.

zuphilip: Hm.. Okay, I guess you are right. This is similar to not to save the link to amazon in the url field, just because the metadata where taken from there.

On the other hand there is a Wikidata property [P675](https://www.wikidata.org/wiki/Property:P675) for the Google-Books-ID, which is in particularly used in giving references to statements.

Should we add the Google-Books-ID in the `extra` field, i.e.
```js
newItem.extra = "Google-Books-ID: " + booksIDMatch[1];
```
? Then @mvolz can take it from there for the Citoid service?

adam3smith: Sounds good to me, yes

owcz: @adam3smith Is there a reason why you only use the URLs that host the full text? I would think that having a link to the desired GB item would be preferable than no URL in the field at all (especially since preview rights do change)

adam3smith: Yes -- because URL is a field to be used in citations, and citations should point to the actual resource something is cited from. A URL would be misleading if you actually read the book in print but got the data from google books. For Zotero, there is an option to attach a link, which we do for google books w/o fulltext (and for other similar catalog-type resources like Amazon).

mvolz:
The reason we change the url in Google books is because Google puts tracking information in its urls, and we didn't want that information being stored in Wikipedia. So this was a hack for that. Putting it in the memo field is fine but we probably wouldn't use that actually. 

I'm assuming the option to attach the link in Zotero doesn't actually go into any field in the citation?

adam3smith:
> I'm assuming the option to attach the link in Zotero doesn't actually go into any field in the citation? 

I'm afraid not (though the translator server can attach files, so this is going somewhere...)

zuphilip:
I think that attachments are currently not present in translation server, cf. https://github.com/infolis/translation-server/commit/4e600bdd98e83db051f461c0bec9d654b5e8ea43 . However, it is also worth to think about a more stable (official) solution.

zuphilip:
@mvolz Why did you close this PR?

Google Books.js
owcz: Correct me if I'm wrong, but there are two use cases here. (1) Google Books is used to lookup ISBNs for metadata, no? And in those cases you don't want to pull the URL as there isn't necessarily any text cited at that time. (2) Google Books often _does_ have full-page view (if not full text) for recent and public domain books, and in those cases it would absolutely be useful to have the URL linked in the citation.

Case #2 is much more frequent for me, but then again I'm less privy to the backend on ISBN lookup. Wouldn't it make sense to handle the two cases differently, though, based on whether the user is citing from the page view (more than the snippet view)? It would even be nice to save the full-page view image to Z for citation...

zuphilip: The standard use case for the translators is to save the bibliographic metadata to make citations and list of references in publications. The citation styles are then responsible for the specific formatting in a published manuscript. It is common to show the url for webpages, blog posts and to show the doi for journal articles, conference proceeding in the list of references. However, it would be really strange if all references to books would have Google or Amazon urls. Have you ever seen a bibliography with a Google or Amazon url for a recent book?

However, I don't understand why you don't want to look at the `extra` field for the Google url. AFAIK Citoid is already parsing the `extra` field for PMID. Moreover, [we began at the WikiCite](https://meta.wikimedia.org/wiki/WikiCite_2016/Report/Group_8#Subtask.232_Improve_ID_import_into_Citoid) to put other IDs in the same way in extra fields such that Citoid can grab it from there, e.g. for the OCLC number https://github.com/zotero/translators/commit/046a7a584ca901744e74f586e3123b5eb9d7facc, https://github.com/zotero/translators/commit/ff293f90aa30334d148cbf7ece96c164b4bc98db.

owcz: I don't mean just for Wikipedia in this case, and perhaps it's more for "book section" than "book"-type citations, but I'll often have a passage of one to ten pages that I'd cite on its own from a chapter and if I recall correctly, the Zotero ingest doesn't take the digital webpage into account. I would think that the point of having full-page text readable online is to be able to share links to passages like that and thus to have them in the citations. I agree that I rarely see books.google.com URLs in bibliographies for books, and I can't say I recall seeing it often for chapters or passages either. For me this is more of keeping my research in the same place. Apart from that, no qualms in particular about using the `extra` field as you mentioned for citation generation.

PR #1079, opened 2016-06-14
Create Eurogamer and USgamer translators

owcz:
eurogamer.net, usgamer.net. Too dissimilar to combine into one translator.


adam3smith:
Looks good, thanks!

PR #1080, opened 2016-06-15
IGN translator rewritten in Framework

owcz:
Rewritten in Framework, replacing antiquated, outdated version

### DIFF
--- a/IGN.js
+++ b/IGN.js
@@ -1,101 +1,134 @@
 {
 	"translatorID": "d210c5a1-73e1-41ad-a3c9-331d5a3ead48",
 	"label": "IGN",
-	"creator": "odie5533",
-	"target": "^https?://[^/]+\\.ign\\.com/",
-	"minVersion": "1.0",
+	"creator": "czar",
+	"target": "^https?://(www\\.)?ign\\.com/articles/",
+	"minVersion": "3.0",
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2015-06-02 19:11:09"
+	"lastUpdated": "2016-06-15 16:37:47"
 }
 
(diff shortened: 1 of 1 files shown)


IGN.js
```diff
@@ -1,103 +1,105 @@
 {
-	"translatorID": "d210c5a1-73e1-41ad-a3c9-331d5a3ead48",
+	"translatorID": "cecafe6f-7fc9-4bdb-ae70-bd0a75a8ca52",
```
zuphilip: You should never change an existing `translatorID` for a translator, if you are updating it.

zuphilip:
I haven't looked at the details, but usually I think that a normal translator would be preferred compared to one that is using the framework.

IGN.js
owcz: Even if it completely replaces the former translator, without re-using any of it?

owcz:
Why? This performs the same function but isn't broken and is easier for others to maintain

IGN.js
adam3smith: yes -- otherwise you won't replace the old translator.

avram:
Probably the wrong forum for this, but I see little reason to discourage Framework translators that work. Yes, they're more limited and can't be customized as much. But they can also lower the bar for contribution and help increase the total coverage that our translators have.

adam3smith:
agree with @avram here: As long as we don't need any clean-up on a translator, I don't see why we shouldn't do it as FW.

IGN.js
```diff
@@ -1,103 +1,105 @@
...
-	}
-	
-	// attach html
-	newItem.attachments.push({title:"IGN Article Snapshot", document:doc});
-	
-	newItem.complete();
-}
+FW.Scraper({
+itemType         : 'blogPost',
+detect           : FW.Url().match(/articles\/\d{4}/),
```
adam3smith: given that you only detect on ign.com/articles/, could you add /articles/ to the target (leave it in detect to be safe)

adam3smith:
Thanks, looks good. FWIW, I think it makes it easier to review if you don't squash your commits, especially now that we can just squash on merging.

owcz:
Didn't know you could do that—noted & thanks!

PR #1082, opened 2016-06-17
changes based on pull request 915

heinonline:
changes based on pull request https://github.com/zotero/translators/pull/915 forked from https://github.com/Juris-M/translators/blob/0e34b56e4f2efe9b5123addd4010f0e989ee36de/HeinOnline.js

### DIFF
--- a/HeinOnline.js
+++ b/HeinOnline.js
@@ -77,12 +77,6 @@ function extractQueryValues(url) {
 	return ret;
 }
 
-// Remangle HeinOnline RIS to form expected by translator
-function fixRIS (ris) {
-	ris=ris.replace(/^([A-Z0-9][A-Z0-9]) /mg, "$1  - ").replace(/^VO  /mg, "VL  ").replace(/^YR  /mg, "DA  ").replace(/^OP  /mg, "EP  ");
-	return ris;
-}
-
 // Not all pages have a downloadable PDF
 function translateRIS(ris, pdfURL) {
 	var trans = Zotero.loadTranslator('import');
@@ -120,7 +114,6 @@ function scrapePage(doc, url) {
 			+ "&base=js";
 		var pdfPageURLs = doc.getElementsByClassName("updatediv");
 		ZU.doGet(risURL, function(ris) {
-			ris = fixRIS(ris);
 			if (pdfPageURLs) {
 				Array.prototype.filter.call(pdfPageURLs, function(pdfPageURL){
 					if (pdfPageURL){
--- a/HeinOnline.js
+++ b/HeinOnline.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsbv",
-	"lastUpdated": "2016-06-26 23:08:21"
+	"lastUpdated": "2016-08-10 09:29:21"
 }
 
 /*
--- a/HeinOnline.js
+++ b/HeinOnline.js
@@ -1,0 +1,224 @@
+{
+	"translatorID": "3dcbb947-f7e3-4bbd-a4e5-717f3701d624",
+	"label": "HeinOnline",
+	"creator": "Frank Bennett",
+	"target": "^https?://(www\\.)?heinonline\\.org/HOL/(?:LuceneSearch|Page|IFLPMetaData)\\?",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsbv",
+	"lastUpdated": "2016-02-22 23:08:21"
+}
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+	Copyright © 2015-2016 Frank Bennett
+	This file is part of Zotero.
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+	GNU Affero General Public License for more details.
+	You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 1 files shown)


HeinOnline.js
```diff
@@ -0,0 +1,224 @@
...
+function getXPathStr(attr, elem, path) {
+	var res = ZU.xpath(elem, path);
+	res = res.length ? res[0][attr] : '';
+	return res ? res : '';
+}
+
+// Extract query values to keys on an object
+function extractQueryValues(url) {
+	var ret = {};
+	ret.base = url.replace(/(.*?)[a-zA-Z]+\?.*/, "$1");
```
adam3smith: since URLs never has two question marks, how is this ever different from 
`url.replace(/[a-zA-Z]+\?.*/, "");` ?

HeinOnline.js
```diff
@@ -0,0 +1,224 @@
...
+	for (var i=0,ilen=query.length;i<ilen;i++) {
+		var pair = query[i].split("=");
+		ret[pair[0]] = pair[1];
+	}
+	return ret;
+}
+
+// Remangle HeinOnline RIS to form expected by translator
+function fixRIS (ris) {
+	var lines = ris.split("\n");
```
adam3smith: reposting aurimas' comment from the old review:
>You don't need to split up everything into single line. You can run the replacements on the whole text with mg flags (m makes ^ match start of line).

HeinOnline.js
```diff
@@ -0,0 +1,224 @@
...
+		var pdfPageURL = doc.getElementsByClassName("updatediv");
+		ZU.doGet(risURL, function(ris) {
+			ris = fixRIS(ris);
+			if (pdfPageURL) {
+				ZU.doGet(pdfPageURL, function(pdfPage) {
+					// Call to pdfPageURL prepares PDF for download via META refresh URL
+					var pdfURL = null;
+					var m = pdfPage.match(/<META.*URL=([^"]+)/);
+					if (m) {
+						var pdfURL = m[1];
```
adam3smith: from old review: no need for `var` here. The variable is declared in line 132

HeinOnline.js
```diff
@@ -0,0 +1,224 @@
...
+	*********
+*/
+
+function detectWeb (doc, url) {
+	if (url.indexOf("/LuceneSearch?") > -1) {
+		if (getSearchResults(doc)) {
+			return "multiple";
+		}
+	} else {
+		return "journalArticle";
```
adam3smith: from older review:
>Make sure either RIS out COinS is available

in other words, make sure that we only detect something when we're not then going to throw an error on import.

HeinOnline.js
```diff
@@ -0,0 +1,224 @@
...
+
+			ZU.processDocuments(urls, scrapePage);
+		});
+	} else {
+		scrapePage(doc, url);
+	}
+}
+
+/** BEGIN TEST CASES **/
+var testCases = [
```
adam3smith: as per aurimas, more tests cases would be great (if you're not using scaffold alredy, you'll definitely want to for that purpose, since it creates test JSON automatically.)]
https://www.zotero.org/support/dev/translators/scaffold

adam3smith:
Great, almost there, just a couple of small things left. You can update your pull request by clicking the pencil icon here: https://github.com/zotero/translators/pull/1082/files and editing/pasting the style.

Any questions or problems, please let me know.

HeinOnline.js
heinonline: The first question mark is not a literal match, instead it is making the .* non-greedy.

adam3smith: I get that, but when would that ever be necessary? Since there's only one ?, the non-greedy match has no effect. Maybe I'm just missing an example, but I'm thinking of a URL like
http://heinonline.org/HOL/Page?<somequerystring>  url.replace(/[a-zA-Z]+\?.*/, ""); chops of everything from Page? and returns the rest and your version keeps everything until HOL/ and returns it. In other words, the two are functionally equivalent, the 2nd one is less efficientl.

heinonline: I see. Yes, you are correct.

HeinOnline.js
heinonline: I don't know how to do what you are asking, I'll look into it.

adam3smith:
```
var COinS = getXPathStr("title", doc, '//span[contains(@class, "Z3988")]');
var RIS = getXPathStr("href", doc, '//form[@id="pagepicker"]//a[contains(@href, "PrintRequest")][1]');
...
else if (COinS || RIS) {
    return "journalArticle"
}
```

or something along those line.

HeinOnline.js
```diff
@@ -2,14 +2,14 @@
 	"translatorID": "3dcbb947-f7e3-4bbd-a4e5-717f3701d624",
 	"label": "HeinOnline",
 	"creator": "Frank Bennett",
-	"target": "^https?://(www\\.)?heinonline\\.org/HOL/(?:LuceneSearch|Page|IFLPMetaData)\\?",
+	"target": "^https?://(www\.)?heinonline\.org/HOL/(?:LuceneSearch|Page|IFLPMetaData)\\?",
```
zuphilip: Undo this change.

The point in a regexp refers to any character and therefore you have to escape it here `\.` in scaffold which will result in double escaped `\\.` here.

heinonline:
Is there something else I need to do before this can be moved forward? Or is this just in a queue to be reviewed when time permits? I don't mean to be pushy I just want to make sure there is nothing I need to do. Thanks.

zuphilip:
I think that there is just a queue for PR in general here. Maybe @adam3smith finds some time to check the update.

BTW there are no notifications if you update a PR and therefore it is a good idea to write a short message after you updated a PR.

heinonline:
Thanks @zuphilip !

Pull request updated to address remaining issues please review.

heinonline:
Just checking in to make sure there is nothing else to be done here. We really appreciate all your help with this.

zuphilip:
I think that currently there is also vacation time:  :high_brightness: :palm_tree: :swimmer:

heinonline:
We recently fixed the RIS file generator and so it no longer needs to be fixed here.

adam3smith:
Great, thanks and sorry for the delay. FYI @fbennett your earlier work now fully implemented.

fbennett:
Super, thanks for the follow-up.

PR #1085, opened 2016-06-21
Revert "Change version to versionNumber for computerProgram"

adam3smith:
Reverts zotero/translators#1084

### DIFF
--- a/BibLaTeX.js
+++ b/BibLaTeX.js
@@ -15,7 +15,7 @@
 		"exportFileData": false,
 		"useJournalAbbreviation": false
 	},
-	"lastUpdated": "2015-09-10 08:45:20"
+	"lastUpdated": "2016-06-21 08:45:20"
 }
 
 
--- a/BibLaTeX.js
+++ b/BibLaTeX.js
@@ -4,7 +4,7 @@
 	"label": "BibLaTeX",
 	"creator": "Simon Kornblith, Richard Karnesky and Anders Johansson",
 	"target": "bib",
-	"minVersion": "4.0.27",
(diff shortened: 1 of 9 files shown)


zuphilip:
Hm.. I see that there is a problem in Ubuntu but in Windows it seems to work. I tried the CSV translator with `"minVersion": "4.0.27",` but it works with `"minVersion": "4.0.26",` and every smaller number. The error message seems quite clear:
```
(2)(+0000001): Translate: Translation using CSV failed: 
string => TypeError: item.uniqueFields is undefined
stack => getValue@CSV:211:1
doExport@CSV:110:7
```
@dstillman Any ideas what is going on? Was there any change for `uniqueFields` in this version?

dstillman:
I don't really know anything about the `uniqueFields` stuff, sorry.

aurimasv:
There are no `uniqueFields` in the API version of Zotero.Item JSON. https://github.com/zotero/zotero/blob/master/test/tests/data/translatorExport.js

zuphilip:
Is `uniqueFields` old code and only supported for legacy translators (pre 4.0.27) via [addCompatibilityMappings](https://github.com/zotero/zotero/blob/4.0/chrome/content/zotero/xpcom/utilities_internal.js#L421)?

Please let me know whether it is true that: By increasing the `minVersion` to `4.0.27` the function `addCompatibilityMappings` will not be called. This could explain the problems with the translators commit...

aurimasv:
>Please let me know whether it is true that: By increasing the `minVersion` to 4.0.27 the function `addCompatibilityMappings` will not be called

That is correct.

PR #1086, opened 2016-06-22
Some tweaks to how DC item types are identified

adam3smith:
- dataset in eprints and DC should become a journal article
- article type in DC should be a journal article, both given common
- usage and for uniformity with eprint
@zuphilip -- pls. review when you get a chance.

### DIFF
--- a/RDF.js
+++ b/RDF.js
@@ -394,7 +394,7 @@ function detectType(newItem, node, ret) {
 			t.dc = type;
 		} else {
 			//on eprints the type fields are often in the form "Journal Article", "Conference Item" etc.
-			type = type.toLowerCase().replace(/\s/g, "")
+			type = type.toLowerCase().replace(/\s/g, "");
 			switch (type) {
 				//eprints
 				//from http://www.ukoln.ac.uk/repositories/digirep/index/Eprints_Type_Vocabulary_Encoding_Scheme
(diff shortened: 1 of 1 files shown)


zuphilip:
The change for the `eprints` vocabulary looks good.

However, I cannot say much about the `dc` vocabulary. As [it will considered only](https://github.com/adam3smith/translators/blob/master/RDF.js#L684-686) if there is no more specified vocabulary,
```
   var itemType = t.zotero || t.bib || t.prism ||t.eprints|| t.og || t.dc || 
         exports.defaultUnknownType || t.zoteroGuess || t.bibGuess || 
         t.prismGuess || t.ogGuess || t.dcGuess 
```
examples from EPrints or Prism are only partially useful. Did [you switch](https://github.com/zotero/translators/pull/1086/files#diff-139a6b4a8ace7b438018890bcdba1de4L448) from `t.dcGuess` to `t.dc` on purpose? I guess it will not change much in the end, but you will maybe know better...

zuphilip:
(Nits from previous coding: On line [397](https://github.com/zotero/translators/pull/1086/files#diff-139a6b4a8ace7b438018890bcdba1de4L397) and [686](https://github.com/zotero/translators/pull/1086/files#diff-139a6b4a8ace7b438018890bcdba1de4L686) a semicolon is missing at the end.)

adam3smith:
Yes, the move to `t.dc` from `t.dcGuess` is on purpose. Item types in `dcGuess` are overwritten as webpage when this is called from the EM translator.
Added the semicolons.

PR #1095, opened 2016-07-22
washingtonpost.js support for /archive/ URLs

owcz:
Some of the old tests (#2,3) are failing with translation issues, so you'll probably want to look at those, Seb (first one updated just fine and I added a fourth). All of the other formatting updates are Scaffold—I only added the new detect for /archive/, added a test, and tried to update the other tests. Also I thought you didn't like to detect on Xpath?


adam3smith:
no, I have no objections to testing for xpaths on detect. URL match is more efficient, but the difference is tiny, so only matters if there is no possible trade-off on precision. 
Will match this shortly; I'd like to fix some of the lost formatting for the license block (odd that Scaffold would break that).

adam3smith:
superseded by #1102

PR #1117, opened 2016-08-23
Rewrite ProjectMuse using EM

adam3smith:
No more RIS available. Books metadata is limited -- no place to be found anywhere.
For PDF attachments, I do overwrite all attachments on purpose; I also leave the PDF URL in on purpose.
@zuphilip whenever you can take a look?

### DIFF
--- a/Project MUSE.js
+++ b/Project MUSE.js
@@ -1,182 +1,116 @@
 {
 	"translatorID": "c54d1932-73ce-dfd4-a943-109380e06574",
 	"label": "Project MUSE",
-	"creator": "Simon Kornblith, Avram Lyon, Sean Takats",
-	"target": "^https?://[^/]*muse\\.jhu\\.edu[^/]*/(login\\?.+ur[il]=/)?(?:journals/[^/]+/(summary/)?[^/]+/[^/]+\\.html|search/results|results|books/\\d+)",
-	"minVersion": "2.1.9",
+	"creator": "Sebastian Karcher",
+	"target": "^https?://[^/]*muse\\.jhu\\.edu/(book/|article/|issue/|results\\?)",
+	"minVersion": "3.0",
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2015-04-12 16:54:33"
+	"lastUpdated": "2016-08-23 03:49:35"
 }
+
+/*
+    ***** BEGIN LICENSE BLOCK *****
+
+    Copyright © 2016 Sebastian Karcher
+
+    This file is part of Zotero.
+
+    Zotero is free software: you can redistribute it and/or modify
+    it under the terms of the GNU Affero General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    Zotero is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+    GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
+    along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+    ***** END LICENSE BLOCK *****
+*/
 
 function detectWeb(doc, url) {
-	var searchRe = new RegExp("(^https?://[^/]+/search/results|/search/save|/toc/|/results(#.+)?$)");
-	if(searchRe.test(url)) {
-	//some old TOCs just display links to pdfs - don't detect those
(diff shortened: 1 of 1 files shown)


Project MUSE.js
```diff
@@ -1,182 +1,116 @@
...
+    // Embedded Metadata
+    translator.setTranslator('951c027d-74ac-47d4-a107-9c3069ab7b48');
+    translator.setDocument(doc);
+    translator.setHandler('itemDone', function (obj, item) {
+        if (abstract) {
+        	item.abstractNote = abstract.replace(/^\s*Abstract/, "").replace(/,\s*$/, "").trim();
+        }
+        if (url.indexOf("/article/") != -1) {
+       		var pdfurl = url.replace(/(\/article\/\d+).*/, "$1") + "/pdf";
+       		Z.debug(pdfurl)
```
zuphilip: Delete or comment out

Project MUSE.js
```diff
@@ -1,182 +1,116 @@
...
-		var m = hostRe.exec(url);
-		var host = m[1];
+function scrape(doc, url) {
+	var abstract = ZU.xpathText(doc, '//div[@class="abstract"]');
+	if (!abstract) abstract = ZU.xpathText(doc, '//div[@class="description"][1]/text()');
+	var translator = Zotero.loadTranslator('web');
+    // Embedded Metadata
+    translator.setTranslator('951c027d-74ac-47d4-a107-9c3069ab7b48');
+    translator.setDocument(doc);
+    translator.setHandler('itemDone', function (obj, item) {
```
zuphilip: The indent seems on these 4 lines wrong. Maybe check also above that you are using tabs and not spaces for the indent (when I copy and paste the code block this goes usually wrong).

Project MUSE.js
```diff
@@ -1,182 +1,116 @@
...
+        	item.abstractNote = abstract.replace(/^\s*Abstract/, "").replace(/,\s*$/, "").trim();
+        }
+        if (url.indexOf("/article/") != -1) {
+       		var pdfurl = url.replace(/(\/article\/\d+).*/, "$1") + "/pdf";
+       		Z.debug(pdfurl)
+       		item.attachments = [{
+						"url": pdfurl,
+						"title": "Full Text PDF",
+						"mimeType": "application/pdf"
+					}]
```
zuphilip: Semicolon is missing here.

BTW I prefer to use here also `item.attachments.push( { ... } );`, which will not destroy any other attachments possibly added at some time.

Project MUSE.js
```diff
@@ -1,182 +1,116 @@
...
-			item.complete();
-			});
-			translator.translate();
-			})
-	}
-	
-	else{
+    translator.getTranslatorObject(function(trans) {
+        trans.doWeb(doc, url);
+    });
```
zuphilip: If you don't need any special handling here, then you can omit this function completely.

Project MUSE.js
```diff
@@ -190,62 +124,53 @@ var testCases = [
...
+				"abstractNote": ", , Jeffrey S. Theis focuses on pastoral literature in early modern England as an emerging form of nature writing. In particular, Theis analyzes what happens when pastoral writing is set in forests — what he terms “sylvan pastoral.”",
```
zuphilip: Why does this start with two commas? Are there some empty nodes?

Project MUSE.js
```diff
@@ -190,62 +124,53 @@ var testCases = [
...
-	},
-	{
-		"type": "web",
-		"url": "http://muse.jhu.edu/books/9780820705057",
+		"url": "http://muse.jhu.edu/book/785",
 		"items": [
 			{
 				"itemType": "book",
-				"title": "Writing the Forest in Early Modern England: A Sylvan Pastoral Nation",
+				"title": "Writing the Forest in Early Modern England",
```
zuphilip: It is a pity, that we lose the subtitle here...

zuphilip:
This looks great. I have just some small comments, see above.

Project MUSE.js
zuphilip: Or do you do this because of https://github.com/zotero/zotero/issues/859 ?

adam3smith: yes, but I didn't text if it's necessary.

zuphilip: `s/text/test/` ? :smirk:

adam3smith:
if you want to take another look @zuphilip ?

zuphilip:
This looks fine for me!

(BTW for calling web translator in translation server the function `setDocument` is currently broken, but `getTranslatorObject` seems a way to circumvent this...)

adam3smith:
> (BTW for calling web translator in translation server the function setDocument is currently broken, but getTranslatorObject seems a way to circumvent this...)

Do I understand that correctly that if I replace `translator.translate()` with 

```
translator.getTranslatorObject(function(trans) {
        trans.doWeb(doc, url);
   });
```

That this will work on the server, but otherwise it won't? In that case, I should probably switch this back, no?

zuphilip:
Yes, but you need additionally also to delete the `translator.setDocument(doc);` line (that is the critical function), cf. https://github.com/zotero/translation-server/issues/26#issuecomment-241850317

PR #1123, opened 2016-08-27
Primo Improvements

adam3smith:
Especially to publisher/place parsing, but also some other things.
Closes https://github.com/zotero/translators/pull/850/files
Please review @zuphilip

### DIFF
--- a/Primo.js
+++ b/Primo.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsbv",
-	"lastUpdated": "2016-04-22 09:32:07"
+	"lastUpdated": "2016-08-27 04:43:48"
 }
 
 /*
@@ -213,7 +213,13 @@ function importPNX(text, url) {
 	
 	var itemType = ZU.xpathText(doc, '//display/type');
 	if(!itemType) {
-		throw new Error('Could not locate item type');
+		if ( ZU.xpathText(doc, '//facets/rsrctype') ) {
+			var itemType = ZU.xpathText(doc, '//facets/rsrctype');
+		} else if ( ZU.xpathText(doc, '//search/rsrctype') ) {
+			var itemType = ZU.xpathText(doc, '//search/rsrctype');
+		} else {
+			throw new Error('Could not locate item type');
+		}
 	}
 	
 	switch(itemType.toLowerCase()) {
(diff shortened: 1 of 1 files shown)


Primo.js
```diff
@@ -525,7 +547,8 @@ var testCases = [
 				"callNumber": "5552.406",
 				"language": "eng",
 				"libraryCatalog": "princeton-primo.com",
-				"publisher": "London",
+				"place": "London",
+				"publisher": "London 1860-1912",
```
adam3smith: Not sure if the publisher here is better or worse, but the new parsing seems overall reasonable and cutting off too much is risky.

Primo.js
```diff
@@ -213,7 +213,13 @@ function importPNX(text, url) {
 	
 	var itemType = ZU.xpathText(doc, '//display/type');
 	if(!itemType) {
-		throw new Error('Could not locate item type');
+		if ( ZU.xpathText(doc, '//facets/rsrctype') ) {
+			var itemType = ZU.xpathText(doc, '//facets/rsrctype');
+		} else if ( ZU.xpathText(doc, '//search/rsrctype') ) {
+			var itemType = ZU.xpathText(doc, '//search/rsrctype');
```
zuphilip: You can write such code more compact by using the OR operator `||`,  i.e.
```js
var itemType = ZU.xpathText(doc, '//display/type') || ZU.xpathText(doc, '//facets/rsrctype') || ZU.xpathText(doc, '//search/rsrctype')
```
(Maybe split into several lines for readability.) Then the variable `itemType` equals to the first expression, it this is not `null`, or otherwise equals to the second expression, if this is not `null`, or otherwise equals to the third expression.

zuphilip: There is also a type field in the `addata` section and especially there is a `addata/ristype`, which should be more standardized. Did you try this out? Actually, I would even try to use it as first option.

Primo.js
```diff
@@ -312,15 +318,24 @@ function importPNX(text, url) {
...
-		item.publisher = pubplace[1].replace(/,\s*c?\d+(\-\d+)?|[\(\)\[\]]|(\.\s*)?/g, "")
-			.replace(/^\s*"|,?"\s*$/g, '');
-	} else if(pubplace) {
-		item.publisher = pubplace[0].replace(/,\s*c?\d+(\-\d+)?|[\(\[].+[\)\]]|(\.\s*)?/g, "")
-			.replace(/^\s*"|,?"?\s*$/g, '');
+	item.place = ZU.xpathText(doc, '//addata/cop');
+	var publisher = ZU.xpathText(doc, '//addata/pub');
+	if(!publisher) publisher = ZU.xpathText(doc, '//display/publisher');
+	if(publisher) {
+    	publisher = publisher.replace(/,\s*c?\d+|[\(\)\[\]]|(\.\s*)?/g, "");
```
zuphilip: Add tab as indent.

Primo.js
```diff
@@ -312,15 +318,24 @@ function importPNX(text, url) {
...
+	    if(pubplace && pubplace[1]) {
+	        var possibleplace = pubplace[0];
+	        if(!item.place ) {
+	            item.publisher = pubplace[1].replace(/^\s*"|,?"\s*$/g, '');
+	            item.place = possibleplace;
+	        }
+	        if(item.place && item.place == possibleplace) {
+	            item.publisher = pubplace[1].replace(/^\s*"|,?"\s*$/g, '');
+	        }
+	    }
```
zuphilip: I forgot why we have here to replace the `"` with some possible white spaces in the beginning and end. Do we have examples for that?

Primo.js
```diff
@@ -361,7 +376,11 @@ function importPNX(text, url) {
 
 	item.edition = ZU.xpathText(doc, '//display/edition');
 	
-	var subjects = ZU.xpath(doc, '//search/subject');
+	var subjects = ""//ZU.xpath(doc, '//search/subject');
```
zuphilip: ?

Primo.js
```diff
@@ -456,7 +475,10 @@ function stripAuthor(str) {
 		// Remove year
 		.replace(/\s*,?\s*\(?\d{4}-?(\d{4})?\)?/g, '')
 		// Remove things like (illustrator). TODO: use this to assign creator type?
-		.replace(/\s*,?\s*[\[\(][^()]*[\]\)]$/, '');
+		.replace(/\s*,?\s*[\[\(][^()]*[\]\)]$/, '')
+		// The full "continuous" name uses no separators, which need be removed
+		// cf. "Luc, Jean André : de (1727-1817)"
+		.replace(/ :/, "");
```
zuphilip: I would suggest a slightly different replacement here for dealing with different possible spacing options:
```js
.replace(/\s+:\s+/, " ");
```

Primo.js
```diff
@@ -761,4 +785,4 @@ var testCases = [
 		]
 	}
 ]
-/** END TEST CASES **/
+/** END TEST CASES **/
```
zuphilip: Add another test case for http://digitale.beic.it/primo_library/libweb/action/display.do?doc=39bei_digitool2018516 ?

zuphilip:
Thank you for continuing this and I agree that especially the publisher and publication place is an improvement. See my comments above.

Primo.js
adam3smith: using `ristype` makes sense, but would require a fair amount of testing across instances, since the current item type detection if heavily tested and I don't trust PNX/Primo to perfectly follow the RIS item types, so I'm leaving this for another time. 
Implemented the simplified syntax, though. Good point there.

Primo.js
adam3smith: I took this from the old PR, so no, no example, but don't see any harm and it looks very purposeful.

Primo.js
adam3smith: oversight left from testing. reverted.

Primo.js
adam3smith: using
`.replace(/\s*:\s+/, " ");` no need to require spaces on both sides.

Primo.js
zuphilip: You are right, they are from code before and they don't look harmful. (But I also cannot think about an example or what could lead to this from any strange cataloguing rule...)

zuphilip:
There was a bump in the forum for this changes: https://forums.zotero.org/discussion/60485?page=1#Item_6

@adam3smith I think this can be squashed+merged now.

Primo.js
```diff
@@ -759,6 +777,34 @@ var testCases = [
...
+						"lastName": "Milano",
+						"creatorType": "author",
+						"fieldMode": 1
+					}
+				],
+				"date": "1500",
+				"language": "ita",
+				"libraryCatalog": "digitale.beic.it",
+				"place": "Milano",
+				"publisher": "Ambrogio : da Caponago",
```
nemobis: This is not good enough, I will have to continue using our custom script. :(

adam3smith: sorry, but the regex you had (and anything I could think to use for the same effect) simply caught too much and is too risky to implement in such a widely used translator.

PR #1127, opened 2016-08-28
Try to remove RIS tags in Wiley abstracts

adam3smith:
@zuphilip -- let me know if you have an idea for something more
elegant. Reported here: https://forums.zotero.org/discussion/61983/error-194384372-nameis-too-long-to-sync/#Item_9

### DIFF
--- a/Wiley Online Library.js
+++ b/Wiley Online Library.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2015-06-04 16:47:36"
+	"lastUpdated": "2016-08-27 22:30:47"
 }
 
 /*
@@ -229,7 +229,14 @@ function scrapeBibTeX(doc, url, pdfUrl) {
 				'&fileFormat=REFERENCE_MANAGER' +
 				'&hasAbstract=CITATION_AND_ABSTRACT';
 	ZU.doPost(postUrl, body, function(text) {
-		//Z.debug(text)
+		//on at least one occasion, Wiley puts RIS within RIS, thus causing havoc.
+		//This is trying to prevent this (see last test).
+		textClean = text.split(/Related Articles\r/);
+		if (textClean.length>1) {
+			Z.debug("cleaning RIS");
+			text = textClean[0] + "ER  -";
+		}
+		//Z.debug(text);
 		var translator = Zotero.loadTranslator('import');
 		//use RIS
 		translator.setTranslator("32d59d2d-b65a-4da4-b0a3-bdd3cfb979e7");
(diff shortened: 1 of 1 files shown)


Wiley Online Library.js
```diff
@@ -1215,6 +1222,66 @@ var testCases = [
 				"seeAlso": []
 			}
 		]
+	},
+	{
+		"type": "web",
+		"url": "http://onlinelibrary.wiley.com.libezproxy2.syr.edu/doi/10.1111/polp.12144/full",
```
zuphilip: I guess we want the url here w/o your proxy, i.e.

http://onlinelibrary.wiley.com/doi/10.1111/polp.12144/full

?

Wiley Online Library.js
```diff
@@ -229,7 +229,14 @@ function scrapeBibTeX(doc, url, pdfUrl) {
 				'&fileFormat=REFERENCE_MANAGER' +
 				'&hasAbstract=CITATION_AND_ABSTRACT';
 	ZU.doPost(postUrl, body, function(text) {
-		//Z.debug(text)
+		//on at least one occasion, Wiley puts RIS within RIS, thus causing havoc.
```
zuphilip: You mean, Wiley puts metadata of related article into the RIS of the article itself? I would add "related articles" somewhere in the comment here...

Wiley Online Library.js
```diff
@@ -229,7 +229,14 @@ function scrapeBibTeX(doc, url, pdfUrl) {
...
+		//This is trying to prevent this (see last test).
```
zuphilip: The "last test" might not be always the same... Maybe, point here directly to the url?

Wiley Online Library.js
```diff
@@ -229,7 +229,14 @@ function scrapeBibTeX(doc, url, pdfUrl) {
...
+		textClean = text.split(/Related Articles\r/);
```
zuphilip: Are you sure the regex works for ```Related Articles: ```? I see there also a colon and space and sometimes `\n`...

zuphilip:
Okay, I see the problem. This really seems to be something Wiley should fix on their end. The solution you are proposing is short but may be fragile. What happens when the abstract with the "Related Articles" things is in the beginning and more important information would follow?

Another option here could be to take the RIS without the abstract and add the abstract separately. This might be more stable, but maybe also more work to been done...

adam3smith:
Yeah, you've convinced me it's too risky. I've reported to Wiley instead.

adam3smith:
Reopening this because this issue has struck again:
https://forums.zotero.org/discussion/62619/error-with-syncing-report-id-1485905285#latest
Wiley never got back to me (except, ironically, to ask if my issue had been resolved to my satisfaction...)
@zuphilip -- let's strategize how to fix this with the least amount of possible side effects. This took forever to troubleshoot and was frustrating for both the user and us.

zuphilip:
Okay, then I would suggest to take the RIS metadata without the abstract and maybe add the abstract in a second step, i.e.
```diff
	var postUrl = '/documentcitationdownloadformsubmit';
 	var body = 'doi=' + encodeURIComponent(doi) + 
 				'&fileFormat=REFERENCE_MANAGER' +
- 				'&hasAbstract=CITATION_AND_ABSTRACT';
+ 				'&hasAbstract=CITATION';
 	ZU.doPost(postUrl, body, function(text) {
```
and then add something like
```js
			//abstract
			var abstract = doc.getElementById("abstract");
			if (!item.abstractNote && abstract) {
				item.abstractNote = abstract.textContent.replace(/^(Abstract|Summary)/i, '');
			}
```
What do you think?

adam3smith:
I liked the idea a lot, but tried it out on the Wiley tests and there are a whole bunch of problems this introduces (missing linebreaks, duplicate abstracts in various languages, etc.). So I don't thinks that'll work, unfortunately...

zuphilip:
Okay, I see. I would suggest to do the following plan:
 * open the bibtex including the abstract where possibly some unwanted related articles with their metadata are mentioned
 * read the BibTeX-file line-by-line
 * look for some starting trigger (e.g. Related Articles) and remember its position
 * go through the other lines and analyze them with respect to the brackets
    * (start with counter=1)
    * if there is an open bracket then increase some counter
    * if there is an closing bracket then decrease the same counter
    * if the counter is zero then remember the ending position
 * delete the content from the starting position to the ending position
 * call BibTeX translator with this optimized content

( I have to rethink this again, but I am distracted now by some conference talks 😃 ;-)

adam3smith:
@zuphilip -- holding off on this. Wiley claims to have fixed this, though last I checked I was still getting the same error.

zuphilip:
Closing this, because it looks that this is fixed or otherwise superseded. CC @adam3smith

PR #1131, opened 2016-09-02
Add Pastebin.com translator

febrezo:
It processes both profiles and pastes and stores the website as an attachment. Test cases included.

PR #1132, opened 2016-09-02
Add Justpaste.it translator

febrezo:
It stores the website as an attachment. Test case included.

### DIFF
--- a/Justpaste.it.js
+++ b/Justpaste.it.js
@@ -2,7 +2,7 @@
 	"translatorID": "386d63c0-e038-4fdb-ae29-786264b9d3f5",
 	"label": "Justpaste.it",
 	"creator": "febrezo",
-	"target": "https?://justpaste\\.it/.+",
+	"target": "^https?://justpaste\\.it/.+",
 	"minVersion": "3.0",
 	"maxVersion": "",
 	"priority": 100,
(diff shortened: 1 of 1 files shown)


adam3smith:
All comments on https://github.com/zotero/translators/pull/1131/files apply here, too

Justpaste.it.js
```diff
@@ -0,0 +1,125 @@
+{
+	"translatorID": "1d1a51d4-60bf-44b8-ae28-8b154d1ed721",
+	"label": "Justpaste.it",
```
zuphilip: This is the same ID as you used in #1131, but the same ID cannot be used for two different translators.

Justpaste.it.js
```diff
@@ -0,0 +1,94 @@
...
+
+	//Setting common data:
+	newItem.websiteTitle = "Justpaste.it";
+	newItem.url = url;
+	var auxTitle = ZU.xpathText(doc, '//title');	
+	newItem.title = auxTitle.substring(0, auxTitle.lastIndexOf(" - justpaste.it"));
+	newItem.websiteType = "Paste Site";
+
+	var auxDate = ZU.xpathText(doc, '//div[@class="col-md-3 col-xs-4 noteBottomFooter"]');
+	if (auxDate.indexOf("ago")>=0) {
```
adam3smith: There are several problems here:
1. The also counts days, so will get the wrong date (e.g. for you test). You could read up on how you can go x days back in a javascript data items -- I'm sure there's something on stackoverflow, I don't know ottomh.
2. months in js date items start at 0, so you need to add 1
3. you get the day using `.getDate()`, not `Day`.

AbeJellinek:
@febrezo, would you be able to rebase on `master`, lint, and make the requested changes?

AbeJellinek:
Closing for lack of activity.

PR #1157, opened 2016-09-28
Add translator for Incunabula Short Title Catalogue (ISTC)

mkittelmann:


ISTC.js
```diff
@@ -0,0 +1,185 @@
+/*
```
adam3smith: I'll need some time to look at this, but I can immediately say that this is missing a translator header. You should be using the file as saved in the Zotero translator folder, not something copy&pasted from Scaffold.

mkittelmann: Thank you for your quick reply. Yes - header added!

ISTC.js
```diff
@@ -0,0 +1,199 @@
+{
+	"translatorID": "29051e64-8eba-4b26-bbf1-0c224bc59497",
+	"label": "ISTC",
+	"creator": "Maike Kittelmann",
+	"target": "^https?://data\\.cerl\\.org/istc/(_search|i[a-z]\\d{8})",
+	"minVersion": "",
```
zuphilip: Set this to the default `3.0` which is AFAIK needed for `Zotero.selectItems`.

ISTC.js
```diff
@@ -0,0 +1,199 @@
...
+
+
+function detectWeb(doc, url) {
+	if ( getSearchResults(doc, true, url) ) {
+		if (url.match('_search?')){
+			return "multiple";
+		}
+		else if (url.match(/i[a-z]\d{8}/)){
+			return 'book';
+		}
```
zuphilip: Is the test in `getSearchResults` needed for a single item (book)? I guess that in your test case the detection will fail... I suggest to try out something like
```js
if (url.indexOf('_search?')>-1 && getSearchResults(doc, true, url)) {
   ...
} else if (url.match(/i[a-z]\d{8}/)) {
   ...
}
```

zuphilip: For simple matches `.indexOf` suffices and is faster than `.match`.

ISTC.js
```diff
@@ -0,0 +1,199 @@
...
+			found = true;
+			items[href] = title;
+		}
+	}
+	else {
+		var title = ZU.trimInternal( ZU.xpath(doc, '//div[contains(@class, "ample-record")]/h3' )[0].textContent );
+		if (!title) return false;
+		if (checkOnly) return true;
+		items[url] = title;
+	}
```
zuphilip: It looks for me that this `else` case is only relevant for the detection. If so then it is IMO better to move this up into `detectWeb`. I guess it is enough to check the existence of the XPath...

zuphilip: Is the check important? I.e. are there cases such that the url check says 'yes' but it is no book described and we need an additional XPath test?

ISTC.js
```diff
@@ -0,0 +1,199 @@
...
+	if ( detectWeb(doc, url) == "multiple" ) {
+		Zotero.selectItems( getSearchResults(doc, false, url), function (items) {
+			if (!items) {
+				return true;
+			}
+			var books = [];
+			for (var i in items) {
+				books.push(i);
+			}
+		ZU.doGet(Object.keys(items), scrape);
```
zuphilip: Missing indent

ISTC.js
```diff
@@ -0,0 +1,199 @@
...
+	} 
+	else if ( detectWeb(doc, url) == "book" ){
+		ZU.doGet( url + '?format=json', scrape );
+	}
+}
+
+
+function scrape( response, obj, url ) {
+		var jsonObject = JSON.parse( response );
+		var item = new Zotero.Item();
```
zuphilip: Try maybe directly
```js
var item = new Zotero.Item("book");
```

ISTC.js
```diff
@@ -0,0 +1,199 @@
...
+		//	url: url.replace('?format=json', ''),
+		//	title: "ISTC",
+		//	mimeType: "text/html",
+		//	snapshot: true
+		// }];		
+		item.complete();
+}
+
+
+function scrape_old( items ) {
```
zuphilip: Is this old code and can be removed?

ISTC.js
```diff
@@ -0,0 +1,199 @@
...
+		item.title = jsonObject.data.title;
+		item.url = url.replace('?format=json', '');
+		item.place = jsonObject.data.imprint[0].imprint_place;
+		item.publisher = jsonObject.data.imprint[0].imprint_name;
+		item.date = jsonObject.data.imprint[0].imprint_date;
+		item.notes = [ jsonObject.data.imprint[0].geo_info.geonames_id || '' ] ;
+		item.callNumber = 'ISTC' + jsonObject._id;
+		item.extra = jsonObject.data.notes[0];
+		item.language = jsonObject.data.language_of_item;
+		item.seeAlso = [ jsonObject.data.references[0].reference_name + ' ' + jsonObject.data.references[0].reference_location_in_source ];
```
zuphilip: I think you cannot save this information into `seeAlso` because that field is for relations to other Zotero items. What is this exactly you want to save here?

ISTC.js
adam3smith: and where you need regular expressions , `.search(/regex/) != -1` is more efficient than `.match().`

ISTC.js
```diff
@@ -0,0 +1,199 @@
...
+
+function scrape( response, obj, url ) {
+		var jsonObject = JSON.parse( response );
+		var item = new Zotero.Item();
+		var name =  jsonObject.data.author;
+		var cleanNameObj = Zotero.Utilities.cleanAuthor(name, "author", true);
+		item.creators[0] = cleanNameObj;
+		item.title = jsonObject.data.title;
+		item.url = url.replace('?format=json', '');
+		item.place = jsonObject.data.imprint[0].imprint_place;
```
adam3smith: I'd like this cleaned up a bit -- should be possible to just save `jsonObject.data` and probably also `jsonObject.data.imprint[0]` as seperate variables and then parse those instead of having these lengthy strings for every item. No functional difference, but much more readable code.

ISTC.js
```diff
@@ -0,0 +1,199 @@
...
+	}
+}
+
+
+function scrape( response, obj, url ) {
+		var jsonObject = JSON.parse( response );
+		var item = new Zotero.Item();
+		var name =  jsonObject.data.author;
+		var cleanNameObj = Zotero.Utilities.cleanAuthor(name, "author", true);
+		item.creators[0] = cleanNameObj;
```
adam3smith: Do we never have more than one author for a book in the catalog?

ISTC.js
```diff
@@ -0,0 +1,199 @@
...
+		item.title = jsonObject.data.title;
+		item.url = url.replace('?format=json', '');
+		item.place = jsonObject.data.imprint[0].imprint_place;
+		item.publisher = jsonObject.data.imprint[0].imprint_name;
+		item.date = jsonObject.data.imprint[0].imprint_date;
+		item.notes = [ jsonObject.data.imprint[0].geo_info.geonames_id || '' ] ;
+		item.callNumber = 'ISTC' + jsonObject._id;
+		item.extra = jsonObject.data.notes[0];
```
adam3smith: push notes to a note, not the extra field, please. I'd suggest concating it with the note above.

ISTC.js
```diff
@@ -0,0 +1,199 @@
...
+		var item = new Zotero.Item();
+		var name =  jsonObject.data.author;
+		var cleanNameObj = Zotero.Utilities.cleanAuthor(name, "author", true);
+		item.creators[0] = cleanNameObj;
+		item.title = jsonObject.data.title;
+		item.url = url.replace('?format=json', '');
+		item.place = jsonObject.data.imprint[0].imprint_place;
+		item.publisher = jsonObject.data.imprint[0].imprint_name;
+		item.date = jsonObject.data.imprint[0].imprint_date;
+		item.notes = [ jsonObject.data.imprint[0].geo_info.geonames_id || '' ] ;
```
adam3smith: if there is something in here, I'd like to see it reflected in a test. I also find pushing to notes (which always exists as an empty array) safer than overwriting the array.

ISTC.js
mkittelmann: Thanks!

ISTC.js
mkittelmann: OK

ISTC.js
mkittelmann: Remove

ISTC.js
mkittelmann: OK, sorry. Attempted to store reference work referrals.

ISTC.js
mkittelmann: No, getSearchResults doesn't fail, it anticipates this case.

ISTC.js
mkittelmann: There is only one "author" field in the database.

adam3smith:
Thanks, looks great. You'll have seen that I ran this through jsbeautify before committing, so please work from the version on the repository should you need to make further changes in the future.

mkittelmann:
Thank you!

PR #1159, opened 2016-10-07
Fairfax Australia.js

marcralew:
This translator scrapes all Australian Fairfax media newspapers. The Age, The Sydney Morning Herald, Brisbane times and WAtoday. If accepted it could replace the pre-existing but non-functioning The Age & Sydney Morning Herald translators.

### DIFF
--- a/Fairfax Australia
+++ b/Fairfax Australia
@@ -1,0 +1,708 @@
+{
+	"translatorID": "5c7892cb-a19e-4b85-8700-d68b547898f9",
+	"label": "Fairfax Australia",
+	"creator": "Martin Lewis",
+	"target": "^https?://www\\.(theage|smh|brisbanetimes|watoday)\\.com\\.au/",
+	"minVersion": "1.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsi",
+	"lastUpdated": "2016-10-07 01:10:10"
+}
+
+/*********************** BEGIN FRAMEWORK ***********************/
+/**
+	Copyright (c) 2010-2013, Erik Hetzner
+
+	This program is free software: you can redistribute it and/or
+	modify it under the terms of the GNU Affero General Public License
+	as published by the Free Software Foundation, either version 3 of
+	the License, or (at your option) any later version.
+
+	This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


Fairfax Australia
```diff
@@ -0,0 +1,708 @@
...
+itemType	: 'newspaperArticle',
+detect		: FW.Xpath('.//h1[@itemprop="name headline"]'),
+title		: FW.Xpath('.//h1[@itemprop="name headline"]').text().trim(),
+attachments : [ {url: FW.Xpath('//div[@class="article__toolbar"]/ul/li[3]/a').key("href"),
+  title: FW.Xpath('//meta[@name="application-name"]/@content').text().append( " Printable"),
+  type:  "text/html",
+}],
+
+date		: FW.Xpath('//*[@id="content"]/main/article/header/div[1]/ul/li[1]/time').text().trim(),
+creators	: FW.Xpath('//h5[@itemprop="name"]').text().split(/, |" and "/).cleanAuthor("Author"),
```
adam3smith: the split doesn't look right to me. I think that should be `.split(/,\s+|\s+and\s+/)`: you can't mix regular expressions with literal expressions.

Fairfax Australia
```diff
@@ -0,0 +1,708 @@
...
+FW.Scraper({
+itemType	: 'newspaperArticle',
+detect		: FW.Xpath('.//h1[@itemprop="name headline"]'),
+title		: FW.Xpath('.//h1[@itemprop="name headline"]').text().trim(),
+attachments : [ {url: FW.Xpath('//div[@class="article__toolbar"]/ul/li[3]/a').key("href"),
+  title: FW.Xpath('//meta[@name="application-name"]/@content').text().append( " Printable"),
+  type:  "text/html",
+}],
+
+date		: FW.Xpath('//*[@id="content"]/main/article/header/div[1]/ul/li[1]/time').text().trim(),
```
adam3smith: this looks very fragile to me -- I'd expect the time element is unique within the article header? so couldn't we just use:
`//*[@id="content"]/main/article/header//time`?

marcralew: Yes, you are quite right. Code modified.

Fairfax Australia
marcralew: You are quite correct about mixing regular & literal expressions. Apologies. Code modified.

adam3smith:
Great, thanks. Only change I made was to set the min version to 3.0

PR #1162, opened 2016-10-20
Fix SlideShare web translator

bencomp:
This update includes a correction of the first test case (based on current data from SlideShare) and fixes the translator to pass the tests.

It does not improve handling of attachments.


adam3smith:
Great, thanks!

PR #1176, opened 2016-11-03
switch DOI EM Priority

adam3smith:
@zuphilip @dstillman I believe we agreed we wanted to do this as a first step, right? Dan in particular, if you could briefly confirm this is OK by you (I did see your example where it does give us worse data, but I still think it's overall much better)?


dstillman:
OK with me.

zuphilip:
Yes, I am looking forward for that. Fine with me. Thank you!

Issue #1160, opened 2016-10-12
[Embedded Metadata.js] Use Identifier.URL or canonical URL when available


niklasholm:
I was saving [this page](http://www.diva-portal.org/smash/record.jsf?pid=diva2:766397) using the Embedded Metadata translator, but I noticed that the URL was set to the URL of the current page as shown in the address bar, rather than the canonical URL. I'm no javascript expert, but it seems that when all else fails the URL of the new item is set to doc.location.href, which I assume is the URL of the current page as I saw in my item. However, when looking at the HTML meta tags i notice there is both a canonical link tag:

`<link rel="canonical" href="http://www.diva-portal.org/smash/record.jsf?pid=diva2:766397" />`

and a meta tag

`<meta name="DC.Identifier.url" content="http://urn.kb.se/resolve?urn=urn:nbn:se:liu:diva-112443">`

which would both be better options. I argue that the translator should look for these tags and use one of those URLs instead before resorting to the document href. In this case the second is preferable since it contains the permanent URL (equivalent of an DOI URL).

adam3smith:
Unless someone sees issues with this, that sounds fine to me. Which of the two should be preferred where both exists?

zuphilip:
I don't think that `DC.Identifier.url` exists in the Dublin Core Specs.

niklasholm:
I do not have any affiliation with diva-portal.org so I cannot change anything but I suppose I can contact the webmaster and make them aware of the issue.  However, I cannot see anything in the code that looks for `DC.Identifier` either, so I don't thing that would work anyway.

Either way, the canonical URL should definitely be used if available and should probably be preferred. The problem with `DC.Identifier` is that it isn't necessarily an URL, it can be any unique identifier, as for example in this case a URN:NBN which is used by multiple European national libraries, see [this RFC][1], which could of course be converted to an URL using a lookup service, however there's no standard way to do this. Quote from a DOI factsheet:

[...]

PR #1181, opened 2016-11-07
EM: use canonical URL, use cover_date

adam3smith:
Closes https://github.com/zotero/translators/issues/1160
Addresses https://forums.zotero.org/discussion/59197/springer-useing-first-online-instead-of-real-publication-date#latest
Supersedes https://github.com/zotero/translators/pull/1167

Springer, update tests.
@zuphilip -- if you could take a quick look? I'm putting the cover date in 2nd place after citation_date but before citation_publication_date. I think tha's the best call, though we can consider moving it to 1st place.

### DIFF
--- a/Embedded Metadata.js
+++ b/Embedded Metadata.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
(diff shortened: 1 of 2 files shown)


zuphilip:
I just glanced at this briefly. This looks good, also I haven't thought about the order, but it seems to work fine.

PR #1203, opened 2016-12-28
[ISTC.js]: Replace (deprecated) for each loops

adomasven:
Also change inconsistent line endings to be in line with the rest of translators using `\n`

### DIFF
--- a/ISTC.js
+++ b/ISTC.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2016-11-01 14:02:29"
+	"lastUpdated": "2016-12-28 14:39:45"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


ISTC.js
```diff
@@ -12,127 +12,128 @@
 	"lastUpdated": "2016-11-01 14:02:29"
```
zuphilip: You have to increase the date here, in order to trigger the distribution to the clients.

adomasven:
>It looks that you changed also some whitespaces because the diff shows that everything changed

Yes. This is my IDE automatically converting newline characters. I scanned all translator files and it seems that this is the only one which had windows style newlines `\r\n` mixed in with `\n`. Needless to say, having mixed newlines can cause all kinds of unexpected issues, so all of those were replaced with `\n`, per my commit comment.

zuphilip:
Okay, that looks good for me. 👍 Changing `\r\n` to `\n` is reasonable and consistent, then. (On Windows the default for git is to allow such line endings but to check them into the Linux way.)

adam3smith:
Great, thanks!

PR #1213, opened 2017-01-04
Add Tesis Doctorals en Xarxa (TDX)

CSUC:


CSUC:
Any comment? we would want to use it as soon as possible, could anyone please check it?

Thank you very much

Tesis Doctorals en Xarxa.js
```diff
@@ -0,0 +1,277 @@
+
+{
+	"translatorID": "dd2d7c63-0c4b-4da4-b1d8-e72c7445c1e0",
+	"label": "Tesis Doctorals en Xarxa",
+	"creator": "CSUC",
+	"target": "^https?://(www\.)?(tdx\.cat|tesisenred\.net)",
```
zuphilip: The regexp in JSON needs to escape the period `.` with double backslash, i.e. use
```
"target": "^https?://(www\\.)?(tdx\\.cat|tesisenred\\.net)",
```

Tesis Doctorals en Xarxa.js
```diff
@@ -0,0 +1,277 @@
...
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsb",
```
zuphilip: By default we add here all, i.e. `"gcsibv"`

Tesis Doctorals en Xarxa.js
```diff
@@ -0,0 +1,277 @@
...
+	"lastUpdated": "2016-09-22 15:33:33"
+}
+
+
```
zuphilip: All translators need a license block (AGPL), see the example here https://github.com/zuphilip/translators/wiki/Common-code-blocks-for-translators#licence-block

adam3smith: @zuphilip which reminds me -- we should write a style guide for translator coding.

zuphilip: Yeah, I think what is currently missing is a `README.md` and a file `CONTRIBUTING` (which we can link from the `README.md`). A style guide could be of this? We can also discuss this further in a new issue, if you would like.

CSUC:
Hi! Thank you! I've done some changes adding the right escape on regex and adding the missed license block.
It's necessary to add the getSearchResults function?

zuphilip:
These things looks good now. The function `getSearchResults` is not strictly needed, but it is often used in translators as a general pattern.

Let me repeat my comment and questions from above for the rest:

> But it seems that you a lot of meta tags and just one or to other elements. Therefore I guess that the current metadata with the Embedded Metadata translator is already pretty good. No? Shouldn't we then better first start with calling EM translator and adjust the remaining things afterwards? Here is my template for such a case https://github.com/zuphilip/translators/wiki/Calling-metadata-translator

CSUC:
Hi @zuphilip, thank you for your help! I've changed the code following the Translator template, I hope I didn't forgot anything else.

Tesis Doctorals en Xarxa.js
```diff
@@ -0,0 +1,309 @@
...
+}
+
+function scrape(doc, url) {
+			
+	var translator = Zotero.loadTranslator('web');
+	translator.setTranslator('951c027d-74ac-47d4-a107-9c3069ab7b48');//https://github.com/zotero/translators/blob/master/Embedded%20Metadata.js
+	translator.setDocument(doc);
+	translator.setHandler('itemDone', function (obj, item) {
+			//title
+			var titles = ZU.xpath(doc, '//meta[@name="citation_title"]');
```
zuphilip: I am quite sure that all these metadata fields are already capture, when you arise her. Can you comment out everything in this function expect the `item.complete();`? What is the result then and which parts are missing then? I would expect that only very few lines will be needed in the end here in this `setHandler` function.

adam3smith: yes, please do that. In addition, for the fields you do need to add or change, using `item.field = ZU.xpathText(doc, xpath)` is much simpler and equivalent to what you're doing with `ZU.xpath`.

Tesis Doctorals en Xarxa.js
```diff
@@ -0,0 +1,309 @@
...
+    along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+    ***** END LICENSE BLOCK *****
+*/
+
+
+function detectWeb(doc, url) {
+	var type = ZU.xpath(doc, '//meta[@name="DC.type"]/@content');
+	if (type.length>0) {
+		if (mappingTable[type[0].textContent]) {
```
adam3smith: while this is generally a good approach, it seems silly to do this as long as the whole excercise will just lead to a `thesis` item type either way. If you think the site will add other items in the feature, I'd be OK with you just commenting this (and the then unused mapping function) out. If this will likely remain a thesis repository, just return thesis.

CSUC:
Done more changes following all the tips

zuphilip:
This looks pretty good now. Thank you! I haven't looked at the details, but I am sure that @adam3smith will do that.

Tesis Doctorals en Xarxa.js
```diff
@@ -0,0 +1,247 @@
...
+			if (rights.length>0) {
+					item.rights = [];
+					for (var t=0; t<rights.length; t++) {
+						if (mappingRights[rights[t].content])
+						item.rights.push(mappingRights[rights[t].content]);
+					}
+			}
+		
+		//deleting all unwanted tags
+		item.reportType=item.letterType=item.manuscriptType=item.mapType=item.websiteType=item.presentationType=item.postType=item.audioFileType=null
```
adam3smith: no need for this. Zotero discards them.

Tesis Doctorals en Xarxa.js
```diff
@@ -0,0 +1,247 @@
...
+			item.thesisType="Ph.D. Thesis"
+		}
+		
+		//add tags -> from rights
+			var rights = ZU.xpath(doc, '//meta[contains(@name, "DC.rights")]');
+			if (rights.length>0) {
+					item.rights = [];
+					for (var t=0; t<rights.length; t++) {
+						if (mappingRights[rights[t].content])
+						item.rights.push(mappingRights[rights[t].content]);
```
adam3smith: you end up with an array here. `item.rights` expects a string. You'd want to run someting like 
`item.rights = item.rights.join(", ")` or some other operation that converts this to a string. 
But I'm also wondering: are there items with multiple rights metatags?

Tesis Doctorals en Xarxa.js
```diff
@@ -0,0 +1,247 @@
...
+	translator.getTranslatorObject(function(trans) {
+       		trans.itemType = "thesis";
+       		trans.doWeb(doc, url);
+    	});
+}
+
+
+/** BEGIN TEST CASES **/
+/** BEGIN TEST CASES **/
+var testCases = [
```
adam3smith: Are these tests created with Scaffold? The most recent version shouldn't be including all those extraneous variables in fields. It would also get spacing right and not insert spaces at the beginning of some lines. 
Please fix.

Tesis Doctorals en Xarxa.js
CSUC: Yes, they have multiple rights metatags, I will include this join sentence.

Tesis Doctorals en Xarxa.js
CSUC: I did it with Scaffold,, I have deleted all spaces at the beggining of some lines as you said

CSUC:
Hi!
Is still not valid? Is there any change pending to be done?

Thank you.

toniher:
Hi! anything that can be done to help on this? After merged and once included in Citoid, this translator will be very useful in some Wikipedias. Thanks!

adam3smith:
Sorry for letting this linger & thanks for contributing.

PR #1222, opened 2017-01-13
Add camelCase url support

JaretK:
Rebased repository and added support for camelCase urls in access medicine translator


adam3smith:
could you bump the timestamp please? Otherwise the translator will not update for users.

adam3smith:
Thanks!

PR #1229, opened 2017-01-19
Update TEI.js to handle changes to idno

martindholmes:
Recent changes to the TEI schema disallow the idno element as a direct child of biblStruct, and constrain where it may appear in analytic and monogr. This patch attempts to place idnos in the correct location to produce valid TEI. Original bug report is here:

https://forums.zotero.org/discussion/comment/268352

### DIFF
--- a/TEI.js
+++ b/TEI.js
@@ -19,7 +19,7 @@
 		"Full TEI Document": false,
 		"Export Collections": false
 	},
-	"lastUpdated": "2016-05-25 15:08:00"
+	"lastUpdated": "2016-01-19 20:45:00"
 }
 
 // ********************************************************************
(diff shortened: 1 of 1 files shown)


TEI.js
```diff
@@ -19,7 +19,7 @@
 		"Full TEI Document": false,
 		"Export Collections": false
 	},
-	"lastUpdated": "2016-05-25 15:08:00"
+	"lastUpdated": "2016-01-19 20:45:00"
```
adam3smith: as much as we may not want it to be true, it is, in fact 2017 ;)

adam3smith:
something went wrong with beautify -- these are spaces, not tabs in the indenting.

TEI.js
```diff
@@ -1,25 +1,25 @@
...
+        "getCollections": "true"
+    },
+    "displayOptions": {
+        "exportNotes": false,
+        "Export Tags": false,
+        "Generate XML IDs": true,
+        "Full TEI Document": false,
+        "Export Collections": false
+    },
+    "lastUpdated": "2016-01-22 20:45:00"
```
adam3smith: and your date is still 2016 -- the translator won't update to users unless you move this to 2017.

martindholmes: Yep, screwed up the date. I copy/pasted what came out of beautify with the default settings directly in -- what settings should I use? Indent with a tab character?

adam3smith: yes. That's the only one you need to change, all others are fine with the default.

adam3smith:
Thanks! I'll ping you if there are complaints about this -- as I said, I don't know TEI specs terribly well, so beyond the fact it makes sense in general, I can only judge the Zotero/js part of the code which is in good shape.

martindholmes:
Thanks. There may be people working with older TEI schemas who are expecting the old behaviour; if so, we'll have to see whether they need a workaround or they can migrate. The old placement of idno was deprecated for two years before being made invalid.

PR #1230, opened 2017-01-20
ci: checkTranslator with a --skip-warn flag

kba:
This should make it possible to run "npm test" in Windows with cmd.exe instead of a POSIX shell.

### DIFF
--- a/.ci/checkTranslator.sh
+++ b/.ci/checkTranslator.sh
@@ -123,8 +123,12 @@ usage () { (( $# > 0 )) && err "$*"; err "Usage: $0 <translator.js>"; }
 main() {
 
     # Add './node_modules/.bin to PATH for jsonlint
-    PATH="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"/node_modules/.bin:"$PATH"
+    PATH=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/node_modules/.bin:"$PATH"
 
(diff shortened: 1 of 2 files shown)


.ci/checkTranslator.sh
```diff
@@ -123,8 +123,12 @@ usage () { (( $# > 0 )) && err "$*"; err "Usage: $0 <translator.js>"; }
 main() {
 
     # Add './node_modules/.bin to PATH for jsonlint
-    PATH="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"/node_modules/.bin:"$PATH"
+    PATH=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/node_modules/.bin:"$PATH"
```
zuphilip: It looks for me that you maybe can delete another pair of `"` after the `cd`. No? At least it then still seems to work fine for me.

adam3smith: does it matter either way? Happy to merge this otherwise.

zuphilip: No, it doesn't matter and is just cosmetic. Feel free to merge. (We are already planing the next commit which will check also the deletion of translators.)

adam3smith:
if you're already planning another commit, I don't want to add any merge conflicts, so could you add `  "multistr": true` to the jshint configuration file, please? laxbreak does not, in fact, remove the EOL warning we had talked about before, though I think we should keep it, too.

PR #1232, opened 2017-01-22
ScienceDirect.js -- More fixes for new format

adam3smith:
Also, fix RIS spacing, fix some code warnings, and run through jsbeautify

### DIFF
--- a/ScienceDirect.js
+++ b/ScienceDirect.js
@@ -9,35 +9,35 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2017-01-13 20:03:14"
+	"lastUpdated": "2017-01-22 19:29:47"
 }
 
 function detectWeb(doc, url) {
 	if (!doc.body.textContent.trim()) return;
-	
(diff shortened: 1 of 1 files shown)


adam3smith:
( cleaner diff at https://github.com/zotero/translators/pull/1232/files?w=1 )

ScienceDirect.js
```diff
@@ -209,31 +211,36 @@ function processRIS(doc, text) {
...
-			
+
 			return pre + name;
 		}
 	);
+	//Z.debug(text)
+	//The RIS sometimes has spaces at the beginning of lines, which break things
+	//as of 20170121 e.g. on http://www.sciencedirect.com/science/article/pii/B9780123706263000508 for A2
+	//remove them
+	text = text.replace(/\n\s+([A-Z][A-Z0-9])\s{2}\-\s/, "\n$1  - ");
```
zuphilip: How about simplifying this with
```js
text = text.replace(/\n\s*/, "\n");
```
? Because the `*` is greedy, it should delete all spaces in the beginning of lines, and it seems less fragile than your regexp.

adam3smith:
thanks for looking it over!

PR #1233, opened 2017-01-22
BibTeX/RIS: Add DOIs to non-supported item types

adam3smith:
BibTeX: also add additional test, udpated tests,  and add better
handling of additional identifiers from notes (which were in my local
bibtex either from an old version by Emiliano or from an attempt by me;
either way they look good).


RIS.js
```diff
@@ -1451,6 +1452,19 @@ function applyValue(item, zField, value, rawLine) {
...
+				if(item.extra) {
+					item.extra += '\nDOI: ' + value;
+				} else {
+					item.extra = 'DOI: ' + value;
+				}
+			}
+			else {
+				item[zField] = value;
+			}
+		break;
```
zuphilip: Alternatively, you can delete the `else` above and just continue without a `break`, which means that for `DOI` all `default` things will be performed and (depending on the `itemType`) also some more, i.e. delete these lines
```js
			else {
				item[zField] = value;
			}
		break;
```

BibTeX.js
```diff
@@ -262,7 +263,14 @@ function setKeywordDelimRe( val, flags ) {
 function processField(item, field, value, rawValue) {
 	if(Zotero.Utilities.trim(value) == '') return null;
 	if(fieldMap[field]) {
-		item[fieldMap[field]] = value;
+		//map DOIs + Label to Extra for unsupported item types
+		if (field == "doi" && item.itemType != "journalArticle" && item.itemType != "conferencePaper") {
```
zuphilip: I like the (more general) function `ZU.fieldIsValidForType("DOI", item.itemType)` which you used in another translator. How about using it here as well?

CrossRef.js
```diff
@@ -269,28 +269,32 @@ function processCrossRef(xmlOutput) {
...
 	if(pages.length) {
 		item.pages = ZU.xpathText(pages, 'c:first_page[1]', ns);
 		var lastPage = ZU.xpathText(pages, 'c:last_page[1]', ns);
 		if(lastPage) item.pages += "-"+lastPage;
 	}
-	
+
 	item.DOI = ZU.xpathText(refXML, 'c:doi_data/c:doi', ns);
+	//add DOI to extra for unsupprted items
+	if (item.DOI && item.itemType != "journalArticle" && item.itemType != "conferencePaper") {
```
zuphilip: I like the (more general) function `ZU.fieldIsValidForType("DOI", item.itemType)` which you used in another translator. How about using it here as well?

BibTeX.js
```diff
@@ -262,7 +263,14 @@ function setKeywordDelimRe( val, flags ) {
 function processField(item, field, value, rawValue) {
 	if(Zotero.Utilities.trim(value) == '') return null;
 	if(fieldMap[field]) {
-		item[fieldMap[field]] = value;
+		//map DOIs + Label to Extra for unsupported item types
```
zuphilip: We could also let the first line as it was and just do some extra work for the doi, i.e. doing the `else` case below always.

adam3smith: I'll leave this here -- I'd like to leave the DOI in the generic field mapping (l. 84) and the only way to get to `else` would be to remove it there.

zuphilip: Not sure we understand each other here. My suggestion would like this:
```js
 	if(fieldMap[field]) {
		//default behaviour
		item[fieldMap[field]] = value;
		//additionally map DOIs + Label to Extra for unsupported item types
		if (field == "doi" && item.itemType != "journalArticle" && item.itemType != "conferencePaper") {
			var label = extraIdentifiers[field];
			item._extraFields.push({field: label, value: value.trim()});
		}
```

adam3smith: Ah I see. I figured it'd be better if the item _isn't_ mapped to item.DOI when it's written to Extra, so that we don't get DOI twice when using bibtex from a web translator that also adds DOIs (I know a couple do). What's the advantage of doing this your way around?

zuphilip: Unsupported fields will not be saved in Zotero such that there will never be two fields containing the DOI information. No?

Anyway, there is no big deal here. I just thought it might be easier to read (and maybe already work correctly in some future version of Zotero item types).

I will review your changes a little later, but I am sure everything is good.

adam3smith: but they're available in the translator object before saving, so something like this in a web translator:
https://github.com/zotero/translators/blob/master/Springer%20Link.js#L158
would cause DOI: 10.1234/123456 to appear twice in the Extra field.

zuphilip: ok

RIS.js
```diff
@@ -1453,18 +1453,14 @@ function applyValue(item, zField, value, rawLine) {
...
 				if(item.extra) {
 					item.extra += '\nDOI: ' + value;
 				} else {
 					item.extra = 'DOI: ' + value;
 				}
 			}
-			else {
-				item[zField] = value;
-			}
-		break;
```
zuphilip: This change I cased with my previous comment, might produce now that the doi data will be saved in `item.extra` as well as in `item.DOI` in the translator obejct as you outlined in https://github.com/zotero/translators/pull/1233#discussion_r97809580 . Thus, we might switch back here again to the else case and the break to prevent that.

RIS.js
```diff
@@ -2069,7 +2065,7 @@ var testCases = [
 				"abstractNote": "Abstract",
 				"archive": "Name of Database",
 				"archiveLocation": "Accession Number",
-				"extra": "Publication Number",
+				"extra": "DOI: 10.1234/123456\nPublication Number",
```
zuphilip: Why did this value vanish now?

adam3smith: I don't understand -- it's still there after the line break.

CrossRef.js
```diff
@@ -293,7 +293,12 @@ function processCrossRef(xmlOutput) {
 	item.DOI = ZU.xpathText(refXML, 'c:doi_data/c:doi', ns);
 	//add DOI to extra for unsupprted items
 	if (item.DOI && item.itemType != "journalArticle" && item.itemType != "conferencePaper") {
-		item.extra = "DOI: " + item.DOI;
```
zuphilip: How about using `!ZU.fieldIsValidForType("DOI", item.itemType)` here as well?

RIS.js
zuphilip: Sorry, my fault. Somehow I switched the meaning of green and red here in my mind. The doi is here, which is good.

RIS.js
adam3smith: I think I have an elegant solution now moving the `break` into the if condition

CrossRef.js
adam3smith: somehow forgot this; done.

adam3smith:
great, thanks

zuphilip:
Thank you @adam3smith for the work! This looks really great.
(And sorry for my sloppy reviews, I am a little distracted by other things which are going on...)

PR #1234, opened 2017-01-22
Clinical key journal articles

mtd91429:
I updated the translator for Clinical Key recently written by @JaretK to handle journal articles. It uses the CrossRef REST API to query  Publisher Item Identifiers as was discussed on the forum with @adam3smith  https://forums.zotero.org/discussion/51257/clinical-key

The translator works from my testing (except the tags feature). However, I suspect that my code format and/or implementation may not be in line with the repository standard and/or sound development practices. Let me know what needs tweaked and I'll do my best.


Clinical Key.js
```diff
@@ -130,6 +140,117 @@ function doWeb(doc, url){
...
+		ref = jsonParsed['message']['items'][0];
+	} catch (e) {
+		Zotero.debug(e);
+		return false;
+	}
+	
+	if (jsonParsed['message']['total-results'] > 1) {
+		//Not sure what to do with multiple results 
+		//Should only return 1 item
+		throw new Error('Returned multiple results');
```
adam3smith: This should be impossible for a DOI query, but if it happens, just parse the first one. Don't throw an error.

Clinical Key.js
```diff
@@ -130,6 +140,117 @@ function doWeb(doc, url){
...
+		return;
+	}
+	
+	if (ref['type'] != 'journal-article') {
```
adam3smith: CrossRef JSON closely mirrors CSL JSON, for which there already is a translator, which you can invoke as we do for DataCite: https://github.com/zotero/translators/blob/master/DataCite.js#L72 -- just like there, you can then simply fix the things that require fixing, but safe yourself all the remapping done below.

Clinical Key.js
```diff
@@ -130,6 +140,117 @@ function doWeb(doc, url){
...
+	}
+	
+}
+
+//Search & Processing based on CrossRef.js translator
+function doSearch(pii, doc){
+	crossRefQuery = 'http://api.crossref.org/works?query=' + pii
+	Zotero.Utilities.HTTP.doGet(crossRefQuery, function(responseText) {
+		processCrossRefREST(responseText, doc);
+		Zotero.done();
```
adam3smith: `Zotero.done` and `Zotero.wait` are no longer needed.

Clinical Key.js
```diff
@@ -130,6 +140,117 @@ function doWeb(doc, url){
...
+		
+		doSearch(pii, doc);
+	}
+	
+}
+
+//Search & Processing based on CrossRef.js translator
+function doSearch(pii, doc){
+	crossRefQuery = 'http://api.crossref.org/works?query=' + pii
+	Zotero.Utilities.HTTP.doGet(crossRefQuery, function(responseText) {
```
adam3smith: while it doesn't make much of a difference, `ZU.doGet` is preferred. (See https://github.com/zotero/translators/issues/1242 for background if you're interested)

Clinical Key.js
```diff
@@ -130,6 +140,117 @@ function doWeb(doc, url){
...
+	else if (contentType == 'journalArticle') {
+		eid = url.split('/');
+		pii = eid.pop().slice(7);
+		
+		/* can use xpath to get pii; guessing the url is faster
+		var eid = ZU.xpathText(doc, "//ul/@data-eid");
+		pii = eid.slice(7);*/
+		
+		if(!/^S(\d{15}X|\d{16})/.test(pii)){
+			throw new Error('PII incorrectly formatted');
```
adam3smith: why not try the xpath above if the URL method fails before giving up. We want this as robust as possible.

Clinical Key.js
```diff
@@ -130,6 +140,117 @@ function doWeb(doc, url){
...
+	if (jsonParsed['message']['total-results'] > 1) {
+		//Not sure what to do with multiple results 
+		//Should only return 1 item
+		throw new Error('Returned multiple results');
+		return;
+	}
+	
+	if (ref['type'] != 'journal-article') {
+		//redundency check to ensure proper citation 
+		throw new Error('Returned unexpected reference type')
```
adam3smith: Also, let's not throw this out if the item type looks wrong. Still better to try importing. I have a CrossRef to CSL JSON mapping table here that might be useful: https://github.com/zotero/translators/pull/1135/files

mtd91429:
Thanks for all the suggestions; I'll get to work on these this weekend. The other change that I started on yesterday, and was hoping for some feedback, has to deal with managing API versions for the CrossRef REST. They have documentation [here](https://github.com/CrossRef/rest-api-doc/blob/master/rest_api.md#how-to-manage-api-versions) specifying version requests. However, I couldn't get it to work ([link to the issue](https://github.com/CrossRef/rest-api-doc/issues/181)). From what I can tell (which may be incorrect), the bug is on their side. In the next commit I'll show you what I've come up with. Thanks again!

Clinical Key.js
mtd91429: In my comment below, I mentioned attempting to specify the API version using an Accept Header in the HTTP request. Currently, that doesn't seem to work. However, if the problem is solved, based on the criteria you describe in #1242, would it be appropriate then to continue using the ZU.HTTP.doGet?

Clinical Key.js
mtd91429: This is a great suggestion; I noticed a few fields that would need handling if not present (ex. 'issue'). The CSL JSON has a much better implementation to handle those issues.

Clinical Key.js
zuphilip: This sounds similar to the lines from ORCID translator, see https://github.com/zotero/translators/blob/master/ORCID.js#L68-L76 which we currently use. (The `minVersion` has to be then `4.0.29.11`.) However, I haven't look into the details here...

Clinical Key.js
```diff
@@ -141,116 +141,95 @@ function doWeb(doc, url){
...
 		doSearch(pii, doc);
 	}
 	
 }
 
 //Search & Processing based on CrossRef.js translator
 function doSearch(pii, doc){
-	crossRefQuery = 'http://api.crossref.org/works?query=' + pii
+	crossRefQuery = 'http://api.crossref.org/works?query=' + pii;
+	//acceptHeader = {'Accept': 'application/vnd.crossref-api-message+json; version=1.0'}
```
mtd91429: Hoping for some feedback here with managing API versions for the CrossRef REST. They have documentation [here](https://github.com/CrossRef/rest-api-doc/blob/master/rest_api.md#how-to-manage-api-versions) specifying version requests. However, I couldn't get it to work ([link to the issue](https://github.com/CrossRef/rest-api-doc/issues/181)). It seems to be server side based on my GET request (see the issue), but I could be wrong.

Clinical Key.js
```diff
@@ -141,116 +141,95 @@ function doWeb(doc, url){
...
 	Zotero.Utilities.HTTP.doGet(crossRefQuery, function(responseText) {
```
mtd91429: Just saw @zuphilip's comment on my previous commit. In the next revision, I'll switch this to `ZU.doGet()` as it should work.

adam3smith: Just to make sure I understand -- I'm waiting for you to make that switch before reviewing again, right? (no rush, I just want to make sure you're not waiting for me)

mtd91429: Kind of (but this is largely my fault; sorry). I guess what I'd specifically like feedback on (which is what I'm actually waiting for) is how to deal with the crossref API version control. Ideally it would just be implemented and we'd move on. However, bc the API version control is not working (or I've missed something), should we implement the translator with the knowledge that it may break in six months? Or, should I spend the time to try and write in an error-catching plan? (I hope that makes sense; I'll try to clarify if not). The error catching plan seems like more effort than it's worth (but I could be short sighted). Similarly, we could just update the translator if the API is fixed and/or updated. Perhaps I'm overthinking it.

adam3smith: ah I see -- I can't get content negotiation working on their API at all either (I tried with `curl`), including when just requesting a format, so I think you're right it's there issue. I doubt they'll push out many major versions of the API, so I'd just no handle API versions at all and wait until it breaks ;)

mtd91429: Sounds like a plan. I'll keep an eye on the ticket I opened on their repository. If it gets fixed, I'll update the code.

mtd91429:
Sorry for the commit proliferation. I thought my actions would squash all the commits into one, but it apparently did the opposite. I'll fix this tonight.

adam3smith:
no need to squash. We can do this with a single button when accepting.

adam3smith:
Thanks -- since none of the "usual suspects" has access to this, let us know how it works and thanks for contributing!

PR #1237, opened 2017-01-26
[De Gruyter] Fix PDF Links

adam3smith:


adam3smith:
https://forums.zotero.org/discussion/63135/zotero-detects-metadata-but-does-not-download-pdf#latest

PR #1238, opened 2017-01-27
Ci check deleted.txt when translators are deleted

zuphilip:
Add another script for checking the `deleted.txt`, which has to be updated when deleting translators. This will work for
* pull requests against the master branch
* locally when working on feature branch

Though, we should just make sure that all deletions of translators are done as PR and not direct pushes, but this might be anyway a good idea.

There are also some small improvements for the `checkTranslator.sh`.

This is joint work with @kba (the shell script ninja) 🙇 .

### DIFF
--- a/.ci/README.md
+++ b/.ci/README.md
@@ -1,0 +1,41 @@
+# Continuous integration tests
+
+The tests here are automatically used by travis
+see [.travis.yml](../.travis.yml).
+
+
+## Requirements
+
+Testing works also locally. The only requirements should be
+an (recent) installation of nodejs. Then, for the dependencies you can
+simply type
+```
+npm install
+```
+ 
+## Using the npm bindings
+
+You can use the npm bindings to run tests on all translators:
+* `npm start` to run check for all errors and warnings with verbose output
+(slow, because only one core is used, such that the output will stay in alphabetical order)
+* `npm test` to run check for errors only with terse output
+(faster, because up to 8 cores are used, such that output will not stay in alphabetical order)
+
+
+## Using the script directly
+
+You can use the bash script `checkTranslator.sh` directly to
+run tests on a single translator, e.g.
+```
+./checkTranslator.sh ../Amazon.js
+```
+Moreover, you can skip the warnings by setting the environment
+variable `SKIP_WARN` to `true`:
+```
+export SKIP_WARN=true
+./checkTranslator.sh ../Amazon.com
+```
+
+To switch back just set the variable to the empty string.
+
+
--- a/.ci/README.md
+++ b/.ci/README.md
@@ -27,12 +27,16 @@ You can use the npm bindings to run tests on all translators:
 You can use the bash script `checkTranslator.sh` directly to
 run tests on a single translator, e.g.
 ```
+# in .ci directory:
 ./checkTranslator.sh ../Amazon.js
+
+# in translators directory:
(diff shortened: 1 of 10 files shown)


.ci/README.md
```diff
@@ -22,20 +22,24 @@ You can use the npm bindings to run tests on all translators:
...
 ./checkTranslator.sh ../Amazon.js
 ```
-Moreover, you can skip the warnings by setting the environment
-variable `SKIP_WARN` to `true`:
+Moreover, you can skip the warnings by using the `--skip-warn`
+option:
 ```
-export SKIP_WARN=true
-./checkTranslator.sh ../Amazon.com
+./checkTranslator.sh --skip-warn ../Amazon.com
```
adam3smith: question: is there are reason the script needs to use the awkward ../Amazon.com syntax or could we add the ../ into the script so that this would be `/./checkTranslator.sh --skip-warn Amazon.com`

adam3smith:
I have one question above, otherwise this looks great to me.

.ci/README.md
dstillman: I think that's a mistake anyway — it assume it should be `../Amazon.com.js`.

We could have it treat the given name as a label (and have that work regardless of the current working directory — so `.ci/checkTranslator.sh Amazon.com` would work too), but it should also take a filename so that you can get filename completion.

zuphilip: Yes, the missing `.js` is a mistake, however the Amazon translator was renamed to [`Amazon.js`](https://github.com/zotero/translators/blob/master/Amazon.js) without the `.com`, because it works on several (localized) Amazon websites. I always use the filename completion (especially if the filename contains spaces) and therefore didn't saw my error earlier.

This works after this PR: `.ci/checkTranslator.sh Amazon.js`.

We can do some of the other "simplifications" but I am not sure how much we want of these. Let me know what you think we should (essentially) support:
1. In the `translators` directory
   a) `.ci/checkTranslator.sh Amazon.js` ✔️ 
   b) `.ci/checkTranslator.sh Amazon` ❔ 
2. In the `.ci` directory
   a) `./checkTranslator.sh ../Amazon.js` ✔️ 
   b) `./checkTranslator.sh Amazon.js`❔ 
   c) `./checkTranslator.sh Amazon`❔❔

dstillman: We should support path and label (filename without `.js`). We don't need to support filename (but if you're in the root, the filename is the same as the path).

So:

1. In the `translators` directory
   a) `.ci/checkTranslator.sh Amazon.js` ✔ 
   b) `.ci/checkTranslator.sh Amazon` ✔ 
2. In the `.ci` directory
   a) `./checkTranslator.sh ../Amazon.js` ✔️ 
   b) `./checkTranslator.sh Amazon.js` ✘ 
   c) `./checkTranslator.sh Amazon`✔

dstillman: So, basically, we just need to test whether path exists as-is, and if it doesn't try `$ROOTDIR/$LABEL.js` (or whatever the variables are).

zuphilip:
Okay, I added the option to reference the translators just by label and updated the README file.

Please have a look at the new version.

adam3smith:
excellent! Thanks

zuphilip:
Thanks to @kba for making my first messy attempts into scripts which work excellently and have well organized code!

PR #1248, opened 2017-03-03
Update wikipedia citation templates format

bunyk:
And fix pipe errors in titles

### DIFF
--- a/Wikipedia Citation Templates.js
+++ b/Wikipedia Citation Templates.js
@@ -354,24 +354,28 @@ function doExport() {
 		}
 		
 		// write out properties
-		Zotero.write((first ? "" : "\r\n") + "{{"+type);
+		Zotero.write((first ? "" : "\r\n") + "<ref>{{"+type);
 		for(var key in properties) {
 			if (!properties[key]) continue;
 			
 			if (key == 'authors') {
 				var index = properties.authors.length > 1;
 				for (var i=0; i<properties.authors.length; i++) {
-					Zotero.write('| last' + (i|| index ? i+1 : '') + ' = ' + properties.authors[i].last);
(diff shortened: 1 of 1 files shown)


adam3smith:
@owcz could I ask you to check if this makes sense? No need to check syntax, but I'd like someone with WP knowledge to take a look at the changed output.

owcz:
Yep, the line breaks are non-standard (sometimes written like that to make the citations easier for humans to read, but not efficient within articles). For standard web citation formats, see [this section](https://en.wikipedia.org/wiki/Template:Cite_web#Examples). As for the `<ref>` tags, that's also a non-standard user preference. The citations don't necessarily need them, depending on how the bibliography is built.

I had actually rewritten this script several years ago but I just never cleaned it up for production and edge cases... Part of the problem here is that there will be a high variance in what different users will want off of the base citation (changing the date format, whether to put an asterisk or `<ref>` in front of the built citation, etc.) I remember there not being good options for dealing with that within Zotero, so I had worked on my own tools outside of it.

adam3smith:
Thanks! So @bunyk -- I'm not convinced we want the other changes, then (I'm open to be convinced, though; I have no stake in this except to get it right). 
Just so I understand the actual fix you're proposing here -- pipes are escaped as {{!}} in the template? That seems odd -- but if that's the case, the solution is fine and we'll take it.

owcz:
The pipe escaping is good, as pipes are used to delineate between params in the Wikipedia citation style. If not escaped, the pipes cause havoc.

bunyk:
@adam3smith Yes, all this actually is copied from my own setup. I made this pull request minimal.

adam3smith:
thanks & sorry for the delay in merging this

PR #1253, opened 2017-03-11
Update Polygon.js for 2018

owcz:
authors was broken in prior format, rewritten in FW

### DIFF
--- a/Polygon.js
+++ b/Polygon.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2018-07-07 19:10:18"
+	"lastUpdated": "2021-05-26 20:01:12"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


owcz:
Some of the site's old "feature" articles (e.g., [this one](http://www.polygon.com/2012/10/4/3456778/the-professional-fan-how-earthbound-led-to-a-marriage-and-a-career)) use non-standard metadata (particularly re: author). Not sure how best to handle them—open to suggestions, or just ignoring them

adam3smith:
I haven't looked at the differences in depth, but this is the type of things you'd find much easiert to address in a non-FW translator, where you could just cycle through a couple of ways of getting an author in turn. Given that you already have the existing structure, it'd makes sense to me to keep the old translator and just swap out the xpaths, then add checks for the other author options.

owcz:
The site's early Features (from several years ago) have inconsistent formatting, hence why I'm not particularly concerned about handling the edge cases. (FW handles 99% of the site's articles and is much easier to maintain.)

@adam3smith Is there no way to do a secondary detection when FW.Scraper runs for "creators" and finds nothing? There is a fairly standard `//meta[@property="author"]/@content` field, but it will only show the first author if there are multiple (hence why I opted for `c-byline__item` instead).

adam3smith:
I think you could do a hook testing for an empty creators array and using the other options if you want to stick to FW -- I find it quickly gets easier to write regular translators once you start using hooks, but up to you.

owcz:
EM handles most of this site now, but updated to handle multis (search, HP, cat pages) and that of its sister sites
Ready for review

AbeJellinek:
Squashed, linted, updated (a little bit), and merged! Thank you :)

PR #1255, opened 2017-03-11
Typo in Libgen translator URL detection

owcz:
See examples—order was switched in URL detection


adam3smith:
thanks -- could you please add a test from the gen.lib.... domain (or substitute one of the libgen.io one)?

adam3smith:
Thanks!

PR #1256, opened 2017-03-11
Add ARTnews.js

owcz:


ARTnews.js
```diff
@@ -0,0 +1,147 @@
...
+				"itemType": "blogPost",
+				"title": "Venice Biennale Taps Christine Macel to Be Artistic Director of 2017 Edition",
+				"creators": [
+					{
+						"firstName": "Andrew",
+						"lastName": "Russeth",
+						"creatorType": "author"
+					}
+				],
+				"date": "01/23/16 10:35 am",
```
adam3smith: It'd be good to transform these dates to YYYY-MM-DD to avoid confusion with different locales. I think something like
`.replace(/(\d{2})\/(\d{2})\/(\d{2})/, "20$3-$2-$2")` should work, but please test.

ARTnews.js
```diff
@@ -0,0 +1,147 @@
...
+	{
+		"type": "web",
+		"url": "http://www.artnews.com/2017/03/10/the-whitney-biennial-arrives-heres-a-round-up-of-coverage-of-artists-in-the-show/",
+		"items": [
+			{
+				"itemType": "blogPost",
+				"title": "The Whitney Biennial Arrives! Here’s a Round-Up of Coverage of Artists in the Show",
+				"creators": [
+					{
+						"firstName": "The Editors of",
```
adam3smith: how common is this as the exact author string? If this comes up a lot, might be worth adding a hook that transform this into a single-field author.

ARTnews.js
zuphilip: Maybe, before that just try out whether `ZU.strToISO(...)` is giving us already a nice formatted date.

@adam3smith I guess, that one of `$2` should be `$1`.

adam3smith: yes, thanks fixed, and good point on strToISO

ARTnews.js
owcz: Not often in my usage, but upon search, looks [often enough](https://duckduckgo.com/?q=site%3Aartnews.com+%22the+editors+of+artnews%22&atb=v16&ia=web). Is there a way to do what you mention within FW or does the whole translator need to be written out?

ARTnews.js
owcz: `strToISO` can handle the string but drops the time ("03/02/17 10:40 pm" => "2017-03-02"). For what it's worth, Zotero did detect and export the date correctly as it was originally ingested. 

@zuphilip What's the best way to call this in FW? `Zotero.Utilities.strToISO(FW.Xpath('//span[@class="date-meta"]').text())` doesn't work.

zuphilip: The date without the time is normally enough for citations when used as a publication date. The time last modified or accessed a resource might be different.

I am not sure whether you can use this function in the FW.

adam3smith: yeah, you only have a list of specified functions available for FW, so this probably doesn't work.
It _should_ be available in a hook, though:
https://www.zotero.org/support/dev/translators/framework#post-processing_with_hooks

ARTnews.js
adam3smith: I know I'm saying this everywhere, but in FW, hooks are the only way to go here:
https://www.zotero.org/support/dev/translators/framework#post-processing_with_hooks

ARTnews.js
owcz: @zuphilip I can call strToISO (`Zotero.Utilities.strToISO("03/02/17 10:40 pm")` works fine) but I don't know if there's a trick to calling it with FW.Xpath

ARTnews.js
owcz: Aha! Just what I wanted. Thanks

owcz: @adam3smith I'm stumped on how to handle single-field authors instead of lastName/firstName and can't find a solution in the documentation. I'm currently shoving the full name into either lastName or firstName. Is there a single-field parameter name I should be using?

adam3smith: I got an e-mail with a question about single field but can't find your post. Did you figure that out or do you still need an answer?

ARTnews.js
```diff
@@ -0,0 +1,155 @@
...
+  type: "text/html" }],
+creators         : FW.Xpath('//span[@class="author-meta-name"]/a').text().cleanAuthor("author"),
+date             : FW.Xpath('//span[@class="date-meta"]').text().replace(/(\d{2})\/(\d{2})\/(\d{2})/, "20$3-$1-$2,"), // "03/02/16 10:35 am" => 2017-03-02, 10:35 pm
+blogTitle        : "ARTnews",
+language         : "en-US",
+abstractNote     : FW.Xpath('//meta[@name="description"]/@content').text(),
+tags             : FW.Xpath('//span[@class="cat-links"]/a').text(),
+hooks : { "scraperDone": function  (item,doc,url) {
+	if (item.creators[0].lastName == "ARTnews") {
+		item.creators[0].lastName = "The Editors of ARTnews";
```
adam3smith: you need to remove the `firstName` here still (I think you can just do that with `delete` but I always get confused about removing items from objects.

ARTnews.js
adam3smith: as in you can't log in with edit access or something else? There was so much vandalism, that Zotero devs need to give you access individually (I thought they did that for you?). Post to the forums with a link to this post.

owcz: I log in but don't have write access again—I'll work it out with Dan but just wonder if it wouldn't be easier with another (more accessible) platform

ARTnews.js
owcz: Done

adam3smith:
Great, thanks!

PR #1260, opened 2017-03-12
Add translation-server support to Twitter.js

owcz:
Per [request from Wikimedia Citoid](https://phabricator.wikimedia.org/T78785), [turn on](https://www.zotero.org/support/dev/translators) `translation-server` support for Twitter.

Does this need any kind of specific testing?


zuphilip:
The current tests with the translation server can be seen here: http://zotero-translator-tests.s3-website-us-east-1.amazonaws.com/testTranslators.html#date=2017-03-12&browser=v&version=4.0.SOURCE

And for Twitter it already indicates, that this should work, but is not activated. Thus, your PR looks good in general 👍 . Moreover, you will use it in Citoid and therefore also see whether there is any problem.

Can you update the test for twitter translator? It looks that date information was improved by some changes....

PR #1261, opened 2017-03-13
Add Destructoid.js

owcz:
Translator for http://destructoid.com, a [prominent](https://en.wikipedia.org/wiki/Wikipedia:WikiProject_Video_games/Sources) video game website

The scraper and multis all work fine and are good to go, but I'm having an odd issue with the test examples. When I run any page with Scaffold's `doWeb`, it is translated just fine with no issue, but only when run as an example test, I cannot use `FW.Xpath` in the `title` field, or else Scaffold throws an error about how no title was provided during the scrape (note: it compiles fine if FW.Xpath is replaced with a dummy, static string, see original commit below). The test examples are correct as written, but I'd be curious if they run without issue on other systems (i.e., if the test issue is specific to my setup).


owcz:
removed FW and ready for review
added support for multiples too (HP, reviews cat) but not search, which isn't even loading for me

owcz:
Site is using LD-JSON but doesn't include author in any standard format, so would need a dedicated translator

PR #1263, opened 2017-03-13
Add Kinja.js

owcz:
Translator for http://kotaku.com, a [prominent](https://en.wikipedia.org/wiki/Wikipedia:WikiProject_Video_games/Sources) video game website

Single-page scraper for regular posts, multis for tag pages, search pages, and main page

EM is mostly okay for the single-page scrape, but doesn't get its date format (`//time/@datetime`) and uses the sub-branding for some pages rather than the site name


owcz:
* ready for review
* removed FW
* relies on EM but adds date import
* added support for all Kinja sites (tech behind formerly Gawker sites): see Target in header
* added multi support for homepage, categories, search

owcz:
does the translator need a new unique ID if it's replacing another?

owcz:
* Site now uses LD-JSON
* Everything but the `multiple` can be handled in EM once #2765 is merged

PR #1265, opened 2017-03-13
Add GameZone.js

owcz:
Translator for http://gamezone.com, a [prominent](https://en.wikipedia.org/wiki/Wikipedia:WikiProject_Video_games/Sources) video game website

Single-page scraper for regular news and reviews posts, multis for game and search pages


owcz:
[EM](https://github.com/zotero/translators/blob/master/Embedded%20Metadata.js) appears to do a sufficient job of this now
not very popular site (actually was thought to have closed this year) so not worth doing a multi just yet

PR #1266, opened 2017-03-13
Add Paste.js

owcz:
Translator for http://paste.com, [a music and entertainment magazine](https://en.wikipedia.org/wiki/Paste_(magazine))

Single-page scraper for regular posts, multis for search and tag pages

### DIFF
--- a/Paste.js
+++ b/Paste.js
@@ -1,0 +1,149 @@
+{
+	"translatorID": "af48bcb9-fae7-463d-a751-dff7b8589336",
+	"label": "Paste",
+	"creator": "czar",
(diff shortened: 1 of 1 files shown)


owcz:
ready for review
* removed [Framework](https://www.zotero.org/support/dev/translators/framework)
* added multi support for homepage, categories, search

print suspended in 2010, so using "blog" rather than "magazine" type 
https://en.wikipedia.org/wiki/Paste_(magazine)

owcz:
Site now uses LD-JSON, so everything but the `multiple` will be handled with adding basic LD-JSON to EM (#2765)

PR #1268, opened 2017-03-14
Add OkayAfrica.js

owcz:
Translator for http://okayafrica.com, blog about African arts and culture


owcz:
ready for review
* removed [Framework](https://www.zotero.org/support/dev/translators/framework) per #1277
* multi support for homepage, categories, search

owcz:
* Site now uses LD-JSON
* Everything but the `multiple` can be handled in EM once #2765 is merged

PR #1271, opened 2017-03-14
Add Artforum.js

owcz:
Translator for art news website http://www.artforum.com. Covers major pages and a multi for search. Note that some pages are undated and others have no author, hence their absence in the test results.

### DIFF
--- a/Artforum.js
+++ b/Artforum.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2018-07-22 18:06:54"
+	"lastUpdated": "2021-09-02 00:33:38"
 }
 
 /*
@@ -38,12 +38,16 @@
 
 
 function detectWeb(doc, url) {
-	if (/-\d{5,}$/.test(url)) {
-		if (doc.querySelector('h3.print-article__issue-title')) return "magazineArticle";
+	if (/-\d{5,}([?#].*)?$/.test(url)) {
+		if (doc.querySelector('h3.print-article__issue-title')) {
+			return "magazineArticle";
+		}
 		return "blogPost";
-	} else if (getSearchResults(doc, true)) {
+	}
+	else if (getSearchResults(doc, true)) {
 		return "multiple";
 	}
+	return false;
 }
 
 
@@ -53,12 +57,17 @@ function scrape(doc, url) {
 	translator.setDocument(doc);
 	
 	translator.setHandler('itemDone', function (obj, item) { // corrections to EM
-		item.itemType = "blogPost";
 		item.publicationTitle = "Artforum";
-		var json_ld = doc.querySelector('script[type="application/ld+json"]');
-		if (json_ld) {
-			json_ld = json_ld.textContent;
-			item.date = json_ld.match(/"datePublished"\s?:\s?"([^"]*)"/)[1].split(":")[0];
+		item.language = 'en-US';
+		var jsonLD = doc.querySelector('script[type="application/ld+json"]');
+		if (jsonLD) {
+			jsonLD = JSON.parse(jsonLD.textContent);
+			item.title = jsonLD.name;
+			item.date = jsonLD.dateModified || jsonLD.datePublished;
+			
+			if (!item.creators.length && jsonLD.author) {
+				item.creators.push(ZU.cleanAuthor(jsonLD.author.name, 'author'));
+			}
 		}
 		var authorMetadata = doc.querySelectorAll('.contrib-link a');
 		for (let author of authorMetadata) {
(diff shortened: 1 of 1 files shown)


owcz:
## ready for review
* removed [Framework](https://www.zotero.org/support/dev/translators/framework) per #1277
* multi support for homepage, many section pages, search
* open to ideas on how to pull the magazine's volume/issue numbers from their subpages

Artforum.js
```diff
@@ -0,0 +1,356 @@
...
+		for (let author of authorMetadata) {
+			item.creators.push(ZU.cleanAuthor(author.text, "author"));
+		}
+		if (url.includes('/print/')) {
+			item.itemType = "magazineArticle";
+			item.ISSN = "0004-3532";
+			var issueDate = doc.querySelector('h3.print-article__issue-title');
+			if (issueDate) {
+				item.date = issueDate.textContent.trim().replace('PRINT ','');
+				/* are you feeling adventurous? could follow link to issue's page and scrape the volume/issue numbers...
```
adam3smith: If you want to do this, I think the way to go is to reconstruct from the issue description: every year has ten issues from September to May, then a Summer issue, so getting the issue is easy. For the volume, you'd have to do something like 
take the year, subtract 1962, except for September, October, November December, where you subtract 1961. Making the extra call seems tedious, though also doable.

Artforum.js
```diff
@@ -0,0 +1,356 @@
...
+		case "multiple":
+			Zotero.selectItems(getSearchResults(doc, false), function (items) {
+				if (!items) {
+					return true;
+				}
+				var articles = [];
+				for (var i in items) {
+					articles.push(i);
+				}
+				Z.debug(articles);
```
adam3smith: remove

Artforum.js
```diff
@@ -0,0 +1,356 @@
...
+				ZU.processDocuments(articles, scrape);
+			});
+			break;
+		case "blogPost":
```
adam3smith: use default: here to also capture magazine articles once you update above

Artforum.js
```diff
@@ -0,0 +1,356 @@
...
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+
+function detectWeb(doc, url) {
+	if (/-\d{5,}$/.test(url)) {
```
adam3smith: You should be able to detect `magazineArticle` by checking for the "Print" thing

Artforum.js
```diff
@@ -0,0 +1,356 @@
...
+	});
+}
+
+
+function getSearchResults(doc, checkOnly) {
+	var items = {};
+	var found = false;
+	// 1st for search page, 2nd for issue ToC, 3rd/4th/5th for sections, 5th+ for homepage
+	var rows = doc.querySelectorAll('h1.results-list__h1, .toc-article__title, .news-list h1, .reviews-list h1, .article-list h1, p.hp-singlefeature-author__writer, h3.hp-news__title, h3.hp-twocolumn__title a, h3.hp-artguide__title, p.hp-bloglist__teaser a');
+	var links = doc.querySelectorAll('h1.results-list__h1 a, .toc-article__title a, .news-list h1 a, .reviews-list h1 a, .article-list h1 a, h3.hp-singlefeature-author__title > a, a.hp-news__article, h3.hp-twocolumn__title a, .hp-artguide div.image-container > a, p.hp-bloglist__teaser a');
```
owcz: is this selector list too expensive? if preferable, i could break the list into subsets and first check against the page URL?

only downside is that this site has many eligible multi pages and i don't want to end up missing one

Artforum.js
owcz: i thought about the former but seemed like more of a hack and prone to potential error if the publication cycle was altered in a year

looking for a model: any recommendations for translators that call through to another page?

adam3smith: This isn't a perfect example but basically what you'd do is along the lines of https://github.com/zotero/translators/blob/master/Stanford%20Encyclopedia%20of%20Philosophy.js#L71 
You scrape all the info you need from the current page, then do a `ZU.doGet` or (more likely here) `ZU.processDocuments` for the new page add the missing information to the item and (and this is important) finalize the item within that routine, i.e. `item.complete()` would be within the `ZU.processDocuments`.

owcz: I couldn't perform querySelector/xpath on the doc so had to go with regex—not sure if that's slower

owcz:
@adam3smith this should be ready for review (as are [the rest of these](https://github.com/zotero/translators/pulls?utf8=%E2%9C%93&q=is%3Aopen+is%3Apr+author%3Aowcz+NOT+ebscohost+))

PR #1274, opened 2017-03-15
Add GamesRadar.js

owcz:
Translator for video game news and reviews site, also the online home of Future Publishing's print magazines. Covers main article types and group pages (search, categories, print mag pages).


owcz:
Site now uses LD-JSON, so everything but the `multiple` will be handled with adding basic LD-JSON to EM (#2765)

PR #1282, opened 2017-03-18
Add The Economic Times.js

sonali0901:
I am learning to write a translator. The following scraping is working fine in Scaffold and the translation is successful. I want to ask how the tests are automatically created by Scaffold?
```
{
	"translatorID": "6ec8008d-b206-4a4c-8d0a-8ef33807703b",
	.....
	"browserSupport": "gcsibv",
	"lastUpdated": "2014-04-04 09:55:12"
}
```
I believe, the above section and the test cases that we need to write for each translator are generated by Scaffold and not coded by us. Can anyone explain how to do that?

### DIFF
--- a/The Economic Times.js
+++ b/The Economic Times.js
@@ -2,7 +2,7 @@
 	"translatorID": "1a9a7ecf-01e9-4d5d-aa19-a7aa4010da83",
 	"label": "The Economic Times",
 	"creator": "Sonali Gupta",
-	"target": "http://economictimes.indiatimes.com",
+	"target": "^https?://economictimes\\.indiatimes\\.com",
 	"minVersion": "3.0",
 	"maxVersion": "",
 	"priority": 100,
@@ -36,7 +36,7 @@
 */
 
 function detectWeb(doc, url) {
-	if (url.indexOf("/topic/") != -1) {
+	if (url.indexOf("/topic/") != -1 && getSearchResults(doc, true)) {
 		return "multiple";
 	}
 	else if (ZU.xpathText(doc, '//article')) {
@@ -89,8 +89,7 @@ function scrape(doc, url) {
 	newItem.title = title;
 
 	//get abstract
-	var abstract = ZU.xpathText(doc, '//meta[@property="og:description"]/@content');
-	newItem.abstractNote = abstract;
+	newItem.abstractNote = ZU.xpathText(doc, '//meta[@property="og:description"]/@content');
 	
 	//get date
 	var xpathdate_author = '//article/div/div[contains(@class, "publish_on") or contains(@class, "byline")]';
@@ -110,7 +109,7 @@ function scrape(doc, url) {
 			authors_org=authors.substring(0,authors.lastIndexOf("|")-1);
 			var regex = /(.*By\s+)(.*)/;
 			authors = authors_org.replace(regex, "$2");
-			newItem.creators.push({lastName:authors,  creatorType: "author",fieldMode: 1})
+			newItem.creators.push({lastName:authors, creatorType: "author", fieldMode: 1})
 		}
 	}
 	
--- a/The Economic Times.js
+++ b/The Economic Times.js
@@ -1,5 +1,21 @@
+{
+	"translatorID": "1a9a7ecf-01e9-4d5d-aa19-a7aa4010da83",
+	"label": "The Economic Times",
+	"creator": "Sonali Gupta",
+	"target": "http://economictimes.indiatimes.com",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2017-03-19 08:42:35"
+}
+
 /*
 	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2017 Sonali Gupta
 	
 	This file is part of Zotero.
 
@@ -20,7 +36,7 @@
 */
 
 function detectWeb(doc, url) {
-	if (doc.location.href.indexOf("/topic/") != -1) {
+	if (url.indexOf("/topic/") != -1) {
 		return "multiple";
 	}
 	else if (ZU.xpathText(doc, '//article')) {
@@ -29,21 +45,17 @@ function detectWeb(doc, url) {
 }
 
 function doWeb(doc, url) {
-	var namespace = doc.documentElement.namespaceURI;
-	var nsResolver = namespace ? function(prefix) {
-	if (prefix == 'x') return namespace; else return null;
-	} : null; 
 
 	//srcaping titles of search results
 	if (detectWeb(doc, url) == "multiple") {
 		var myXPath = '//main/descendant::h3';
-		var myXPathObject = doc.evaluate(myXPath, doc, nsResolver, XPathResult.ANY_TYPE, null); 
-		var items = {};
-		var headers;
+		var myXPathObject = ZU.xpathText(doc, '//main/div[1]/div[2]/a/h2',',','*');
+		myXPathObject = myXPathObject.concat(ZU.xpathText(doc, myXPath,',','*'));
+		var array = myXPathObject.split('*')
+		var items = new Object();
 		var articles = new Array();
-		var k=0;
-		while (headers = myXPathObject.iterateNext()) {
-			items[k++] = headers.textContent;
+		for (var i in array){
+			items[i]=array[i];
 		}
 		Zotero.selectItems(items, function (items) {
 			if (!items) {
@@ -53,7 +65,7 @@ function doWeb(doc, url) {
 				Zotero.debug(i);
 				articles.push(i);
 			}
-			Zotero.Utilities.processDocuments(articles, getEdition);
+			Zotero.Utilities.processDocuments(articles, scrape);
 		});
 	}	
 	else{
@@ -62,8 +74,9 @@ function doWeb(doc, url) {
 }
 
 function scrape(doc, url) {
+	
 	newItem = new Zotero.Item("newspaperArticle");
-	newItem.url = doc.location.href;
+	newItem.url = url;
 	newItem.publicationTitle = "The Economic Times";
 
 	//get headline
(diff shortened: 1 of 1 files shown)


adam3smith:
Which documentation are you using for learning? A lot of the things you're doing in the translator can be done much simpler with current Zotero code.

As for Scaffold, the header is written automatically. You should just save the translator from Scaffold and then open it from it's location on the harddisk, i.e. in the translators folder in the Zotero data director.

The translator tests are created automatically when you click on "New Web" in the "Testing" pane in translator while looking at a relevant page. You'll then want to click "Save" again in Testing once you have all tests and make sure to save your translator (using the "save" icon at the top of Scaffold).

sonali0901:
I went through the HWZT guide http://niche-canada.org/member-projects/zotero-guide/chapter1.html
I took help from other pre-written translators in the same category.
I think I'll learn as I explore, thanks for the quick reply :+1:

adam3smith:
yeah, HWZT is almost 10 years old, pretty much everything it says related to code is outdated, so we'll ask you to redo a lot of the things that you base on that. Do check recently updated translators, @zuphilip 's common code blocks (and mistakes): https://github.com/zuphilip/translators/wiki/Common-code-blocks-for-translators as well as the translator coding documentation at https://www.zotero.org/support/dev/translators/coding

The Economic Times.js
```diff
@@ -0,0 +1,105 @@
...
+	if (doc.location.href.indexOf("/topic/") != -1) {
+		return "multiple";
+	}
+	else if (ZU.xpathText(doc, '//article')) {
+		return "newspaperArticle";
+	}
+}
+
+function doWeb(doc, url) {
+	var namespace = doc.documentElement.namespaceURI;
```
adam3smith: e.g. you don't need this for any regular translator

The Economic Times.js
```diff
@@ -0,0 +1,105 @@
...
+	var nsResolver = namespace ? function(prefix) {
+	if (prefix == 'x') return namespace; else return null;
+	} : null; 
+
+	//srcaping titles of search results
+	if (detectWeb(doc, url) == "multiple") {
+		var myXPath = '//main/descendant::h3';
+		var myXPathObject = doc.evaluate(myXPath, doc, nsResolver, XPathResult.ANY_TYPE, null);
```
adam3smith: doing this type of thing with `ZU.xpath` is much easier

The Economic Times.js
```diff
@@ -0,0 +1,105 @@
...
+		});
+	}	
+	else{
+		scrape(doc, url);
+	}
+}
+
+function scrape(doc, url) {
+	newItem = new Zotero.Item("newspaperArticle");
+	newItem.url = doc.location.href;
```
adam3smith: don't use this if you have the URL from the `url` variable

sonali0901: Understood.

sonali0901:
The test creation for  search pages is failing in Scaffold, reporting that the date is null. When each search result is passed for scraping through `Zotero.Utilities.processDocuments(articles, scrape);` , why is it showing date null? When a single article(through its url) is scraped the tests run fine. Any idea?

zuphilip:
I guess you have to restrict the search results to articles only. Currently I also find "slidesshows" e.g. http://economictimes.indiatimes.com/slideshows/nation-world/many-memorable-firsts-in-2017-up-assembly-elections/many-firsts/slideshow/57577668.cms which will give you an error.

The Economic Times.js
```diff
@@ -0,0 +1,176 @@
...
+}
+
+function doWeb(doc, url) {
+
+	//srcaping titles of search results
+	if (detectWeb(doc, url) == "multiple") {
+		var myXPath = '//main/descendant::h3';
+		var myXPathObject = ZU.xpathText(doc, '//main/div[1]/div[2]/a/h2',',','*');
+		myXPathObject = myXPathObject.concat(ZU.xpathText(doc, myXPath,',','*'));
+		var array = myXPathObject.split('*')
```
zuphilip: Try something like this instead:
```js
	var items = {};
	var found = false;
	//TODO: adjust the xpath
	var rows = ZU.xpath(doc, '//main//a[(h2 or h3) and contains(@href, "/articleshow")]');
	for (var i=0; i<rows.length; i++) {
		//TODO: check and maybe adjust
		var href = rows[i].href;
		//TODO: check and maybe adjust
		var title = ZU.trimInternal(rows[i].textContent);
		if (!href || !title) continue;
		if (checkOnly) return true;
		found = true;
		items[href] = title;
	}
	return found ? items : false;
}


function doWeb(doc, url) {
		Zotero.selectItems(getSearchResults(doc, false), function (items) {
			if (!items) {
				return true;
			}
			var articles = [];
			for (var i in items) {
				articles.push(i);
			}
		});
	} else {
		scrape(doc, url);
	}
}
```
This is some stable code which is used in a lot of other translators as well. Check the `TODO`s and adjust them possibly.

The Economic Times.js
```diff
@@ -0,0 +1,176 @@
...
+	//get abstract
+	var abstract = ZU.xpathText(doc, '//meta[@property="og:description"]/@content');
+	newItem.abstractNote = abstract;
+	
+	//get date and day
+	var date = ZU.xpathText(doc, '//article/div[4]/div[1]');
+	if(!date) date = ZU.xpathText(doc, '//article/div[1]/div[2]');
+	date = date.substring(date.indexOf("|") + 1);
+	var date_day = date.replace('Updated:','');
+	newItem.date = date_day;
```
zuphilip: This looks very fragile, because it depends on the exact order of the nodes. Usually it is better to use some attributes to filter on the correct elements (especially `class`). Moreover, if the variables like `date` can still be `null`, we should test that before continue.

I suggest here something like:
```js
	var date = ZU.xpathText(doc, '//article/div/div[contains(@class, "publish_on") or contains(@class, "byline")]');
	if (date) {
		newItem.date = ZU.strToISO(date);
	}
```
Note: The `ZU.strToISO` is deleting most of the other noise around the date and it seems already to work good for a few examples I tested.

owcz:
> HWZT is almost 10 years old, pretty much everything it says related to code is outdated

@adam3smith Should we remove it from the wiki then?

avram:
The original HWZT isn't on the wiki, so we'd have to ask Adam Crymble to take it down or put a warning on the page. There is a version of it on the wiki (https://www.zotero.org/support/dev/how_to_write_a_zotero_translator_plusplus); that version is also outdated and would need major work to be a fully appropriate introduction to translator writing.

The Economic Times.js
```diff
@@ -0,0 +1,192 @@
+{
+	"translatorID": "1a9a7ecf-01e9-4d5d-aa19-a7aa4010da83",
+	"label": "The Economic Times",
+	"creator": "Sonali Gupta",
+	"target": "http://economictimes.indiatimes.com",
```
zuphilip: This is a regular expression and we can therefore be more strict about the beginning, but already now allow possible `https` urls, and escape the points, i.e.
```diff
-"target": "http://economictimes.indiatimes.com",
+"target": "^https?://economictimes\\.indiatimes\\.com",
```
Note: In Scaffold you only need to escape them once, i.e. `\.`.

The Economic Times.js
```diff
@@ -0,0 +1,192 @@
...
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+function detectWeb(doc, url) {
+	if (url.indexOf("/topic/") != -1) {
```
zuphilip: We should also check here that there are any results for multiples, i.e.
```js
if (url.indexOf("/topic/") != -1 && getSearchResults(doc, true)) {
```

The Economic Times.js
```diff
@@ -0,0 +1,192 @@
...
+	newItem.publicationTitle = "The Economic Times";
+
+	//get headline
+	var title = ZU.xpathText(doc, '//article/h1');
+	if (!title) title = ZU.xpathText(doc, '//meta[@property="og:title"]/@content');
+	newItem.title = title;
+
+	//get abstract
+	var abstract = ZU.xpathText(doc, '//meta[@property="og:description"]/@content');
+	newItem.abstractNote = abstract;
```
zuphilip: There is no need for an extra variable `abstract` here and you can write these two lines in one line.

The Economic Times.js
```diff
@@ -0,0 +1,192 @@
...
+	if(!authors.length){
+		authors = ZU.xpathText(doc, xpathdate_author);
+		if(authors){
+			authors_org=authors.substring(0,authors.lastIndexOf("|")-1);
+			var regex = /(.*By\s+)(.*)/;
+			authors = authors_org.replace(regex, "$2");
+			newItem.creators.push({lastName:authors,  creatorType: "author",fieldMode: 1})
+		}
+	}
+
```
zuphilip: Is the language always English? Then you can add this as a constant value. Maybe also an ISSN exists?

The Economic Times.js
```diff
@@ -0,0 +1,192 @@
...
+	for (var i in authors){
+		newItem.creators.push(ZU.cleanAuthor(authors[i].textContent, "author"));
+	}
+	if(!authors.length){
+		authors = ZU.xpathText(doc, xpathdate_author);
+		if(authors){
+			authors_org=authors.substring(0,authors.lastIndexOf("|")-1);
+			var regex = /(.*By\s+)(.*)/;
+			authors = authors_org.replace(regex, "$2");
+			newItem.creators.push({lastName:authors,  creatorType: "author",fieldMode: 1})
```
zuphilip: Add space before `fieldMode`, delete one of the two spaces before `creatorType`.

The Economic Times.js
sonali0901: Yes it is always English. ISSN doesn't exist. I will make the desired changes.

The Economic Times.js
sonali0901: I thought since this section is generated automatically, so no changes were to be made. Thanks for letting me know.

zuphilip: I don't think you can generate the target regexp automatically...

zuphilip: Let me say it better: You input the information on the first tab in Saffold and all these information will be automatically written into this part here.

sonali0901:
@zuphilip There is an option to choose from different languages, but that redirects to other URL. For example http://hindi.economictimes.indiatimes.com/ or http://gujarati.economictimes.indiatimes.com/
The HTML structure of these pages is also not similar. So I think keeping the language as English should be fine.

adam3smith:
Thanks you & sorry for the delay with this.

PR #1284, opened 2017-03-20
Change ISBN search URL

adam3smith:
make sure it detects for both individual and multiple search results. Possibly fixes, but definitely improves, https://forums.zotero.org/discussion/64880/isbn-problem

### DIFF
--- a/Open WorldCat.js
+++ b/Open WorldCat.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 12,
 	"browserSupport": "gcsbv",
-	"lastUpdated": "2017-01-03 11:14:23"
+	"lastUpdated": "2017-03-19 23:26:57"
 }
 
 /**
@@ -246,7 +246,6 @@ function doWeb(doc, url) {
 				oclcID = extractOCLCID(canonicalURL.href);
 			}
 		}
-		
 		if(!oclcID) throw new Error("WorldCat: Failed to extract OCLC ID from URL: " + url);
 		scrape([oclcID]);
 	}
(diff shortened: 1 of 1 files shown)


adam3smith:
@zuphilip if you could take a look when you have a little time?

Open WorldCat.js
```diff
@@ -341,7 +340,17 @@ function fetchIDs(isbns, ids, callback) {
 				} else {
 					Z.debug("Could not extract OCLC ID for ISBN " + isbn);
 				}
-			} else {
+			}
+			//but sometimes we have single items
+			else if (getFirstContextObj(doc)) {
```
zuphilip: The `getFirstContextObj` is AFAIK just checking for the presence of COinS data, but this is also available in search pages, e.g. http://www.worldcat.org/search?qt=worldcat_org_all&q=mannheim

Open WorldCat.js
```diff
@@ -341,7 +340,17 @@ function fetchIDs(isbns, ids, callback) {
...
+				var canonicalURL = ZU.xpath(doc, '/html/head/link[@rel="canonical"][1]')[0];
```
zuphilip: I would suggest:
```js
var canonicalURL = ZU.xpathText(doc, '/html/head/link[@rel="canonical"]/@href');
```
(It looks IMO not likely that a website has several canonical links.)

Open WorldCat.js
zuphilip: Isn't it here enough to check for the presence of this canonical url, i.e. something like
```js
} else {
   var canonicalURL = ZU.xpathText(doc, '/html/head/link[@rel="canonical"]/@href');
   if (canonicalURL) {
      oclcID = extractOCLCID(canonicalURL);
      if(!oclcID) throw new Error("WorldCat: Failed to extract OCLC ID from URL: " + url);
         scrape([oclcID]);
      }
   } else {
      Z.debug("No search results found for ISBN " + isbn);
   }
}
```
?

Open WorldCat.js
```diff
@@ -327,11 +326,11 @@ function fetchIDs(isbns, ids, callback) {
 	}
 	
 	var isbn = isbns.shift();
-	var url = "http://www.worldcat.org/search?qt=results_page&q=bn%3A"
+	var url = "http://www.worldcat.org/search?qt=results_page&q=isbn%3A"
```
zuphilip: I assume that you tested this change.

adam3smith: yes, that's tested. It behaves slightly

PR #1287, opened 2017-03-29
Create KitapYurdu.com.js

has-hus:
Translator for an online Turkish book store.

### DIFF
--- a/KitapYurdu.com.js
+++ b/KitapYurdu.com.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2017-03-29 20:52:49"
+	"lastUpdated": "2017-05-02 20:55:39"
 }
 
 /*
@@ -33,10 +33,6 @@
 	***** END LICENSE BLOCK *****
 */
 
-/*
-Known issues:
-- Turkish char problem: using ZU.capitalizeTitle with publisher names with "I" returns "i" instead of "ı".
-*/
 
 function detectWeb(doc, url) {
 	if (url.indexOf('/kitap/')>-1||url.indexOf('product_id')>-1) {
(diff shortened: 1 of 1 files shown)


adam3smith:
This looks quite neat, thanks. Could you add some tests as well as an (AGPL3) license?

KitapYurdu.com.js
```diff
@@ -0,0 +1,113 @@
+{
+	"translatorID": "d3f35d5a-55da-4e07-be7d-b4d2a821279f",
+	"label": "KitapYurdu.com",
+	"creator": "Hasan Huseyin DER",
+	"target": "^http?://www\\.kitapyurdu\\.com/",
```
zuphilip: This should start with `"^https?://"` to work for `http` as well as `https`.

KitapYurdu.com.js
```diff
@@ -0,0 +1,113 @@
...
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "g",
```
zuphilip: Activate here all browsers and platforms, normally they work as well.

KitapYurdu.com.js
```diff
@@ -0,0 +1,113 @@
...
+	
+	var edition = ZU.trimInternal(doc.querySelector('[itemprop=bookEdition]').textContent);
+	item.edition = edition.split('.')[0];
+	
+	var publisher = ZU.trimInternal(doc.querySelector('[itemprop=publisher]').textContent);
+	item.publisher = ZU.capitalizeTitle(publisher,true);
+	
+	item.date = ZU.trimInternal(doc.querySelector('[itemprop=datePublished]').textContent);
+	item.ISBN = ZU.trimInternal(doc.querySelector('[itemprop=isbn]').textContent);
+	item.numPages = ZU.trimInternal(doc.querySelector('[itemprop=numberOfPages]').textContent);
```
zuphilip: Did you try the Embedded Metadata translator as a starting point? It looks that there is some metadata available there.

has-hus:
Thanks for the inputs.

@zuphilip I assume something's missing in the site's html, Embedded Metadata translator did not extract book info, just gave webpage...

adam3smith:
Yeah, the site has almost no embedded metadata in the header, so the Zotero EM translator is of no help here. It does have schema.org microdata embedded, but we're not currently trying to get that in the embedded metadata translator, so I think this is the right approach.

KitapYurdu.com.js
```diff
@@ -0,0 +1,303 @@
...
+		item.creators.push(ZU.cleanAuthor(creator, "translator"));
+	}
+	
+	var editors = ZU.xpath(doc, '//tr[contains(., "Editor")]|//tr[contains(., "Derleyici")]');
+	for (var i=0; i<editors.length; i++) {
+		var creator = cleanCreatorTitles(editors[i].textContent);
+		item.creators.push(ZU.cleanAuthor(creator, "editor"));
+	}
+	
+	var edition = ZU.trimInternal(doc.querySelector('[itemprop=bookEdition]').textContent);
```
adam3smith: I'm a bit worried this is fragile to the absence of an edition. `ZU.trimInternal` and `textContent` will both throw an error when used on null. I'd do this in two steps:
```
var edition = doc.querySelector('[itemprop=bookEdition]');
if (edition) edition = ZU.trimInternal(edition.textContent);
```
same for all the following fields. Or you could write a function that does this and returns either the trimemd text or `null`. Setting an item element to null is fine.

has-hus: @adam3smith Thanks. 

I tried to update them as you suggested.

adam3smith:
great, thanks!

PR #1289, opened 2017-04-03
Handle Newsbeat for BBC

sonali0901:
This is with reference to issues reported at https://phabricator.wikimedia.org/T94663
I want to update BBC translator so that it handles Newsbeat.
I want help in writing regex for http://www.bbc.co.uk/newsbeat
Articles on Newsbeat can't be marked as newspaper articles and giving them type as blog posts also don't suit much. I want suggestions on what should I return in detectWeb.

### DIFF
--- a/BBC.js
+++ b/BBC.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2016-08-20 19:58:56"
+	"lastUpdated": "2017-04-02 07:33:14"
 }
 
 /*
@@ -112,6 +112,12 @@ function scrape(doc, url) {
 				item.date = date.toISOString();
 			} else {
 				item.date = ZU.xpathText(doc, '//meta[@property="rnews:datePublished"]/@content');
+				if(!item.date) {
+					item.date = ZU.xpathText(doc, '//p[@class="timestamp"]');
+					if (item.date) {
+						item.date = ZU.strToISO(item.date);
+					}
+				}
 			}
 		}
 		
@@ -132,6 +138,18 @@ function scrape(doc, url) {
 				}
 				item.creators.push(ZU.cleanAuthor(authors[i], "author"));
 			}
+		}
+		else
+		{
+			var authorString = ZU.xpathText(doc, '//p[@class="byline"]');
+			var title = ZU.xpathText(doc, '//em[@class="title"]');
+			if (authorString) {
+				authorString = authorString.replace(title, '').replace('By', '');
+				var authors = authorString.split('&');
+				for (var i=0; i<authors.length; i++) {
+					item.creators.push(ZU.cleanAuthor(authors[i], "author"));
+				}
+			}	
 		}
 		
 		item.language = "en-GB";
--- a/BBC.js
+++ b/BBC.js
@@ -55,28 +55,16 @@ function detectWeb(doc, url) {
 }
 
 function getSearchResults(doc, checkOnly) {
-	var namespace = doc.documentElement.namespaceURI;
-	var nsResolver = namespace ? function(prefix) {
-	if (prefix == 'x') return namespace; else return null;
-	} : null;
-
 	var items = {};
 	var found = false;
 	var rows = ZU.xpath(doc, '//a[h3[@class="title-link__title"]]');
 	//for NewsBeat
-	if(!rows.length)
-	{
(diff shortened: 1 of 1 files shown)


zuphilip:
AFAIS there is no issue or other container for the articles on newsbeat. Maybe, `blogPost` is then still the best option?

sonali0901:
I have updated this PR to handle newsbeat. Test cases added.

BBC.js
```diff
@@ -133,6 +141,20 @@ function scrape(doc, url) {
...
+			var authorString = ZU.xpathText(doc, '//p[@class="byline"]');
+			var title = ZU.xpathText(doc, '//em[@class="title"]');
+			if (authorString) {
+				authorString = authorString.replace(title, '').replace('By', '');
+				var authors = authorString.split('&');
+				for (var i=0; i<authors.length; i++) {
+					item.creators.push(ZU.cleanAuthor(authors[i], "author"));
+				}
+			}	
+		}
```
adam3smith: How about hard-coding the the blogname for newsbeat. 
```
if (url.indexOf("/newsbeat/article") != -1) {
  item.blogTitle = "BBC Newsbeat";
}
```

If that makes sense, please add&update test.

sonali0901:
@adam3smith @zuphilip Is there a field for blogPost similar to section field in newspaperArtcile where one can save the blog type. For Newsbeat, blogs are categorized in sections life Music, Entertainment, etc. It looks sensible to store such a category with a blog entry.
Is there any way to know what all fields are available for an itemType(like title, publication date, etc for an article) and the use of those fields? Conventionally, what information should be stored in extras, notes and so on?

adam3smith:
Full list of fields is here : https://aurimasv.github.io/z2csl/typeMap.xml
On my phone, more later

adam3smith:
For your other question -- there are no exact guidelines on what goes into which field for every item type, unfortunately. It's a ton to document and wouldn't always be clear. There is no section field for blogs and I'd argue it probably doesn't make sense. I've never seen blogs cited with a section, whereas that's quite common for newspaper citations. You could add that information as a tag, but not sure that makes sense here. 

I'd generally leave extra alone except for storing information to CSL variables that don't have a corresponding field in Zotero. E.g. you can store things like
`original-date: 1776` or for books or reports with DOIs
`DOI: 10.1234/434566` in the Extra field
You can be fairly liberal in the use of notes -- though try to use one, max 2 notes per item -- as long as you're adding information that's clearly valuable (e.g. a ToC for a book)

adam3smith:
@sonali0901 -- I think this is just waiting on an answer on my last question (and potentially a slight tweak). Everything else looks good, no?

sonali0901:
@adam3smith Yes, in addition to this, I noticed that I forgot to push the changes that handled search pages of Newsbeat and now those changes are lost. So need to fix that too. I was a bit occupied because of academics but will finalize this patch by 16th.

adam3smith:
no rush, just making sure you're not waiting on us.

sonali0901:
While working on this, I found out that the original translator is failing one of its test cases: http://www.bbc.com/news/world-africa-37066738

BBC.js
```diff
@@ -46,20 +46,37 @@ function detectWeb(doc, url) {
...
+	var namespace = doc.documentElement.namespaceURI;
+	var nsResolver = namespace ? function(prefix) {
+	if (prefix == 'x') return namespace; else return null;
+	} : null;
+
 	var items = {};
 	var found = false;
 	var rows = ZU.xpath(doc, '//a[h3[@class="title-link__title"]]');
-	for (var i=0; i<rows.length; i++) {
+	//for NewsBeat
```
adam3smith: do you have an example URL or could create a test here? I'm not quite happy with the code yet (would like to avoid the `ZU.xpath` together with the `doc.evaluate`) but I can't see where this applies exactly.

sonali0901: I did this because for newsbeat, rows[i].href wasn't catching the url. Rather there was only one key, baseURL(url of the search page) and no other url information. So to retrieve the url of a specific title, I used this evaluate method. The test case for this is the harry potter search in newbeat.

adam3smith: thanks, sorry, hadn't realized that test was a newsbeat search. I understand why you used the code; I just want to see if we can make this more robust.

adam3smith:
OK, have a look & see if this makes sense. Passes all tests and is a good bit simpler. With this we should be good to merge

sonali0901:
It's perfect. Cool how you did that. I unnecessarily wrote those lines of code.

adam3smith:
Thanks!

PR #1290, opened 2017-04-09
Update The Chronicle of Higher Education.js

Jmuccigr:
Add `www.` to start of target. Leaving it optional in case the site drops it in the future.


adam3smith:
thanks, looks good. I've seen your error report for this on Chrome, but don't see any obvious cause. I'm inclined to wait until the next version of the Chrome connector comes out.

PR #1291, opened 2017-04-14
Update The Globe and Mail.js

### LINKED ISSUE #162: The Globe and Mail - site updated, needs complete rewrite

sonali0901:
Fixes #162 
There is a limit on numbers of articles that I can access in a day with the free subscription policy (and I ran out of it)
Pending work: code for extracting author details in case an article is taken from other source needs to be rewritten.

### DIFF
--- a/The Globe and Mail.js
+++ b/The Globe and Mail.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2017-04-28 19:23:09"
+	"lastUpdated": "2017-04-29 17:01:57"
 }
 
 /*
@@ -128,10 +128,9 @@ function scrape(doc, url) {
 		newItem.notes.push( {note: note} );
 	}
 
-	newItem.attachments = ({
+	newItem.attachments.push({
 		document: doc,
 		title: "The Globe and Mail Snapshot",
-		mimeType: "text/html"
 	});
 	newItem.complete();
 }
@@ -173,27 +172,11 @@ var testCases = [
 				"libraryCatalog": "The Globe and Mail",
 				"section": "news",
 				"url": "http://www.theglobeandmail.com/news/toronto/doug-ford-says-hes-not-yet-sure-about-his-political-future/article21428180/",
-				"attachments": {
-					"document": {
-						"Globe and Mail columnist Gary Mason.": {},
-						"</caption><cutline_leadin>iStockphoto": {},
-						"Specialists on the floor of the New York Stock Exchange watch Gary Cohn, director of the White House National Economic Council, Wednesday, April 26.": {},
-						"location": {
-							"href": "http://www.theglobeandmail.com/news/toronto/doug-ford-says-hes-not-yet-sure-about-his-political-future/article21428180/",
-							"origin": "http://www.theglobeandmail.com",
-							"protocol": "http:",
-							"host": "www.theglobeandmail.com",
-							"hostname": "www.theglobeandmail.com",
-							"port": "",
-							"pathname": "/news/toronto/doug-ford-says-hes-not-yet-sure-about-his-political-future/article21428180/",
-							"search": "",
-							"hash": ""
-						},
-						"_html5shiv": 1
-					},
-					"title": "The Globe and Mail Snapshot",
-					"mimeType": "text/html"
-				},
+				"attachments": [
+					{
+						"title": "The Globe and Mail Snapshot"
+					}
+				],
 				"tags": [
 					"Christine Elliott",
 					"Doug Ford",
--- a/The Globe and Mail.js
+++ b/The Globe and Mail.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2017-04-22 13:49:53"
+	"lastUpdated": "2017-04-28 19:23:09"
 }
 
 /*
@@ -100,12 +100,12 @@ function scrape(doc, url) {
 
 	//get author
 	var authors = ZU.xpath(doc, '//meta[@itemprop="author"]/@content');
-	for (var i in authors){
+	for (var i = 0; i<authors.length; i++){
 		newItem.creators.push(ZU.cleanAuthor(authors[i].textContent, "author"));
 	}
 
 	var publishers = ZU.xpath(doc, '//article/header//div[@itemprop="publisher"]//p');
-	for (var i in publishers ){
+	for (var i = 0; i<publishers.length; i++){
 		newItem.creators.push( {
 		"lastName" : publishers[i].innerText,
 		"creatorType" : "contributor",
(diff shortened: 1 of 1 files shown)


mvolz:
Great! 

A little confused by the WIP in the subject- is this ready to merge or not? Generally "WIP" is a way to say "not yet ready to be merged or reviewed" on gerrit. On github it's generally not used because the pull request itself implicitly implies the commit is ready for review. 

Also a little more description in the body about what works now would be nice as per https://www.mediawiki.org/wiki/Gerrit/Commit_message_guidelines#Body (obviously this is not a gerrit commit but descriptive commit messages are good in any repository).

sonali0901:
@mvolz   I use Work In Progress(WIP) to signify that it is not ready to be merged yet(for this particular PR I need to fix author section) but can be reviewed so that improvements can be made simultaneously(code related and as you suggested, description)

zuphilip:
This looks like a completely rewritten translator code. Thus, you can replace the author field in the top with your name name and add the AGPL licencense block. Moreover, `minVersion` is by default now `3.0`.

The Globe and Mail.js
```diff
@@ -1,223 +1,159 @@
...
-	var namespace = doc.documentElement.namespaceURI;
-	var nsResolver = namespace ? function(prefix) {
-		if (prefix == 'x') return namespace; else return null;
-	} : null;
+	//get author or organization
+	//get author or organization
+	var authors = ZU.xpath(doc, '//meta[@itemprop="author"]/@content');
+	for (var i in authors){
+		newItem.creators.push(ZU.cleanAuthor(authors[i].textContent, "author"));
+		newItem.creators[i].publisher = ZU.xpathText(doc, '//article/header//div[@itemprop="publisher"]//p').replace("\n",'');
```
zuphilip: This looks wrong.

I guess this should be just
```
newItem.publisher = ZU.xpathText(doc, '//article/header//div[@itemprop="publisher"]//p');
```
and outside the loop. The `replace` is IMO not needed but otherwise you have to check first that this is not `null`.

The Globe and Mail.js
```diff
@@ -1,223 +1,159 @@
...
+	var title = ZU.xpathText(doc, '//article/h1');
+	if (!title) title = ZU.xpathText(doc, '//meta[@property="og:title"]/@content');
+	newItem.title = title;
+
+	//get abstract
+	newItem.abstractNote = ZU.xpathText(doc, '//meta[@property="og:description"]/@content');
 	
-}
+	//get date
+	var xpathdate_author = '//time[@itemprop="datePublished"]';
```
zuphilip: Rename this variable (it has nothing to do with the author here) or maybe just write these two lines together.

The Globe and Mail.js
```diff
@@ -1,223 +1,159 @@
...
+				"language": "en-ca",
+				"libraryCatalog": "The Globe and Mail",
+				"section": "news",
+				"url": "http://www.theglobeandmail.com/news/toronto/doug-ford-says-hes-not-yet-sure-about-his-political-future/article21428180/",
+				"attachments": {
+					"url": "http://www.theglobeandmail.com/news/toronto/doug-ford-says-hes-not-yet-sure-about-his-political-future/article21428180/",
+					"title": "The Globe and Mail Snapshot",
+					"mimeType": "text/html"
+				},
+				"tags": [],
```
zuphilip: Should be easy to add the tags from e.g.
```html
<meta name="keywords" content="Rob Ford;John Tory;Jim Flaherty;Ontario Progressive Conservative;Christine Elliott;Doug Ford;leadership;">
```

The Globe and Mail.js
sonali0901: I used publisher in the loop because for every author there is a publisher tag, in case there are multiple authors of an article, they may have different entries in publishers. For some articles, the publisher tag is Reuters where the author contains content as "Wire Service". Also, there are articles where the publisher tag has content "Special to The Globe and Mail". How to handle such cases?
I have updated the code as you suggested but the newItem.publisher is not visible in the output of the test cases( a '-' sign appears before it in Scaffold)

sonali0901:
There is a section titled "More Related to this Story" holding few urls at the bottom of each article.
Can those links be added in "see also"?

The Globe and Mail.js
zuphilip: @sonali0901 Did you consider this comment?

zuphilip:
No, related items are only Zotero items and have to be addressed by their id. It is not possible to add an url there. I would suggest a note for the more related to this story. A note can also contain HTML elements and therefore e.g. links.

The field `publisher` is not available for the itemType newspaperArticle and therefore cannot be used. I would suggest to use then contributor, which also can be multiple, i.e. something like (not tested):
```js
var publishers = ZU.xpathText(doc, '//article/header//div[@itemprop="publisher"]//p');
for (var i in publishers ){
   newItem.creators.push( {
      "lastName" : publishers[i].textContent, 
      "creatorType" : "contributor",
      "fieldmode" : 1
   );
}
```
Sorry, for the wrong suggestion previously.

The Globe and Mail.js
```diff
@@ -155,7 +171,11 @@ var testCases = [
 					"Rob Ford",
 					"leadership"
 				],
-				"notes": [],
+				"notes": [
+					"http://www.theglobeandmail.com/news/toronto/doug-ford-says-ontario-pc-leadership-bid-is-on-the-table/article21358827/",
+					" http://www.theglobeandmail.com/news/toronto/uniting-a-divided-toronto-will-be-a-key-task-for-torys-transition-team/article21359883/",
+					" http://www.theglobeandmail.com/news/toronto/a-chuckling-ford-on-his-mayoralty-it-will-definitely-be-remembered/article21414526/"
+				],
```
zuphilip: The `notes` contains an array of objects with a parameter `note` where the content is saved. Actually, I would suggest to just create one note with all urls. Compare for example the [translator for beck-online](https://github.com/zotero/translators/blob/master/beck-online.js).

The Globe and Mail.js
```diff
@@ -1,223 +1,189 @@
...
+	//get date
+	var xpathdate = '//time[@itemprop="datePublished"]';
+	var date = ZU.xpathText(doc, xpathdate);
+	if (date) {
+		newItem.date = ZU.strToISO(date);
+	}
+
+	//get author
+	var authors = ZU.xpath(doc, '//meta[@itemprop="author"]/@content');
+	for (var i in authors){
```
adam3smith: `ZU.xpath` return an array. You always want to loop through arrays sequentially, i.e. `for (var i = 0; i<authors.length; i++)` both to avoid odd behavior and for efficiency.

The Globe and Mail.js
```diff
@@ -1,223 +1,189 @@
...
+		newItem.creators.push(ZU.cleanAuthor(authors[i].textContent, "author"));
+	}
+
+	var publishers = ZU.xpath(doc, '//article/header//div[@itemprop="publisher"]//p');
+	for (var i in publishers ){
```
adam3smith: ditto here.

The Globe and Mail.js
```diff
@@ -1,223 +1,189 @@
...
+
+	var xpathtags = ZU.xpathText(doc, '//meta[@name="keywords"]/@content');
+	newItem.tags = xpathtags.split(";").filter(function(tag) {return tag.length != 0});
+
+	var related_links = ZU.xpathText(doc, '//article//li/p/a/@href');
+	if(related_links)
+		newItem.notes = related_links.split(",").filter(function(tag) {return tag.length != 0});
+
+	newItem.attachments = ({
+		url: url,
```
adam3smith: use `document: doc,` instead if you're just going to attach the page already open/in memory. That's both more efficient and may perform more reliably in edge cases.

sonali0901: I have applied this change, but I don't think it is providing the desired result. Kindly check the test case one to see the result. This test case didn't run successfully in Scaffold, rather returned the status that item 0 didn't match. Many fields in the attachments had a '-' sign before them. Can you explain or provide any resource to understand the use of this?

adam3smith: Documentation is here: https://www.zotero.org/support/dev/translators/coding#attachments (the last bit). 
But the test indeed doesn't look right. I have never seen that behavior in Scaffold, would have to check. What's the exact set-up you're running that on?

The Globe and Mail.js
```diff
@@ -1,223 +1,221 @@
...
+	if(related_links)
+		note = addNote(note, related_links.split(",").filter(function(tag) {return tag.length != 0}))
+	if (note.length != 0) {
+		newItem.notes.push( {note: note} );
+	}
+
+	newItem.attachments = ({
+		document: doc,
+		title: "The Globe and Mail Snapshot",
+		mimeType: "text/html"
```
adam3smith: try if this works better w/o mimeType -- I don't think it should matter, but you also don't need a mimeType when saving the current doc.

The Globe and Mail.js
zuphilip: The `attachments` are a list which is initiated with the empty list `[]` and you want to add another element, i.e. try `newItem.attachments.push({...})` and maybe double check with another translator code here. Does this help?

adam3smith: ah yes, that's probably it, good catch @zuphilip

sonali0901: I think this worked well.

adam3smith:
Thanks!

PR #1293, opened 2017-05-02
Update Library Catalog (PICA).js

symac:
ABES has been updating the cataloguing recommandations to FRBR/RDA and on some records, the display has been updated with new labels like : http://www.sudoc.fr/200278649 (Publication instead of Editeur(s) for example). The purpose of this patch is to fix the issue that happens actually which is that publisher information are not found by Zotero for these newly created/updated records.


adam3smith:
1) you'll need to bump the update date
2) could you at 1 or 2 tests for this?

symac:
Thanks for the input @adam3smith . I have bumped the date and added 1 test, hope it will be fine.

adam3smith:
Thanks!

PR #1299, opened 2017-05-18
Add saving from PDF pages for IACR

adam3smith:


ePrint IACR.js
```diff
@@ -108,7 +107,13 @@ function doWeb(doc, url) {
 			ZU.processDocuments(articles, scrape);
 		});
 	} else {
-		scrape(doc, url)
+		if (url.search(/\.pdf$/)!= -1) {
+			//go to the landing page to scrape
+			//we use http to prevent (somewhat mysterious) same-origin violations leading to permission denied errors
+			url = url.replace(/\.pdf$/, "").replace(/^https/, "http")
```
adam3smith: @zuphilip @adomasven -- testing this on 4.0.29 in Firefox -- I'm a little confused as to why I need to switch to http, but it's the only way I was able to make that work

adam3smith:
I don't think any harm can come from taking this. If you have any ideas on the https/http issue, still happy to look into that more.

ePrint IACR.js
zuphilip: I don't have an idea why `http` is different from `https` here. If this hack is working, then this fine for me for the moment.

Issue #1277, opened 2017-03-16
Deprecate/integrate FW


dstillman:
@owcz has been writing lots of new translators using [FW](https://gitlab.com/egh/zotero-transfw). That's great (thanks!), but before we go too much further with that, I wanted to get @adam3smith's and @zuphilip's thoughts on FW. I know @simonster was never really a fan of it, both because it's limited in what it can do and because he thought that for people who knew basic JS it might actually be harder than writing a translator without it, but you guys have worked with translators more recently, so we'd be interested in your perspective. If we do think there are benefits, though, I think we should consider finding a way to achieve those in the main architecture rather than having this large chunk of code duplicated into lots of different translators.

adam3smith:
@owcz is very insistent that FW translators are easier to maintain/write. That certainly used to be the case -- my first 10 translators were all FW -- but I'm curious (actually) why they still think that's the case. 

Now, though, with the `ZU.xpathText` function as well as @zuphilip 's [Common code blocks](https://github.com/zuphilip/translators/wiki/Common-code-blocks-for-translators), I find regular translators just as quick, including for scrapers, and they don't have any of the limitations (and ideosyncracies, which I assume Simon disliked), so I could certainly live without FW easily (and I don't use it anymore). 

In my view, we could just put up a regular translator skeleton for scrapes (along the line of what Philipp already has for other types) that emulates FW with regular js code. Using a good template also has the advantage of increasing readability/maintainability. I find the translators that follow Philipp's codeblocks particularly easy to work with.

Philipp will chime in himself, I'm sure, but I know he's never been a fan of FW.

owcz:
I agree with [the principle](https://gitlab.com/egh/zotero-transfw) of FW: removing boilerplate and making the translator process more approachable to those who may be put off by curly braces. I wouldn't say I'm insistent about it—I certainly see [its limitations](https://github.com/owcz/translators/wiki/Common-Framework-templates)—but in my estimation, it would be easier to train (and assist/review) contributors to focus on hunting XPaths for FW translators than it would be to familiarize each with JS and the Zotero libraries (especially without more attention to their documentation).

[...]

PR #1308, opened 2017-05-26
Kstudy translator

yunusong:
Translator for the Kstudy database.

Help appreciated for attachments, changing authors to a single field, and opening a javascript url at "journal/list_name.asp" detect. But any suggestions to improve the translator is welcome.

Thank you.

### DIFF
--- a/KStudy.js
+++ b/KStudy.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2017-05-27 14:00:17"
+	"lastUpdated": "2018-01-01 22:27:16"
 }
 
 /*
@@ -37,9 +37,9 @@
 */
 
 function detectWeb(doc, url) {
-	if (url.indexOf('/journal/thesis_name.asp')>-1) {
+	if (url.indexOf('/thesis/thesis-view.asp')>-1) {
 		return "journalArticle";
-	} else if ((url.indexOf('/journal/list_name.asp')>-1 || url.indexOf('/search/result_kiss.asp')>-1) && getSearchResults(doc, true)) {
+	} else if ((url.indexOf('/journal/journal-view.asp')>-1 || url.indexOf('/search/result_kiss.asp')>-1) && getSearchResults(doc, true)) {
 		return "multiple";
 	}
 }
@@ -47,14 +47,9 @@ function detectWeb(doc, url) {
 function getSearchResults(doc, checkOnly) {
 	var items = {};
 	var found = false;
-	var rows = ZU.xpath(doc, '//li[@class="title"]/a');
+	var rows = ZU.xpath(doc, '//div[contains(@class, "thesis-info")]/h5/a[contains(@href, "/thesis")]');
 	for (var i=0; i<rows.length; i++) {
 		var href = rows[i].href;
-		if (href.indexOf("javascript")>-1) {
-			var tnameAndKey = href.match(/detailView\(\'([\S]+)\'\,\'([\S]+)\'/);
-			var journalhref = "../journal/thesis_name.asp?tname=" + tnameAndKey[1] + "&key=" + tnameAndKey[2];
-			href = journalhref;
-		}
 		var title = ZU.trimInternal(rows[i].textContent);
 		if (!href || !title) continue;
 		if (checkOnly) return true;
@@ -84,52 +79,62 @@ function doWeb(doc, url) {
 
 function scrape(doc, url) {
 	var item = new Zotero.Item("journalArticle");
-	item.title = ZU.xpathText(doc, '//div[@class="agency_box"]/div[1]/span');
+	item.title = ZU.xpathText(doc, '//section[@class="pub-info"]/h3');
 	item.language = "ko-KR";
 	
-	var creators = ZU.xpathText(doc, '//div[@class="agency_detail"]/ul/li[1]');
+	var creators = ZU.xpathText(doc, '//div[@class="writers"]');
 	if (creators) {
 		var creatorsList = creators.split(',');
 		for (var i=0; i<creatorsList.length; i++) {
-			item.creators.push(ZU.cleanAuthor(creatorsList[i], "author", true));
+			item.creators.push({
+				lastName: creatorsList[i],
+				fieldMode: true,
+				creatorType: "author"
+			});
+		}
+	}
+
+	var container = ZU.xpathText(doc, '//li[label[text()="간행물"]]');
+	// e.g. container = 간행물 : 국제어문 54권0호
+	if (container && container.includes(':')) {
+		var containerValue = container.split(':')[1];
+		var containerParts = containerValue.match(/(.*)\s+(\d+)\D(\d+)/);
+		if (containerParts) {
+			item.publicationTitle = containerParts[1];
+			item.volume = containerParts[2];
+			item.issue = containerParts[3] !== '0' ? containerParts[3] : null;
 		}
 	}
 	
-	var details = ZU.xpathText(doc, '//div[@class="agency_detail"]/ul/li[2]');
-	//e.g. details = "한국공자학회, <공자학> 27권0호 (2014), pp.179-210"
-	//Z.debug(details);
-	if (details) {
-		item.date = ZU.strToISO(details);
-		var publicationTitle = details.match(/<([^>]+)>/);
-		if (publicationTitle) {
-			item.publicationTitle = publicationTitle[1];
-		}
-		var pagerange = details.match(/p+\.(\d+-\d+)/);
-		if (pagerange) {
-			item.pages = pagerange[1];
-		}
-		var volumeIssue = details.match(/>\s+(\d+)\D(\d+)/);
-		if (volumeIssue) {
-			item.volume = volumeIssue[1].replace(/^0/, '');
-			item.issue = volumeIssue[2].replace(/^0/, '');
-		}
+	var date = ZU.xpathText(doc, '//li[label[text()="발행년월"]]');
+	// e.g. date = 발행년월 : 2012년 04월
+	if (date && date.includes(':')) {
+		var dateValue = date.split(':')[1];
+		item.date = dateValue.trim().replace(/\D+/g, '-').replace(/-$/, '');
 	}
 	
-	item.abstractNote = ZU.xpathText(doc, '//div[@class="con_detail"]/div[4]/ul');
-	if (item.abstractNote) {
-		item.abstractNote = ZU.trimInternal(item.abstractNote);
+	var pages = ZU.xpathText(doc, '//li[label[text()="페이지"]]');
+	// e.g. pages = 페이지 : 43-93(51pages)
+	if (pages && pages.includes(':')) {
+		var pagesValue = pages.split(':')[1];
+		item.pages = pagesValue.split('(')[0].replace('-', '–');
+	}
+	
+	var abstract = ZU.xpathText(doc, '//comment()[contains(., "초록 보기")]/following-sibling::section[1]');
+	if (abstract) {
+		item.abstractNote = ZU.trimInternal(abstract);
 	}
 
-	var pdfJSurl = ZU.xpathText(doc, '//div[@class="search_box"]/div[@class="choice"]/span[1]/a/@href');
-  	var pdfurlKeys = pdfJSurl.match(/Select1\(\'(\S+)\'\,\s?\'(\S+)\'\,\s?(\S+)\,\s?(\S+)\,\s?(\S+)\)\;/);
-  	var pdfUrl = pdfurlKeys[1].replace(/^(.*\/).*/, "$1") + "viewer.asp?code_num=" + pdfurlKeys[2];
-
+	//var pdfJSurl = ZU.xpathText(doc, '//div[@class="search_box"]/div[@class="choice"]/span[1]/a/@href');
+  	//var pdfurlKeys = pdfJSurl.match(/Select1\(\'(\S+)\'\,\s?\'(\S+)\'\,\s?(\S+)\,\s?(\S+)\,\s?(\S+)\)\;/);
+  	//var pdfUrl = pdfurlKeys[1].replace(/^(.*\/).*/, "$1") + "viewer.asp?code_num=" + pdfurlKeys[2];
+/*
 	item.attachments.push({
 		url : pdfUrl,
 		title : "KStudy Full Text PDF",
 		mimeType : "application/pdf",
 	});	
-	
+*/
 	item.complete();
 
 }
@@ -138,7 +143,7 @@ function scrape(doc, url) {
 var testCases = [
 	{
 		"type": "web",
-		"url": "http://kiss.kstudy.com/journal/thesis_name.asp?tname=kiss2002&key=3297333",
+		"url": "http://kiss.kstudy.com/thesis/thesis-view.asp?key=3297333",
 		"items": [
 			{
 				"itemType": "journalArticle",
@@ -146,14 +151,15 @@ var testCases = [
 				"creators": [
 					{
 						"lastName": "조우진 ( Woo Jin Cho )",
+						"fieldMode": true,
 						"creatorType": "author"
 					}
 				],
-				"date": "2014",
-				"abstractNote": "초록 <한국어 초록> 본 논문의 목적은 소옹의 선천역학(先天易學)을 바탕으로 한 송대(宋代) 상수역(象數易)의 흐름을 살펴보고 왕부지의 비판적 입장을 고찰하는데 있다. 필자는 이러한 과정에서 상수역의 본래 모습을 확인하고자 한다. 왕부지는 자신의 역학체계를 바탕으로 선천과 후천의 개념, 선천의 전수과정, 선천도(先天圖)의 체계, 선천역학의 원리, 괘(卦)의 형성의 과정 등을 신랄하게 비판하였다. 그의 비판 논리는 경전의 내용을 바탕으로 하는 실증주의적 사고방식에 근거하고 있다. 왕부지의 입장에서 보자면 소옹의 선천역학과 관련된 이론이나 내용은 경전에 전혀 찾아볼 수 없는 것이며, 도가(道家)의 술수가들에 의해 전수된 것에 불과할 따름이다. 왕부지는 소옹의 선천역학을 비판하는 근거를 여러 가지로 제시하는데, 가장 결정적인 것은 점(占)과 관련된다. 선천역학의 핵심원리인 가일배법(加一倍法)은 아래로부터 위로 쌓아서 괘(卦)를 만들어가는 과정으로 점치는 것과 같은 것이다. 그래서 왕부지는 소옹의 선천역학을 술수학일 뿐만 아니라 점역(占易)에 치우친 것이라고 비판하면서 ‘점학일리(占學一理)’를 주장한다. <영어 초록> The object of this paper is to examine the current of Xiangshuyi(象數易) in Song dynasty(宋代) based on Shao-Yong(邵雍)’s Sunchenyeokhak(先天易學) and consider Wang-Fuzhi(王夫之)’s critical position. By this critical examination, we can ascertain the true features of Xiangshuyi. Wang-Fuzhi severely criticised the concept of Sunchen(先天) and Huchen(後天), the transmission process of Sunchen, the system of Sunchentu(先天圖), and the principle of Sunchenyeokhak, and the formation process of Gua(卦) through his Yeokhak(易學) system. His criticism is especially based on positivistic thinking on the contents of Jingzhuan(經傳). For Wang-Fuzhii, the content and theory concerning Shao-Yong’s Sunchenyeokhak aren’t found in Jingzhuan(經傳) and are nothing but brought by Zhushujia(術數家) in Daojia(道家). Wang-Fuzhi presents the several evidences of criticising Shao-Yong’s Xiangshuyi. Among them, the most decisive evidence is related to Zhan(占). Jiayibeifa(加一倍法) as the key principle of Sunchenyeokhak is a process of making Gua by stacking one on top of another and this process is the same as that of Zhan. In conclusion, Wang-Fuzhi maintains Zhanxueyili(占學一理) criticizing that Shao-Yong’s Sunchenyeokhak is only Zhushuxue and is biased toward Zhanyi.",
-				"language": "kr",
+				"date": "2014-11",
+				"abstractNote": "초록 보기 본 논문의 목적은 소옹의 선천역학(先天易學)을 바탕으로 한 송대(宋代) 상수역(象數易)의 흐름을 살펴보고 왕부지의 비판적 입장을 고찰하는데 있다. 필자는 이러한 과정에서 상수역의 본래 모습을 확인하고자 한다. 왕부지는 자신의 역학체계를 바탕으로 선천과 후천의 개념, 선천의 전수과정, 선천도(先天圖)의 체계, 선천역학의 원리, 괘(卦)의 형성의 과정 등을 신랄하게 비판하였다. 그의 비판 논리는 경전의 내용을 바탕으로 하는 실증주의적 사고방식에 근거하고 있다. 왕부지의 입장에서 보자면 소옹의 선천역학과 관련된 이론이나 내용은 경전에 전혀 찾아볼 수 없는 것이며, 도가(道家)의 술수가들에 의해 전수된 것에 불과할 따름이다. 왕부지는 소옹의 선천역학을 비판하는 근거를 여러 가지로 제시하는데, 가장 결정적인 것은 점(占)과 관련된다. 선천역학의 핵심원리인 가일배법(加一倍法)은 아래로부터 위로 쌓아서 괘(卦)를 만들어가는 과정으로 점치는 것과 같은 것이다. 그래서 왕부지는 소옹의 선천역학을 술수학일 뿐만 아니라 점역(占易)에 치우친 것이라고 비판하면서 ‘점학일리(占學一理)’를 주장한다. The object of this paper is to examine the current of Xiangshuyi(象數易) in Song dynasty(宋代) based on Shao-Yong(邵雍)’s Sunchenyeokhak(先天易學) and consider Wang-Fuzhi(王夫之)’s critical position. By this critical examination, we can ascertain the true features of Xiangshuyi. Wang-Fuzhi severely criticised the concept of Sunchen(先天) and Huchen(後天), the transmission process of Sunchen, the system of Sunchentu(先天圖), and the principle of Sunchenyeokhak, and the formation process of Gua(卦) through his Yeokhak(易學) system. His criticism is especially based on positivistic thinking on the contents of Jingzhuan(經傳). For Wang-Fuzhii, the content and theory concerning Shao-Yong’s Sunchenyeokhak aren’t found in Jingzhuan(經傳) and are nothing but brought by Zhushujia(術數家) in Daojia(道家). Wang-Fuzhi presents the several evidences of criticising Shao-Yong’s Xiangshuyi. Among them, the most decisive evidence is related to Zhan(占). Jiayibeifa(加一倍法) as the key principle of Sunchenyeokhak is a process of making Gua by stacking one on top of another and this process is the same as that of Zhan. In conclusion, Wang-Fuzhi maintains Zhanxueyili(占學一理) criticizing that Shao-Yong’s Sunchenyeokhak is only Zhushuxue and is biased toward Zhanyi.",
+				"language": "ko-KR",
 				"libraryCatalog": "KStudy",
-				"pages": "179-210",
+				"pages": "179–210",
 				"publicationTitle": "공자학",
 				"shortTitle": "투고논문",
 				"volume": "27",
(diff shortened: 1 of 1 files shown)


fbennett:
I've raised my hand, so ... a question first for the translator gurus listening in. The translator is based on the Framework, but it's going to need some out-of-the-ordinary post-processing of name field content, at least one intermediate xmlhttprequest call to acquire a ToC for one form of "multiple" content, and another to acquire attachments.

Is it possible to shoehorn arbitrary JS code into a Framework translator, or should things be recoded from scratch?

avram:
So FW does have some capacity to hook into the processing and run arbitrary JS, but I suspect it's better to rewrite than to do much complex work in those hooks

zuphilip:
> Is it possible to shoehorn arbitrary JS code into a Framework translator, or should things be recoded from scratch?

One can add more JS to the functions `detectWeb`, `doWeb` for example https://github.com/zotero/translators/blob/346e1647206ddcfaa1a7ee7adae4d772f38e91b9/Foreign%20Policy.js#L17-L23 ; or define a new function and call it somewhere in the FW chain.

Personally, I prefer the non-FW translators and here it seems that anything else would be pretty complex as well. @yunusong I can rewrite what you have as a normal translator without the FW, if this helps.

fbennett:
@zuphillip if you have time, and if @yunusong agrees, that would be great (Philip is much more skilled with translators than I am).

With a non-FW translator as a base, I could then splice in conditional operations that would wake up in Juris-M, for transliterated names and whatnot.

zuphilip:
@fbennett Here is a non-FW version of this code: https://github.com/zuphilip/translators/blob/yunusong-patch-2/KStudy.js

yunusong:
Thanks @zuphilip @fbennett !

Author field is not that big of a deal, actually. I tried adding ".remove(/\(/).remove(\)/)" after ".split(/,/)" so that it won't take ")" as the last name. Then it took the romanized last name as the last name, and that was good enough for me. Separating last name and first name is almost never necessary in Korean academia.

On the other hand, I don't mind you guys rewriting the code at all. If it helps improving the code, I am all for it. What I would really like is the translator to automatically fetch the pdf files. The xpath for link to pdf files is "//div[@class="choice"]/span[1]/a/@href" and an example link is 
"javascript:SiteSelect1('http://210.101.116.18/kiss10/inFTP_Journal.asp', '87300332.pdf', 2201, 25169, 1);"
I tried rewriting the code with the framework myself, but I didn't have any better luck with the attachments. If you can help me with the attachments, that would be wonderful. Thank you.

yunusong:
This looks great. You should replace the original file with this one on the pull request and also add your name to the creator. Thank you for this. 

    2017년5월27일 (토요일) 오전 10:02에 Philipp Zumstein <notifications@github.com>님이 쓴 메시지:
 

 @fbennett Here is a non-FW version of this code: https://github.com/zuphilip/translators/blob/yunusong-patch-2/KStudy.js—

zuphilip:
@yunusong I created a PR towards your fork and branch https://github.com/yunusong/translators/pull/1 . After accepting this, we should see it here as well.

KStudy.js
```diff
@@ -0,0 +1,163 @@
+{
+	"translatorID": "b298ca93-0010-48f5-97fb-e9923519a380",
+	"label": "KStudy",
+	"creator": "Yunwoo Song","Philipp Zumstein"
```
zuphilip: Okay, but this needs to be one string containing both names, i.e.
```
"Yunwoo Song, Philipp Zumstein"
```

yunusong: Okay

yunusong:
I located the JS file which contains the function for downloading pdf.
The function goes as
function SiteSelect1(ftproot, a_imag, inst_key, publ_key,isDownLoad) {
	var url = "../search/download.asp?ftproot=" + ftproot + "&inst_key=" + inst_key + "&a_imag=" + a_imag + "&isDownLoad=" + isDownLoad + "&publ_key=" + publ_key
	window.open(url, "", "toolbar=0,location=0,directories=0,status=0,menubar=0,scrollbars=1,resizable=1,copyhistory=1,top=170,left=250,width=540,height=500");
}

So I coded item.attachments as follows.

  	var pdfJSurl = ZU.xpathText(doc, '//div[@class="search_box"]/div[@class="choice"]/span[1]/a/@href');
  	var pdfurlKeys = pdfJSurl.match(/Select1\(\'(\S+)\'\,\s?\'(\S+)\'\,\s?(\S+)\,\s?(\S+)\,\s?(\S+)\)\;/);
  	var pdfurl = "../search/download.asp?ftproot=" + pdfurlKeys[1] + "&inst_key=" + pdfurlKeys[3] + "&a_imag=" + pdfurlKeys[2] + "&isDownLoad=" + pdfurlKeys[5] + "&publ_key=" + pdfurlKeys[4];
	item.attachments.push({
		title : item.title,
		url : pdfurl,
		mimeType : "application/pdf",
		snapshot : false
	})

But Zotero still won't fetch the pdf.

I tried fetching the resulting url by assigning to an unused slot. eg. issn. as
item.ISSN = pdfurl;
An example of a resulting url is as follows.
../search/download.asp?ftproot=http://210.101.116.17/kiss8/inFTP_Journal.asp&inst_key=7074&a_imag=40323420.pdf&isDownLoad=1&publ_key=25408

I tried pasting this address directly on a browser and it will start downloading the pdf.
So I am wondering what is still blocking Zotero from storing the pdf.

The odd thing I noticed about the original function was that it does not end the "var url =" statement with a semicolon. The semicolon only appears at the end of the second line. Would this have something to do with the problem? Thank you guys for the comments.

KStudy.js
```diff
@@ -1,7 +1,7 @@
 {
 	"translatorID": "b298ca93-0010-48f5-97fb-e9923519a380",
 	"label": "KStudy",
-	"creator": "Yunwoo Song","Philipp Zumstein"
+	"creator": "Yunwoo Song, Philipp Zumstein"
```
zuphilip: This is not valid JSON. You need to add a comma at the end of this line.

zuphilip:
Can you try `snapshot : true` in your code above?

fbennett:
I'll take a look on Monday (I need to be in the University net to get access). As an offhand thought in the meantime, assuming that the translator is loading properly, there may be a redirect that the translator is not able to follow without some further jiggery-pokery. (You can follow the transactions by opening Developer Tools in the browser and then refreshing on the download URL.)

yunusong:
Thank you @fbennett !
 @zuphilip I just tried "snapshot : true" but nothing changed. Zotero still shows a big red X for attachments, and the test page at scaffold will not show any value for "url" under attachments. I also tried "snapshot : false," but no pdf either.

adam3smith:
Thanks everyone -- please let me know once this is ready for final review. As for FW, given https://github.com/zotero/translators/issues/1277, I'd prefer us avoiding new FW translators.

yunusong:
@fbennett I have one more question.
When I erased all other attributes of the attachment and just left the url as "url : pdfurl" (which gives a long url that would have been the result of the function SiteSelect1), Zotero saves the snapshot of the page that leads to downloading, and the browser downloads the pdf file to my download folder in my computer.

I think I need the translator to access this final url directly. The gist of the html code that calls the url is as follows.

<body onLoad="document.down.submit();">
<form name="down" method="post" action="http://210.101.116.17/kiss8/inftp_journal.asp">
	<input type="hidden" name="a_code" value="40323420.pdf">
	<input type="hidden" name="code" value="7624112825580073101">
	<input type="hidden" name="isDownLoad" value="1">
</form>
</body>

If you paste this html code into a new text and open it in a browser, you will see that it downloads a pdf file even without requiring proxy. Do you think you can work around this code and have the zotero to fetch the final url? I really appreciate your help.

fbennett:
Yes, that's the plan. Likely it can be done by nesting another call that
runs inside the pop-up.

yunusong:
Can you guys tell me what is wrong with my code? I don't think I quite understand how to use ZU.doPost. But I am quite certain that I have to use it because the actual page that leads to pdf has method="post".

So in the scrape function I added the code for attachments as follows.

  	`var` pdfJSurl = ZU.xpathText(doc, '//div[@class="search_box"]/div[@class="choice"]/span[1]/a/@href');
  	`var` pdfurlKeys = pdfJSurl.match(/Select1\(\'(\S+)\'\,\s?\'(\S+)\'\,\s?(\S+)\,\s?(\S+)\,\s?(\S+)\)\;/);
  	`var` pdflinkurl = "../search/download.asp?ftproot=" + pdfurlKeys[1] + "&inst_key=" + pdfurlKeys[3] + "&a_imag=" + pdfurlKeys[2] + "&isDownLoad=" + pdfurlKeys[5] + "&publ_key=" + pdfurlKeys[4];
	
	`item.attachments.push({`
		url : ZU.processDocuments(pdflinkurl, getpdf),
		snapshot : true,
	})

I was hoping that this would process the getpdf function on the page that opens. And for getpdf function, I coded it as follows.

`function` getpdf(doc, url) {
	`var` theurl = ZU.xpathText(doc, '//form[@name="down"]/@action');
	`var` a_code = ZU.xpathText(doc, '//input[@name="a_code"]/@value');
	`var` code = ZU.xpathText(doc, '//input[@name="code"]/@value');
	`var` isDownload = ZU.xpathText(doc, '//input[@name="isDownload"]/@value');
	`var` postData = [a_code, code, isDownload];
	`var` headerData = {"a_code": a_code, "code": code, "isDownload": isDownload};
	ZU.doPost(theurl, a_code + code + isDownload, function(pdfurl2) {});	
}

I checked how other translators use the doPost, and they were very different from how I wrote it, but I don't quite understand how to fix it.

fbennett:
Picking up now (fell asleep last night, but I turned on the office proxy yesterday, so I'm "at work" this morning). Will fiddle around a bit and see what I can come up with.

yunusong:
Ok, I think I figured out how to use ZU.doPost. But I still couldn't get Zotero to save pdf. This is the code I added at the end of the scrape function. The weird thing is, though, when I run test in Scaffold, the result will show "Translation Successful" AND my browser will download the pdf to my computer, even though I only ran test. And when I actually test it with Zotero, it will not fetch pdf. Any thoughts on what I should change?

	var pdfJSurl = ZU.xpathText(doc, '//div[@class="search_box"]/div[@class="choice"]/span[1]/a/@href');
  	var pdfurlKeys = pdfJSurl.match(/Select1\(\'(\S+)\'\,\s?\'(\S+)\'\,\s?(\S+)\,\s?(\S+)\,\s?(\S+)\)\;/);
  	var pdflinkurl = "../search/download.asp?ftproot=" + pdfurlKeys[1] + "&inst_key=" + pdfurlKeys[3] + "&a_imag=" + pdfurlKeys[2] + "&isDownLoad=" + pdfurlKeys[5] + "&publ_key=" + pdfurlKeys[4];
	ZU.processDocuments(pdflinkurl, function(doc, url) {
		var actionurl = ZU.xpathText(doc, '//form[@name="down"]/@action');
		var inputData = "a_code=" + ZU.xpathText(doc, '//input[@name="a_code"]/@value') + "&code=" + ZU.xpathText(doc, '//input[@type="code"]/@value') + "&isDownload=" + ZU.xpathText(doc, '//input[@name="isDownload"]/@value');
		ZU.doPost(actionurl, inputData, function(pdfdoc) {
			item.attachments.push({
				url : pdfdoc,
			});
		});
	});

  	item.complete();

fbennett:
Cracked it. I have code that attaches a readable PDF. As it turned out, no further call (to `doGet`, `doPost` or `processDocuments`) was necessary. Tried various things, but what did the trick was to use the `viewer.asp` that is the ultimate target of a view request. It's a POST call on the site, but setting the post data as a query string on the URL (which I presume the translator calls with GET) worked anyway. The only post data needed is `code_num`, which is just the server-side filename of the PDF. So it turns out to be super simple, after all our guesswork.

I'll file a pull request in a bit.

fbennett:
Erm, I don't know now to pitch in with a pull request on a pull request (of if that's possible), so here's the code block:
``` javascript
    var pdfJSurl = ZU.xpathText(doc, '//div[@class="search_box"]/div[@class="choice"]/span[2]/a/@href');
    var pdfurlKeys = pdfJSurl.match(/Select1\(\'(\S+)\'\,\s?\'(\S+)\'\,\s?(\S+)\,\s?(\S+)\,\s?(\S+)\)\;/);
    var pdfUrl = pdfurlKeys[1].replace(/^(.*\/).*/, "$1") + "viewer.asp?code_num=" + pdfurlKeys[2];

    item.attachments.push({
	url: pdfUrl,
	title: "KISS PDF",
	mimeType:"application/pdf"
    });
```
Tested it with a few other pages, and it seems to work reliably.

adam3smith:
@fbennett -- you could actually make the pull request against yunusong's patch-2 branch and they'd then merge it, but I can also just integrate this into the translator to simplify things.

fbennett:
@adam3smith I'll give that a try. Protocol.

adam3smith:
beautiful, worked perfectly. Is this ready for review then?

fbennett:
Ha. @yunusong beat me to it. Glad to hear it works!

yunusong:
Yes, I think it is ready for review. Do I need to click close this myself?

adam3smith:
no, don't close. Re review here on the thread as soon as I can.

KStudy.js
```diff
@@ -0,0 +1,173 @@
...
+		});
+	} else {
+		scrape(doc, url);
+	}
+}
+
+function scrape(doc, url) {
+	var item = new Zotero.Item("journalArticle");
+	item.title = ZU.xpathText(doc, '//div[@class="agency_box"]/div[1]/span');
+	item.language = "kr";
```
adam3smith: this should be `ko-KR` please, see e.g. http://www.localeplanet.com/icu/ko-KR/

KStudy.js
```diff
@@ -0,0 +1,173 @@
...
+	
+	var creators = ZU.xpathText(doc, '//div[@class="agency_detail"]/ul/li[1]');
+	if (creators) {
+		var creatorsList = creators.split(',');
+		for (var i=0; i<creatorsList.length; i++) {
+			item.creators.push(ZU.cleanAuthor(creatorsList[i], "author", true));
+		}
```
adam3smith: To get authors into single field mode, add here:
```
(var i=0; i<item.creators; i++) {
  if (!item.creators[i].firstName) {
    item.creators[i].fieldMode = true;
  }
}
```

KStudy.js
```diff
@@ -0,0 +1,173 @@
...
+		item.abstractNote = ZU.trimInternal(item.abstractNote);
+	}
+
+	var pdfJSurl = ZU.xpathText(doc, '//div[@class="search_box"]/div[@class="choice"]/span[1]/a/@href');
+  	var pdfurlKeys = pdfJSurl.match(/Select1\(\'(\S+)\'\,\s?\'(\S+)\'\,\s?(\S+)\,\s?(\S+)\,\s?(\S+)\)\;/);
+  	var pdfUrl = pdfurlKeys[1].replace(/^(.*\/).*/, "$1") + "viewer.asp?code_num=" + pdfurlKeys[2];
+
+	item.attachments.push({
+		url : pdfUrl,
+		title : "KISS PDF",
```
adam3smith: let's make that `KStudy Full Text PDF`, please (we want "full text" in there for uniformity and the actual name of the translator, too (though there's some disagreement on that).

zuphilip:
Please have a look at the new version which I pushed directly to @yunusong's branch:
* The website seems to be redesigned and therefore quite a lot of code had to change. I update one test case and added two more. Can the Korean speakers check that the test cases still make sense?
* All comments by @adam3smith are handled now. Review of the new code would be appreciated.
* Saving the PDF seems not to work anymore and I don't have subscription to that database. @fbennett Let me know if you want to work on that again.

adam3smith:
I'm also seeing a different format: http://kiss.kstudy.com/public/public3-article.asp?key=60001761 we could test PDF downloading on this perhaps unless that's a different way for downloading (and then we should still do it for those pages).

yunusong:
I am seeing that a lot of things are not working properly. Some pages for journal article does not show the journal article button, and some show a folder icon. I think I need work on this a little, but can you tell me how to use scaffold now? I see that it no longer works with firefox, but I haven’t figured out how to use it yet. 

    2018년1월1일 (월요일) 오후 6:21에 Sebastian Karcher <notifications@github.com>님이 쓴 메시지:
 

 I'm also seeing a different format: http://kiss.kstudy.com/public/public3-article.asp?key=60001761 we could test PDF downloading on this perhaps unless that's a different way for downloading (and then we should still do it for those pages).—

zuphilip:
The example from @adam3smith looks like a working paper and we could maybe save them as `report`.

@yunusong Scaffold can be used in Zotero 5.0 directly:
1. Download the latest version from https://github.com/zotero/scaffold/releases
2. Open Zotero (Standalone)
3. Go to the add-ons within Zotero and install the previously downloaded xpi file.
4. Possibly restart Zotero

Okay, then I will let you work further. I can help you further, you just have to advice me because of my not understanding any Korean.

yunusong:
Thank you, I just installed Scaffold.
I saw that you already made the important changes. I think it works fine for what is in there, although I couldn't check if saving pdf works or not because I am not at school. The test results look good.
The "different format" you found seems like a government report, written by employees in various government departments. They call it "copyrighted public records."  Should we add another category in detectWeb for that? And what would it be? "report"?  There is also another category, which they call it "national knowledge." I'm not sure how else to categorize this, but this also seems to be government reports. But they have volume/issue numbers. I would like to know what you think about these. 

    2018년1월1일 (월요일) 오후 6:59에 Philipp Zumstein <notifications@github.com>님이 쓴 메시지:
 

 The example from @adam3smith looks like a working paper and we could maybe save them as report.@yunusong Scaffold can be used in Zotero 5.0 directly:   
   - Download the latest version from https://github.com/zotero/scaffold/releases
   - Open Zotero (Standalone)
   - Go to the add-ons within Zotero and install the previously downloaded xpi file.
   - Possibly restart Zotero
Okay, then I will let you work further. I can help you further, you just have to advice me because of my not understanding any Korean.—

zuphilip:
Great that it worked. Okay, the itemType `report` sounds fine for these cases. There is a `reportNumber` and `seriesTitle` which we then can use. I am not sure about the issue number. Do you have an example?

Let me know if you want to do more changes. Otherwise I will probably continue tomorrow to work here after 😴  🛌 .

yunusong:
Can you help me with one more thing? I installed scaffold, but it doesn’t seem to recognize any of my browsers.  current URL keep shows up as “blank,” and even if I manually copy and paste the address, it doesn’t seem to work. Thanks for the help. 

    2018년1월1일 (월요일) 오후 7:25에 Philipp Zumstein <notifications@github.com>님이 쓴 메시지:
 

 Great that it worked. Okay, the itemType report sounds fine for these cases. There is a reportNumber and seriesTitle which we then can use. I am not sure about the issue number. Do you have an example?Let me know if you want to do more changes. Otherwise I will probably continue tomorrow to work here after 😴 🛌 .—

zuphilip:
Scaffold has its own browser now and you can simply copy and paste whichever URL you wanted to see into the `Current URL` field. After copy and paste you have to hit enter and wait for the page to load. Does this work?

scaffold-url

zuphilip:
Okay, I am continuing to work here now. I will let you know when I have a new version...

zuphilip:
I generalized the translator for the example(s) found by @adam3smith but choosed in the end `journalArticle` for that documented in new test cases. However, it seems for me, that the PDF download we see in the public part is maybe tailored for that and could be fragile at all:

kiss-public-download

Thus, I decided to skip the PDF download for the moment. Maybe, we can start with the translator as it is now. Any improvement for PDF download could be added later by someone with access to the database.

zuphilip:
Please have a look at the new version.

socheres:
@zuphilip below some more test cases and remarks.
author in Korean should import in a "single field" in Zotero., because the first letters(Silben) of Korean names are usually surname and the other two letters are first name. Also, there is no middle name in korean.
Authors in latinized form with comma is separated correctly.
one more suggestions for issn.
`//e.g. issn = ISSN(Print) 
	var issn = ZU.xpathText(doc, '//li[label[text()="ISSN(Print)"]]');
	if(issn){
		item.ISSN = ZU.cleanISSN(issn);
	}`

test cases:

http://kiss.kstudy.com/public/public2-article.asp?key=50789039

```
18:30:10 Returned item:
           {
             "itemType": "journalArticle"
             "creators": [
               {
                 "firstName": "Yoon Tae"
                 "lastName": "Kim"
                 "creatorType": "author"
               }
               {
                 "firstName": "Hyun Suk"
                 "lastName": "Park"
                 "creatorType": "author"
               }
             ]
             "notes": []
             "tags": []
             "seeAlso": []
             "attachments": []
             "title": "KOLMOGOROV DISTANCE FOR MULTIVARIATE NORMAL APPROXIMATION"
             "language": "ko-KR"
             "publicationTitle": "Korean Journal of mathematics (강원경기수학회)"
             "volume": "23"
             "issue": "1"
             "date": "2015"
             "pages": "1–10"
             "abstractNote": "초록 보기"
             "libraryCatalog": "kiss"
           }
18:30:10 Translation successful
```

 

http://kiss.kstudy.com/thesis/thesis-view.asp?key=3550330

```
18:31:51 Returned item:
           {
             "itemType": "journalArticle"
             "creators": [
               {
                 "lastName": "( Yoon Tae Kim ) "
                 "fieldMode": true
                 "creatorType": "author"
               }
               {
                 "lastName": " ( Hyun Suk Park )"
                 "fieldMode": true
                 "creatorType": "author"
               }
             ]
             "notes": []
             "tags": []
             "seeAlso": []
             "attachments": []
             "title": "Convergence rate of a test statistics observed by the longitudinal data with long memory"
             "language": "ko-KR"
             "publicationTitle": "CSAM(Communications for Statistical Applications and Methods)"
             "volume": "24"
             "issue": "5"
             "date": "2017-09"
             "pages": "481–492"
             "abstractNote": "초록 보기 This paper investigates a convergence rate of a test statistics given by two scale sampling method based on Ait-Sahalia and Jacod (Annals of Statistics, 37, 184-222, 2009). This statistics tests for longitudinal data having the existence of long memory dependence driven by fractional Brownian motion with Hurst parameter H ∈ (1/2, 1). We obtain an upper bound in the Kolmogorov distance for normal approximation of this test statistic. As a main tool for our works, the recent results in Nourdin and Peccati (Probability Theory and Related Fields, 145, 75-118, 2009; Annals of Probability, 37, 2231-2261, 2009) will be used. These results are obtained by employing techniques based on the combination between Malliavin calculus and Stein`s method for normal approximation."
             "libraryCatalog": "kiss"
           }
18:31:51 Translation successful
```

 

http://kiss.kstudy.com/public/public2-article.asp?key=50375984

 

```
18:33:10 Returned item:
           {
             "itemType": "journalArticle"
             "creators": [
               {
                 "firstName": "Ja-Son Y."
                 "lastName": "Kim"
                 "creatorType": "author"
               }
               {
                 "firstName": "Tae-Joon"
                 "lastName": "Park"
                 "creatorType": "author"
               }
               {
                 "firstName": "Jin-Sol"
                 "lastName": "Lee"
                 "creatorType": "author"
               }
               {
                 "firstName": "Ji-Yong"
                 "lastName": "Chun"
                 "creatorType": "author"
               }
               {
                 "firstName": "Joon-Seol"
                 "lastName": "Bae"
                 "creatorType": "author"
               }
               {
                 "firstName": "Byung-Lae"
                 "lastName": "Park"
                 "creatorType": "author"
               }
               {
                 "firstName": "Hyun-Sub"
                 "lastName": "Cheong"
                 "creatorType": "author"
               }
               {
                 "firstName": "Hyo-Suk"
                 "lastName": "Lee"
                 "creatorType": "author"
               }
               {
                 "firstName": "Yoon-Jun"
                 "lastName": "Kim"
                 "creatorType": "author"
               }
               {
                 "firstName": "Hyoung-Doo"
                 "lastName": "Shin"
                 "creatorType": "author"
               }
             ]
             "notes": []
             "tags": []
             "seeAlso": []
             "attachments": []
             "title": "Association Analysis of SERPINB5 Polymorphisms with HBV Clearance and HCC Occurrence in a Korean Population"
             "language": "ko-KR"
             "publicationTitle": "Genomics & informatics (한국유전체학회)"
             "volume": "8"
             "issue": "1"
             "date": "2010"
             "pages": "1–8"
             "abstractNote": "초록 보기"
             "libraryCatalog": "kiss"
           }
18:33:10 Translation successful
```

KStudy.js
```diff
@@ -131,11 +133,23 @@ function scrape(doc, url) {
 		var containerParts = containerValue.match(/(.*)\s+(\d+)\D\s*(\d+)/);
 		if (containerParts) {
 			item.publicationTitle = containerParts[1];
-			item.volume = containerParts[2];
+			if (/\d{4}/.test(containerParts[2])) {
+				containerParts[2] = null;
```
zuphilip: Add a comment why you are doing this. Moreover, I guess you don't need to delete the value in the first case, but rather simply want to concentrate on the second case. Is this correct? Then you could instead try something like
```js
if (containerParts[2].length<4) { // or the regexp if really needed
   item.volume = containerParts[2];
}
```

yunusong: Oh, okay thank you for pointing that out.
The reason I added that line was that I found that sometimes the pages have published year in place of the volume number. So sometimes it would appear as Vol. 2013, and I didn't want that to happen, but your correction makes much more sense. Thank you. I'll fix that.

zuphilip:
@yunusong For your question about the multiple in the public part: You need to tweak also the CSS selector for the different entries, i.e. this line
```js
var rows = ZU.xpath(doc, '//div[contains(@class, "thesis-info")]/h5/a[contains(@href, "/thesis")]');
```
(The last check is not true `contains(@href, "/thesis")`; but change it wisely/carefully in order not to influence the other multiples negatively.)

zuphilip:
@socheres Thank you for the review! The ISSN is a good idea to add as well. I am not sure that I understand the naming correctly. Did you see currently an error in the splitting of author names? Example?

socheres:
@zuphilip authors name splitting works only correct in case 1, not in 2 or 3. 
1.[Kim, Moon Tae , Park, Hyun Suk](http://kiss.kstudy.com/public/public2-article.asp?key=50789039)
2.[( Yoon Tae Kim ) , ( Hyun Suk Park )](http://kiss.kstudy.com/thesis/thesis-view.asp?key=3550330)
3.[김윤태 ( Yuntae Kim )](http://kiss.kstudy.com/thesis/thesis-view.asp?key=3483357)

(Kim|Park|김) = last name

In case 2,3 they should be - if possible - in a full name field, like in Zotero where i can switch author field from double  in single field. 
3. 김 = last name
    윤태 = first name
   Kim = last name
   Yuntae = first name

socheres:
i think the website structure - authors -  is too fragile and i think the effort is not worth it, because all the metadata on http://kiss.kstudy.com/ can be downloaded - third button below 인용하기 - in RIS.

zuphilip:
Uh... RIS, yes we should use that in the translator (it is the blue button with the arrow to the right). I will look into this now...

zuphilip:
Please have a look at the new version, which is pretty much a rewrite, but much simpler by using the RIS data. 🚀

KStudy.js
```diff
@@ -0,0 +1,509 @@
...
+		});
+	} else {
+		scrape(doc, url);
+	}
+}
+
+
+function scrape(doc, url) {
+	var key = url.match(/\bkey=(\d+)\b/)[1];
+	var risURL = "http://kiss.kstudy.com/p-common/export_endnote.asp";
```
adam3smith: lets make this relative in case the site starts offering https

KStudy.js
```diff
@@ -0,0 +1,509 @@
...
+			ZU.processDocuments(articles, scrape);
+		});
+	} else {
+		scrape(doc, url);
+	}
+}
+
+
+function scrape(doc, url) {
+	var key = url.match(/\bkey=(\d+)\b/)[1];
```
adam3smith: we should check for this in detect, then, to make sure we have a key

KStudy.js
```diff
@@ -0,0 +1,509 @@
...
+		translator.setHandler("itemDone", function(obj, item) {
+			if (url.includes('/public/public3-article')) {
+				item.itemType = "report";
+			}
+			if (item.issue && item.issue == "0") {
+				delete item.issue;
+			}
+			item.language = "ko-KR";
+			if (item.pages) {
+				item.pages = item.pages.replace('-', '–');
```
adam3smith: we don't usually do this -- do we actually prefer en-dashes over hyphens in the database?

KStudy.js
```diff
@@ -0,0 +1,509 @@
...
+			}
+		]
+	},
+	{
+		"type": "web",
+		"url": "http://kiss.kstudy.com/public/public2-article.asp?key=50789039",
+		"items": [
+			{
+				"itemType": "journalArticle",
+				"title": "KOLMOGOROV DISTANCE FOR MULTIVARIATE NORMAL APPROXIMATION",
```
adam3smith: let's title case all caps titles

zuphilip: I tried to think about this before. I don't know how to check easily whether the title is all-uppercase but not a Korean title. Any ideas?

dstillman: You should be able to just use `ZU.capitalizeTitle(title, true)`.

KStudy.js
dstillman: I don't think there's any harm in doing so, though we should probably just be automatically converting hyphens in page ranges to en dashes on save.

KStudy.js
zuphilip: Thank you. This also changes the title `치과용 콘빔 CT를 이용한 상악 정중과잉치의 3차원 분석`, which I wanted to avoid. I guess some test like `[\u0000-\u00FF]*` should work for Latin alphabet only titles...

dstillman: How does it change it?

dstillman: So it lowercases the 't'. I haven't looked at this function too closely, but that's plausibly just a bug that we should fix.

zuphilip: Yes, it makes `Ct`.

But I prefer to use `capitalizeTitle` only when we have all-uppercase titles. Otherwise also names like `USA` or `EU` will be strangely capitalize. See also the used pattern in other translators like in https://github.com/zotero/translators/blob/e04633baeaac2c62eddb4435e517ae7d51c89a73/Web%20of%20Science.js#L131-L133

KStudy.js
```diff
@@ -0,0 +1,509 @@
...
+						"lastName": "Park",
+						"creatorType": "author",
+						"fieldMode": 1
+					},
+					{
+						"lastName": "Hyun Suk",
+						"creatorType": "author",
+						"fieldMode": 1
+					}
+				],
```
socheres: romanized korean names with comma create two last names:

>[ 저자 : Kim, Yoon Tae, Park, Hyun Suk](http://kiss.kstudy.com/public/public2-article.asp?key=50789039)

Korean names are correct imported in "filedMode 1", but it will be nice to have, if korean names can be separated in last and first name like the function "ZU.cleanAuthor" for korean names.

KStudy.js
```diff
@@ -0,0 +1,509 @@
...
+					{
+						"lastName": "Park",
+						"creatorType": "author",
+						"fieldMode": 1
+					},
+					{
+						"lastName": "Hyun Suk",
+						"creatorType": "author",
+						"fieldMode": 1
+					}
```
socheres: line403-423

KStudy.js
dstillman: `capitalizeTitle` won't change "USA" and "EU" unless the string is fully capitalized. But it will convert things that are in sentence case to title case, which we don't want to do unnecessarily (if the capitalizeTitles pref is left off).

```
>> Zotero.Utilities.capitalizeTitle("The USA and also the EU", true)
"The USA and Also the EU"
```

```
>> Zotero.Utilities.capitalizeTitle("THE USA AND ALSO THE EU", true)
"The Usa and Also the Eu"
```

If we want to avoid messing up things like the `Ct`, we could perhaps say that `capitalizeTitle` shouldn't change words if they contain any characters outside `[\u0000-\u00FF]`. I'm not positive that's always what we want, though.

socheres: it will be great if the English title can be normalized, taking capitalization e.g. proper noun into account, and other words such as USA, EU in capitalize form. There are some similar projects (https://github.com/nlp-compromise/compromise), but I haven't found a "killer Javascript function" yet.

dstillman: If a site gives a title in all-caps, as seems to be the case here, we generally just title case it according to some basic rules and leave the fixing of proper nouns and acronyms up to users. (Along other things, we don’t even necessarily know the language in order to reliably exclude words.)

The issue here just seems to be (I haven’t looked at this closely) that some Korean titles include English characters, and we want to avoid changing the case of those, either in the translator or in a more generalized way in `capitalizeTitle`.

zuphilip: @dstillman Okay, my mistake. But the code pattern can still make sense then, to prevent making title case, when sentence case is correctly given.

I test that the majority of the title is from Latin alphabet. But your suggestion for `capitalizeTitle` on the words

> not change words if they contain any characters outside [\u0000-\u00FF]

sounds reasonable for me.

@socheres The function `ZU.capitalizeTitle` can be used to "normalize" English titles into title case  in translators, see https://github.com/zotero/zotero/blob/master/chrome/content/zotero/xpcom/utilities.js#L921 . However, it will add a capital letter to more or less every word with only a small number of exceptions cf. https://github.com/zotero/zotero/blob/2baa537542fcd5bc6e2826eb0eef6b34c571dcfc/chrome/content/zotero/xpcom/utilities.js#L922-L924 . To do anything similar e.g. for German and thereby differentiating between nouns and verbs etc. might be out of scope for Zotero.

KStudy.js
zuphilip: Hm... my mistake then. How is the page range then shown in citation styles? Maybe the replacement could also happens at some later step... I will delete it on the translator code, beause this would then be a larger issue.

KStudy.js
socheres: thank you @avram @zuphilip for explaining. i totally understand that fixing crappy data records is out of scope for zotero. it was just feedback - nice to have feature - from the users/scientists. publisher who create metadata have to deal it.

KStudy.js
zuphilip: Okay, you are right. The RIS data contains wrongly four authors in the `AU` tags. But I still can reuse some of the previous code for scraping it directly in these cases. Stay tuned..

zuphilip:
I did changes according to the reviews of @adam3smith and @socheres . Sometimes it is needed to scrape the information for the authors directly from the page. I hope that we have the author information now correctly.

Please have a look at the new version.

socheres:
it looks :thumbsup: 
source: > [( Chang-young Hong ) , ( Su-yeon Lee ) , ( Myungkil Kim ) , ( In-gyu Choi )](http://kiss.kstudy.com/thesis/thesis-view.asp?key=3480580)
`{
             "itemType": "journalArticle"
             "creators": [
               {
                 "firstName": "Chang-young"
                 "lastName": "Hong"
                 "creatorType": "author"
               }
               {
                 "firstName": "Su-yeon"
                 "lastName": "Lee"
                 "creatorType": "author"
               }
               {
                 "firstName": "Myungkil"
                 "lastName": "Kim"
                 "creatorType": "author"
               }
               {
                 "firstName": "In-gyu"
                 "lastName": "Choi"
                 "creatorType": "author"
               }`

source > [저자 : Kim, Yoon Tae, Park, Hyun Suk](http://kiss.kstudy.com/public/public2-article.asp?key=50789039)
`{
             "itemType": "journalArticle"
             "creators": [
               {
                 "firstName": "Yoon Tae"
                 "lastName": "Kim"
                 "creatorType": "author"
               }
               {
                 "firstName": "Hyun Suk"
                 "lastName": "Park"
                 "creatorType": "author"
               }
             ]`
source > [김윤태 ( Yuntae Kim )](http://kiss.kstudy.com/thesis/thesis-view.asp?key=3483357)
`{
             "itemType": "journalArticle"
             "creators": [
               {
                 "lastName": "김윤태"
                 "creatorType": "author"
                 "fieldMode": 1
               }`:clap:

yunusong:
I checked the update, and things look really great. Thanks for the work.

adam3smith:
Great, thanks everyone!

zuphilip:
✨  Thank you all @yunusong, @fbennett, @avram, @dstillman, @socheres, @adam3smith for the work and the review! That was a massive collaboration on a Zotero translator!

PR #1311, opened 2017-05-30
Create DBpia.js

yunusong:
A translator for another Korean Studies database.
I coded it based on @zuphilip 's common code, so I included his name in the creators.
I think this translator is ready for review as is.

### DIFF
--- a/DBpia.js
+++ b/DBpia.js
@@ -2,14 +2,14 @@
 	"translatorID": "0c31f371-e012-4b1c-b793-f89ab1ae2610",
 	"label": "DBpia",
 	"creator": "Yunwoo Song, Philipp Zumstein",
-	"target": "^https?://[^/]+\\.dbpia\\.co\\.kr",
+	"target": "^https?://[^/]+\\.dbpia\\.co\\.kr/",
 	"minVersion": "3.0",
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
-	"browserSupport": "gcsi",
-	"lastUpdated": "2017-05-30 18:55:18"
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2021-09-14 00:23:40"
 }
 
 /*
@@ -37,25 +37,25 @@
 */
 
 function detectWeb(doc, url) {
-	if (url.indexOf('/Journal/ArticleDetail/')>-1) {
+	if (url.includes('/journal/articleDetail')) {
 		return "journalArticle";
-	} else if ((url.indexOf('/SearchResult/')>-1 || url.indexOf('/Journal/ArticleList/')>-1) && getSearchResults(doc, true)) {
+	} else if ((url.includes('/search/') || url.includes('/journal/articleList/')) && getSearchResults(doc, true)) {
 		return "multiple";
 	}
+	return false;
 }
 
 function getSearchResults(doc, checkOnly) {
 	var items = {};
 	var found = false;
-	var rows = ZU.xpath(doc, '//div[@class="titleWarp"]/a');
-	for (var i=0; i<rows.length; i++) {
-		var href = rows[i].href;
-		var c = i + 1;
-		var title = ZU.trimInternal(rows[i].textContent);
+	var rows = doc.querySelectorAll('h5 > a[href*="/journal/articleDetail"]');
+	for (let row of rows) {
+		var href = row.href;
+		var title = ZU.trimInternal(row.textContent);
 		if (!href || !title) continue;
 		if (checkOnly) return true;
 		found = true;
-		items[c] = title;
+		items[href] = title;
 	}
 	return found ? items : false;
 }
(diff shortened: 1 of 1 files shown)


zuphilip:
This looks interesting! You don't have to add my name when you just use the common code blocks, which were formed over several PRs and are influenced by a lot of more people. Sometimes the creator field is more useful to have someone to ask also later about some of the details in the translator code.

yunusong:
Oh okay. Thanks for letting me know. But I will leave it as is this time. If I create another one, I will write just my name.

DBpia.js
```diff
@@ -0,0 +1,186 @@
...
+	"translatorID": "0c31f371-e012-4b1c-b793-f89ab1ae2610",
+	"label": "DBpia",
+	"creator": "Yunwoo Song, Philipp Zumstein",
+	"target": "^https?://[^/]+\\.dbpia\\.co\\.kr",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsi",
```
zuphilip: You can toggle all options here on.

DBpia.js
```diff
@@ -0,0 +1,186 @@
...
+	item.publicationTitle = ZU.xpathText(doc, '//meta[@name="citation_journal_title"]/@content');
+
+	var pagerange = ZU.xpathText(doc, '//div[@class="book_info"]/dl/dt[1]').match(/\s([\d+-\d+])\s/);
+		if (pagerange) {
+			item.pages = pagerange[1];
+		}
+		
+	item.volume = ZU.xpathText(doc, '//meta[@name="citation_volume"]/@content');
+	item.issue = ZU.xpathText(doc, '//meta[@name="citation_issue"]/@content');
+	item.ISSN = ZU.xpathText(doc, '//meta[@name="citation_issn"]/@content');
```
zuphilip: This looks like the Embedded Metadata translator should already give you pretty good data. I would suggest to first call EM translator and then add the missing fields only.

DBpia.js
```diff
@@ -0,0 +1,186 @@
...
+		}
+	}
+	
+	item.date = ZU.xpathText(doc, '//meta[@name="citation_publication_date"]/@content');
+	item.publicationTitle = ZU.xpathText(doc, '//meta[@name="citation_journal_title"]/@content');
+
+	var pagerange = ZU.xpathText(doc, '//div[@class="book_info"]/dl/dt[1]').match(/\s([\d+-\d+])\s/);
+		if (pagerange) {
+			item.pages = pagerange[1];
+		}
```
zuphilip: Fix indenting of these three lines.

DBpia.js
```diff
@@ -0,0 +1,186 @@
...
+				var n = num[i];
+				scrapeMulti(doc, url, n);
+			}
+		});
+	} else {
+		scrape(doc, url);
+	}
+}
+
+function scrapeMulti(doc, url, n) {
```
zuphilip: Do you need here an extra function? Can we not already save this url which we want to scrape from, already in the `doWeb`?

zuphilip:
@socheres Can you review the examples for this new suggested translator?

socheres:
@zuphilip @yunusong this translator is very useful for korean studies. thanks both of you. I will review the translator for dbpedia soon.

zuphilip:
For calling EM in another translator you can look here for a template and some examples: https://github.com/zuphilip/translators/wiki/Calling-metadata-translator

DBpia.js
```diff
@@ -0,0 +1,162 @@
...
+			}
+			var num = [];
+			for (var i in items) {
+				num.push(i);
+			}
+			for (var i=0; i<num.length; i++) {
+				var n = num[i];
+				var link = ZU.xpathText(doc, '//div[@class="subSearchResultList"]/div[' + n + ']/ul[1]/li[4]/dl/dt/div/a/@href')
+				if (link.indexOf('book')==-1) {
+					ZU.processDocuments(link, scrape);
```
adam3smith: It's preferred (and more stable) if you don't run `processDocuments` multiple times inside a loop, but instead make link an array of links and then run `processDocuments` once that's populated. I *think* that should work here.

DBpia.js
```diff
@@ -0,0 +1,162 @@
...
+		scrape(doc, url);
+	}
+}
+
+function scrape(doc, url) {
+	var translator = Zotero.loadTranslator('web');
+	translator.setTranslator('951c027d-74ac-47d4-a107-9c3069ab7b48');
+	translator.setHandler('itemDone', function (obj, item) {
+		var pdfurl = ZU.xpathText(doc, '//div[@class="btn_box"]/a[@title="PDF Download"]/@href');
+		item.attachments.push({
```
adam3smith: please test for pdfurl before doing this and give the attachment a title.

DBpia.js
```diff
@@ -0,0 +1,162 @@
...
+			if (!items) {
+				return true;
+			}
+			var num = [];
+			for (var i in items) {
+				num.push(i);
+			}
+			for (var i=0; i<num.length; i++) {
+				var n = num[i];
+				var link = ZU.xpathText(doc, '//div[@class="subSearchResultList"]/div[' + n + ']/ul[1]/li[4]/dl/dt/div/a/@href')
```
yunusong: The reason I did it this way is that if the search results show books along with journal articles, I can't have the translator process the links for the books just like I can have it for journal articles. (The links for books redirect to a completely different website, and sometimes just to the main page.) So I had to use the metadata on the search result page.

However, the other metadata (like author and date) for books are not under the same node as the title and url. They are not even in sibling nodes but in like "uncle" nodes or "cousin" nodes or something. If such terminology even exits.
eg. the title is at //dl/dt/div/a
but the author is at //dl/dd[1]
then publisher at //dl/dd[2] etc.

 So I had to preserve the information from the `getSearchResult` which item was selected, and then use that information to specify the nodes with other metadata. So I had to go back all the way up to the div with "subSearchResultList" title, and then specify their count. As far as I could see on the html, that count on div was the only difference between each entry in the list.  The tags for author/date/publisher weren't marked with anything. I also think it looks very complicated. If you have better idea to simplify it, I am more than open to incorporate it.

DBpia.js
adam3smith: by "test for pdfurl" I mean wrapping lines 110 to 115 into `if (pdfurl)`. The translator will otherwise throw an error when it cant find the url and that'd be unnecessarily fragile.

DBpia.js
```diff
@@ -0,0 +1,162 @@
...
+				} else {
+					var item = new Zotero.Item("book");
+					var title = ZU.xpathText(doc, '//div[@class="subSearchResultList"]/div[' + n + ']/ul[1]/li[4]/dl/dt/div/a');
+					item.title = title;
+					var creators = ZU.xpathText(doc, '//div[@class="subSearchResultList"]/div[' + n + ']/ul[1]/li[4]/dl/dd[1]');
+					if (creators) {
+						var creatorsList = creators.split(',');
+						for (var i=0; i<creatorsList.length; i++) {
+							item.creators.push(ZU.cleanAuthor(creatorsList[i], "author", true));
+						}
```
adam3smith: were you still intending to do this for both translators? Feel free to ask if my instructions weren't clear.

yunusong: If you mean "for this" by having the author field into a single field, I no longer care. When I tried saving an article with Zotero, full names of the authors were in "Last Name". This is good enough for me. With or without empty "first name" doesn't really matter. If, for some coding reasons, you prefer one way or another, you can change it, but I don't mind leaving it as is.

AbeJellinek:
Thank you and sorry for the long delay in getting this merged!

Issue #917, opened 2015-07-09
JSON-LD export/import


adam3smith:
and then also include into embedded metadata translator. Suggested by Faolan here:
https://forums.zotero.org/discussion/50327/is-unapi-deadbecause-their-website-sure-seems-to-be/#Item_8
seems like a great idea.

dstillman:
So we'd like to move forward on this, and we'd be happy to sponsor the work on this if someone is interested. The primary goal would be to be able to generate JSON-LD for items via the Zotero API (via the API's support for export translators) and import that back into Zotero clients.

One thing that's not clear to me is the state of the various bibliographic ontologies — bib.schema.org, BIBO, Zotero RDF… There's a JS library that can generate JSON-LD from RDF, so we might be able to piggyback off the existing RDF translators, but as far as I know there hasn't really been much uptake of BIBO, so I don't know what makes the most sense. I'd add that, while we obviously want a round-trip to be close to lossless, I don't think it's a fixed requirement — meaning that I don't think we need to use API JSON (which we don't really want to be an exchange format) or Zotero RDF (same).

Thoughts?

/cc @rmzelle @aurimasv @fcheslack others?

adam3smith:
That's wonderful. Very much agree this should be a priority.

This has been on things I've been wanting to do for quite a while (and learning more about JSON-LD is going to be quite useful for me), so I'd be interested, but I do have a regular day job now, so I'm happy to leave this to someone with more time.

I always imagined writing this from scratch rather than draw on RDF, in particular because once we specify the RDF to JSON-LD mappings, I'm not sure how much time is saved.

I agree that Bibo hasn't really caught on at all, so I wouldn't feel great on using that. Zotero RDF doesn't even have a written specification. 
For an ontology, I think schema.org makes most sense. I'm not even sure we need the bib extension, as the regular schema is very thorough for bibliographic data. So JSON-LD with schema.org support would be my way to go. I'm pretty sure that's what we're most likely to find on the import side, too.

If we want a new, rich RDF format, that should almost certainly be BibFrame, but that's very heavily oriented towards linked data, which doesn't jibe well with Zotero's current data model.

Those my not yet fully coherent thoughts.
Making sure @zuphilip sees this, as he knows a ton about metadata formats.

zuphilip:
Nice to hear the support on this. Which vocabulary to use might also depend on our goal: do we want the most widespread vocabulary or do we want the most detailed vocabulary?

[...]

PR #1314, opened 2017-06-02
[DISCUSS] Schema.org translator

zuphilip:
This schema.org translator was done by @tgr during the Hackday at WikiCite 17.

This PR is to have some place to discuss how to continue from here.

### DIFF
--- a/Schema.org.js
+++ b/Schema.org.js
@@ -1,0 +1,242 @@
+{
+	"translatorID": "856d17bd-14b5-4a29-9565-6d3abd9fb077",
+	"label": "Schema.org",
+	"creator": "Gergő Tisza",
+	"target": "",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 300,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2017-05-25 14:20:52"
+}
+
+/*
+ ***** BEGIN LICENSE BLOCK *****
+
+ Copyright © 2017 Gergő Tisza
+ http://zotero.org
+
+ This file is part of Zotero.
+
+ Zotero is free software: you can redistribute it and/or modify
+ it under the terms of the GNU Affero General Public License as published by
+ the Free Software Foundation, either version 3 of the License, or
+ (at your option) any later version.
+
+ Zotero is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU Affero General Public License for more details.
+
+ You should have received a copy of the GNU Affero General Public License
+ along with Zotero.  If not, see <http://www.gnu.org/licenses/>.
+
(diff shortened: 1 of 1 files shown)


tgr:
Thanks Philip!

This is very much WIP (only looks for JSON-LD, probably does not recognize all the possible formats, only uses a small subset of attributes, only tested on a singe site (bbc.com)). Before going forward with it, there are some general problems which should be considered:
* [JSON-LD](https://json-ld.org/spec/latest/json-ld/) documents can be very complex (multiple contexts, multiple ways to declare the type, named graphs, aliasing, nesting...) and parsing them by hand is not really maintainable. Something like [jsonld.js](https://github.com/digitalbazaar/jsonld.js) should be made available to translators.
* microdata and RDFa are more straightforward but would ideally be parsed to the exact same JS object as their JSON-LD equivalent, to keep the business logic readable. Haven't looked close enough to say whether that's easy or not. [RDFaJSON](https://niklasl.github.io/rdfa-lab/js/rdfa-json.js) and [microdata-node](https://github.com/Janpot/microdata-node) are two tools that claim to do this.
* A generic schema.org parser needs to coexist with page-specific translators. Right now the code in the patch will produce better results for many BBC pages than the dedicated bbc.com translator, but will be ignored due to that transator's lower priority. The current (not so great) workaround for that is that many transators internally invoke the generic translator (Embedded Metadata). Shoud the Schema.org translator be squashed into that one, or is there a better solution?

owcz:
@tgr the plan (#917) was to include it in EM, not a separate translator

Even basic handling would be appreciated in the short-run, since it's been so sorely needed for so long

zuphilip:
I remembered that there is also another approach by @dgerber in https://github.com/dgerber/translators/tree/microdata which looks promising.

dgerber:
To keep things DRY and manageable, I think it's important to separate two steps:
1. the *parsing* of specific formats -- JSON-LD, Microdata, RDFa, turtle, html headers, RDF/XML (this one is implemented in `Zotero.RDF`), etc.
2. the "semantic" translation, mapping vocabularies / ontologies -- Dublin core, schema.org, etc. -- to Zotero items.

Currently, the most complete vocabulary translation is done in RDF.js (and called from EM), which happens to also parse RDF/XML. So [that branch](https://github.com/dgerber/translators/tree/microdata) puts schema.org-related code (mostly) in RDF.js, and the parsing in Microdata.js. (I remember I could not find a lean, easily integrable and correct js microdata parser then, but not which library had which issue.)

It would be nice to make all generic metadata easily available to page specific translators. However some web pages have different/conflicting metadata in different formats (e.g. JSON-LD, repeated but with errors as microdata), so simply merging by default in `Embedded Metadata.js` what is parsed from JSON-LD or Microdata would break a few translators.

+1 to use a proper JSON-LD lib like jsonld.js.

zuphilip:
More examples for schema.org for publications:
* https://www.dukeupress.edu/steeped-in-heritage

HughP:
I have been working on expressing academic CVs in schema.org JSON+LD would those examples be helpful to test against? I know there's what one might build as a test and then there is real world data. I'm not sure if more item types and examples would be helpful.

PR #1325, opened 2017-06-08
Translator for Oxford Reference

### LINKED ISSUE #796: Add Oxford Reference translator
http://www.oxfordreference.com/

Unfortunately no good metadata format available, but the metadata seems to be structured ok on the page and designated by unique classes.

E.g. http://www.oxfordreference.com/view/10.1093/acref/9780199546572.001.0001/acref-9780199546572-e-0009

Re https://forums.zotero.org/discussion/41765/oxford-reference/

sonali0901:
Need some inputs on how to identify item type as I couldn't figure out any way to uniquely identify books and books sections.
Fixes #796

### DIFF
--- a/Oxford Reference.js
+++ b/Oxford Reference.js
@@ -38,14 +38,12 @@
 function detectWeb(doc, url) {
 	if (url.indexOf("/search") != -1 && getSearchResults(doc, true))
 		return "multiple";
-	else
-	{
+	else {
 		var body = doc.getElementsByTagName("body")[0];
 		if ((body.className).indexOf('dctype-oxencycl-entry') > -1) {
-   			return "bookSection";
-		}
-		else if ((body.className).indexOf('dctype-book') > -1) {
-   			return "book";
+			return "bookSection";
+		} else if ((body.className).indexOf('dctype-book') > -1) {
+			return "book";
 		}
 	}
 }
@@ -54,7 +52,7 @@ function getSearchResults(doc, checkOnly) {
 	var items = {};
 	var found = false;
 	var rows = ZU.xpath(doc, '//h3[@class="source"]/a[span[@class="title"]]');
-	for (var i=0; i<rows.length; i++) {
+	for (var i = 0; i < rows.length; i++) {
 		var href = rows[i].href;
 		var title = ZU.trimInternal(rows[i].textContent);
 		if (!href || !title) continue;
@@ -67,7 +65,7 @@ function getSearchResults(doc, checkOnly) {
 
 function doWeb(doc, url) {
 	if (detectWeb(doc, url) == "multiple") {
-		Zotero.selectItems(getSearchResults(doc, false), function (items) {
+		Zotero.selectItems(getSearchResults(doc, false), function(items) {
 			if (!items) {
 				return true;
 			}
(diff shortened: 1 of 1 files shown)


zuphilip:
I looked at your two examles and I would suggest to try to differentetiate the two type by looking at the classes in `body`, i.e. try something like this (not tested):
```js
var body = document.getElementsByTagName("body")[0];
var item;
if (body.className).indexOf('dctype-oxencycl-entry') > -1) {
   item = new Zotero.Item("encyclopediaArticle");
} else { // class is then 'dctype-book'
   item = new Zotero.Item("book");
}
```

sonali0901:
@zuphilip For pages that are locked and only fully visible after subscription have limited content in the abstract note (for eg. Accutane in test cases). Is it fine to leave it that way?

Oxford Reference.js
```diff
@@ -0,0 +1,226 @@
+{
+	"translatorID": "62415874-b53c-4afd-86e8-814e18a986f6",
+	"label": "Oxford Reference",
+	"creator": "Sonali Gupta",
+	"target": "http://www.oxfordreference.com/",
```
zuphilip: Start with `^https?` and escape the points, which means `\.` in Scaffold or `\\.` in the textfile directly.

Oxford Reference.js
```diff
@@ -0,0 +1,226 @@
...
+	
+	item.complete();	
+}/** BEGIN TEST CASES **/
+var testCases = [
+	{
+		"type": "web",
+		"url": "http://www.oxfordreference.com/view/10.1093/acref/9780199546572.001.0001/acref-9780199546572-e-0009",
+		"items": [
+			{
+				"itemType": "bookSection",
```
zuphilip: The title of the containing book should also be saved, i.e. here it should be "A-Z of Plastic Surgery".

zuphilip: For chapters we should also save the title of the book. This seems missing here.

Oxford Reference.js
```diff
@@ -0,0 +1,226 @@
...
+				"creators": [
+					{
+						"firstName": "Andrew",
+						"lastName": "Hodges",
+						"creatorType": "author"
+					}
+				],
+				"date": "2008",
+				"ISBN": "9780199546572",
+				"abstractNote": "Isotretinoin. The synthetic retinoid derivative 13-cis-retinoic acid (Accutane) used for severe Acne vulgaris. The dose is 1",
```
zuphilip: That is IMO fine, but maybe add the three dots as well...

Oxford Reference.js
```diff
@@ -0,0 +1,226 @@
...
+			}
+		]
+	},
+	{
+		"type": "web",
+		"url": "http://www.oxfordreference.com/view/10.1093/acref/9780199608218.001.0001/acref-9780199608218",
+		"items": [
+			{
+				"itemType": "book",
+				"title": "Concise Oxford Companion to English Literature - Oxford Reference",
```
zuphilip: We should clean the title such that the output is without the " - Oxford Reference".

Oxford Reference.js
```diff
@@ -0,0 +1,226 @@
...
+		else
+   			return "book";
+	}
+}
+
+function getSearchResults(doc, checkOnly) {
+	var items = {};
+	var found = false;
+	var rows = ZU.xpath(doc, '//span[@class="titlePart"]/a');
+	var rowsExtendedTitle = ZU.xpath(doc, '//span[@class="title"]');
```
sonali0901: I did this because with the first XPath only the tile of the word we searched was visible. For eg, I searched Shalimar the Clown and all the entries in the pop up had Salman Rushdie has the title. So I used the second XPath to concatenate the name of the book from where the reference came to make it easierl for the user to choose which one to save.

zuphilip: I think it would then be better to try something like:
```js
var rows = ZU.xpath(doc, '//span[@class="titlePart"]');
...
   var href = ZU.xpathText(rows[i], './a/@href');
   var title = ZU.trimInternal(rows[i].textContent);
```
(Note: this is untested code.)

sonali0901: The two spans have different class ids. I will check this though.

zuphilip: Okay, I looked closer now.

How about the xpath
```js
var rows = ZU.xpath(doc, '//h3[@class="source"]/a[span[@class="title"]]');
```
?

Oxford Reference.js
```diff
@@ -0,0 +1,226 @@
...
+	{
+		authors = authors.split(',');
+		for (var i in authors){
+			item.creators.push(ZU.cleanAuthor(authors[i], "author"));
+		}
+		
+	}
+	
+	item.attachments = ({
+		url: url,
```
zuphilip: (fixed comment here)

Oxford Reference.js
sonali0901: This xpath is giving the title as the book title(and not the section) in the pop-up window and the href is the book section page.

adam3smith:
Thanks!

PR #1327, opened 2017-06-11
Fix risURL formatting

mtd91429:
This removes the '/' from the end of the replacement string that generated a malformed url


mtd91429:
From the bug reported here: https://forums.zotero.org/discussion/66022

PR #1338, opened 2017-06-19
Improvements for Dataset & DOI handling

adam3smith:
Note for SAE: this appears currently broken due to invalid HTML
@zuphilip if you have time to review, that'd be great

### DIFF
--- a/Embedded Metadata.js
+++ b/Embedded Metadata.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2017-06-06 18:09:35"
+	"lastUpdated": "2017-06-16 18:52:30"
 }
 
 /*
@@ -144,7 +144,7 @@ function getPrefixes(doc) {
 			}
 		}
 	}
-	
+
 	//also look in html and head elements
 	var prefixes = (doc.documentElement.getAttribute('prefix') || '')
 		+ (doc.head.getAttribute('prefix') || '');
@@ -196,7 +196,7 @@ function processFields(doc, item, fieldMap, strict) {
 function completeItem(doc, newItem) {
 	// Strip off potential junk from RDF
 	newItem.seeAlso = [];
-	
+
 	addHighwireMetadata(doc, newItem);
 	addOtherMetadata(doc, newItem);
 	addLowQualityMetadata(doc, newItem);
@@ -205,7 +205,7 @@ function completeItem(doc, newItem) {
 	if(CUSTOM_FIELD_MAPPINGS) {
 		processFields(doc, newItem, CUSTOM_FIELD_MAPPINGS, true);
 	}
-	
+
 	newItem.complete();
 }
 
@@ -259,14 +259,13 @@ function init(doc, url, callback, forceLoadRDF) {
 
 			if(_prefixes[prefix]) {
 				var prop = tag.substr(delimIndex+1, 1).toLowerCase()+tag.substr(delimIndex+2);
-
 				//bib and bibo types are special, they use rdf:type to define type
 				var specialNS = [_prefixes['bib'], _prefixes['bibo']];
 				if(prop == 'type' && specialNS.indexOf(_prefixes[prefix]) != -1) {
 					value = _prefixes[prefix] + value;
 					prefix = 'rdf';
 				}
-				
+
 				// This debug is for seeing what is being sent to RDF
 				//Zotero.debug(_prefixes[prefix]+prop +"=>"+value);
 				statements.push([url, _prefixes[prefix]+prop, value]);
@@ -275,7 +274,7 @@ function init(doc, url, callback, forceLoadRDF) {
 				if(lcValue.indexOf('blogger') != -1
 					|| lcValue.indexOf('wordpress') != -1
 					|| lcValue.indexOf('wooframework') != -1
-				) {	
+				) {
 					generatorType = 'blogPost';
 				}
 			} else {
@@ -308,7 +307,7 @@ function init(doc, url, callback, forceLoadRDF) {
 			}
 		}
 	}
-	
+
 	if(statements.length || forceLoadRDF) {
 		// load RDF translator, so that we don't need to replicate import code
 		var translator = Zotero.loadTranslator("import");
@@ -317,13 +316,12 @@ function init(doc, url, callback, forceLoadRDF) {
 			_haveItem = true;
 			completeItem(doc, newItem);
 		});
-		
+
 		translator.getTranslatorObject(function(rdf) {
 			for(var i=0; i<statements.length; i++) {
-				var statement = statements[i];			
+				var statement = statements[i];
 				rdf.Zotero.RDF.addStatement(statement[0], statement[1], statement[2], true);
 			}
-
 			var nodes = rdf.getNodes(true);
 			rdf.defaultUnknownType = hwType || hwTypeGuess || generatorType ||
 				//if we have RDF data, then default to webpage
@@ -462,13 +460,13 @@ function addHighwireMetadata(doc, newItem) {
 		newItem.pages = firstpage +
 			( ( lastpage && ( lastpage = lastpage.trim() ) )?'-' + lastpage : '' );
 	}
-	
+
 	//fall back to some other date options
 	if(!newItem.date) {
 		newItem.date = getContentText(doc, 'citation_online_date')
 			|| getContentText(doc, 'citation_year');
 	}
-	
+
 	//prefer ISSN over eISSN
 	var issn = getContentText(doc, 'citation_issn') ||
 			getContentText(doc, 'citation_eIssn');
@@ -493,16 +491,16 @@ function addHighwireMetadata(doc, newItem) {
 
 	//add snapshot
 	newItem.attachments.push({document:doc, title:"Snapshot"});
-	
+
 	//store PMID in Extra and as a link attachment
 	//e.g. http://www.sciencemag.org/content/332/6032/977.full
 	var PMID = getContentText(doc, 'citation_pmid');
 	if(PMID) {
 		if(newItem.extra) newItem.extra += '\n';
 		else newItem.extra = '';
-		
+
 		newItem.extra += 'PMID: ' + PMID;
-		
+
 		newItem.attachments.push({
 			title: "PubMed entry",
 			url: "http://www.ncbi.nlm.nih.gov/pubmed/" + PMID,
@@ -525,16 +523,16 @@ function addOtherMetadata(doc, newItem) {
 		try {
 			var parsely = JSON.parse(parselyJSON);
 		} catch(e) {}
-		
+
 		if(parsely) {
 			if(!newItem.title && parsely.title) {
 				newItem.title = parsely.title;
 			}
-			
+
 			if(!newItem.url && parsely.url) {
 				newItem.url = parsely.url;
 			}
-			
+
 			if(!newItem.date && parsely.pub_date) {
 				var date = new Date(parsely.pub_date);
 				if(!isNaN(date.getUTCFullYear())) {
@@ -545,11 +543,11 @@ function addOtherMetadata(doc, newItem) {
 					}, true);
 				}
 			}
-			
+
 			if(!newItem.creators.length && parsely.author) {
 				newItem.creators.push(ZU.cleanAuthor(''+parsely.author, 'author'));
 			}
-			
+
 			if(!newItem.tags.length && parsely.tags && parsely.tags.length) {
 				newItem.tags = parsely.tags;
 			}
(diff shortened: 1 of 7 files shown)


Zenodo.js
```diff
@@ -114,38 +116,79 @@ function doWeb(doc, url) {
 	}
 }
 
+var zoteroType = {
+	"figure": "artwork",
+	 "article": "report"
+}
```
zuphilip: I would suggest to move this to the place below where it is actually used. Otherwise I will have forgotten what this is when reading the code below...

zuphilip: What about data set?

Zenodo.js
```diff
@@ -114,38 +116,79 @@ function doWeb(doc, url) {
...
+					item.creators[i].firstName = item.creators[i].lastName.replace(/.+?,\s*/, "");
+					item.creators[i].lastName = item.creators[i].lastName.replace(/,.+/, "");
+				}
+			}
+			
+			item.url = url;
+			if (abstract) item.abstractNote = abstract;
+			if (item.itemType == "document" && zoteroType[type]) {
+				item.itemType = zoteroType[type];
+			}
```
zuphilip: Maybe add a comment here, what these lines are doing and why.

Embedded Metadata.js
```diff
@@ -795,15 +793,28 @@ function finalDataCleanup(doc, newItem) {
...
 	//Cleanup DOI
 	if (newItem.DOI){
 		newItem.DOI =newItem.DOI.replace(/^doi:\s*/, "");
 	}
-	
+
+	// Add DOI to non-supported item types
+	if (newItem.DOI && !ZU.fieldIsValidForType("DOI", newItem.itemType)) {
+		if (newItem.extra){
+			newItem.extra += "\nDOI: " + newItem.DOI;
```
zuphilip: Maybe also check that we only do this once, i.e.
```js
if (item.extra.search(/(^|\n)DOI:/) == -1) {
```
as below.

Figshare.js
```diff
@@ -109,14 +109,19 @@ function scrape(doc, url) {
...
 				DOI = DOI.replace(/^doi:?\s*/i, "")
-				if (item.itemType == "journalArticle") {
+				if (ZU.fieldIsValidForType("DOI", item.itemType)) {
 					item.DOI = DOI;
 				}
-				else if (DOI) {
-					item.extra = "DOI:" + DOI;
+				else {
+					if (item.extra) {
+						if (item.extra.search(/^DOI:/) == -1) {
```
adam3smith: I haven't tested, but shouldn't ^DOI find the string at the beginning of any new line?

RDF.js
```diff
@@ -967,6 +967,16 @@ function importItem(newItem, node) {
 	
 	// type
 	var type = getFirstResults(node, [n.dc+"type", n.dc1_0+"type", n.dcterms+"type"], true);
+	
+	/**CUSTOM ITEM TYPE  -- Currently only Dataset **/
+	if (type && type.toLowerCase() == "dataset") {
+		if (newItem.extra) {
+			newItem.extra += "\ntype: dataset";
+		}
+		else newItem.extra = "type: dataset";
```
adam3smith: again, I need to test this, but I'm pretty sure type is correct here, as this is used by citeproc and hence relies on CSL json (which uses type, not itemType).

RDF.js
```diff
@@ -967,6 +967,16 @@ function importItem(newItem, node) {
...
+	}
+
+
```
adam3smith: because the RDF translator handles all other dublin core code.

Zenodo.js
```diff
@@ -37,41 +37,43 @@
...
 					return "report";
-				case "Theses":
+				case "thesis":
 					return "thesis";
-				case "Datasets":
+				case "dataset":
 					//change when dataset as itemtype is available
-				case "journal-articles":
-				case "preprints":
+					return "document";
```
adam3smith: so, I'm not sure about this: since we are using Extra anyway, there is no strong reason to use journal article anymore, is there? And in that case, using Document makes sense because it's rarely used otherwise -- the main reason being that it doesn't have its own CSL item type, but since we're changing the item type via Extra, that's not a problem. What do you think?

Zenodo.js
adam3smith: see above -- I'm leaving that as document on purpose (though feel free to disagree

Zenodo.js
```diff
@@ -114,38 +116,79 @@ function doWeb(doc, url) {
...
+			else {
+				item.attachments.push({url:url, title: "Zenodo Snapshot", mimeType: "text/html"})
+			}
+			for (var i = 0; i<tags.length; i++) {
+				item.tags.push(tags[i].content);
+			}
+			
+			//something is odd with zenodo's author parsing; fix it
+			for (var i = 0; i< item.creators.length; i++) {
+				if (!item.creators[i].firstName && item.creators[i].lastName.indexOf(",")!=-1) {
```
adam3smith: https://www.zenodo.org/record/569304/export/csl#.WUgtNsaQy70
but it's also in one of the tests. I don't understand the problem -- have double-checked data entry for this one and we did that exactly as recommended.

Zenodo.js
```diff
@@ -327,50 +353,50 @@ var testCases = [
...
+				"extra": "Argonne is operated by UChicago Argonne, LLC, for the U.S. Department of Energy, Office of Biological and Environmental Research under contract DE-AC02-06CH11357. The SGC is a registered charity (number 1097737) that receives funds from AbbVie, Bayer Pharma AG, Boehringer Ingelheim, Canada Foundation for Innovation, Eshelman Institute for Innovation, Genome Canada through Ontario Genomics Institute, Innovative Medicines Initiative (EU/EFPIA) [ULTRA-DD grant no. 115766], Janssen, Merck & Co., Novartis Pharma AG, Ontario Ministry of Economic Development and Innovation, Pfizer, São Paulo Research Foundation-FAPESP, Takeda, and the Wellcome Trust.\nDOI: 10.5281/zenodo.45756\ntype: dataset",
```
adam3smith: presumably it's in CSL "note" -- move to a note here?

Figshare.js
zuphilip: Hm... this could also be...

RDF.js
zuphilip: I haven't thought this through, but RDF is usually called by EM. Is there also another (common) use case for RDF translator?

Zenodo.js
zuphilip: Okay, I haven't seen this example. But I cannot find it in any test case...

(BTW I read the comment above also, that there is _always_ something wrong with zenodo's author metadata, which is not the case.)

Zenodo.js
zuphilip: A note sounds good.

RSC Publishing.js
```diff
@@ -73,18 +73,21 @@ function scrape(doc, type) {
...
-			if (item.DOI){
-				if (item.extra) item.extra += "\nDOI: " + item.DOI;
-				else item.extra = "DOI: " + item.DOI;
+		if (item.DOI && !ZU.fieldIsValidForType("DOI", item.itemType)){
+			if (item.extra) {
+				if (item.extra.search(/^DOI:/) == -1) {
+					item.extra += "\nDOI: " + DOI;
+				}
+			} else {
+				item.extra = "DOI: " + DOI;
```
adam3smith: it's needed because this is recognized by EM as a journal article so it doesn't move the DOI. We change that above.

Figshare.js
adam3smith: No, you were right; changing

RDF.js
adam3smith: here I was right: testing shows that "type: dataset" works, `itemType: dataset` doesn't. Changed in the documentation

RDF.js
adam3smith: Common no. General -- other RDF imports, including Zotero RDF, DC RDF, and Bibliontology RDF. Arguably those should all recognize the (valid) DC type dataset.

Zenodo.js
adam3smith: done

Zenodo.js
```diff
@@ -158,23 +201,21 @@ var testCases = [
...
+				"date": "June 3, 2016",
```
adam3smith: done

Zenodo.js
```diff
@@ -158,23 +201,21 @@ var testCases = [
...
 				"extra": "DOI: 10.5281/zenodo.54766",
-				"libraryCatalog": "zenodo.org",
-				"url": "http://zenodo.org/record/54766",
+				"libraryCatalog": "Zenodo",
+				"university": "Zenodo",
```
adam3smith: fixed

Zenodo.js
adam3smith: fixed

Embedded Metadata.js
adam3smith: Since there is no way for the Extra field to contain anything other than note (given the CSL import from Zotero) I just empty the field above and can do without all this code.

Zenodo.js
```diff
@@ -114,38 +116,79 @@ function doWeb(doc, url) {
...
-	var schemaType = ZU.xpathText(doc, '//div[@id="wrap"]//div[@itemscope]/@itemtype');
+	var abstract = ZU.xpathText(doc, '//meta[@name="description"]/@content');
+	var doi = ZU.xpathText(doc, '//meta[@name="citation_doi"]/@content');
+	var schemaType = ZU.xpathText(doc, '//div[contains(@class, "record-detail")]//div[@itemscope]/@itemtype');
+	var pdfURL = ZU.xpathText(doc, '//meta[@name="citation_pdf_url"]/@content');
+	var tags = ZU.xpath(doc, '//meta[@name="citation_keywords"]')
+	var cslURL = url.replace(/#.+/, "").replace(/\?.+/, "").replace(/\/export\/.+/, "") + "/export/csl"
+	//Z.debug(cslURL)
+	//	Z.debug(schemaType)
 	// Embedded Metadata
```
adam3smith: done

Zenodo.js
```diff
@@ -114,38 +116,79 @@ function doWeb(doc, url) {
...
-		trans.doWeb(doc, url);
+			//workaround while we don't have proper item type for data
+			if (schemaType.indexOf("Dataset") != -1) {
+				if (item.extra) {
+					item.extra += "\ntype: dataset";
+				}
+				else {
+					item.extra = "type: dataset";
+				}
+			}
```
adam3smith: I looked at this -- the CSL JSON importer doesn't even get this information. We'd have to go into the Zotero core code and I don't think that's necessarily a good idea given that they are workarounds.

Zenodo.js
adam3smith: done

Zenodo.js
```diff
@@ -37,41 +37,43 @@
 
 function detectWeb(doc, url) {
 	if (url.indexOf('/record/')>-1) {
-		var collections = ZU.xpath(doc, '//div[contains(@class, "metadata")]//a[@class="navtrail"]');
+		var collections = ZU.xpath(doc, '//div[@itemscope]//span[@class="pull-right"]/span[contains(@class, "label-default")]');;
```
adam3smith: done

Zenodo.js
```diff
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2016-06-13 10:18:29"
+	"lastUpdated": "2017-06-19 03:23:09"
```
adam3smith: done

Figshare.js
```diff
@@ -109,14 +109,19 @@ function scrape(doc, url) {
...
+							item.extra += "\nDOI: " + DOI;
+						}
+					} else {
+						item.extra = "DOI: " + DOI;
+					}
```
adam3smith: turns out the whole passage isn't necessary -- the DOI is in the RIS and the RIS translator does handle this now.

adam3smith:
Thanks, updated with your comments.

RSC Publishing.js
zuphilip: We could also change the `itemType` in the translatorObject cf. https://github.com/zotero/translators/blob/master/Embedded%20Metadata.js#L815 and then (I assume) it is not necessary anymore. I.e.
```js

	translator.getTranslatorObject(function(trans) {
		trans.itemType = type;
		trans.doWeb(doc, url);
	});
```

Zenodo.js
```diff
@@ -116,36 +118,86 @@ function doWeb(doc, url) {
...
-			item.extra = "itemType: dataset\n" + "DOI: " + item.DOI;
-		}
-		item.complete();
-	});
-
-	translator.getTranslatorObject(function(trans) {
-		trans.itemType = type;
-		trans.doWeb(doc, url);
+			
+			if (!item.DOI) {
```
zuphilip:
```js
if (!item.DOI && doi) {
```

zuphilip:
This looks good. Just some suggestions/comments/replies.

PR #1348, opened 2017-07-02
Fix bug seriesTitle and series

helieh:
seriesTitle and series do not show after export item and then import that.

### DIFF
--- a/Zotero RDF.js
+++ b/Zotero RDF.js
@@ -244,7 +244,7 @@ function generateItem(item, zoteroType, resource) {
 		// set series type
 		Zotero.RDF.addStatement(series, rdf+"type", n.bib+"Series", false);
 		// add relationship to resource
-		Zotero.RDF.addStatement((containerElement ? containerElement : resource), n.dcterms+"isPartOf", series, false);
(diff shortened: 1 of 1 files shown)


Zotero RDF.js
```diff
@@ -244,7 +244,7 @@ function generateItem(item, zoteroType, resource) {
 		// set series type
 		Zotero.RDF.addStatement(series, rdf+"type", n.bib+"Series", false);
 		// add relationship to resource
-		Zotero.RDF.addStatement((containerElement ? containerElement : resource), n.dcterms+"isPartOf", series, false);
+		Zotero.RDF.addStatement(resource, n.dcterms+"isPartOf", series, false);
```
adam3smith: Thanks but I don't think this is the right fix -- if there is a container (i.e. a book or a journal) the series term should be applied to that with the isPartOf term, not to the article/chapter. This needs to get fixed on import, not on export.

adam3smith: ticket created to track this: https://github.com/zotero/translators/issues/1350

adam3smith:
I'm going to close this here, since import is in a different translator, but it'd be great if you'd take that one, obviously (I believe that should be `RDF.js`)

PR #1349, opened 2017-07-02
use marcxml for importing metadata from BSB OpacPlus

bkroll:
Just a few lines to use the MarcXML data provided through the OPACplus of Bavarian State Library instead of COinS which usually provides more detailed results and avoids punctuation problems if books have a main title and a subtitle.

This is my first try at writing Zotero translators, so any suggestions for improvements would be helpful. I'm looking forward to expand the translator for other media types or listings as well, if time permits.

### DIFF
--- a/BayerischeStaatsbibliothekOpacPlus.js
+++ b/BayerischeStaatsbibliothekOpacPlus.js
@@ -1,0 +1,37 @@
+{
+	"translatorID": "6b09b92d-f47e-4573-8a00-74245f7c6c5e",
+	"label": "Bayerische Staatsbibliothek (OpacPlus)",
+	"creator": "bkroll",
+	"target": "^https://opacplus.bsb-muenchen.de/metaopac/singleHit.do.*",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "g",
+	"lastUpdated": "2017-07-02 19:51:29"
+}
+
+function detectWeb(doc, url) {
+	mediaTypeIcon = doc.getElementById("mediaTypeIcon");
+	mediaTypeNode = mediaTypeIcon.getElementsByTagName("img");
+	mediaType = mediaTypeNode[0].getAttribute("alt");
+
+	if (mediaType == "Buch") {
+		return "book";
+	}
+}
+
+/** get MarcXml representation from opac dataset and use MarcXml translator **/
+function doWeb(doc, url) {
+	marcxmlLinkNode = doc.getElementById("marcLink");
+	marcxmlLinkString = marcxmlLinkNode.getAttribute("href");
+	Zotero.Utilities.doGet("https://opacplus.bsb-muenchen.de" + marcxmlLinkString, translateMarcxml);
+}
+
+function translateMarcxml(marcxmlString, responseObject, responseUrl) {
+	var marcxmlTranslator = Zotero.loadTranslator("import");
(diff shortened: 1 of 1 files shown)


zuphilip:
@bkroll Can you influence the catalogue of the BSB? My idea is that possibly this could be handled with the [unAPI translator](https://github.com/zotero/translators/blob/master/unAPI.js), when the marclink will changed such that it is detected.

For your translator here, some test cases would be good to have and can be easily added and updated with [Scaffold](https://github.com/zotero/scaffold).

adam3smith:
we also really like multiples (i.e. search results) to work for translators... Should be terribly hard to do here.

bkroll:
@zuphilip Not at all, this is just a private project.  But ok, I'll be adding some tests to the translator.

@adam3smith I'll have a look at multiples. The relevant ID number seems to be available for each list element in the HTML source, so requesting a MarcXml version of the item's metadata should work.

BayerischeStaatsbibliothekOpacPlus.js
```diff
@@ -0,0 +1,37 @@
+{
+	"translatorID": "6b09b92d-f47e-4573-8a00-74245f7c6c5e",
+	"label": "Bayerische Staatsbibliothek (OpacPlus)",
+	"creator": "bkroll",
+	"target": "^https://opacplus.bsb-muenchen.de/metaopac/singleHit.do.*",
```
AbeJellinek, line 5: Escape `.`s, remove the `.*` at the end.

BayerischeStaatsbibliothekOpacPlus.js
```diff
@@ -0,0 +1,37 @@
...
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "g",
+	"lastUpdated": "2017-07-02 19:51:29"
+}
+
+function detectWeb(doc, url) {
+	mediaTypeIcon = doc.getElementById("mediaTypeIcon");
```
AbeJellinek, line 16: These shouldn't be global variables - add `var`

BayerischeStaatsbibliothekOpacPlus.js
```diff
@@ -0,0 +1,37 @@
...
+	mediaTypeNode = mediaTypeIcon.getElementsByTagName("img");
+	mediaType = mediaTypeNode[0].getAttribute("alt");
+
+	if (mediaType == "Buch") {
+		return "book";
+	}
```
AbeJellinek, line 22: Add a `return false` if `mediaType == "Buch"` is false.

BayerischeStaatsbibliothekOpacPlus.js
```diff
@@ -0,0 +1,37 @@
...
+}
+
+/** get MarcXml representation from opac dataset and use MarcXml translator **/
+function doWeb(doc, url) {
+	marcxmlLinkNode = doc.getElementById("marcLink");
+	marcxmlLinkString = marcxmlLinkNode.getAttribute("href");
+	Zotero.Utilities.doGet("https://opacplus.bsb-muenchen.de" + marcxmlLinkString, translateMarcxml);
```
AbeJellinek, line 29: `doGet` will handle a relative URL fine, so no need to concatenate it with the domain here.

AbeJellinek (changes requested):
Hey, any updates on this? Left a few comments (in addition to what others have mentioned). I would love to get this merged.

AbeJellinek:
Closing this for lack of activity.

Issue #1353, opened 2017-07-05
Update import translators for 5.0


dstillman:
Import translators in 5.0 should account for the possibility of an `item.complete()` that returns a promise. This allows for items to be saved to disk individually using the asynchronous database access in 5.0. These translators should also return a promise from `doImport()`.

5.0 will assume that all import translators with a `lastUpdated` from today or later support asynchronous saving. For translators that haven't been updated, items will be added to a queue and saved together at the end of the import process, which for large imports can hang the UI during the import process.

The native Promise object is available in all modern browsers, but native promises aren't currently available in child sandboxes or in some older browsers (e.g., Safari under 10.8), so translators should use `typeof Promise == 'undefined'` to determine whether `doImport` can return a promise. (I don't think we can use a polyfill because `setTimeout()` also isn't available in the sandbox.)

For now, this won't affect web translators, including web translators that call import translators. Those will still queue items and save them at the end.

I'll post a PR shortly with an updated BibTeX translator as an example.

twsh:
Is someone updating BibLaTeX.js? If not, I will see if I can figure it out.

dstillman:
Yes, as long as they're not somehow called as child translators (which shouldn't be the case if you just directly triggering them with import/export processes).

retorquere:
Where it says "the possibility of an `item.complete()` that returns a promise", can I read that as "if `async: true` is set, `item.complete()` *will* return a promise"?

Can I still import items into collections as usual?

retorquere:
Can you walk me through what the async importers do, conceptually? I'm looking at the Zotero CSL-JSON translator as that seems the simplest translator that does async import, but I'm not fully grokking what is going on. I think that if `async` is set to `true` in the config options, `item.complete` will return a promise. Is that correct? Is `Promise.coroutine` available in the translators? Would I be mostly set by just wrapping `doImport` with `Promise.coroutine` and yielding on `item.complete()`?

PR #1354, opened 2017-07-05
Update translators for 5.0

dstillman:
Addresses #1353

### DIFF
--- a/BibTeX.js
+++ b/BibTeX.js
@@ -7,6 +7,7 @@
 	"maxVersion": "",
 	"priority": 200,
 	"configOptions": {
+		"async": true,
 		"getCollections": true
 	},
 	"displayOptions": {
@@ -18,7 +19,7 @@
 	"inRepository": true,
 	"translatorType": 3,
 	"browserSupport": "gcsv",
-	"lastUpdated": "2017-03-11 13:39:28"
+	"lastUpdated": "2017-07-05 19:32:38"
 }
 
 function detectImport() {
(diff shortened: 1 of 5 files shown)


dstillman:
OK, these are all I'm going to update for the moment, unless there are any others that people think are commonly used in large imports. I've tested them all in 5.0 and some of them in 4.0, but if anyone wants to test them, feel free. I'll merge tomorrow if no one has any concerns. (Append ?w=1 for a cleaner diff.)

BibTeX.js
```diff
@@ -889,30 +890,67 @@ function beginRecord(type, closeChar) {
...
+			if(read == "@") {
+				type = "";
+			} else if(type !== false) {
+				if(type == "comment") {
+					processComment();
+					type = false;
+				} else if(read == "{") {		// possible open character
+					// This might return a promise if an item was saved
+					// TODO: When 5.0-only, make sure this always returns a promise
+					let maybePromise = beginRecord(type, "}");
```
zuphilip: Can/should we use `let` in translator code which has to run in various browsers?

dstillman: Ugh, Safari 9. We'll be able to stop worrying about that (and hopefully start using various other features) after we drop support for 4.0, because 1) the new Safari connector requires macOS 10.11 anyway and 2) Zotero 5.0 requires 10.9, and all machines that can run 10.8 or later can be upgraded to 10.11. I think all other browser versions we care about support it (and have for a long time). But I've reverted these for now. Thanks.

PR #1355, opened 2017-07-06
Add "edition" to the typeMapping

mvolz:
Add edition to be treated like a book. Example item that is an edition: https://www.wikidata.org/wiki/Q14944010

### DIFF
--- a/Wikidata.js
+++ b/Wikidata.js
@@ -43,6 +43,7 @@ var typeMapping = {
 	"Q686822" : "bill",
 	"Q17928402" : "blogPost",
 	"Q571" : "book",
+	"Q3331189" : "book", // Edition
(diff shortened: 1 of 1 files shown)


zuphilip:
Looks good! But you should also bump the time-stamps for triggering an update to the Zotero users.

mvolz:
Thanks, done :).

adam3smith:
Thanks @mvolz 

@zuphilip the timestamp is another candidate for automated tests. If we could check whether anything other than tests have changed and then require a timestamp update?

zuphilip:
@adam3smith Good idea. I guess it should be possible by adapting the `checkDeletedTxt.sh`. Can you create an issue for this?

PR #1362, opened 2017-07-08
NYTimes: capture PDFs of archives

### LINKED ISSUE #1361: NYTimes: PDF capture
As mentioned in #1305 (@zuphilip)

* http://www.nytimes.com/1970/09/02/archives/aspen-colo-the-philosophers-at-bay.html?_r=0
* http://query.nytimes.com/gst/abstract.html?res=9406EFDF153DE532A25751C1A96F9C946791D6CF&legacy=true

Both have direct PDF links that should be captured when available. The current translator links to the PDFs (correctly, it appears) but shows a red X when attempting to download the file. I'd like to fix this myself (I have access) but I'm having major debugging issues.

I wasn't getting debug messages in my browser console and the debug messages in the Z Connector extension/add-on weren't giving me helpful information. I reset translators in Firefox and now Scaffold is not showing any available translators so I'm not sure if I should backup, move to the 5.0 beta, and start over. But is Scaffold working with 5.0? Is there some other place that logs the error when a translator shows a red X for an attachment?

I'm also not sure how to tell whether Z is resolving the PDF URL correctly. I see that it grabs the right URL (http://query.nytimes.com/mem/archive/pdf?res=9406EFDF153DE532A25751C1A96F9C946791D6CF) but it still isn't able to grab the file, though I can click through (or enter the URL directly) to get to it just fine. I also don't know whether there are any conflicts with the base embedded metadata translator, which is saving the HTML snapshot. Ideas?

owcz:
fixes #1361, formats libraryCatalog

### DIFF
--- a/NYTimes.com.js
+++ b/NYTimes.com.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2017-06-17 18:44:24"
+	"lastUpdated": "2017-07-09 03:45:18"
 }
 
 /*
@@ -72,6 +72,8 @@ function scrape(doc, url) {
 		item.language = ZU.xpathText(doc, '//meta[@itemprop="inLanguage"]/@content') || "en-US";
 		if (item.date) {
 			item.date = ZU.strToISO(item.date);
+		} else {
+			item.date = doc.querySelector('time').getAttribute('datetime');
 		}
 		if (item.itemType == "blogPost") {
 			item.blogTitle = ZU.xpathText(doc, '//meta[@property="og:site_name"]/@content');
@@ -81,7 +83,9 @@ function scrape(doc, url) {
 		}
 		//Multiple authors are just put into the same Metadata field
 		//we have to split this here
-		var authors = ZU.xpathText(doc, '//meta[@name="author"]/@content');
+		var authors = doc.querySelector('meta[name="author"], span[class^="Byline-bylineAuthor--"]').textContent;
+		if (authors == authors.toUpperCase()) // convert to title case if all caps
+			authors = ZU.capitalizeTitle(authors, true);
 		if (authors) {
 			item.creators = [];
 			var authorsList = authors.split(/,|\band\b/);
@@ -141,7 +145,6 @@ function scrape(doc, url) {
 	});
 	
 }
-
 
 function getSearchResults(doc, checkOnly) {
 	var items = {};
@@ -271,7 +274,6 @@ var testCases = [
 				],
 				"date": "2013-06-19",
 				"abstractNote": "At their core, are America’s problems primarily economic or moral?",
-				"libraryCatalog": "NYTimes.com",
 				"blogTitle": "Opinionator",
 				"language": "en-US",
 				"url": "https://opinionator.blogs.nytimes.com/2013/06/19/our-broken-social-contract/",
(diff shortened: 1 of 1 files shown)


NYTimes.com.js
```diff
@@ -90,14 +94,46 @@ function scrape(doc, url) {
...
 		item.url = ZU.xpathText(doc, '//link[@rel="canonical"]/@href') || url;
-		var pdfurl = ZU.xpathText(doc, '//div[@id="articleAccess"]//span[@class="downloadPDF"]/a[contains(@href, "/pdf")]/@href');
+		item.libraryCatalog = "NYTimes.com";
+		// Convert title/tags of NYT archives when in all caps: 
+		if (item.title == item.title.toUpperCase())
+				item.title = ZU.capitalizeTitle(item.title, true);
+		var allcaps = false;
+		for (i=0; i < item.tags.length; i++) {
+			if (item.tags[i] == item.tags[i].toUpperCase())
+				allcaps = true;
```
adam3smith: why go through this trouble? Just do `item.tags[i] = ZU.capitalizeTitle(item.tags[i], true);` here right away. no need for the allcaps variable and the second loop. (This also works at a tag level whereas your solution tries to fix all tags if one is allcaps -- which could just be an acronym).

NYTimes.com.js
```diff
@@ -90,14 +94,46 @@ function scrape(doc, url) {
...
+							title: "NYTimes Archive PDF",
+							mimeType: 'application/pdf',
+							url: authenticatedPDFURL
+						});
+					} else {
+						Z.debug("Could not find authenticated PDF URL");
+						item.complete();
+					}
+				},
+				function() {
```
adam3smith: why do you need this function? Why not just put its contens as part of the if above, i.e. after line 122.

NYTimes.com.js
owcz: I should have had the cases switches (assume all caps until proven otherwise)—I meant to avoid the acronym situation you mentioned. If I were to wipe `allcaps`/the second loop, wouldn't that lead to that exact situation (title-casing acronyms), as `ZU.capitalizeTitle` wouldn't be able to distinguish acronyms from all-caps tags?

adam3smith: ah yes, good point. Then what you suggest makes sense. Start with `allcaps = true` and then set it to false if there's one case where the tag is not identical to the uppercased one. You should then add `break;` so you don't unnecessarily run through all other tags.

NYTimes.com.js
owcz: There was an issue with the item completing before the PDF was retrieved but I'm not sure if that was this function or the `else` that follows it.

No preference, but I thought I might as well use `ZU.processDocuments`' [`onDone`](https://www.zotero.org/support/dev/translators/coding#saving_multiple_items) if it was built for this. `EBSCOhost.js` used `onDone` similarly too

adam3smith: ah I see; yeah, that makes sense then.

owcz:
Still doesn't handle http://www.nytimes.com/1985/10/20/books/in-the-spirit-of-68.html author
(or, for that matter, any of the NYT Book Review archives: http://www.nytimes.com/books/98/12/06/specials/holmes-footsteps.html )

adam3smith:
yeah, let's keep these separate. Feel free to create an issue for those. I'll try to review and merge this one asap.

adam3smith:
Thanks & sorry for the delay

PR #1363, opened 2017-07-08
Add Sacramento Bee

owcz:
tests text()/attr() polyfill from #1277

### DIFF
--- a/Sacramento Bee.js
+++ b/Sacramento Bee.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2017-07-08 22:56:15"
+	"lastUpdated": "2017-07-09 02:19:40"
 }
 
 /*
@@ -58,7 +58,10 @@ function scrape(doc, url) {
 	item.title = attr(doc,'[property="og:title"]','content');
 	item.date = text(doc,'.published-date');
 	item.abstractNote = text(doc,'#content-body- p');
-	item.tags = attr(doc,'meta[name="keywords"]','content').split(", ");
+	var keywords = attr(doc,'meta[name="keywords"]','content');
+	if (keywords) { // so as not to perform a split when keyword string is null
+		item.tags = keywords.split(", ");
+	}
 	item.attachments.push({
 		title: "The Sacramento Bee snapshot",
 		mimeType: "text/html",
@@ -67,7 +70,7 @@ function scrape(doc, url) {
 
 	// Authors
 	var authorMetadata = doc.querySelectorAll('.ng_byline_name');
-	if (authorMetadata) {
+	if (authorMetadata.length) { // querySelectorAll always retuns a NodeList, so test against length instead
 		// when authors are split between multiple selectors, combine them
 		var authorString = '';
 		for (i=0; i < authorMetadata.length; i++) {
@@ -88,7 +91,7 @@ function doWeb(doc, url) { // news & search pages supported, Feature support to 
 	if (detectWeb(doc, url) == "multiple") {
 		var items = {};
 		articles = [];
-		var titles = doc.querySelectorAll('h2.title a');
+		var titles = doc.querySelectorAll('#site-search-results .title a, .media-body .title a, #story-list .title a, .col-sm-7 .title a');
 		for (var i = 0; i < titles.length; i++) {
 			items[titles[i].href] = titles[i].textContent;
 		}
@@ -241,6 +244,16 @@ var testCases = [
 	{
 		"type": "web",
 		"url": "http://www.sacbee.com/search/?q=mcdavid",
+		"items": "multiple"
+	},
+	{
+		"type": "web",
+		"url": "http://www.sacbee.com/sports/",
+		"items": "multiple"
+	},
+	{
+		"type": "web",
+		"url": "http://www.sacbee.com/sports/nfl/san-francisco-49ers/",
 		"items": "multiple"
 	}
 ]
--- a/Sacramento Bee.js
+++ b/Sacramento Bee.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2017-07-09 02:19:40"
+	"lastUpdated": "2017-07-09 05:05:18"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


Sacramento Bee.js
```diff
@@ -0,0 +1,247 @@
...
+	item.tags = attr(doc,'meta[name="keywords"]','content').split(", ");
+	item.attachments.push({
+		title: "The Sacramento Bee snapshot",
+		mimeType: "text/html",
+		document: doc
+	});
+
+	// Authors
+	var authorMetadata = doc.querySelectorAll('.ng_byline_name');
+	if (authorMetadata) {
```
dstillman: `querySelectorAll()` always returns a `NodeList`, even if it's empty, so you would want to use `authorMetadata.length` here.

Sacramento Bee.js
```diff
@@ -0,0 +1,247 @@
...
+function scrape(doc, url) {
+	var item = new Zotero.Item("newspaperArticle");
+	item.websiteTitle = "The Sacramento Bee";
+	item.language = "en-US";
+	item.ISSN = "0890-5738";
+	item.url = url;
+	item.title = attr(doc,'[property="og:title"]','content');
+	item.date = text(doc,'.published-date');
+	item.abstractNote = text(doc,'#content-body- p');
+	item.tags = attr(doc,'meta[name="keywords"]','content').split(", ");
```
dstillman: `attr()` will return null if the selector isn't found, which would produce an error, so it'd be safer to assign it to a `keywords` variable and then do `if (keywords) { item.tags = keywords.split(", "); }`.

Sacramento Bee.js
```diff
@@ -0,0 +1,260 @@
...
+	***** END LICENSE BLOCK *****
+*/
+
+// attr()/text()
+function attr(doc,selector,attr,index){if(index>0){var elem=doc.querySelectorAll(selector).item(index);return elem?elem.getAttribute(attr):null}var elem=doc.querySelector(selector);return elem?elem.getAttribute(attr):null}function text(doc,selector,index){if(index>0){var elem=doc.querySelectorAll(selector).item(index);return elem?elem.textContent:null}var elem=doc.querySelector(selector);return elem?elem.textContent:null}
+
+function detectWeb(doc, url) {
+	if (url.match(/article\d+/)) {
+		return "newspaperArticle";
+	} else if (url.match(/\/(news|sports|entertainment)\//)) {
```
adam3smith: I'd be more comfortable if you implement a version of a `getSearchResults` function (using `querySelectorAll` instead of `ZU.xpath`, of course). These matches cover a broad set of pages and we really want to avoid false positives. There's a reason @zuphilip puts them in all translators.

Also, use `.search(/regex/)!= -1` for efficient tests of strings against regexs

Sacramento Bee.js
```diff
@@ -0,0 +1,260 @@
...
+		return "multiple";
+	} else if (url.indexOf("/search/?q=") != -1) {
+		return "multiple";
+	} else return null;
```
adam3smith: (no need for an explicit return null or false here)

Sacramento Bee.js
```diff
@@ -0,0 +1,260 @@
...
+	item.url = url;
+	item.title = attr(doc,'[property="og:title"]','content');
+	item.date = text(doc,'.published-date');
+	item.abstractNote = text(doc,'#content-body- p');
+	var keywords = attr(doc,'meta[name="keywords"]','content');
+	if (keywords) { // so as not to perform a split when keyword string is null
+		item.tags = keywords.split(", ");
+	}
+	item.attachments.push({
+		title: "The Sacramento Bee snapshot",
```
adam3smith: capitalize "snapshot" per convention.

Sacramento Bee.js
```diff
@@ -0,0 +1,260 @@
...
+						"lastName": "Walsh",
+						"creatorType": "author"
+					},
+					{
+						"firstName": "Sam",
+						"lastName": "Stanton",
+						"creatorType": "author"
+					}
+				],
+				"date": "January 08, 2015 1:09 PM",
```
adam3smith: use `ZU.strToISO` to avoid importing English dates into non-English locales.

Sacramento Bee.js
owcz: `.search` is more efficient than `.match`?

adam3smith: yes, because it just has to return true/false vs. an actual match. See https://jsperf.com/exec-vs-match-vs-test-vs-search/5 (`indexOf` is dramatically more efficient, so you could also use that for this one since you don't really need a regex, just three separate strings)

adam3smith: mozilla actually makes this distinction explicitin in the [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/search)

>When you want to know whether a pattern is found and also its index in a string use search() (if you only want to know it exists, use the similar test() method on the RegExp prototype, which returns a boolean); for more information (but slower execution) use match() (similar to the regular expression exec() method).

Sacramento Bee.js
```diff
@@ -0,0 +1,274 @@
...
+	***** END LICENSE BLOCK *****
+*/
+
+// attr()/text()
+function attr(doc,selector,attr,index){if(index>0){var elem=doc.querySelectorAll(selector).item(index);return elem?elem.getAttribute(attr):null}var elem=doc.querySelector(selector);return elem?elem.getAttribute(attr):null}function text(doc,selector,index){if(index>0){var elem=doc.querySelectorAll(selector).item(index);return elem?elem.textContent:null}var elem=doc.querySelector(selector);return elem?elem.textContent:null}
+
+function detectWeb(doc, url) {
+	if (url.search(/article\d+/) != -1) {
+		return "newspaperArticle";
+	} else if (url.search(/(\/((news|sports|entertainment)\/)|(search\/\?q=))|sacbee\.com\/?$/) != -1  && getSearchResults(doc, true)) {
```
adam3smith: @owcz -- this here is half the reason we want the `getSearchResults` function (or something like it): by including it in `detectWeb`, you make sure you don't get false positives. Even if you're quite confident that you have the site covered as it is now, imagine they change their CMS -- with things as you had them, you could get false positives in all sorts of places, including on article pages where the generic metadata translator might perform OK otherwise.

adam3smith:
This is good to merge, but I'll hold off for a little to see if the discussion in #1277 changes anything.

Sacramento Bee.js
owcz: I went to test this case myself and [it looks like](https://jsperf.com/zotero/1) `test()` becomes (15%) more efficient than multiple `indexOf()`s when comparing for multiple strings

Sacramento Bee.js
```diff
@@ -40,9 +40,9 @@
 function attr(doc,selector,attr,index){if(index>0){var elem=doc.querySelectorAll(selector).item(index);return elem?elem.getAttribute(attr):null}var elem=doc.querySelector(selector);return elem?elem.getAttribute(attr):null}function text(doc,selector,index){if(index>0){var elem=doc.querySelectorAll(selector).item(index);return elem?elem.textContent:null}var elem=doc.querySelector(selector);return elem?elem.textContent:null}
 
 function detectWeb(doc, url) {
-	if (url.search(/article\d+/) != -1) {
+	if (/article\d+/.test(url) != false) {
```
adam3smith: no need for `!= false` -- that's what `if` does aready.

adam3smith:
Cool, thanks!

PR #1366, opened 2017-07-10
Create Vice.js

owcz:
Handles every text-focused content site in the network. Some subsites (mainly i-D, Amuse i-D, and Vice News) vary significantly from the others, so some selectors were uglier than I'd prefer so as to accommodate. The rest more or less follow the same format. While i-D uses OpenGraph and might have benefited from using EM, the rest needed custom coding, so I thought it was best to build i-D support into a single Vice translator.

Need help with two issues:

1. Every snapshot I verified leads to a 404 page. Is there a workaround or should I just remove the snapshots altogether?
2. Some multis from the network sites that follow the same format have a weird issue:
https://broadly.vice.com/en_us
https://broadly.vice.com/en_us/topic/occult
https://waypoint.vice.com/en_us
https://waypoint.vice.com/en_us/latest
https://thump.vice.com/en_us/search?q=venetian
For instance, try any one of the above (all multis) and the saved articles will lack both author and tags, even though both do save properly if you run the scraper on the individual page (without the multi). There's something about the multi that leaves those two parameters out.

    Scaffold is strange too. Add any one of those site's individual articles to the test, click "save", and it scrapes fine, as it normally does in the browser. But then "run" any one of those same individual articles and it'll ignore the site's authors and tags (same as above).

    I tried to isolate the issue by removing everything that wasn't necessary for, say, broadly.vice.com's multi, but the scraper still returned null for each page's author/tags (even though the scraper returned the right results if run on the same pages individually). So not sure if it's an issue with the multi code or something with text()/attr() (#1277)

### DIFF
--- a/Vice.js
+++ b/Vice.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2017-07-10 08:03:12"
+	"lastUpdated": "2017-07-17 05:38:35"
 }
 
 /*
@@ -71,13 +71,7 @@ function scrape(doc, url) {
 			item.tags.push(keywords_i_D[i].getAttribute('content'))
 		}		
 	}
-	
-	// archives appear to 404 on load?
-	item.attachments.push({
-		title: (attr(doc,'meta[property="og:site_name"]','content') || "Vice")+ " Snapshot",
-		mimeType: "text/html",
-		document: doc
-	});
+	// see PR#1366: dynamic pages don't archive well, so skipping
 
 	// Authors – I haven't found an article with more than one author, but share if you do
 	var authors = attr(doc,'.contributor__name .contributor__link','title') || attr(doc,'meta[property="og:article:author"]','content')/* i-D */ || text(doc,'.post__speaker a')/* Vice News? */ || text(doc,'.header-info-module__info span')/* Amuse i-D */;
@@ -140,12 +134,6 @@ var testCases = [
 				"blogTitle": "Vice",
 				"language": "en",
 				"url": "https://www.vice.com/en_us/article/padaqv/anti-g20-activists-told-us-what-they-brought-to-the-protest-in-hamburg",
-				"attachments": [
-					{
-						"title": "Vice Snapshot",
-						"mimeType": "text/html"
-					}
-				],
 				"tags": [
 					"G20",
 					"Hamburg",
@@ -179,12 +167,6 @@ var testCases = [
 				"blogTitle": "Waypoint",
 				"language": "en",
 				"url": "https://waypoint.vice.com/en_us/article/bjxjbw/nina-freemans-games-really-get-millennial-romance",
-				"attachments": [
-					{
-						"title": "Waypoint Snapshot",
-						"mimeType": "text/html"
-					}
-				],
 				"tags": [
 					"Awkward Teenage Rituals",
 					"How Do You Do It",
@@ -218,12 +200,6 @@ var testCases = [
 				"blogTitle": "Vice",
 				"language": "de",
 				"url": "https://www.vice.com/de/article/59pdy5/wie-kaputte-handys-und-profite-die-g20-gegner-antreiben",
-				"attachments": [
-					{
-						"title": "Vice Snapshot",
-						"mimeType": "text/html"
-					}
-				],
 				"tags": [
 					"G20",
 					"Hamburg",
@@ -266,12 +242,6 @@ var testCases = [
 				"blogTitle": "i-D",
 				"language": "en",
 				"url": "http://i-d.vice.com/en_us/article/anish-kapoor-has-been-banned-from-using-yet-another-rare-paint",
-				"attachments": [
-					{
-						"title": "i-D Snapshot",
-						"mimeType": "text/html"
-					}
-				],
 				"tags": [
 					"Anish Kapoor",
 					"Art",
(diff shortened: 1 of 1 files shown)


Vice.js
```diff
@@ -0,0 +1,365 @@
...
+		for (var i=0; i<keywords.length; i++) {
+			item.tags.push(keywords[i].textContent)
+		}
+	} else if (keywords_i_D.length) { 
+		for (var i=0; i<keywords_i_D.length; i++) {
+			item.tags.push(keywords_i_D[i].getAttribute('content'))
+		}		
+	}
+	
+	// archives appear to 404 on load?
```
owcz: Snapshots on every page I checked will succeed, but when later checked, they all 404. Not sure if there's a workaround or if it would be better to leave the snapshots out altogether.

zuphilip: Saving this website also fails when I try to do this in the browser. They are using a lot of JavaScript and some iframe things which seems not so friendly for saving purposes. Moreover, I haven't found a link to a "print friendly version" which sometimes exists. I suggest to skip saving snapshots here.

zuphilip:
I guess that the problems you are describing are connected to some dynamic loading of the website or building the website out of some JSON together with some JavaScript calls...

When coming from a multiple to a single article, then I see that some JSON data is loaded:

vice-loading

Actually, it looks promising to use such API calls directly and work with that JSON data rather than scrape from the page directly. In some the examples it is enough to add `?json=true` to the url, e.g. https://www.vice.com/en_us/article/padaqv/anti-g20-activists-told-us-what-they-brought-to-the-protest-in-hamburg?json=true . Let me know if this helps.

owcz:
Hm, it appears inconsistent:

Works for
https://www.vice.com/en_us/article/padaqv/anti-g20-activists-told-us-what-they-brought-to-the-protest-in-hamburg?json=true
https://waypoint.vice.com/en_us/article/bjxjbw/nina-freemans-games-really-get-millennial-romance?json=true
https://waypoint.vice.com/en_us/latest?page=2&json=true
https://www.vice.com/de/article/59pdy5/wie-kaputte-handys-und-profite-die-g20-gegner-antreiben?json=true
https://www.vice.com/de?json=true
https://thump.vice.com/en_us/search?q=venetian&json=true
Not
https://news.vice.com/story/voters-may-soon-toughen-up-americas-weakest-police-shootings-law?json=true
As above, I think i-D and i-D Amuse are on a different system
http://i-d.vice.com/en_us/article/anish-kapoor-has-been-banned-from-using-yet-another-rare-paint?json=true
https://amuse-i-d.vice.com/exclusive-around-the-world-with-goshas-favourite-artist?json=true

Do you see JSON beneath news.vice.com and the i-D sites?

zuphilip:
The news site on vice uses the [WordPress API](https://developer.wordpress.org/rest-api/reference/posts/), i.e. e.g. https://news.vice.com/wp-json/wp/v2/posts/24702

adam3smith:
Is this ready for review or are you still tweaking this?

owcz:
Still tweaking, thanks

owcz:
okay, ready for review
* rewritten to use JSON
* falls back to EM for News, i-D, and Amuse, and supplements with JSON-LD for News until it's officially built into EM

Vice.js
```diff
@@ -0,0 +1,438 @@
+{
+	"translatorID": "131310dc-854c-4629-acad-521319ab9f19",
+	"label": "Vice",
+	"creator": "czar",
+	"target": "^https?://((www|broadly|creators|i-d|amuse-i-d|impact|motherboard|munchies|news|noisey|sports|thump|tonic|waypoint)\\.)?vice\\.com",
```
adam3smith: I'd be OK with the simpler `^https://(.+?\\.)vice\\.com` unless you need to specifically exclude certain subdomains

Vice.js
```diff
@@ -0,0 +1,438 @@
...
+			tags[tags.indexOf(tag)] = ZU.capitalizeTitle(tag,true);	
+		}
+	}
+    return tags;
+}
+
+
+function detectWeb(doc, url) {
+	if (/\/(article|story)\//.test(url)) {
+		return "blogPost";
```
adam3smith: I would have thought of Vice as an online magazine, not as a series of blogposts -- not going to insist, but consider switching this.

To wits: Wikipedia has Vice as a magazine and it appears to have a ISSN

Vice.js
```diff
@@ -0,0 +1,438 @@
...
+	} else if (/\.vice\.com\/?($|\w\w(\_\w\w)?\/?$)|\/(search\?q=)|topic\/|category\/|(latest|read)($|\?page=)/.test(url) && getSearchResults(doc, true) ) {
```
adam3smith: I don't think you want that beginning `\.` before vice.com, do you? (doesn't match https://vice.com)

Vice.js
```diff
@@ -0,0 +1,438 @@
...
+		return "multiple";
+	} else if (attr(doc,'meta[property="og:type"]','content') == "article") { /* Amuse i-D */
+		return "blogPost";
+	}
+}
+
+
+function scrape(doc, url) {
+	var jsonURL = url+'?json=true';
```
adam3smith: You'll almost certainly want to clean the URL by doing `url.replace(/#.+/, "")` -- otherwise internal page navigation as well as social media origin info will get included in the jsonURL and will likely break it.

Vice.js
owcz: I could see potentially adding it for the base vice.com even though I don't know its overlap with the print circulation
But I'd consider all of its many subdomains to be more blog- than magazine-like in nature

Vice.js
owcz: my understanding is that all vice.com addresses redirect to www.vice.com but I suppose the translator's target detection would have caught any bad addresses by this point

Vice.js
owcz: https://motherboard.vice.com/en_us/article/43pngb/how-to-make-your-own-medicine-four-thieves-vinegar-collective?utm_campaign=sharebutton appears to work fine and I haven't seen `#` parameters used, but will add anyway

adam3smith:
Thanks!

PR #1371, opened 2017-07-16
Handle old pages of BBC

### LINKED ISSUE #1364: BBC: handle older pages
Pages such as http://news.bbc.co.uk/2/hi/uk_news/politics/1451221.stm are unhandled in current translator

sonali0901:
Fixes #1364
@mvolz @owcz  I need some ideas for the title of older pages. The title provided by metadata is not accurate and if I extract it through `ZU.xpathText(doc, '//meta[@name="Headline"]/@content')` then it would affect results of other pages. Any workaround? Refer to last test case to see the issue.

### DIFF
--- a/BBC.js
+++ b/BBC.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2017-07-16 12:37:59"
+	"lastUpdated": "2017-07-18 11:06:53"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


zuphilip:
AFAIS you can distinguish these two cases quite easily by analyzing the url. Thus, I suggest to use some conditional code, i.e. something like
```js
if (url.substr(-4)==".stm") {
   //only for old pages of BBC
   item.title = ZU.xpathText(doc, '//meta[@name="Headline"]/@content');
   item.section = ZU.xpathText(doc, '//meta[@name="Section"]/@content');
}
```

adam3smith:
What @zuphilip says, though looking at both this and the detectWeb, I think we don't want to require the .stm is at the end of the URL. E.g. some link shorteners add something like ?utm_campaing=mycampaignname to the end of URLs and there's really no reason we should break this. I think the safest would be to (again both here and in detect) clean the URL by doing `url.replace(/[\?#].+/, "")`

PR #1373, opened 2017-07-19
RIS: Make ignoreUnknown true by default

adomasven:
Hidden prefs are being restored in the Zotero Connectors. However, the connector won't include any default options (unlike Zotero for Firefox did for select few).

This PR adjusts `ignoreUnknown` to be `true` by default (as it used to be provided by Z4FX and is [documented](https://www.zotero.org/support/preferences/hidden_preferences#translator_preferences)) and adjusts usage of `getHiddenPref`.

### DIFF
--- a/RIS.js
+++ b/RIS.js
@@ -18,7 +18,7 @@
 	"inRepository": true,
 	"translatorType": 3,
 	"browserSupport": "gcsv",
-	"lastUpdated": "2017-07-05 19:32:38"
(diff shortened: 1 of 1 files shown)


adam3smith:
@adomasven you need to bump the translator timestamp so that this actually propagates to users. Otherwise looks good, feel free to merge

adomasven:
Timestamp updated. I don't have merge rights.

adam3smith:
Thanks!

PR #1381, opened 2017-07-28
Fix ScienceDirect PDF downloads

dstillman:
This loads the intermediate page and extracts the PDF URL from the meta refresh. The most annoying part of this is that the intermediate page URL from the JS drop-down doesn't seem to be in the DOM after the page has loaded, but it's in the initial HTML, so this refetches the HTML. There's probably a way to get the intermediate URL without doing that. (If we were really desperate we could simulate a click, but that's probably not necessary.)

This also uses a new method for determining whether the user has access to the PDF or would get a "Your organization may be charged for transactional access to this non-subscribed article." page (which has a Continue button, but presumably we don't want to touch that).

This could use some testing, particularly for the access-checking part — there may be a better way to test for that, or there may be other versions of that access box. (I didn't want to test for the text strings (e.g., "Get Full Text Elsewhere") because I'm not sure if they're localized.)

### DIFF
--- a/ScienceDirect.js
+++ b/ScienceDirect.js
@@ -66,6 +66,14 @@ function getPDFLink(doc, onDone) {
 		return;
 	}
 	
+	// If intermediate page URL is available, use that directly
+	var intermediateURL = attr(doc, '.PdfEmbed > object', 'data');
+	if (intermediateURL) {
+		//Zotero.debug("Embedded intermediate URL: " + intermediateURL);
+		parseIntermediatePDFPage(intermediateURL, onDone);
+		return;
+	}
+	
 	// For pages that use the new JS drop-down, the DOM doesn't seem to have the intermediate page URL
 	// after the page load, but it's in the initial HTML, so fetch the page again. This is awful, and
 	// we should try to find a better way of getting the URL (which is obviously available to JS in
@@ -79,24 +87,30 @@ function getPDFLink(doc, onDone) {
 		var intermediateURL = attr(doc, '.pdf-download-btn-link', 'href');
 		//Zotero.debug("Intermediate URL: " + intermediateURL);
 		if (intermediateURL) {
-			// Get the PDF URL from the meta refresh on the intermediate page
-			ZU.doGet(intermediateURL, function (html) {
-				doc = dp.parseFromString(html, 'text/html');
-				var pdfURL = attr(doc, 'meta[HTTP-EQUIV="Refresh"]', 'CONTENT');
-				//Zotero.debug("Meta refresh URL: " + pdfURL);
-				if (pdfURL) {
-					// Strip '0;URL='
-					var matches = pdfURL.match(/\d+;URL=(.+)/);
-					pdfURL = matches ? matches[1] : null;
-				}
-				onDone(pdfURL);
-				return;
-			});
+			parseIntermediatePDFPage(intermediateURL, onDone);
 			return;
 		}
 		onDone();
 	});
 }
+
+
+function parseIntermediatePDFPage(url, onDone) {
+	// Get the PDF URL from the meta refresh on the intermediate page
+	ZU.doGet(url, function (html) {
+		var dp = new DOMParser();
+		var doc = dp.parseFromString(html, 'text/html');
+		var pdfURL = attr(doc, 'meta[HTTP-EQUIV="Refresh"]', 'CONTENT');
+		//Zotero.debug("Meta refresh URL: " + pdfURL);
+		if (pdfURL) {
+			// Strip '0;URL='
+			var matches = pdfURL.match(/\d+;URL=(.+)/);
+			pdfURL = matches ? matches[1] : null;
+		}
+		onDone(pdfURL);
+	});
+}
+
 
 function getISBN(doc) {
 	var isbn = ZU.xpathText(doc, '//td[@class="tablePubHead-Info"]\
--- a/ScienceDirect.js
+++ b/ScienceDirect.js
@@ -84,9 +84,9 @@ function getPDFLink(doc, onDone) {
 		try {
 			pdfLink.click();
 			intermediateURL = attr(doc, '.PdfDropDownMenu li a', 'href');
-			var clickEvent = document.createEvent('MouseEvents');
+			var clickEvent = doc.createEvent('MouseEvents');
 			clickEvent.initEvent('mousedown', true, true);
-			document.dispatchEvent(clickEvent);
+			doc.dispatchEvent(clickEvent);
 		}
 		catch (e) {
 			Zotero.debug(e, 2);
--- a/ScienceDirect.js
+++ b/ScienceDirect.js
@@ -9,8 +9,11 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2017-06-22 16:54:54"
+	"lastUpdated": "2017-07-28 00:42:47"
 }
+
+// attr()/text() v2
+function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null}function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null}
 
 function detectWeb(doc, url) {
 	if (!doc.body.textContent.trim()) return;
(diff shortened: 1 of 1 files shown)


ScienceDirect.js
```diff
@@ -48,26 +51,51 @@ function detectWeb(doc, url) {
...
-			//Z.debug(classes);
-			if (mainPdf[0].title!="Purchase PDF" && classes.indexOf('excerptLink')==-1 && classes.indexOf('purchaseSprite')==-1 && classes.indexOf('ref-article-pdf')==-1 && classes.indexOf('cLink')==-1) {
-				pdfLink = mainPdf[0].href;
-			}
+	
+	// For pages that use the new JS drop-down, the DOM doesn't seem to have the intermediate page URL
+	// after the page load, but it's in the initial HTML, so fetch the page again. This is awful, and
+	// we should try to find a better way of getting the URL (which is obviously available to JS in
+	// the page).
+	var url = doc.location.href;
```
zuphilip: Add url as another parameter to the function `getPDFLink` instead?

ScienceDirect.js
```diff
@@ -48,26 +51,51 @@ function detectWeb(doc, url) {
...
-			var classes = mainPdf[0].class || '';
-			//Z.debug(classes);
-			if (mainPdf[0].title!="Purchase PDF" && classes.indexOf('excerptLink')==-1 && classes.indexOf('purchaseSprite')==-1 && classes.indexOf('ref-article-pdf')==-1 && classes.indexOf('cLink')==-1) {
-				pdfLink = mainPdf[0].href;
-			}
+	
+	// For pages that use the new JS drop-down, the DOM doesn't seem to have the intermediate page URL
+	// after the page load, but it's in the initial HTML, so fetch the page again. This is awful, and
+	// we should try to find a better way of getting the URL (which is obviously available to JS in
+	// the page).
```
zuphilip: The url of the intermediate page should be able to extract directly with the methods before, i.e. combine
```js
ZU.xpathText(doc, '//div[@class="PdfEmbed"]/object/@data');
ZU.xpath(doc, '//div[@id="articleToolbar" or @class="Toolbar"]//a[contains(@href, "main.pdf")]');
```

ScienceDirect.js
```diff
@@ -48,26 +51,51 @@ function detectWeb(doc, url) {
...
+		// TODO: Switch to HTTP.request() and get a Document from the XHR
+		var dp = new DOMParser();
+		var doc = dp.parseFromString(html, 'text/html');
+		var intermediateURL = attr(doc, '.pdf-download-btn-link', 'href');
+		//Zotero.debug("Intermediate URL: " + intermediateURL);
+		if (intermediateURL) {
+			// Get the PDF URL from the meta refresh on the intermediate page
+			ZU.doGet(intermediateURL, function (html) {
+				doc = dp.parseFromString(html, 'text/html');
+				var pdfURL = attr(doc, 'meta[HTTP-EQUIV="Refresh"]', 'CONTENT');
```
zuphilip: Nice!

ScienceDirect.js
dstillman: Not sure what you mean. Where in the current DOM are you seeing the intermediate page URL before "Download PDF" is clicked? A div.PdfDropDownMenu is added once it's clicked, but it's not available before that.

ScienceDirect.js
dstillman: Why?

ScienceDirect.js
```diff
@@ -48,26 +51,51 @@ function detectWeb(doc, url) {
...
+		return;
 	}
-	if (!pdfLink) {
-		pdfLink = ZU.xpathText(doc, '//div[@class="PdfEmbed"]/object/@data');
+	
+	// Some pages still have the PDF link available
+	var pdfURL = attr(doc, '#pdfLink', 'href');
+	if (pdfURL) {
+		onDone(pdfURL);
+		return;
```
dstillman: Good catch. It's bizarre how inconsistent this site is.

dstillman: Though now I'm confused, because I'm seeing id="pdfLink" for all the links that open the menu. Was the initial commit of this actually working? I added the pdfLink part after getting the intermediate-page stuff working, so maybe I broke it after that? Anyway, seems OK now after ignoring '#'.

dstillman:
Yeah, I think the ones that were passing for you were just pages that didn't actually require the intermediate page code.

Basically, I see:

  * "Download PDF"/"Download full text in PDF" w/menu and intermediate page (meta refresh to CDN) ([example](http://www.sciencedirect.com/science/article/pii/S0006349512000835)/[example](http://www.sciencedirect.com/science/article/pii/0584853976801316))
  * "Download PDF" with direct link (302 to CDN) ([example](http://www.sciencedirect.com/science/article/pii/S016748890800116X))
  * "Download PDF" with old PDF icon and no access ([example](http://www.sciencedirect.com/science/article/pii/B9780123706263000508) (for me, at least))

The first ones have `#` for the ``#pdfLink`` ``href`` and use the intermediate page code, the second ones use ``#pdfLink`` directly, and the third ones block on ``.accessContent``/``.access-options-link-text``.

(We should really have a way to annotate tests.)

ScienceDirect.js
zuphilip: If I open for example http://www.sciencedirect.com/science/article/pii/S2212671614001085 and then search in the DOM for `bbc70ac3c091bd4f313ecf311710367b` I see two hits:

firefox_2017-07-28_12-03-13

Does the DOM in your browser look different?

ScienceDirect.js
zuphilip: I remember that properties of `document` might not work for translation-server.

ScienceDirect.js
zuphilip: I tested the adapted version with ignoring the '#'.

ScienceDirect.js
dstillman: OK, that page works (and maybe all "Download full text in PDF" + menu pages). I was testing on the (newer?) "Download PDF" + menu pages, like http://www.sciencedirect.com/science/article/pii/S0006349512000835, which don't appear to have the link until you click on the menu.

I've updated it to get the immediate URL directly for the .PdfEmbed case. If you have an example of the other we can add support for that too (but we can probably push this out without that, since the only consequence is slower saving).

ScienceDirect.js
dstillman: We use that all over, so if something doesn't work in translation-server we should just fix it there — no need for redundant function arguments.

ScienceDirect.js
zuphilip: I thinks that was an example for the other case: http://www.sciencedirect.com/science/article/pii/S0014305710004763?via%3Dihub , but this seems to work now differently. (I just wanted to make sure that we take the previous code into account here, before it will vanish.)

dstillman:
OK, it's pretty ugly, but try that out on the menu pages — it simulates a mouse click to get the intermediate URL from the menu without doing an extra doGet (and then simulates a mousedown on the body to close the menu). Seems to work for me in all browsers (though PDF saving seems broken in Safari for unrelated reasons).

ScienceDirect.js
```diff
@@ -69,23 +69,45 @@ function getPDFLink(doc, onDone) {
...
+	// for the intermediate page, which doesn't seem to be available in the DOM after the page load.
+	// This is an awful hack, and we should look out for a better way to get the URL, but it beats
+	// refetching the original source as we do below.
+	var pdfLink = doc.querySelector('#pdfLink');
+	if (pdfLink) {
+		// Just in case
+		try {
+			pdfLink.click();
+			intermediateURL = attr(doc, '.PdfDropDownMenu li a', 'href');
+			var clickEvent = document.createEvent('MouseEvents');
```
zuphilip: Is this really `document` or rather `doc`? (Same question two lines below.)

I see in the debug output for [the first test case](http://www.sciencedirect.com/science/article/pii/S0896627311004430#bib5) these lines:
```
(2)(+0000144): ReferenceError: document is not defined
    ===== Stack Trace =====
    getPDFLink@C:\Users\zumstein\AppData\Roaming\Mozilla\Firefox\Profiles\eux6gyz4.coding\zotero\translators\ScienceDirect.js:87:8
    processRIS/<@C:\Users\zumstein\AppData\Roaming\Mozilla\Firefox\Profiles\eux6gyz4.coding\zotero\translators\ScienceDirect.js:357:3
    @resource://gre/modules/RemoteAddonsParent.jsm:781:66
    Zotero.Item.prototype.complete@resource://gre/modules/RemoteAddonsParent.jsm:781:415
    completeItem@C:\Users\zumstein\AppData\Roaming\Mozilla\Firefox\Profiles\eux6gyz4.coding\zotero\translators\RIS.js:1705:9
    importNext@C:\Users\zumstein\AppData\Roaming\Mozilla\Firefox\Profiles\eux6gyz4.coding\zotero\translators\RIS.js:1798:23
    startImport@C:\Users\zumstein\AppData\Roaming\Mozilla\Firefox\Profiles\eux6gyz4.coding\zotero\translators\RIS.js:1751:3
    doImport/<@C:\Users\zumstein\AppData\Roaming\Mozilla\Firefox\Profiles\eux6gyz4.coding\zotero\translators\RIS.js:1727:4
    doImport@C:\Users\zumstein\AppData\Roaming\Mozilla\Firefox\Profiles\eux6gyz4.coding\zotero\translators\RIS.js:1726:10
    @resource://gre/modules/RemoteAddonsParent.jsm:781:66
    processRIS@C:\Users\zumstein\AppData\Roaming\Mozilla\Firefox\Profiles\eux6gyz4.coding\zotero\translators\ScienceDirect.js:368:2
    scrape/<@C:\Users\zumstein\AppData\Roaming\Mozilla\Firefox\Profiles\eux6gyz4.coding\zotero\translators\ScienceDirect.js:473:5
    =======================
```

dstillman: Oops, sorry — I guess `document` works in the connectors.

zuphilip: Now, it works fine! (FYI I am testing with Scaffold and Zotero 4.0 in an old Firefox.)

dstillman:
Thanks!

Issue #1032, opened 2016-03-24
Citation keys and HTML markup in BibLaTeX translator


brisk022:
In biological sciences, species and gene names should be italicised. In titles, this is done with HTML tags. The translator then changes it to \textit{} in the output file. However, if the italicised text appears at the beginning of the title, the tags are included in the key, e.g. `Do <i>Arabidopsis halleri</i> from...` is translated to `bert_<i>arabidopsis_2002`. This causes problems downstream in particular with pandoc. Would it make sense to strip HTML before getting the first word of the title?

As a hack, I added a very permissive replacement. However, considering that '<' and '>' are allowed, perhaps something more restrictive would be more appropriate.

`var citeKeyHtmlRe = /<[^>]+>/g;` followed by
`return item["title"].toLowerCase().replace(citeKeyHtmlRe, "").replace(citeKeyTitleBannedRe, "").split(/\s+/g)[0];`

twsh:
I have the same issue with Pandoc. This commit to the bibtex.js translator fixed it for BibTeX: https://github.com/zotero/translators/commit/6d3e71da1b166d2457de8784336db99952b488e3

Is there any reason not to add to biblatex.js what was added there to the `citeKeyTitleBannedRe` variable in bibtex.js i.e. a new regex group `(<\/?(i|b|sup|sub|sc|span style=\"small-caps\"|span)>)`?

adam3smith:
Yes, that would be exactly the right approach. Happy to take a PR with that

adam3smith:
closed in https://github.com/zotero/translators/pull/1395

PR #1395, opened 2017-08-28
Remove markup from entry keys

twsh:
Markup will be excluded from keys. This matches the behaviour of BibTeX.js as of https://github.com/zotero/translators/commit/6d3e71da1b166d2457de8784336db99952b488e3

This addresses https://github.com/zotero/translators/issues/1032 and makes the BibLaTeX export compatible with [Pandoc](http://pandoc.org/).


adam3smith:
great, thanks!

PR #1409, opened 2017-09-12
Create FAO Publications.js (need help on multiple item pages)

swifterslb:
I create a translator for FAO Publications (http://fao.org/publications/ or http://fao.org/documents/). The translator works well for single item pages but I cannot get it pass through multiple item pages (e.g. http://www.fao.org/documents/en/). Please review the code and point out the errors. I have no previous experience in JavaScript (or any serious coding) so please tolerate the clumsy codes.

Thank you very much.

### DIFF
--- a/FAO Publications.js
+++ b/FAO Publications.js
@@ -2,14 +2,14 @@
 	"translatorID": "4883f662-29df-44ad-959e-27c9d036d165",
 	"label": "FAO Publications",
 	"creator": "Bin Liu <lieubean@gmail.com>",
-	"target": "^https?://www\\.fao\\.org/documents|publications/*",
+	"target": "^https?://www\\.fao\\.org/documents|publications/",
 	"minVersion": "3.0",
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
-	"translatorType": 12,
-	"browserSupport": "gcsib",
-	"lastUpdated": "2017-09-12 09:00:08"
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2017-09-16 21:24:44"
 }
 
 /*
@@ -33,7 +33,7 @@ function detectWeb(doc, url) {
 	//Just differentiate single and multiple. Correct itemType (either book or conferencePaper) will be determined in scrape().
 	if (url.indexOf('card') !== -1) {
 		return 'book';
-	} else {
+	} else if (getSearchResults(doc, true)) {
 		return 'multiple';
 	}
 }
@@ -81,7 +81,7 @@ function scrape(doc, url) {
 		newItem.url = pdfUrl.slice(pdfUrl.indexOf('www'));
 		//language: according to the last letter of PDF file name
 		var lang = pdfUrl.charAt(pdfUrl.indexOf('pdf')-2);
-		if (lang === 'a'){
+		if (lang === 'a') {
 			newItem.language = 'Arabic';
 		} else if (lang === 'c') {
 			newItem.language = 'Chinese';
@@ -104,7 +104,7 @@ function scrape(doc, url) {
 
 		//Variables that appear neither in all document pages nor at same positions in the pages.
 		var metaText = ZU.xpath(doc, '//*[@id="mainN1"]')[0].innerText.split('\n'); //scrape text of meta area and split into an array based on line breaks.
-		//get what variables are listed in the page, save to object existingTextVariable
+		//get what variables are listed in the page, save to object existingMeta
 		var textVariable = { //declarations for metadata names as appeared in document pages in different languages
 				date: ['سنة النشر', '出版年代', 'Year of publication', 'Année de publication', 'Год издания', 'Fecha de publicación'],
 				publisher: ['Publisher', 'Издательство'],
@@ -130,8 +130,7 @@ function scrape(doc, url) {
 				}
 			}
 		}
-		Z.debug(existingMeta);
-		
+
 		for (var key in existingMeta) {
 			var metaResult = cleanMeta(existingMeta[key]);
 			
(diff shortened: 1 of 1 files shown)


swifterslb:
Just did more tests and it appears that the multiple-item problem is browser-specific. Using Chrome (60.0.3112.113 x64) and Zotero Standalone (4 or 5), the saving multiple items function works well. Using Firefox (ESR, 52.3.0 x64), neither Zotero Standalone nor Zotero as FF add-on can save multiple items. Relevant error messages are the same as in Scaffold:

```
[JavaScript Error: "metaResult.match is not a function" {file: "jar:file:///C:/Users/liub/Downloads/PortableApps/PortableApps/FirefoxPortableESR/Data/profile/extensions/zotero@chnm.gmu.edu.xpi!/inject/translate_inject.js line 63 > eval" line: 148}]

[JavaScript Error: "ZU.xpath(...)[0] is undefined" {file: "jar:file:///C:/Users/liub/Downloads/PortableApps/PortableApps/FirefoxPortableESR/Data/profile/extensions/zotero@chnm.gmu.edu.xpi!/inject/translate_inject.js line 63 > eval" line: 74}]
```

adam3smith:
(Just to note that I have seen this and your note and zotero-dev and will get to it asap, but it might take as long as the weekend of the 23rd. I'll try to do it faster, though.)

FAO Publications.js
```diff
@@ -0,0 +1,489 @@
...
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+	***** END LICENSE BLOCK *****
+*/
+
+function detectWeb(doc, url) {
+	//Just differentiate single and multiple. Correct itemType (either book or conferencePaper) will be determined in scrape().
+	if (url.indexOf('card') !== -1) {
+		return 'book';
+	} else {
```
zuphilip: Test for multiples also that you have some results, i.e. with `getSearchResults(doc, true)`.

FAO Publications.js
```diff
@@ -0,0 +1,489 @@
...
+		found = true;
+		items[href] = title;
+	}
+	return found ? items : false;
+}
+
+function doWeb(doc, url) {
+	if (detectWeb(doc, url) == "multiple") {
+		Zotero.selectItems(getSearchResults(doc, false), function (items) {
+			Z.debug(items);
```
zuphilip: Delete or comment out this debug message in the end.

FAO Publications.js
```diff
@@ -0,0 +1,489 @@
...
+			ZU.processDocuments(articles, scrape);
+		});
+	} else {
+		scrape(doc, url);
+	}
+}
+
+
+//********** BEGIN TEST CASES **********
+var testCases = [
```
zuphilip: How did you do the test cases? It looks that you did not use [Scaffold](https://github.com/zotero/scaffold) for that...

zuphilip:
> The translator works well for single item pages but I cannot get it pass through multiple item pages (e.g. http://www.fao.org/documents/en/).

It seems that the website is loading dynamically data also after Zotero thinks that the website if fully loaded. There is the method `Z.monitorDOMChanges` for such cases, but it can only be used in the detection step and we would need it also in the scraping itself. A single item usually works, because the detection is only depending on the url and the data is then loaded before someone clicks the Zotero button. However, if you do the same from a multiple page, then Zotero tries to immediately after the website is loaded to scrape its data and fails, because at this time the document is still pretty empty. @dstillman Anything else we can in this situation?

From watching the network connections a little it seems that this website uses some (internal?) API. You could try to use that as well, but I don't know how complicated this would be...

foa-api

adomasven:
My knowledge of translator quality-control rules is limited, so maybe this is something we don't do at all, but this is my suggestion.

> It seems that the website is loading dynamically data also after Zotero thinks that the website if fully loaded. There is the method `Z.monitorDOMChanges` for such cases, but it can only be used in the detection step and we would need it also in the scraping itself.

Seem like the website loads the search results asynchronously via an XHR. Since `detectWeb()` calls have to return synchronously the options are limited here. You could return "multiple" if e.g. `document.querySelector('.searchbox')` returns an element (i.e. finding the searchbox, which means it's a search results page with likely multiple items) and then use `Z.monitorDOMChanges` in `doWeb()` to wait for the search results to load. This is slightly problematic if the results do not return anything, but still better than naught?

dstillman:
> A single item usually works, because the detection is only depending on the url and the data is then loaded before someone clicks the Zotero button. However, if you do the same from a multiple page, then Zotero tries to immediately after the website is loaded to scrape its data and fails, because at this time the document is still pretty empty. @dstillman Anything else we can in this situation?

Sorry, missed this.

Here's what I said on the dev list to a [similar question](https://groups.google.com/d/topic/zotero-dev/EnC5teZkSP8/discussion):

> There's a utility function (monitorDOMChanges()) that might help here, but you shouldn't use it — we're in the process of changing processDocuments() to use XMLHttpRequest + DOMParser instead of using a hidden browser, which will mean that processDocuments() will only ever have access to the initial page structure (and be much faster, since it won't have to actually load full webpages in the background and run JavaScript).
>
> The key thing in these cases is that the fact that the page is generated client-side means that you also have access to the same data it's using to generate the page, either inline as JSON or via an API request the page is making. Judging by the URL [which used a #], this page is likely doing the latter. So you'd want to use the Network pane of the browser dev tools to see what requests it's making and make the same request from the translator, probably for JSON that you can then use directly.
>
> Whether you still get the data from the page for a single-page save is up to you. The API request might be cached by the browser, so requesting it again from the translator might not result in an additional network request even for a single-page save, and it would let you use the same code and also get properly structured data instead of having to scrape the page. But whether that makes sense depends on the site.

adam3smith:
Thanks @dstillman -- I think it makes sense to accept the translator without multiples for now (which is why I disabled them) and look more closely at how we can get multiples to work (which may take some time). It might even make sense to wait until the change to XMLHttpRequest is made.

dstillman:
Yeah, that's fine — just wanted to note that `monitorDOMChanges()` won't even be an option for multiples going forward.

adam3smith:
Thank you!

Issue #1092, opened 2016-07-11
Generic webpage translator


dstillman:
As suggested on https://github.com/zotero/translation-server/issues/32, and further bolstered by https://github.com/zotero/zotero/issues/1059, we should create a translator that saves the basic data (title, URL, access date) on all webpages.

Some follow-up work will be needed in the client to show the gray icon for this translator ID, and probably some other things.

To allow this to be rolled out to 4.0 clients without causing trouble, we should figure out a way to return a value from `detectWeb` only in 5.0. Not sure if we make the Zotero version available now, but if we want to avoid that (e.g., for other consumers of translators), we could do some sort of feature detection.

(Ideally we could just use a `minVersion` here, but as far as I know the client won't ignore translators with later `minVersion`s when running detection, which would seem to make a lot of sense.)

zuphilip:
How about the other idea to extend the EM translator for this case? It looks for me that the function [addLowQualityMetadata](https://github.com/zotero/translators/blob/master/Embedded%20Metadata.js#L557-611) in the EM translator is similar to what you want to achieve. Thus, it might be already enough to extend the `detectWeb` in EM to always output `webpage` as a last option.

adam3smith:
+1 to @zuphilip 's question. Also, I don't understand why this must be limited to Zotero 5+ -- what am I missing?

dstillman:
> How about the other idea to extend the EM translator for this case?

It's what I say in the other thread: "even in the single-save-button era I still think there's value in setting different expectations for EM and &lt;title/&gt;".

> I don't understand why this must be limited to Zotero 5+ -- what am I missing?

Without client changes, the color icon would appear on every page, even for title/URL/accessDate, and there'd be a confusingly redundant set of options in the context menu (which hard-codes web-page saving right now).

zuphilip:
> It's what I say in the other thread: "even in the single-save-button era I still think there's value in setting different expectations for EM and &lt;title/&gt;".

Well, you explained that the _different colors_ serve some purpose. But if the EM translator can extract some data on a page (and it will save also a snapshot of that page), then I cannot think of any use for a lower quality website translator on this page. I guess, that we can also somehow color the icon for EM differently if we are in some low quality data case, maybe just if `detectionWeb` returns `website`.

dstillman:
We can just have the generic translator not show up in the context menu when the EM translator triggers. The point is that we can't distinguish between EM and generic data within a single translator, so they have to be separate.

[...]

PR #1434, opened 2017-10-06
Make EM the generic translator. Addresses #1092

adomasven:
As per [this comment](https://github.com/zotero/translators/issues/1092#issuecomment-333475693):

> - If EM only contains `webpage` data, run DOI translator's detectWeb within EM and if there are DOIs present, return `undefined`, otherwise return `webpage(Gray?)`.
> - Change special-case code within the Connector to always allow running the EM translator. If EM returns `undefined` then list EM options with and without snapshot as the final two options in the context menu, otherwise list the EM translator according to its priority.
> 
> If there are DOIs present, EM will not overshadow the DOI translator and otherwise will overtake. If both rich EM data and DOI are present then both can co-exist. This way we can avoid any changes or special translation handling within zotero and translation server and have _a translator_ for every page. It sacrifices code clarity in the intermediate term, but is a workable solution for the short-term until someone has the time and spirit to commit to the bigger change.

The changes in the connector are coming soon, but we should get this EM translator in circulation early. The consequence of this update is that connector will detect EM translator for all previous pages that only displayed webpage saving options.

Alternatively, we should at least merge the part in `importRDF()` that will allow the EM translator to save even when `init()` doesn't detect a viable item type, such that when push out the updated Connector and it tries to save with EM translation doesn't break.

Awaiting comments from @dstillman regarding connector related stuff.


dstillman:
My main concern here is duplicating the page scan for DOI. It's probably quite fast, but we obviously want to be very careful with translators that run on every page load, so we should at least do some profiling to see how many milliseconds it would be adding.

But I agree that we can merge the `importRDF` part first (and that will let us start using it in translation-server too, which is a priority). Have you tested the saving here? Any unexpected differences between the data from EM and the existing webpage option for a generic page?

adomasven:
> Have you tested the saving here? Any unexpected differences between the data from EM and the existing webpage option for a generic page?

I have tested it for functionality and it works! Otherwise, yes, there are somewhat significant, but expected differences between [EM](https://github.com/adomasven/translators/blob/21dbb16f0f9e6a54ef5109a8fa58f75a3ba8a0c6/Embedded%20Metadata.js#L587-L646) and the [current code](https://github.com/adomasven/zotero-connectors/blob/0850e4a2504a98fb5fe482764488cc9cbc84bc0b/src/common/translate_item.js#L73-L82) we have.

adomasven:
Okay, so I've pushed just the change that allows EM to save even when it doesn't detect any significant tags. Good to merge.

adomasven:
> My main concern here is duplicating the page scan for DOI. It's probably quite fast, but we obviously want to be very careful with translators that run on every page load, so we should at least do some profiling to see how many milliseconds it would be adding.

[See the scan code](https://github.com/adomasven/translators/blob/144099adc7867e70853f83dcb2680798e5a87df6/DOI.js#L23-L62). It is an XPath query and then running a regexp for each returned result. The load time we're adding for each page is proportional to the length of the page, at O(n log n). On [this webpage](http://www.egms.de/static/de/journals/mbi/2015-15/mbi000336.shtml) (which is one of the testcases for DOI `detectWeb()` takes 0.9ms in Chrome. I think we're okay.

adomasven:
@adam3smith please take a look at this when you have time

adam3smith:
Should I be seeing this?

> The changes in the connector are coming soon, but we should get this EM translator in circulation early. The consequence of this update is that connector will detect EM translator for all previous pages that only displayed webpage saving options.

Because that doesn't appear to be the case. e.g. still getting the grey Webpage icon on
http://www.example.com/

dstillman:
No, we took that out for the moment — the only thing necessary for this version is make sure we're OK applying the EM [low-quality save](https://github.com/adomasven/translators/blob/21dbb16f0f9e6a54ef5109a8fa58f75a3ba8a0c6/Embedded%20Metadata.js#L587-L646) to all webpages, in place of the current special-case webpage code (which just saves document.title, URL, and access date), noting that these would include pages where no effort was made to embed metadata. I haven't looked at the whole function, but I would say, for example, that we don't want to assign the hostname as Library Catalog for all webpages (or at all for EM saves). And I'm not sure if including the `keywords` field as tags is a good idea.

zuphilip:
> And I'm not sure if including the keywords field as tags is a good idea.

The tags will be emptied if EM is not called from another translator, see https://github.com/adomasven/translators/blob/21dbb16f0f9e6a54ef5109a8fa58f75a3ba8a0c6/Embedded%20Metadata.js#L821-L823

adomasven:
> The tags will be emptied if EM is not called from another translator, see https://github.com/adomasven/translators/blob/21dbb16f0f9e6a54ef5109a8fa58f75a3ba8a0c6/Embedded%20Metadata.js#L821-L823

We should probably do the same thing for `libraryCatalog`.

Otherwise the function adds:
- Title
- Creators (new for generic saves)
- Abstract note from `<meta name="description"/>` (new)
- Url
- Access date

Creators and abstract note quality will vary wildly. I'm not sure if even the worst possible abstract is worse than no abstract. A bad creator, on the other hand is worse than no creator when citing. So maybe add creators only for saves from another translator too?

dstillman:
We can look around, but I think we want to keep the author meta tag — I’m sure it’s often wrong, but it’s a legit field that, unlike `keywords`, doesn’t have a long history of (worthless) SEO abuse. Seems like people should be able to use it to specify the author of a generic webpage.

adomasven:
I've updated this to exclude libraryCatalog for non-child-EM-translator saves. There will be a [follow up](https://github.com/adomasven/translators/commit/5b439bb77bad385790c7fcfb065cb88a4ce7d306) after we merge some changes into the connector.

Awaiting approval.

adam3smith:
I'm sorry I'm dense here (and by all means feel free to merge this if it's blocking other things). But if you want me to look at it more closely, could someone give me a page where to see the changed behavior at work?

adomasven:
Ugh, sorry. The main change was supposed to be [this](https://github.com/zotero/translators/pull/1434/files#diff-f63cc1658234102c99332f39311ee592R363), although it seems that `new Zotero.Item(undefined)` actually defaults to 'webpage' either way, so aside from removing "libraryCatalog" (which, FWIW, doesn't even apply to 'webpage' items, i think?), the PR actually doesn't change anything. And it means that it doesn't block the connector changes either.

I'll push the initial changes to this PR and we can start over, although we need to merge the connector code first, so it will have to wait. Once again, sorry to bother you.

adam3smith:
no worries, thanks for clarifying.

PR #1437, opened 2017-10-07
Fix PDFs and more reliably find items

adam3smith:
cc @zuphilip


adam3smith:
Superseded by https://github.com/zotero/translators/pull/1438

PR #1440, opened 2017-10-08
New translator for www.springer.com

psisquared2:
I have written a translator for www.springer.com (not to be confused with link.springer.com). The embedded metadata translator is only capable of extracting the title of the book on that website. I have also implemented various fallbacks to the embedded metadata translator in case the website changes.

Example website: http://www.springer.com/gb/book/9780387952697

### DIFF
--- a/Springer Books.js
+++ b/Springer Books.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 12,
 	"browserSupport": "gcsbv",
-	"lastUpdated": "2017-10-22 19:04:17"
+	"lastUpdated": "2017-10-23 17:56:11"
 }
 
 function detectWeb(doc, url) {
@@ -89,18 +89,29 @@ function scrapeBook(doc, url) {
 			item.complete; return;
 		}
 		Z.debug("no creators found by Embedded Metadata translator");
-		//Split title into title and (presumably) subtitle, verify the subtitle from
+		//Split title into title and subtitle, verify both from
 		//a different field and complete the subtitle if neccessary;
 		if (item.title.indexOf(" - ") != -1) {
-			//match to the first " - "
-			titleParts = item.title.match(/^(.*?)(?:\s-\s)(.*)/);
-			//verify that the second entry is a subtitle via the bibliography part
-			var subtitleXpath = ZU.xpath(doc, '//div[@class="product-bibliographic"]/dl/dd/div/div/dl/dd');
-			for (var i=0; i<subtitleXpath.length; i++) {
-				if (shortenedStringTest(titleParts[2], subtitleXpath[i].textContent)) {
-					item.shortTitle = titleParts[1];
-					item.title = titleParts[1] + " - " + subtitleXpath[i].textContent;
+			//Step 1: decompose item.title into the "title" field from the bibliography
+			//and the (potentially abbreviated) subtitle
+			var titlesXpath = ZU.xpath(doc, '//div[@class="product-bibliographic"]/dl/dd/div/div/dl/dd');
+			var titleCand = null;
+			var subtitleCand = null;
+			for (var i=0; i<titlesXpath.length; i++) {
+				titleCand = titlesXpath[i].textContent;
+				if (titleCand == item.title.substring(0, titleCand.length)) {
+					subtitleCand = item.title.substring(titleCand.length).match(/^\s-\s(.+)$/);
 					break;
+				}
+			}
+			//Step 2: find the non-abbreviated subtitle from the "subtitle" field
+			if (subtitleCand) {
+				for (var i=0; i<titlesXpath.length; i++) {
+					if (shortenedStringTest(subtitleCand[1], titlesXpath[i].textContent)) {
+						item.shortTitle = titleCand;
+						item.title = titleCand + " - " + titlesXpath[i].textContent;
+						break;
+					}
 				}
 			}
 		}
--- a/Springer Books.js
+++ b/Springer Books.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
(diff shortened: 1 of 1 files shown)


psisquared2:
Thank you for the feedback. I have addressed the issues you mentioned, added support for multiples and also improved a few minor quirks on the way. Please let me know if you see further issues or room for improvement.

One question: In case the publisher is officially listed as "Springer-Verlag Berlin Heidelberg" (which happens a lot), do we want to split this into publisher="Springer-Verlag" and place="Berlin Heidelberg"? Seems a little weird to have two locations in "place", but it's probably the best we can do. The only other option I see is to revert to not filling the "place" field at all.

Springer Books.js
```diff
@@ -0,0 +1,293 @@
...
+			item.complete; return;
+		}
+		Z.debug("no creators found by Embedded Metadata translator");
+		//Split title into title and subtitle, verify both from
+		//a different field and complete the subtitle if neccessary;
+		if (item.title.indexOf(" - ") != -1) {
+			//Step 1: decompose item.title into the "title" field from the bibliography
+			//and the (potentially abbreviated) subtitle
+			var titlesXpath = ZU.xpath(doc, '//div[@class="product-bibliographic"]/dl/dd/div/div/dl/dd');
+			var titleCand = null;
```
zuphilip: I don't think this title e.g. `Theoretical Physics 7 - Quantum Mechanics - Methods and | Wolfgang Nolting | Springer` is a good text to extract any information from. How about the information at the bottom, e.g.

springer-bibliographic

? One should also check for the remaining lines here...

Springer Books.js
```diff
@@ -0,0 +1,293 @@
...
+	
+}
+
+//returns true if origStr is a shorter version of newStr
+function shortenedStringTest(origStr, newStr) {
+	if (origStr.length > newStr.length) return false;
+	if (origStr.length < 10)
+		return (origStr == newStr.substring(0, origStr.length));
+	else
+		return (origStr.substring(0,9) == newStr.substring(0,9));
```
zuphilip: I don't understand this function. The `9` looks like an arbitrary number here. Can you clarify?

psisquared2: This function is a plausibility check if one string is a shortened version of a different one, e.g. `Quantum Mechanics - Methods and ...` and `Quantum Mechanics - Methods and Applications`. It essentially compares the first 10 characters, or less if the string is shorter. See below why this is needed. One could, alternatively, check the first half of both strings; checking the full string fails in some cases where a "..." appears in the shortened string.

Springer Books.js
psisquared2: The problem is that the structure of this table is not consistent across different books, and only the author, publisher, and ISBN can be accessed in a stable way. I decided to first parse the information from the metadata because that is guaranteed to be correct, but might be abbreviated. In a second step I search the information at the bottom for the complete title, subtitle, and author and check these against the data I have from step 1. That is where the function `shortenedStringTest` from above is used. I found that this is the most stable thing one can do.

Springer Books.js
```diff
@@ -0,0 +1,293 @@
...
+
+function getSearchResults(doc, checkOnly) {
+	var items = {};
+	var found = false;
+	var search_entries = ZU.xpath(doc, '//div[@id="result-list"]/div');
+	for (var i=0; i<search_entries.length; i++) {
+		var item_type = search_entries[i].getAttribute("class");
+		//test if we actually have a usable result
+		if (!item_type || item_type.search(/result-item result-item-\d+ result-type-book.*/) == -1) continue;
+		var title_link = ZU.xpath(search_entries[i], "h4/a");
```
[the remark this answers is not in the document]
psisquared2: This html looks e.g. like
```html
<div class="result-item result-item-3 result-type-book">
  <some stuff>
  <h4>
    <a href="/us/book/978-981-10-0217-5">National Test</a>
  </h4>
```
and the xpath looks for a top level `h4` with an `a` below. Would you suggest a different xpath for that?

Springer Books.js
```diff
@@ -0,0 +1,293 @@
...
+	translator.setDocument(doc);
+	translator.setHandler("itemDone", function(obj, item) {
+		//extract information from the title field:
+		//item.title = <book title> | <first creator> | Springer
+		var title_entries = item.title.split(/\s\|\s/);
+		item.title = title_entries[0];
+		//forward compatibility: check if the embedded metadata translator
+		//has found a creator (i.e. relevant metadata)
+		if (item.creators.length > 0) {
+			item.complete; return;
```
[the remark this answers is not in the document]
psisquared2: My idea was that if Springer ever changes their website and decides to include more metadata, my translator would fall back to the embedded metadata translator.

adam3smith:
@zuphilip leaving this for you to review

zuphilip:
@adam3smith I don't think I can a review the next two weeks, sorry...

zuphilip:
@psisquared2: I simplified several things and updated the translator and pushed directly to your fork. Please have a look at the changes. This would be fine for me now to merge.

psisquared2:
@zuphilip Thank you for the improvements. The changes look fine. However, the tests failed on my machine (both Scaffold 3.3.0 release and the latest commit, which had identical outputs). I updated the tests to match Scaffold. Was there a reason for this discrepancy?

zuphilip:
Okay, thank you!

> I updated the tests to match Scaffold. Was there a reason for this discrepancy?

I did the test within your branch with all other files at this stage. I guess that you are using a newer EM file with the changes 7081d6b524772e408189266e83dd1ed069e99a6f . That would explain the language info. I don't know exactly why the spaces difference, but I guess we can neglect it for the moment.

zuphilip:
Thank you very much, @psisquared2 ! 🎉 

(We squashed the different commits together. Thus you might need to reset your master branch in order to have a clean base again.)

PR #1445, opened 2017-10-17
Hathi Trust missing colon

owcz:


zuphilip:
Thank you very much @owcz !

PR #1452, opened 2017-10-20
Update AMS MathSciNet.js target (#1451)

### LINKED ISSUE #1451: AMS MathSciNet -- Target needs update
According to report https://forums.zotero.org/discussion/68245/ams-mathscinet-broken#latest
all that's needed is switching www. --> mathscinet. in target, though haven't checked

gjankowiak:
changed initial subdomain from www. to mathscinet.. Also updated test cases.
Fixes #1451.

### DIFF
--- a/AMS MathSciNet.js
+++ b/AMS MathSciNet.js
@@ -161,7 +161,55 @@ var testCases = [
 	},
 	{
 		"type": "web",
-		"url": "http://mathscinet.ams.org/mathscinet-getitem?mr=2663710",
+		"url": "https://mathscinet.ams.org/mathscinet/search/publications.html?pg1=ISSI&s1=308850",
+		"items": "multiple"
+	},
+	{
+		"type": "web",
+		"url": "https://mathscinet.ams.org/mathscinet-getitem?mr=2767535",
+		"items": [
+			{
+				"itemType": "bookSection",
+				"title": "On implementation of the Markov chain Monte Carlo stochastic approximation algorithm",
+				"creators": [
+					{
+						"firstName": "Yihua",
+						"lastName": "Jiang",
+						"creatorType": "author"
+					},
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks, this looks great. It also looks like the page now defaults to https, so we should update that (it's already covered in the target, but needs updating on line 100 and in the tests)

gjankowiak:
I can take care of that, but not for the test going through libezproxy2.syr.edu, I don't have credentials and the certificate there doesn't match the subdomains. Shall I leave it out?

adam3smith:
That's a mistake on my part, actually. It should never have gone into the test. Just delete that bit from the test while you're at it please.

gjankowiak:
Ok, will do. I will leave the HTTP/HTTPS alternative in the target, it shouldn't hurt.

AMS MathSciNet.js
```diff
@@ -161,97 +161,12 @@ var testCases = [
 	},
 	{
 		"type": "web",
-		"url": "http://www.ams.org/mathscinet/search/publications.html?pg1=ISSI&s1=308850",
+		"url": "https://mathscinet.ams.org/mathscinet/search/publications.html?pg1=ISSI&s1=308850",
 		"items": "multiple"
 	},
 	{
 		"type": "web",
-		"url": "http://www.ams.org.libezproxy2.syr.edu/mathscinet-getitem?mr=2767535",
```
zuphilip: I wouldn't delete this test completely but rather change the url to a non-proxified url, i.e. https://mathscinet.ams.org/mathscinet-getitem?mr=2767535 . Otherwise this looks good.

adam3smith:
Thank you!

gjankowiak:
Thank you for fixing this quickly!

PR #1459, opened 2017-10-29
Fix notes in RIS

adam3smith:
They were exporting even with the preference unchecked.


dstillman:
Could you `git checkout -p` all the whitespace changes?

adam3smith:
done here, too

Issue #1456, opened 2017-10-23
Zotero RDF loses Extra field on roundtrip


adam3smith:

zuphilip:
Okay, here is an example

![zotero-rdf-extra](https://user-images.githubusercontent.com/5199995/32698050-1bbc414c-c79e-11e7-9a57-490acf9fa4a3.png)

exported to ZoteroRDF as
```rdf
<rdf:RDF
 xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
 xmlns:z="http://www.zotero.org/namespaces/export#"
 xmlns:dcterms="http://purl.org/dc/terms/"
 xmlns:bib="http://purl.org/net/biblio#"
 xmlns:foaf="http://xmlns.com/foaf/0.1/"
 xmlns:dc="http://purl.org/dc/elements/1.1/"
 xmlns:prism="http://prismstandard.org/namespaces/1.2/basic/">
    <bib:Article rdf:about="http://doi.acm.org/10.1145/2629489">
        <z:itemType>journalArticle</z:itemType>
        <dcterms:isPartOf rdf:resource="urn:issn:0001-0782"/>
        <bib:authors>
            <rdf:Seq>
                <rdf:li>
                    <foaf:Person>
                        <foaf:surname>Vrandečić</foaf:surname>
                        <foaf:givenname>Denny</foaf:givenname>
                    </foaf:Person>
                </rdf:li>
                <rdf:li>
                    <foaf:Person>
                        <foaf:surname>Krötzsch</foaf:surname>
                        <foaf:givenname>Markus</foaf:givenname>
                    </foaf:Person>
                </rdf:li>
            </rdf:Seq>
        </bib:authors>
        <dc:identifier>
            <dcterms:URI>
               <rdf:value>http://doi.acm.org/10.1145/2629489</rdf:value>
            </dcterms:URI>
        </dc:identifier>
        <dc:rights>Open Access</dc:rights>
        <bib:pages>78–85</bib:pages>
        <dc:date>September 2014</dc:date>
        <dc:description>OpenCitations ID: 3529929</dc:description>
        <dcterms:dateSubmitted>2017-08-12 12:50:03</dcterms:dateSubmitted>
        <z:libraryCatalog>ACM Digital Library</z:libraryCatalog>
        <z:language>eng</z:language>
        <dcterms:abstract>This collaboratively edited knowledgebase provides a common source of data for Wikipedia, and everyone else.</dcterms:abstract>
[...]

PR #1461, opened 2017-10-29
RDF: Import dc:description as Extra

adam3smith:
Closes https://github.com/zotero/translators/issues/1456
Fixes Zotero RDF roundtriping. 
Also remove some whitespace

### DIFF
--- a/RDF.js
+++ b/RDF.js
@@ -5,7 +5,7 @@
 	"creator": "Simon Kornblith",
 	"target": "rdf",
 	"minVersion": "2.1.9",
-	"maxVersion": null,
+	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
 	"browserSupport": "gcsibv",
@@ -13,7 +13,7 @@
 		"async": true,
 		"dataMode": "rdf/xml"
 	},
-	"lastUpdated": "2017-10-20 21:45:00"
+	"lastUpdated": "2017-10-27 21:45:00"
 }
 
 /*
--- a/RDF.js
+++ b/RDF.js
@@ -1,49 +1,49 @@
 {
 	"translatorID": "5e3ad958-ac79-463d-812b-a86a9235c28f",
+	"translatorType": 1,
 	"label": "RDF",
 	"creator": "Simon Kornblith",
 	"target": "rdf",
 	"minVersion": "2.1.9",
-	"maxVersion": "",
+	"maxVersion": null,
 	"priority": 100,
+	"inRepository": true,
+	"browserSupport": "gcsibv",
 	"configOptions": {
 		"async": true,
 		"dataMode": "rdf/xml"
 	},
-	"inRepository": true,
-	"translatorType": 1,
-	"browserSupport": "gcsibv",
-	"lastUpdated": "2017-10-20 21:15:21"
+	"lastUpdated": "2017-10-20 21:45:00"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


adam3smith:
@dstillman if you could take a very quick look at https://github.com/zotero/translators/pull/1461/files?w=1 to make sure there isn't something fundamentally objectional at the approach I'm taking, that'd be super helpful.

dstillman:
I'll look at the rest, but can you remove whitespace changes to start? (Shouldn't be in the same commit, and indenting empty lines is our standard code style anyway. We'll try to get an eslint setup everywhere to help with things like this.)

adam3smith:
Done

zuphilip:
(
> We'll try to get an eslint setup everywhere to help with things like this.

@dstillman The script here https://github.com/zotero/translators/blob/master/.ci/checkTranslator.sh#L108-L119 can produce some warnings w.r.t. JSHint, but we skip them in Travis CI currently here in translators, because most of the (old) translator would not pass that. And maybe the rules would need to be adjusted.
)

RDF.js
```diff
@@ -965,6 +969,12 @@ function importItem(newItem, node) {
 	// abstract
 	newItem.abstractNote = getFirstResults(node, [n.eprints+"abstract", n.prism+"teaser", n.prism2_0+"teaser", n.prism2_1+"teaser", n.og+"description",
 		n.bibo+"abstract", n.dcterms+"abstract", n.dc+"description.abstract", n.dcterms+"description.abstract", n.dc1_0+"description"], true);
+
+	//extra for Zotero RDF
```
zuphilip: Add one space after `//`.

RDF.js
```diff
@@ -1044,7 +1054,7 @@ function importItem(newItem, node) {
 	// description/attachment note
 	if(newItem.itemType == "attachment") {
 		newItem.note = getFirstResults(node, [n.dc+"description", n.dc1_0+"description", n.dcterms+"description"], true);
-	} else if (!newItem.abstractNote) {
+	} else if (!newItem.abstractNote && !isZoteroRDF) {
 		newItem.abstractNote = getFirstResults(node, [n.dc+"description", n.dcterms+"description"], true);
 	}
```
zuphilip: Can we move the lines from above here, such that all cases for `dc:description` are handled in one place? Something like (not tested):
```js
 	// dc:description
 	if (newItem.itemType == "attachment") {
 		newItem.note = getFirstResults(node, [n.dc+"description", n.dc1_0+"description", n.dcterms+"description"], true);
 	} else if (isZoteroRDF) {
		// for ZoteroRDF dc:description holds the data from extra
		newItem.extra = newItem.extra ? newItem.extra + getFirstResults(node, [n.dc+"description"], true) : getFirstResults(node, [n.dc+"description"], true);
	} else if (!newItem.abstractNote) {
 		newItem.abstractNote = getFirstResults(node, [n.dc+"description", n.dcterms+"description"], true);
 	}
```
Maybe even test it directly here without an additional variable.

zuphilip:
@dstillman Any additional comments from your side?

adam3smith:
@dstillman bump -- only a couple of lines to check over, should be quick.

zuphilip:
@dstillman I will merge this the next days. Please let me know soon when you don't agree...

zuphilip:
Thank you!

Issue #1400, opened 2017-08-30
BibLaTeX import


twsh:
There is a BibLaTeX export translator, and a BibTeX import/export translator. This means that if one imports a BibLaTeX file BibLaTeX field names are ignored. For example, the correct way to give the title of the journal for an article in BibLaTeX is 'journaltitle', rather than 'journal'. This won't be mapped to publication by Zotero. Both translators target .bib extensions, so it's not clear that just adding import to BibLaTeX.js would be the right thing to do. One idea that occurred to me would be to add some common BibLaTeX field names to the BibTeX translator. But then it wouldn't really be a BibTeX translator any more. Another option would be to have two exports and one unified import.

Does anybody have any suggestions about this? I think I will try to get something working for my own interest.

adam3smith:
sorry this never received an answer -- I don't think it would be hard to simply allow biblatex fields on import in the regular bibtex.js translator. This should be clearly marked in the code for readability, but otherwise, as long as they have different names, I don't think that'll cause any issues. 

IIRC there are a couple of fields that actually have different meanings in bibtex and biblatex -- we'd probably want to find a good heuristic (based on other fields present such as e.g. the dat format) to decide what to do for those fields. 

We haven't gotten a lot of requests for biblatex import so this isn't high priority for me, but I'd be very happy to take and review pull requests.

PR #1467, opened 2017-11-08
Allow import from BibLaTeX in BibTeX translator

twsh:
Allow import from biblatex: journaltitle fields are mapped to publication, location fields are mapped to address.

### DIFF
--- a/BibTeX.js
+++ b/BibTeX.js
@@ -12,14 +12,14 @@
 	},
 	"displayOptions": {
 		"exportCharset": "UTF-8",
-		"exportNotes": true,
+		"exportNotes": false,
 		"exportFileData": false,
 		"useJournalAbbreviation": false
 	},
 	"inRepository": true,
 	"translatorType": 3,
 	"browserSupport": "gcsv",
-	"lastUpdated": "2017-07-05 19:32:38"
+	"lastUpdated": "2017-11-08 20:02:00"
 }
 
 function detectImport() {
@@ -272,6 +272,8 @@ function processField(item, field, value, rawValue) {
 		}
 	} else if(inputFieldMap[field]) {
 		item[inputFieldMap[field]] = value;
+    } else if(field == "location") {
+        item.place = value;
 	} else if(field == "subtitle") {
 		if (!item.title) item.title = '';
 		item.title = item.title.trim();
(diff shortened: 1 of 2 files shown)


twsh:
Fixes https://github.com/zotero/translators/issues/1400

adam3smith:
Thanks - I've seen this, trying to think if there's a way we can mark field import as biblatex more clearly in the code and so make it easier to add/adjust others going forward rather than spreading the relevant code out throughout the translator.

We could e.g. have a separate biblatex mapping table that's used on import only.

twsh:
Yes, that would be clearer. But 'journaltitle' has to be outside the table anyway, just as 'journal' is in the original. So only 'location' could be moved.

BibTeX.js
```diff
@@ -12,14 +12,14 @@
 	},
 	"displayOptions": {
 		"exportCharset": "UTF-8",
-		"exportNotes": true,
+		"exportNotes": false,
```
zuphilip: Undo

BibTeX.js
```diff
@@ -272,6 +272,8 @@ function processField(item, field, value, rawValue) {
 		}
 	} else if(inputFieldMap[field]) {
 		item[inputFieldMap[field]] = value;
+    } else if(field == "location") {
+        item.place = value;
```
zuphilip: That is unnecessary, because there is already https://github.com/twsh/translators/blob/9de98ca98b5548c53987ed839f20bd20600ab61f/BibTeX.js#L173

BibTeX.js
```diff
@@ -292,6 +294,12 @@ function processField(item, field, value, rawValue) {
 		} else {
 			item.publicationTitle = value;
 		}
+	} else if(field == "journaltitle") {
+		if(item.publicationTitle) {
+			item.journalAbbreviation = value;
+		} else {
+			item.publicationTitle = value;
+		}
```
zuphilip: I don't think we need this logic here. The logic make sense for bibtex, where it is possible to have `journal` and `fjournal` and any of them are processed first. However, in BibLaTeX there is `shortjournal` and `journaltitle` for that and `journaltitle` is always mapped to `publicationTitle`. Thus, you can add it to the `inputFieldMap` in line https://github.com/twsh/translators/blob/9de98ca98b5548c53987ed839f20bd20600ab61f/BibTeX.js#L167 maybe with a comment that indicates that the following fields are for importing BibLaTeX.

zuphilip:
I created an alternative PR for this issue: #1504.

adam3smith:
superseded in https://github.com/zotero/translators/commit/7cc5ecbf0da8301920ffd225a6f71572f62fcacf
thanks!

PR #1472, opened 2017-11-09
Added translator for ISMRM proceedings

fsantini:
ISMRM (International Society of Magnetic Resonance in Medicine) hosts the proceedings of the annual conference in HTML format since 2016. This is a scraper translator that creates a zotero entry for abstracts, either from a search page or from a single abstract page.

### DIFF
--- a/ISMRMProceedings.js
+++ b/ISMRMProceedings.js
@@ -1,0 +1,148 @@
+{
+	"translatorID": "64f80311-85ec-4089-ae6d-665b535fb3b3",
+	"label": "ISMRM Proceedings",
+	"creator": "Francesco Santini",
+	"target": "http://indexsmart.mirasmart.com/ISMRM[0-9]+/",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "g",
+	"lastUpdated": "2017-11-09 15:10:54"
+}
+
+var places = {
+	'2016': 'Singapore',
+	'2017': 'Honolulu (HI, USA)',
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks. So at a high level -- this uses a lot of legacy code that makes it a) hard to maintain and b) not work well in some cases (in particular multiple saves). What did you base this on? 

The best current guidance is at https://www.mediawiki.org/wiki/Citoid/Creating_Zotero_translators
If you need Xpaths, use `ZU.xpath` and `ZU.xpathText` instead of `doc.evaluate`.  See https://www.zotero.org/support/dev/translators/coding

We'd also want tests and a license statement, please.

fsantini:
Thank you for the feedback. I based it on the tutorial from the official zotero pages... It's hard for a new developer to understand what guidelines to follow, and honestly I wouldn-t have imagined that the best practice was actually indicated on a different domain than the zotero one. I will cancel the pull request and rewrite the translator according to the page you indicated then.

⁣Sent from BlueMail ​

adam3smith:
The bit at wikimedia is just great because they just updated it to reflect best practices.

You mean you followed this? https://www.zotero.org/support/dev/how_to_write_a_zotero_translator_plusplus We should probably take it down entirely, but it does have a note that it's outdated at the top. Or is there something else that mentions `doc.evaluate` code everywhere?

zuphilip:
> I will cancel the pull request and rewrite the translator according to the page you indicated then.

@fsantini Any update on this?

fsantini:
No sorry, the proper documentation is not easy to find and motivation is low ATM as it works fine for me... If anybody wants to have a stab at it, you're of course very welcome. I'm closing this PR now. Thank you all for the feedback.

PR #1479, opened 2017-11-17
Update Springer Link.js

mbwhitfield:
Add "multiple" detection for the new format Springer Link book pages.

### DIFF
--- a/Springer Link.js
+++ b/Springer Link.js
@@ -403,7 +403,7 @@ var testCases = [
 	},
 	{
 		"type": "web",
-		"url": "http://link.springer.com/book/10.1007/978-3-540-88682-2/page/1",
+		"url": "http://link.springer.com/book/10.1007/978-3-540-88682-2",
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks -- note that you have to update the timestamp for translator changes to promulgate to users. We'd recommend just using Scaffold for any translator editing -- does this automatically:
https://github.com/zotero/scaffold

mbwhitfield:
Thanks; I think I tried scaffold a long time ago. I'll install it again if I work on another one.

Also just to mention, most of these books don't have html pages for front matter and back matter (e.g. https://link.springer.com/book/10.1007%2F978-3-642-17841-2). I would have liked to figure out how to point getResultList at the pdf link when the chapter title doesn't have a link, but I couldn't figure out how to make that happen. I know plenty of translators have the same issue. It's usually not a big deal, but occasionally you end up having to manually pull down something important, like an index.

PR #1480, opened 2017-11-18
Add ProQuest Ebook Central

adam3smith:
Reported/requested here: https://forums.zotero.org/discussion/67872/proquest-ebook-central#latest

The test won't run anywhere but at Syracuse -- the Ebook Central libraries are all taylored to institutions -- but I figured it'd be better to have some tests than none and they can be localized by replacing the Syracuse-specific part.

Everything else should be straightforward


zuphilip:
Thank you very much!

adam3smith:
Really enjoyed working with the built-in templates in Scaffold -- very nice work on those.

PR #1486, opened 2017-11-28
Rewrite National Library of Norway for new site

danmichaelo:
A new single page app site was released about a month ago. Unfortunately there's not much to scrape from the site itself anymore, but there's still the RIS export. The internal JSON API also looks promising, but there's been no announcement about the API being intended for public use and there's no documentation, so at least for now, I think the best idea is to stick with the RIS export as the base, and then enrich it with some data from the JSON API that we don't get from the RIS export.

Note: for now the only test cases are "books", since other media types aren't yet identified correctly. When viewing an item page, there is just no metadata in the DOM before you open a small "contents" overlay, and therefore no way to detect the type initially. There are a few meta tags added to head, but at the moment they don't contain the media type. Hopefully that can be changed at some point, but I don't expect anytime soon.

### DIFF
--- a/National Library of Norway.js
+++ b/National Library of Norway.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2016-12-28 12:25:50"
+	"lastUpdated": "2017-11-29 23:36:00"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


National Library of Norway.js
```diff
@@ -124,38 +199,82 @@ function scrape(doc, url) {
...
+				item.ISBN = m.identifiers.isbn10[0];
+			}
+
+			if (item.accessDate) {
+				// Better leave this to Zotero
+				delete item.accessDate;
+			}
+
+			if (m.identifiers.urn) {
+				item.url = 'https://urn.nb.no/' + apiResponse.metadata.identifiers.urn;
```
adam3smith: just making sure: all items with a URN are actual digital objects (like e-books) at that URL, right? If so, this is great. If not, we'd want to attach them as links, though.

National Library of Norway.js
```diff
@@ -35,42 +35,99 @@
...
 
 function doWeb(doc, url) {
-	if (detectWeb(doc, url) == "multiple") {
-		var items = ZU.getItemArray(doc, doc, '/nbsok/nb/');
+	if (detectWeb(doc, url) == 'multiple') {
+		var linkItems = ZU.xpath(doc, '//mat-card/a[contains(@href, "items")]');
+		var items = {};
+		linkItems.forEach(function(linkItem) {
+			var link = linkItem.getAttribute('href');
+			var title = '?';
```
adam3smith: what's the goal here? Are there empty titles?

National Library of Norway.js
```diff
@@ -35,42 +35,99 @@
...
+			var elems = ZU.xpath(linkItem, './/label[@class="title"]');
+			var dds = ZU.xpath(linkItem, './/dd').slice(0, 3);  // Limit to first three
+
+			// The search result list has two different layouts
+			if (elems.length) {
```
adam3smith: These seem complicated -- it wouldn't be easier to just do a `ZU.xpathText` over a reasonable xpath?

National Library of Norway.js
```diff
@@ -35,42 +35,99 @@
...
+					title += ' / ' + elem.textContent;
+				});
+			} else if (dds.length) {
+				// List view
+				title = dds[0].textContent;
+				dds.forEach(function(dd) {
+					title += ' / ' + dd.textContent;
+				});
+			}
+			items[link] = title;
```
adam3smith: test for presence of title & link here to be safe.

National Library of Norway.js
```diff
@@ -178,27 +297,99 @@ var testCases = [
...
 					"Foods",
 					"Gudbrandsdalen",
 					"History",
-					"Kokebøker",
 					"Lesja",
 					"Mattradisjoner",
 					"Merkedager",
 					"Preservation",
-					"Religiøse fester"
+					"Religiøse fester",
```
adam3smith: these are a lot of tags... are they actually useful?

National Library of Norway.js
```diff
@@ -178,27 +297,99 @@ var testCases = [
...
+					"slakting",
+					"stomp",
+					"takkebakst",
+					"tradisjoner",
+					"tradisjonsmat",
+					"tresking",
+					"vilt",
+					"ysting"
 				],
 				"notes": [
```
adam3smith: could we concat notes? Multiple notes, each with one sentence, aren't very useful.

National Library of Norway.js
danmichaelo: Will clean this up

National Library of Norway.js
danmichaelo: Will do

National Library of Norway.js
danmichaelo: That's my understanding, in the form of digitized material. Most Norwegian books have been scanned, but those not out of copyright are IP restricted to either Norway or the venues of the National Library, so you might end up with just a metadata record in practice, but I think the principle is that the URN lead to the digitized version.

National Library of Norway.js
danmichaelo: Not sure anymore. The list used to be quite short (probably controlled indexing), but now it seems to be more automatic extraction. I’m removing it.

National Library of Norway.js
danmichaelo: Ok

National Library of Norway.js
```diff
@@ -178,27 +297,99 @@ var testCases = [
...
+			}
+		]
+	},
+	{
+		"type": "web",
+		"url": "https://www.nb.no/items/280ecacb2d42ca595bc9970becc987f7?page=0&searchText=",
+		"items": [
+			{
+				"itemType": "book",
+				"title": "Barnehagedeltakelse og foreldrebetaling etter sosial bakgrunn : tabellnotat",
```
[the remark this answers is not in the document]
danmichaelo: This is actually from the RIS translator. Btw. NB also provides MODS, so I'm checking if that serialization is better.

danmichaelo:
#1488 and #1489 should best be merged before this. I might have to update my test cases afterwards.

National Library of Norway.js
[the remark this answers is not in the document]
danmichaelo: Fixed

National Library of Norway.js
danmichaelo: Done

National Library of Norway.js
```diff
@@ -98,64 +145,134 @@ function extractNumPages(str) {
...
+						"firstName": "Ludvig",
+						"creatorType": "author"
+					},
+					{
+						"lastName": "Rahbek",
+						"firstName": "K. L.",
+						"creatorType": "author"
+					},
+					{
+						"lastName": "Holberg, Ludvig, 1684-1754",
```
danmichaelo: This repetition is strange. Checking the MODS translator..

adam3smith:
@danmichaelo now with #1488 merged but slightly altered, could you check on the tests and either confirm that they pass or update? Otherwise this is ready to merge.

danmichaelo:
Done! I replaced the first test case with a less complicated one. It's not unlikely that better results could be obtained by using the (undocumented) default JSON serialization rather than the MODS serialization, but I'm fairly happy with the result now, and I prefer to use a standard format when possible :)

adam3smith:
Perfect, thanks (note, btw. that there's no need to rebase&squash on your part anymore. github just allows us to do this now)

PR #1488, opened 2017-11-29
MODS: Three changes

danmichaelo:
1. 
Suggesting to change the mapping for 'bibliography' from 'bookSection' to 'book'.

There are many kinds of bibliographies of course, but for the test cases provided it makes more sense to use 'book' than 'bookSection'.

I also had trouble with books that were *part of* a bibliography: ([source](https://api.nb.no/catalog/v1/metadata/a4cd69796dd3312c9780f5982a4a31f1/mods))

```xml
   <relatedItem displayLabel="Del av Bibliografien" type="host">
      <titleInfo>
         <title>Norbok</title>
      </titleInfo>
      <genre authority="marcgt">bibliography</genre>
   </relatedItem>
```

These were also imported as 'bookSection' because the MODS translators checks the genre of the host item if it doesn't find a genre in the main item.

---

2.
Include part number and title in title.

For e.g. multivolume books, the volume title is important to include. Added a new test case to demonstrate this.

---

3.
Removing the inference of "editor" from any person connected to the host item. This might work in some cases, but seems to risky to be applied as a general rule. For one example where this went wrong, see [here](https://github.com/zotero/translators/pull/1486/commits/06608d3ceb656891564226e44761d08cff6e7720#diff-ad2b8d29c41cf0a9eb388ac7edf97e78R256).


adam3smith:
@zuphilip this looks good to me, but I'd like a 2nd opinion if you have a minute.

adam3smith:
@zuphilip let me know if you're going to have time to look at this -- if not, I'll probably just go ahead & merge.

MODS.js
```diff
@@ -15,7 +15,7 @@
 	"inRepository": true,
 	"translatorType": 3,
 	"browserSupport": "gcsv",
```
zuphilip: @adam3smith While we are looking a this translator, can/should we activate all browserSupport?

MODS.js
```diff
@@ -1008,9 +1013,6 @@ function doImport() {
 				}
 			}
 			
-			// creators (might be editors)
-			processCreators(host, newItem, "editor");
```
zuphilip: You wrote that this is sometimes true, but not always. It looks unfortunate to delete these information completely. Can we save such persons more general as `"contributor"`s? Then, it would be much easier to correct it manually to an `"editor"` in the cases where it applies.

Can you give an example here again? (The url you provided does not work anymore.)

MODS.js
adam3smith: sure -- does browserSupport actually do anything on import translators, though?

zuphilip: Hm... the testing page looks different if `v` is missing, but otherwise maybe there is no consequence. However, I would still prefer to have them more consistently everywhere.

MODS.js
zuphilip: Okay, this is the example: https://github.com/zotero/translators/pull/1488/commits/a595853955f7e9acfcb5c2710773dd79387f32b6#diff-20e6379d9c22904e95b3cb80adab4353R1940 . However, this change will also influence the other test cases negatively.

Note that the "editor" here is just the default value, but in most cases we have better information, e.g.
```xml
  <relatedItem type="host">
    <name type="personal">
      <namePart type="given">Ash</namePart>
      <namePart type="family">Amin</namePart>
      <role>
        <roleTerm type="text">editor</roleTerm>
      </role>
    </name>
   ....
```
This is examined in the `processCreator` function, also marc codes and only when this still leads to nowhere, the default value is taken: https://github.com/zotero/translators/blob/master/MODS.js#L772-L792

zuphilip:
@adam3smith Please have a look at the new version.

BTW I am not sure how accurate the mapping of genres to itemTypes is overall, and whether it is a good idea to look also at the genres of the host elements thereby... I hope that "bibliography" is not that common and maybe "book" is a little better guess for that.

MODS.js
danmichaelo: Sorry I forgot to get back to you on this. The example is here: [MODS](https://api.nb.no/catalog/v1/metadata/URN:NBN:no-nb_digibok_2014012024006/mods) / [MARCXML](https://api.nb.no/catalog/v1/metadata/URN:NBN:no-nb_digibok_2014012024006/).

I can see the usefulness of including it if you're describing a chapter in an edited book, but would the title of the edited book be included as well then? You don't happen to have an example of a full mods record that demonstrates this?

zuphilip: Here is a full example (it is one of the test cases):
```xml
<mods xmlns:xlink="http://www.w3.org/1999/xlink" version="3.4" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://www.loc.gov/mods/v3" xsi:schemaLocation="http://www.loc.gov/mods/v3 http://www.loc.gov/standards/mods/v3/mods-3-4.xsd">
	<titleInfo>
		<title>Fifth-Grade Boys' Decisions about Participation in Sports Activities</title>
	</titleInfo>
	<name type="personal">
		<namePart type="family">Conley</namePart>
		<namePart type="given">Alice</namePart>
		<role>
			<roleTerm type="text">author</roleTerm>
		</role>
	</name>
	<typeOfResource>text</typeOfResource>
	<relatedItem type="host">
		<titleInfo>
			<title>Non-subject-matter Outcomes of Schooling</title>
		</titleInfo>
		<name>
			<namePart>Good, Thomas L.</namePart>
			<role>
				<roleTerm>editor</roleTerm>
			</role>
		</name>
		<note type="statement of responsibility">ed. Thomas L. Good</note>
		<originInfo>
			<issuance>continuing</issuance>
		</originInfo>
		<part>
			<detail type="title">
				<title>Special issue, Elementary School Journal</title>
			</detail>
			<detail type="volume">
				<number>99</number>
			</detail>
			<detail type="issue">
				<number>5</number>
				<caption>no.</caption>
			</detail>
			<extent unit="pages">
				<start>131</start>
				<end>146</end>
			</extent>
			<date>1999</date>
		</part>
	</relatedItem>
</mods>
```

danmichaelo: Ouch, I was sure I had all the test passing :O  Ok, I'm fine with keeping it as it is now then!

adam3smith:
Great, thanks

PR #1489, opened 2017-11-29
MODS: Include part number and title

danmichaelo:
For e.g. multivolume books, the volume title is important to include. Added a new test case to demonstrate this.


danmichaelo:
Included this in #1488 instead

PR #1494, opened 2017-12-09
Add Air University Journals

adam3smith:
ASJP and SSQ -- lot's of scraping, and authors and abstract can be
inconsistent, but rest should be solid.

### DIFF
--- a/Air University Journals.js
+++ b/Air University Journals.js
@@ -1,0 +1,145 @@
+{
+	"translatorID": "e317b4d4-03cf-4356-aa3c-defadc6fd10e",
+	"label": "Air University Journals",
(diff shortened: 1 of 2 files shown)


zuphilip:
The AGPL was missing and I added it in your name. Please have a look. I added also the functions `text` and `attr` which were missing. Moreover, I improved the authors handling, e.g. to extract all three authors in `Maj Gen (sel) Bradley D. Spacy, USAF / Mr. Edwin H. Oshiba, USAF / Capt Nicholas J. Thomas, USAF, PE`. Okay?

adam3smith:
Cool, thanks!

Issue #1485, opened 2017-11-26
Bookmarks translator fails to unescape HTML character references in title names.


ckush:
When importing bookmarks from a .html file, if the title of a bookmark has some characters as numeric references, such as &#39; or &amp;,  the Bookmarks.js translator fails to convert it to the original character.

It find that changing...

`openItem.title = title;`

in Bookmarks.js to...

`openItem.title = ZU.unescapeHTML(title);`

solves the problem. 

I realise this is a minor change but I'm new to programming and would love it if I'm allowed to go through the process of submitting this change myself, for educational purposes.

Below is a .html bookmarks file you can test this with. Note that collection names are imported fine, the problem is only in bookmark titles.

```
<!DOCTYPE NETSCAPE-Bookmark-file-1>
<!-- This is an automatically generated file.
     It will be read and overwritten.
     DO NOT EDIT! -->
<META HTTP-EQUIV="Content-Type" CONTENT="text/html; charset=UTF-8">
<TITLE>Bookm &amp; ardsd ks</TITLE>
<H1>Book &amp; marks</H1>
    <DT><H3 ADD_DATE="1502137359" LAST_MODIFIED="1502137366">Progra &amp; mming</H3>
    <DL><p>
        <DT><A HREF="http://code.google.com/edu/languages/google-python-class/introduction.html" ADD_DATE="1329052843">Python Introduction - Google&#39;s Python Class - Google Code</A>
    </DL>
</DL>
```

adam3smith:
That sounds right -- and yes, absolutely, I'd be happy to take a pull request from you that fixes this. Let me know if you need any directions.

ckush:
OK I'm going to make a pull request. I think I understand what to do but if I make a mistake please excuse this noob.

zuphilip:
Solved by #1496.

PR #1496, opened 2017-12-12
Fixed Bookmarks.js HTML character references in title.

ckush:
Importing bookmarks in Zotero left symbols like `&amp;` and `&#39;` in title. This fix changes them to the original character.

This issue was discussed here: https://github.com/zotero/translators/issues/1485

### DIFF
--- a/Bookmarks.js
+++ b/Bookmarks.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 3,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2017-06-03 11:41:00"
+	"lastUpdated": "2017-12-17 17:43:00"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


zuphilip:
@ckush Okay, that looks fine. But you need to update the `lastUpdated`-timestamp at the top of this file as well. Otherwise this will not get distributed.

ckush:
@zuphilip I have updated the timestamp.

Also, I realised I created I new branch `patch-1`. I don't remember how this happened, as I'm very new to all this. Ideally, was I supposed to keep it to `master`?

adam3smith:
patch-1 is automatically created when you create forks from github online and it's the recommended way to work so that your local master branch is always "clean"

zuphilip:
🎉 Thank you @ckush !

PR #1503, opened 2017-12-23
Add new Google Research translator

GuyAglionby:
There was some embedded metadata but it didn't capture everything + it didn't capture the PDF, so added a translator :)

### DIFF
--- a/Google Research.js
+++ b/Google Research.js
@@ -1,0 +1,395 @@
+{
+	"translatorID": "43293374-bf1d-4f6a-82de-cbd9a572ae7d",
+	"label": "Google Research",
+	"creator": "Guy Aglionby",
+	"target": "^https?://(research\\.google\\.com/(pubs|search\\.html)|static\\.googleusercontent\\.com/media/research\\.google\\.com/en)",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2018-01-27 23:57:56"
+}
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2017 Guy Aglionby
+	
+	This file is part of Zotero.
+
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+// See also https://github.com/zotero/translators/blob/master/BibTeX.js
+var bibtex2zoteroTypeMap = {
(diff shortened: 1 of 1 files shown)


adam3smith:
did you close this on purpose?

GuyAglionby:
Oh -- no, sorry about that. Reopened.

I will commit changes later on to take into account the feedback on the other commit

adam3smith:
OK, will wait with the review then, thanks!

GuyAglionby:
Hi, sorry for the delay but I've now rewritten the translator so it can handle author pages, search results, and the pages itself.

Just a weird thing I'd like to explain: some of the search results are direct links to the paper's PDF in the format https://research.google.com/pubs/archive/#####.pdf . This ID (the `#####`) can be used to find the publication's info page containing all the necessary info to create a Zotero item, which is what I have done. However, in some cases this page doesn't exist, so the extra GET checks that.

I'm not happy about how this is done on line 140 -- ideally we'd use `responseObj.status`, however when I try this I get a Permission Denied error, which I understand is CSRF protection. I don't know if there's any way to get around this, and would appreciate any pointers :)

Any other comments welcomed, thanks in advance!

dstillman:
> However, in some cases this page doesn't exist, so the extra GET checks that.

Can you provide an example of a page where that happens?

GuyAglionby:
Sorry, not sure how this slipped my mind. First result of https://research.google.com/search.html?q=simple%20risk%20bounds is https://research.google.com/pubs/archive/35621.pdf , where the corresponding https://research.google.com/pubs/pub35621.html is a 404.

Google Research.js
```diff
@@ -0,0 +1,395 @@
...
+						"creatorType": "author"
+					},
+					{
+						"firstName": "Sudheendra",
+						"lastName": "Vijayanarasimhan",
+						"creatorType": "author"
+					}
+				],
+				"date": "2016",
+				"itemID": "45619",
```
adam3smith: let's drop the itemID (i.e. `delete item.itemID`)

Google Research.js
```diff
@@ -0,0 +1,395 @@
...
+					}
+				],
+				"tags": [],
+				"notes": [],
+				"seeAlso": []
+			}
+		]
+	},
+	{
+		"type": "web",
```
adam3smith: could you also add a test for search results (or am I just overlooking that)?

GuyAglionby:
This may be a silly question, but if I add a test as follows:

```
{
	"type": "web",
	"url": "https://research.google.com/search.html?q=algorithm",
	"defer": true,
	"items": "multiple"
}
```

It fails with `TranslatorTester: Google Research Test 1: failed (Translation failed: no items returned)`. It seems like this is because the search results are loaded by JS after the page is loaded, which is why I put the `defer` in there, but that doesn't seem to be actually making the test wait. I think I'm missing something?

I'm also happy to add a test case for the direct PDF link, but am having the same issue with the test failing as with the [CiteSeer parser](https://github.com/zotero/translators/pull/1535#issuecomment-361066935).

Happy to make the other change, will push with the test cases as and when. Thanks!

adam3smith:
Yeah, the test framework sometimes hits limits, do what @zuphilip did for the other translator, i.e. comment out the test so that we have it ready for manual testing as needed. Once you remove the itemID, this is ready to go. Thanks!

GuyAglionby:
Amended as requested. Thanks!

GuyAglionby:
Hey there, just wanted to double check that I wasn't missing any other requested changes before this is merged?

adam3smith:
I don't think so. I'm just behind with review & merge

GuyAglionby:
No problem at all -- I just wanted to be sure. Cheers!

adam3smith:
superseded by https://github.com/zotero/translators/pull/1741

PR #1522, opened 2018-01-14
Fix Sage authors after they broke RIS

adam3smith:


adam3smith:
https://forums.zotero.org/discussion/69864/translator-saves-author-names-in-one-field-instead-of-two#latest

zuphilip:
Thank you!

PR #1524, opened 2018-01-20
Updated AustLII translater to support the new fancy website layout.

jpwarren:
Tweaked the AustLII/NZLII translator to support the new fancy website that AustLII recently launched.

The site now has some sort of weird proxy/load-balancer setup that doesn't alias the servers to www, so you get redirected to things like www7.austlii.edu.au and that was breaking the translator site detection.

Also changed the location of the citation title to the `<head/><title/>` element as that's easy to find and contains the same information that used to be in the `<h2/>` element.

Added a new test to verify the translator behavior works on cases in a different area.


jpwarren:
Alas, should have added these second updates to the translator on a feature branch, sorry.

Added some logic to support legislation and sections of legislation on the new AustLII layout.

AustLII and NZLII.js
```diff
@@ -1,121 +1,229 @@
 {
 	"translatorID": "5ed5ab01-899f-4a3b-a74c-290fb2a1c9a4",
 	"label": "AustLII and NZLII",
-	"creator": "Bill McKinney and Sebastian Karcher",
-	"target": "^https?://www\\.(austlii\\.edu\\.au|nzlii\\.org)",
-	"minVersion": "2.1.9",
+	"creator": "Justin Warren, Philipp Zumstein",
+	"target": "^https?://www\\d?\\.(austlii\\.edu\\.au|nzlii\\.org)",
```
zuphilip: Do we have these numbers also for the `nzlii` part? Otherwise we might move it into the bracket to the `austlii` part only.

zuphilip:
Thank you @jpwarren . I continued and rewrote several things which where outdated from the old code here. Moreover, I tried to rely on more stable structure like the ribbon instead of some pure text parsing which is likely to break at some point. I pushed the changes into your fork directly. Please have a look, especially because law sources are always tricky.

zuphilip:
(related forum post: https://forums.zotero.org/discussion/69747/austlii-translator/p1 )

PR #1534, opened 2018-01-28
Fix incorrect type returned for ResearchGate

GuyAglionby:
The inclusion of `(PDF Available)` in the text that the translator was using to determine type was causing it to fail, so I have fixed this. No failing test added as the existing one for URL `https://www.researchgate.net/publication/315673297_Linguistic_Landscape_Bibliography_on_Zotero` was already failing.

Any comments welcome. I think using the RIS data is still a valid approach -- the BibTeX export they provide isn't any more reliable.

Thanks :)

### DIFF
--- a/ResearchGate.js
+++ b/ResearchGate.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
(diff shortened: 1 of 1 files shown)


zuphilip:
Thank you @GuyAglionby ! Everything looks fine and the test case now detect again the correct case.

PR #1542, opened 2018-02-02
Restart ASCE translator

Hanlin-Dong:
Currently, the ASCE translator is turned off, due to some bugs on "multiple" pages. However, the "article" page translates well. So I strongly recommend to restart this translator by annotating the "multiple" related things, and keep the "article" related things for a temporary use.

### DIFF
--- a/ASCE.js
+++ b/ASCE.js
@@ -44,9 +44,10 @@ function getTitles(doc) {
 }
 
 function detectWeb(doc, url) {
-/*	if (url.match(/\/doi\/abs\/10\.|\/doi\/full\/10\./)) {
+	if (url.match(/\/doi\/abs\/10\.|\/doi\/full\/10\./)) {
 		return "journalArticle";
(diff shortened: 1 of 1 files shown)


adam3smith:
This is a misunderstanding on why this was disabled. The problem wasn't multiple pages but import from individual item pages resulting in 1700(!) items being imported: See https://forums.zotero.org/discussion/65088/asce-translator-large-problem-1700-items-imported-instead-of-1 for details. We'd need to make absolutely sure that's actually no longer the case, in particular because the fallbacks for ASCE are decent, if not perfect.

Hanlin-Dong:
OK, now I see that. I just tried on that article (Discrete choice ...), and everything worked well in my Windows Zotero 5.0.34, and Chrome 64.0.3282.119. Hope you can fix that issue soon. Thank you.

AbeJellinek:
The ASCE translator has been completely rewritten and works fine now, so I'm closing this as redundant.

PR #1545, opened 2018-02-03
Adding a translator for idref.fr

symac:
Hello,
[idref.fr](http://www.idref.fr) is the website allowing to get information about authorities (identifiers) available in various French catalogs (mainly [SUDOC](http://www.sudoc.abes.fr/) and [HAL](https://hal.archives-ouvertes.fr/) for the moment). These linked catalogs already have existing translators so this translator is mainly calling the correct one based on the list of items existing on a page. 

Currently the translator is working as expected ([https://www.idref.fr/199676100](https://www.idref.fr/199676100) for example), but I have not been able to run the populated test, I imagine that this is related to the way I call selectItems but I have not been able to understand what the issue is), I hope the Zotero team will be able to fix this slight issue before merging.

Happy to update my code if any advice is given regarding things that would not be done according to Zotero's expectations.

### DIFF
--- a/Idref.js
+++ b/Idref.js
@@ -65,7 +65,7 @@ function scrape(doc, url) {
 	
 	translator.setHandler('itemDone', function (obj, item) {
 		Z.debug(item);
-		//item.complete();
+		item.complete();
 	});
 	
 	translator.getTranslatorObject(function(trans) {
--- a/Idref.js
+++ b/Idref.js
@@ -2,14 +2,14 @@
 	"translatorID": "271ee1a5-da86-465b-b3a5-eafe7bd3c156",
 	"label": "Idref",
 	"creator": "Sylvain Machefert",
-	"target": "^https?://www\\.idref\\.fr/",
+	"target": "^https?://www\\.idref\\.fr/.*",
 	"minVersion": "3.0",
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "g",
-	"lastUpdated": "2018-01-12 10:00:51"
+	"lastUpdated": "2018-01-13 21:39:39"
 }
 
 function detectWeb(doc, url) { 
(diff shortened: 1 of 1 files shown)


Idref.js
```diff
@@ -0,0 +1,83 @@
+{
+	"translatorID": "271ee1a5-da86-465b-b3a5-eafe7bd3c156",
+	"label": "Idref",
+	"creator": "Sylvain Machefert",
+	"target": "^https?://www\\.idref\\.fr/.*",
```
adam3smith: no need for the `.*` -- the target needs to match against the URL, not the other way around.

Idref.js
```diff
@@ -0,0 +1,83 @@
...
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "g",
```
adam3smith: `gcsibv`

Idref.js
```diff
@@ -0,0 +1,83 @@
...
+	"lastUpdated": "2018-02-03 09:39:41"
+}
+
```
adam3smith: Add licenese

Idref.js
```diff
@@ -0,0 +1,83 @@
...
+function doWeb(doc, url)
+{
+	if (detectWeb(doc, url) == "multiple") {
+		var resultsTitle = ZU.xpath(doc, '//div[@id="perenne-references-docs"]/span[contains(@class, "detail_value")]');
+		var resultsHref = ZU.xpath(doc, '//div[@id="perenne-references-docs"]/span[contains(@class, "detail_label")]/a/@href');
+		
+		items = {};
+		for (var i in resultsTitle) {
```
adam3smith: iterate through arrays using either `for (let i=0; i<resultsTitle.lenght; i++)` or `for (result of resultsTitle)`

Idref.js
```diff
@@ -0,0 +1,83 @@
...
+		});
+
+	}	
+}
+
+function scrape(doc, url) {
+	Z.debug("scraping : " + url);
+	var translator = Zotero.loadTranslator('web');	
+	
+	if (url.indexOf("archives-ouvertes")!=-1){
```
adam3smith: we've switched to using `includes` throughout for literal matches for readability (about same efficiency), so `if (url.includes("archives-ouvertes"))` etc.

Idref.js
```diff
@@ -0,0 +1,83 @@
...
+	if (detectWeb(doc, url) == "multiple") {
+		var resultsTitle = ZU.xpath(doc, '//div[@id="perenne-references-docs"]/span[contains(@class, "detail_value")]');
+		var resultsHref = ZU.xpath(doc, '//div[@id="perenne-references-docs"]/span[contains(@class, "detail_label")]/a/@href');
+		
+		items = {};
+		for (var i in resultsTitle) {
+			href = resultsHref[i].textContent;
+			// We need to replace the http://www.sudoc.fr/XXXXXX links are they are redirects and aren't handled correctly from subtranslator
+			href = href.replace(/http:\/\/www\.sudoc\.fr\/(.*)$/, "http://www.sudoc.abes.fr/xslt/DB=2.1//SRCH?IKT=12&TRM=$1");
+			items[href] = resultsTitle[i].textContent;
```
adam3smith: Why not filter out the URLs that you know we won't be able to import here? Better to not offer those results than to fail on them.

symac:
Thanks @adam3smith . I have updated the file following your comments, should be better now. Tell me if you see anything else that could be improved.

Idref.js
```diff
@@ -2,62 +2,85 @@
...
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2018-02-12 13:03:06"
 }
 
+/*
+   IdRef Translator
+   Copyright (C) 2018 Sylvain Machefert - web@geobib.fr
+
+   This program is free software: you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
```
zuphilip: All Zotero translators should be licensed under AGPL (Affero), see https://github.com/zuphilip/translators/wiki/Common-code-blocks-for-translators for the relevant block.

symac: now fixed. I had copied another translator from the repo without having in mind this information regarding the license to be used. AGPL is also fine for me, thanks for the pointer.

symac:
@adam3smith @zuphilip what is the next step regarding this translator? I believe I have fixed the requests both of you did previously, or do I just need to wait (which is fine, no hurry, just want to be sure I am not missing something that I could do).

adam3smith:
Thanks!

symac:
@adam3smith can you confirm that it is working as expecting for you? I have been told there were some issues and when testing, the translator is not showing in Firefox. If I try to visit [this page](https://www.idref.fr/199676100) in FF, nothing happens, if I visit the same page with Scaffold inside Zotero it is ok. 

I have just resinstalled everything (Zotero, Scaffold, Firefox connector). Any idea ? don't know if you have access to debug reports but in case, I have just created one under reference D1823926315.

adam3smith:
Yes, there's something weird going on with detect. Reloading the page sometimes works for me but not always.

Issue #1487, opened 2017-11-29
CRAN translator is blocked by their use of an HTML frame


katrinleinweber:
Zotero doesn't use [R-Packages.js](https://github.com/zotero/translators/blob/master/R-Packages.js) when browsing [CRAN](https://cran.r-project.org/) "the normal way" (red). One has to open / deep-link the package‘s individual page for the translator to find all the meta-data (green).

![cran s frames prevent zotero from easily extracting citations 1](https://user-images.githubusercontent.com/9948149/33364429-aef7175e-d4e4-11e7-8cb9-6d0bbecf9cbf.PNG)
![cran s frames prevent zotero from easily extracting citations 2](https://user-images.githubusercontent.com/9948149/33364431-afe47184-d4e4-11e7-9fe9-f3efb7f0de11.PNG)

Please let me know whether I should report this issue to CRAN as well. My grasp of JavaScript is weak, but I presume it could be used traverse into the frame.

katrinleinweber:
Dear @gaborcsardi and @jeroen, because of your involvement in [METACRAN](https://github.com/metacran), might this be relevant for you?

adam3smith:
Zotero no longer recognizes URLs of embedded frames like it used to, which has likely caused this regression. This will need to get addressed in the translator (unless, of course, CRAN could be convinced to do away with those hideous iframes, but I'd assume that's just not a priority for them, understandably)

jeroen:
I don't know why I get tagged here but you should can get package metadata from the [DESCRIPTION](https://cran.r-project.org/web/packages/scientoText/DESCRIPTION) file in the root of the package. The most important fields are also indexed in the [PACKAGES](https://cran.r-project.org/src/contrib/PACKAGES) file.

CRAN won't change their webpage or URL to accommodate your scraper.

katrinleinweber:
I read over at https://ropensci.org/blog/2013/06/14/goals-for-year/ that R package citations were one success metric for rOpenSci. Therefore, I started wondering about which side here has the best lever to make importing citation info as easy for users as possible (single click, for example). Seems like in this case it's Zotero, correct?

dstillman:
> Zotero no longer recognizes URLs of embedded frames like it used to

Just as a reminder for whoever works on this, this can be re-enabled for a given translator with the [targetAll](https://github.com/zotero/zotero/issues/1021#issuecomment-243371694) property, but if the frames are on the same domain then the frames can just be referenced from the root document.

[...]

PR #1552, opened 2018-02-08
support other CRAN mirrors

katrinleinweber:
Hi!

Since [CRAN.RStudio.com](https://cran.rstudio.com/) is a very popular mirror, please consider expanding the translator to it. Related to #1487.

Also, would a minimal reg-ex to cover even other mirrors be advisable? `cran\..+/web/packages` might work.

Cheers :-)


adam3smith:
Generally happy with this & extending it to other mirrors. Note you'll have to bump the timestamp for updates to propagate (Scaffold, which we recommend, would do that automatically)

R-Packages.js
```diff
@@ -2,7 +2,7 @@
 	"translatorID": "24a10ebf-ada1-4b8d-8f76-5a29e24d3e78",
 	"label": "R-Packages",
 	"creator": "Sebastian Karcher",
-	"target": "r-project\\.org/web/packages/(.+/index\\.html|available_packages_by_(name|date)\\.html)",
+	"target": "(r-project\\.org|rstudio\\.com)/web/packages/(.+(/index\\.html)?|available_packages_by_(name|date)\\.html)",
```
adam3smith: once you make index.html optional, you can just delete the entire regex after packages: `.+(/index\\.html)?` literally matches anything.

adam3smith:
What's the form of the other mirrors? I thought we already covered those?

katrinleinweber:
Oh! That was an unwarranted assumption it seems. I was thinking of https://www.r-pkg.org/pkg/ggplot2 but they seem to have an entirely different site structure.

katrinleinweber:
I used https://duckduckgo.com/?q=cran+web+packages and https://www.google.de/search?q=cran+web+packages to find other popular mirrors.

https://mran.microsoft.com/ & https://www.r-pkg.org/ are out-of-scope for me currently.

katrinleinweber:
I feel bad duplicating the test code so much. Does Scaffold support extracting multiple URLs and comparing each result against the same `items` set? Or against each other?

adam3smith:
Don't worry about it -- we only very rarely test different sites with the same expected result, so not worth implementing dedicated functionality for that scenario, and the size of these tests is just completely negligible (and it's not like it's code that actually needs to be updated/maintained by humans)

katrinleinweber:
I delivered it with my bare hands, but it's good to know I don't have to take care of it.  🤣

katrinleinweber:
related to https://github.com/force11/force11-sciwg/issues/9

zuphilip:
I rewrote it w/o the FW and update some things. Please have a look. Should be fine for merge.

katrinleinweber:
Manual test imports look fine for me (Zotero 5.0.35.1 & Firefox 52.6.0). Thank you :-)

R-Packages.js
```diff
@@ -125,26 +170,255 @@ var testCases = [
...
+						"lastName": "Firth",
+						"creatorType": "author"
+					},
+					{
+						"firstName": "Steve",
+						"lastName": "Taylor",
+						"creatorType": "author"
+					},
+					{
+						"firstName": "R. Core",
```
adam3smith: we should recognize the R Core Team and turn it into an institutional author. Presumably that'll show up on a fair number of packages, so should be worth it.

zuphilip: Done in the new version.

katrinleinweber:
Thank you :-)

zuphilip:
Okay, it is merged now. Thank you all for the contributions here!

zuphilip:
@katrinleinweber Can we close the corresponding issue or is there something still open? (I haven't followed the conversion there closely.)

katrinleinweber:
#1487 is "related", but not completely corresponding to this ;-)

PR #1559, opened 2018-02-15
Add Oxford-EEBO-TCP translator

pierwill:
Not the prettiest or most flexible translator, but it's better than nothing for this important archive.

### DIFF
--- a/Oxford-EEBO-TCP.js
+++ b/Oxford-EEBO-TCP.js
@@ -1,0 +1,232 @@
+{
+	"translatorID": "62151d94-6fdf-442f-be29-94bbd24fcd09",
+	"label": "Oxford-EEBO-TCP",
+	"creator": "pierwill",
+	"target": "[^]*[ota|tei.it|purl].ox.ac.uk/[tcp|id]/*",
+	"minVersion": "5.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2018-02-15 05:20:51"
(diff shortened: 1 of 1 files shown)


Oxford-EEBO-TCP.js
```diff
@@ -0,0 +1,232 @@
...
+
+function doWeb(doc, url) {
+    
+    newItem = new Zotero.Item("book");
+    
+    if (url.includes("http://tei.it.ox.ac.uk/tcp/Texts-HTML/free/") == true) {
+	html(doc, url);
+    } else {
+	catalogue(doc, url);
+    }
```
zuphilip: The following scraping from that page looks very fragile. Moreover, there should be the corresponding TEI file available, e.g. http://tei.it.ox.ac.uk/tcp/Texts-TEI/free/A20/A20631.xml which has the information in structured form and we should therefore use instead.

zuphilip: Stopped here with the review therefore.

Oxford-EEBO-TCP.js
```diff
@@ -0,0 +1,232 @@
+{
+	"translatorID": "62151d94-6fdf-442f-be29-94bbd24fcd09",
+	"label": "Oxford-EEBO-TCP",
```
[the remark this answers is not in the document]
adam3smith: (EEBO = Early English Books Online -- there are different versions of this; Gale has one too, e.g.)

pierwill:
Thanks for the look! I learned a lot writing this from scratch. Nice to finally start learning to contribute to something I use every day.

I'll look into parsing the TEI.

PR #1560, opened 2018-02-15
PDF downloads from SSRN

### LINKED ISSUE #523: SSRN translator does not download PDFs
SSRN translator will not download PDFs (even thought the URL for the PDF is correct), because Zotero does not pass `Referer` header in the request.

@simonster Could we maybe make Zotero add `Referer` header based on item.URL (which I realize does not have to match the page where the PDF link is) or maybe from an additional field in the attachment item (e.g. `item.attachments[0].referer`)?

Edit: or first the latter and fall back on the former

Edit2: The url should also be proxified if needed

psisquared2:
Now that https://github.com/zotero/zotero/commit/6cef0bda80f4a969b0d20818017a4b60ae269028 has been commited and HTTP referrers are supported, automatic PDF downloads from SSRN are possible.

Be aware that SSRN.com has very strict requirements on repeated downloads - if you test the translator a few times on the same article, PDF downloads may fail because you hit SSRN's download limit.

This should close #523 and possibly https://github.com/zotero/zotero/issues/772.

Other changes: The URL of one one 'multiple' test case was outdated and no longer worked in a browser, so I removed it. I have also also updated the other test cases using version 3.3.0 of Scaffold.

### DIFF
--- a/SSRN.js
+++ b/SSRN.js
@@ -8,8 +8,8 @@
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
-	"browserSupport": "gcsbv",
-	"lastUpdated": "2018-02-15 19:29:55"
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2018-02-16 06:39:03"
 }
 
 /*
@@ -43,8 +43,7 @@ function detectWeb(doc,url) {
 }
 
 
-function doWeb(doc,url)
-{
+function doWeb(doc,url) {
 	if (detectWeb(doc, url) == "multiple") {
 		var hits = {};
 		var urls = [];
@@ -65,7 +64,7 @@ function doWeb(doc,url)
 			ZU.processDocuments(urls, scrape);
 		});
 	} else {
-		scrape(doc, url)
+		scrape(doc, url);
 	}
 }
 
@@ -93,10 +92,11 @@ function scrape(doc, url) {
 		//Extract the correct PDF URL from the download button
 		var download_xpath = ZU.xpath(doc, '//a[@id="downloadPdf"]');
 		if (download_xpath.length > 0) {
-			item.attachments.push(
-				{title: "Full Text PDF",
-				url: download_xpath[i].href,
-				mimeType: "application/pdf"});
+			item.attachments.push({
+				title: "Full Text PDF",
+				url: download_xpath[0].href,
+				mimeType: "application/pdf"
+			});
 		}
 
 		item.complete();
--- a/SSRN.js
+++ b/SSRN.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsbv",
-	"lastUpdated": "2017-06-26 04:35:21"
+	"lastUpdated": "2018-02-15 19:29:55"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


zuphilip:
I fixed the index and did some clean up which I pushed directly into your branch.

However, the PDF download did not work for me in Chrome. Does this work for you outside Scaffold?

psisquared2:
Thank you for the changes. I have compiled commit https://github.com/zotero/zotero/commit/00e52ae62b2e84d7fa770ca2f61d1b3b36f156bf of Zotero under Linux Ubuntu 64 bit. Using this own build of Zotero standalone and your version of the SSRN translator, translation including PDF downloads works without problems in both Firefox and Chromium.

Did you use the latest commit of Zotero standalone? Do you get a download error for the PDF, or does the translator not recognize the PDF attachment? Did you accidentally run into SSRN's download limit (which can happen very easily)?

zuphilip:
Okay, right. I am at the latest release of Zotero 5.0.35.1 and I guess the new patch will become available in the next version.

We should wait then for that, because also deployment of the new translator code for everyone else only makes sense when they already have the new Zotero version.

psisquared2:
Oh sorry, I wasn't aware that translators are deployed in a different way. In that case we shoud put this request on hold until the next patch.

psisquared2:
I just tested the translator in version 5.0.37 and I found no issues. Could you double check and merge if you find no issues either?

adam3smith:
Thanks, works for me, too.

zuphilip:
Thank you @psisquared2 !

owcz:
Yes, once https://papers.ssrn.com/sol3/papers.cfm?abstract_id=2699251&download=yes expires, I get a "Page not found" too.

To reproduce on https://papers.ssrn.com/sol3/papers.cfm?abstract_id=2699251:
1. latest Chrome on macOS 10.13.5, logged out of SSRN (no account)
2. SSRN detects properly on https://papers.ssrn.com/sol3/papers.cfm?abstract_id=2699251 (shows SSRN translator in Connector extension)
3. Throws this error: 
ss
4. If I refresh and attempt again: 
screen shot 2018-07-31 at 5 48 37 am

My next step was to click the orange "Download This Paper" button for this page: 
screen shot 2018-07-31 at 5 50 10 am where I have to click "or continue to Download Anonymously" to proceed. Which spits me back to the article's original page: https://papers.ssrn.com/sol3/papers.cfm?abstract_id=2699251&download=yes and while the extra parameter triggers a PDF download, it also makes the SSRN translator not detect the URL. If I remove the parameter, refresh and try again at https://papers.ssrn.com/sol3/papers.cfm?abstract_id=2699251, I get 
screen shot 2018-07-31 at 5 52 51 am

psisquared2:
> Does the first (failed) translation trigger the rate limiter?

I guess it does. The problem at point 3 definitely does not happen for me. It also seems to be unrelated to the issue discussed in this pull request, though point 4 is definitely related.

The fact that you are seeing the "What happened?" screen means you hit the download limit. Clicking on "Download anonymously" leads you to a captcha, which will (As far as I am aware) prevent Zotero downloads from such a page. Even if we allowed the SSRN translator to detect the page with `&download=yes`, I am 99% sure it still wouldn't work.

So to summarize, we need to debug your issue under point 3. Could you provide a log? The fact that the Embedded Metadata translator also crashes suggests that this is not a problem of the SSRN translator.

zuphilip:
I tried some of the urls here and the first time it worked for me fine. After clicking around I am now in the "unusual download pattern" mode and can't save papers manually or with Zotero. However, I have never seen something like step 3 you described here.

PR #1567, opened 2018-02-26
Add translator for sverigesradio.se

sebastian-berlin-wmse:
This is the website for the Swedish public service radio.

### DIFF
--- a/Sveriges radio.js
+++ b/Sveriges radio.js
@@ -1,0 +1,375 @@
+{
+	"translatorID": "caa8f42c-9dbf-446e-963b-6ee18e3133d2",
+	"label": "Sveriges radio",
+	"creator": "Sebastian Berlin",
+	"target": "^https?://sverigesradio\\.se/sida/(artikel|sok).aspx",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2018-02-28 13:17:24"
+}
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2018 Sebastian Berlin
+	
+	This file is part of Zotero.
+
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+
+// attr()/text() v2
+function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null;}function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null;}
+
+
+function detectWeb(doc, url) {
+	if (url.includes('/sida/artikel')) {
+		return "newspaperArticle";
+	} else if (url.includes('/sida/sok') && getSearchResults(doc, true)) {
+		return "multiple";
+	}
+}
+
+
+function getSearchResults(doc, checkOnly) {
+	var items = {};
+	var found = false;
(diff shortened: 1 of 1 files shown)


Sveriges radio.js
```diff
@@ -0,0 +1,709 @@
+{
+	"translatorID": "caa8f42c-9dbf-446e-963b-6ee18e3133d2",
+	"label": "Sveriges radio",
+	"creator": "Sebastian Berlin",
+	"target": "^https?://sverigesradio\\.se/sida/artikel.aspx",
```
adam3smith: This doesn't seem right -- it wouldn't detect multiples if your code below is correct (please add some multiples as tests, if possible)

Sveriges radio.js
```diff
@@ -0,0 +1,709 @@
...
+	// Embedded Metadata
+	translator.setTranslator('951c027d-74ac-47d4-a107-9c3069ab7b48');
+	// translator.setDocument(doc);
+	
+	translator.setHandler('itemDone', function (obj, item) {
+		item.publicationTitle = "Sveriges Radio";
+
+		item.creators = [];
+		var nameNodes = ZU.xpath(doc, '//p[@class="byline"]/text()');
+		for(var node of nameNodes) {
```
adam3smith: `let` is prefered over `var` for loops and similar temp uses

Sveriges radio.js
```diff
@@ -0,0 +1,709 @@
...
+			var nameString = node.wholeText;
+			if(nameString.trim() === "") {
+				// Empty text node are added between authors; we want to skip
+				// those.
+				continue;
+			}
+			var names = nameString.split(" ");
```
adam3smith: It seems like you're just generating a crude version of `ZU.cleanAuthor` here -- I'd recommend using that.

Sveriges radio.js
```diff
@@ -0,0 +1,709 @@
...
+				lastName: lastName,
+				creatorType: "author"
+			};
+			item.creators.push(author);
+		}
+
+		// The title from the meta is in the format:
+		// Australierna säger ja till samkönade äktenskap - Nyheter (Ekot)
+		var titleString =
+			ZU.xpath(doc, '//meta[@name="twitter:title"]')[0]
```
adam3smith: use `attr` and CSS selector here instead for simplicity

Sveriges radio.js
```diff
@@ -0,0 +1,709 @@
...
+			.getAttribute("content");
+		var titleParts = titleString.split(" - ");
+		item.title = titleParts[0];
+		item.section = titleParts[1];
+
+		var dateString = 
+			ZU.xpath(doc, '//meta[@name="displaydate"]')[0]
```
adam3smith: same here

Sveriges radio.js
```diff
@@ -0,0 +1,709 @@
...
+			.getAttribute("content");
+		item.date = dateString.replace(/(....)(..)(..)/, "$1-$2-$3");
```
adam3smith: `(\d{4})(\d{2})(\d{2})` is more precise and more readable

Sveriges radio.js
```diff
@@ -0,0 +1,709 @@
...
+
+		item.complete();
+	});
+
+	translator.getTranslatorObject(function(trans) {
+		trans.itemType = "newspaperArticle";
+		// TODO map additional meta tags here, or delete completely
```
adam3smith: remove all TODOs

Sveriges radio.js
```diff
@@ -0,0 +1,709 @@
...
+	{
+		"type": "web",
+		"url": "https://sverigesradio.se/sida/artikel.aspx?programid=83&artikel=6821850",
+		"items": [
+			{
+				"itemType": "newspaperArticle",
+				"title": "Australierna säger ja till samkönade äktenskap",
+				"creators": [
+					{
+						"firstName": "Ekot",
```
adam3smith: using Zotero's built in function will solve this, but the right way to handle single-name and institutional authors is to use `lastName` only and set `fieldMode: true`

Sveriges radio.js
```diff
@@ -0,0 +1,709 @@
...
+				"libraryCatalog": "sverigesradio.se",
+				"publicationTitle": "Sveriges Radio",
+				"section": "Nyheter (Ekot)",
+				"url": "http://sverigesradio.se/sida/artikel.aspx?programid=83&artikel=6821850",
+				"attachments": [
+					{
+						"title": "Snapshot"
+					}
+				],
+				"tags": [
```
adam3smith: These tags look generic and not helpful. I'd clean them out.

Sveriges radio.js
sebastian-berlin-wmse: Extended regex to include search page.

Sveriges radio.js
sebastian-berlin-wmse: I missed that function. Replaced the logic I had written, but some preprocessing was still required due to how the byline string can be formatted.

Sveriges radio.js
sebastian-berlin-wmse: I wasn't sure what tags did, so I kept them as they were. As you say, at least some of them are quite general and don't seem to relate to the article. Happy to get rid of them.

adam3smith:
Great, thanks!

PR #1571, opened 2018-03-03
Re-added support for old-style AustLII pages

jpwarren:
Some sites (like lawcite) link to the old-style AustLII pages, so we need to continue to support that format as well as the new shiny ribbon styles pages.

Otherwise you have to manually change the URL to use the new shiny ribbon styling for Zotero to correctly detect and process the page/citation.


AustLII and NZLII.js
```diff
@@ -113,6 +116,8 @@ function scrape(doc, url) {
 			var voliss = text(doc, 'head>title');
 			// e.g. C & M [2006] FamCA 212 (20 January 2006)
 			newItem.caseName = voliss.replace(/\s?\[.*$/, '');
+			newItem.title = newItem.caseName;
+
```
zuphilip: Why did you change that? For a `case` the title is actually the `caseName`, thus the title here will be deleted before saving into the database. Please change back these two lines.

zuphilip:
@jpwarren I fixed the merge conflict and have a question/request. Otherwise this looks fine.

AustLII and NZLII.js
jpwarren: Because without a `.title`, the Zotero connector in Chrome fails with an error, and the new item doesn't get created in Zotero.

I agree it shouldn't be necessary, but right now it is.

zuphilip: Uff, yes indeed. There is currently an error for cases also with Firefox, e.g. http://www8.austlii.edu.au/cgi-bin/viewdoc/au/cases/cth/AICmr/2017/134.html @dstillman Is it intended that a missing `title` but present `caseName` leads to an error?

adam3smith:
@zuphilip any reason not to merge this while we're figuring out the title/caseName issue?

zuphilip:
No, go ahead.

Am 04.03.2018 17:56 schrieb "Sebastian Karcher" <notifications@github.com>:

PR #1574, opened 2018-03-04
One more SD PDF fix

adam3smith:
it looks like we now have the PDF URL in the metaheader and can just use that, saving a ton of unnecessary calls and also making this work where it previously didn't (e.g. https://www.sciencedirect.com/science/article/pii/S1053811917307280?via%3Dihub still didn't work for me previously)


zuphilip:
Okay, thank you. The given url did work for me before and after commit, so it seems that SD is serving different contents to us.

PR #1576, opened 2018-03-05
Fix data format in semantic scholar translator

GuyAglionby:
Semantic Scholar changed their website format so the data is now base64 and URI encoded, so I have amended the translator to deal with this.

Only comment is that Scaffold (via Zotero 5.0.35.1,  Scaffold 3.3.0) doesn't have `atob` as a function so the tests fail, though I have manually tested the changes in Chrome.

Thanks!

### DIFF
--- a/Semantic Scholar.js
+++ b/Semantic Scholar.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2017-12-26 22:39:31"
+	"lastUpdated": "2018-03-05 00:00:22"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


adam3smith:
are you sure the problem is that Scaffold doesn't have atob? That'd be very odd -- it should have all javascript available in Firefox 54, which certainly includes atob. 
I suspect the tests are failing because the built-in browser can't handle the semantic scholar website.

adam3smith:
in any way, this is fine, thanks!

GuyAglionby:
I get the below error, but it may be an artefact of the browser not loading the website (which is new & strange behaviour)

```4:54:36 Translation using Semantic Scholar failed: 
         string => ReferenceError: atob is not defined
         stack => parseDocument/<@Semantic Scholar:95:5
...
```

dstillman:
Translators in the Zotero client run in a sandbox, which didn't have `atob`. I've added it for the next version.

PR #1577, opened 2018-03-06
Rewrite ACL translator

GuyAglionby:
Hi there. While the previous translator works on some of the proceedings pages ([http://www.aclweb.org/anthology/P/P93/](e.g.)), it doesn't work from the direct PDF view of papers, nor does it work for proceedings pages where there is no bibtex link ([e.g.](http://aclweb.org/anthology/Y/Y16/)). I've re-written and modernised the code so it works for both of these cases.

Comments welcomed, thanks in advance!

### DIFF
--- a/ACLWeb.js
+++ b/ACLWeb.js
@@ -1,183 +1,195 @@
 {
 	"translatorID": "f4a5876a-3e53-40e2-9032-d99a30d7a6fc",
 	"label": "ACL",
-	"creator": "Nathan Schneider",
-	"target": "^https?://(www[.])?aclweb\\.org/anthology/[^#]+",
-	"minVersion": "1.0.7",
+	"creator": "Nathan Schneider, Guy Aglionby",
+	"target": "^https?://(www\\.)?aclweb\\.org/anthology/[^#]+",
+	"minVersion": "3",
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
-	"browserSupport": "gcsbv",
-	"lastUpdated": "2013-09-16 00:20:13"
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2018-03-06 13:32:50"
 }
 
-// based on ACM translator
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2018 Nathan Schneider, Guy Aglionby
+	This file is part of Zotero.
+
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero.  If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
 function detectWeb(doc, url) {
-  var namespace = doc.documentElement.namespaceURI;
-	var nsResolver = namespace ? function(prefix) {
-		if (prefix == 'x') return prefix; else return null;
-	} : namespace;
-
-	var bibXpath = "//a[./text() = 'bib']"
-	if(doc.evaluate(bibXpath, doc, nsResolver, XPathResult.ANY_TYPE, null).iterateNext()) {
-	  return "multiple"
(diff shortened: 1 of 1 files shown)


ACLWeb.js
```diff
@@ -1,184 +1,196 @@
...
+	"lastUpdated": "2018-03-06 13:32:50"
 }
 
+/*
+	***** BEGIN LICENSE BLOCK *****
 
-function scrapeIndex(doc, items) {
-	var results;
-	var doImport;
+	Copyright © 2018 Nathan Schneider, Guy Aglionby
```
zuphilip: Write only your name here, because this is a complete rewrite.

ACLWeb.js
```diff
@@ -1,184 +1,196 @@
 {
 	"translatorID": "f4a5876a-3e53-40e2-9032-d99a30d7a6fc",
 	"label": "ACL",
-	"creator": "Nathan Schneider",
-	"target": "^https?://(www[.])?aclweb\\.org/anthology/[^#]+",
-	"minVersion": "1.0.7",
+	"creator": "Nathan Schneider, Guy Aglionby",
+	"target": "^https?://(www\\.)?aclweb\\.org/anthology/[^#]+",
+	"minVersion": "3",
```
zuphilip: 3.0

ACLWeb.js
```diff
@@ -1,184 +1,196 @@
...
 		});
-		translator.translate();
+		delete item.itemID;
+		item.complete();
+	});
+	translator.translate();
+}
 
+function scrapeProceedings(doc, id) {
+	let newItem = new Zotero.Item("conferencePaper");
```
zuphilip: There are also journal articles (at character `J` and maybe others), e.g. http://aclweb.org/anthology/J/J17/. Can we try to distinguish those?

ACLWeb.js
```diff
@@ -1,184 +1,196 @@
...
+	
+	let paragraphXpath = '//p[a[text()="' + id + '"]]/';
+	
+	let pdfURL = ZU.xpath(doc, paragraphXpath + 'a')[0].href;
```
zuphilip: I suggest to filter here on the href attribute instead of relying on the order, i.e.
```js
let pdfURL = ZU.xpathText(doc, paragraphXpath + 'a[contains(@href, "pdf")]/@href');
```

ACLWeb.js
```diff
@@ -1,184 +1,196 @@
...
-
-		if (doImport)
-			callTranslator(bib, type, attachments);
-
+function extractFullProceedings(doc) {
+	let unwantedTitles = ['Front Matter', 'Author Index', 'Keyword Index'].map(function(title) { 
+		return 'not(contains(., "' + title + '"))';
+	}).join(' and ');
+	
+	let baseXpath = '//div[@id="content"]/p[i[' + unwantedTitles + ']]/'
```
zuphilip: Add semi-colon

ACLWeb.js
```diff
@@ -1,184 +1,196 @@
...
+	
+	let items = {};
+
+	for (let i = 0; i < ids.length; i++) {
+		let articleAuthors = authors[i].split('; ');
+		let authorSurname = articleAuthors[0].split(' ').pop();
+		let etAl = '';
+		if (articleAuthors.length > 1) {
+			etAl = ' et al.';
+		}
```
zuphilip: You can also write this with the ternary operator:
```js
let etAl = articleAuthors.length > 1 ? ' et al.' : '';
```

ACLWeb.js
```diff
@@ -1,184 +1,196 @@
...
+						scrapeBibtex(responseString, url);
+					});
+				} else {
+					scrapeProceedings(doc, id);
 				}
-				authorLasts.push(last);
+			});
+		});
+	} else if(url.endsWith('.bib')) {
+		// e.g. http://www.aclweb.org/anthology/P10-4014.bib
```
zuphilip: This looks like a good test case we also should add.

GuyAglionby: This test fails with an error

```
21:58:35 TranslatorTester: Running ACL Test 3
21:58:35 TranslatorTester: Translating http://www.aclweb.org/anthology/P10-4014.bib
21:58:35 TranslatorTester: ACL Test 3: failed (Translation failed to initialize: TypeError: doc is null)
```

I have double checked that it works in Chrome though, is that sufficient?

zuphilip: Ah, okay. Yes, fine then.

GuyAglionby:
Thanks for all of your comments! I have made the requested changes.

I've seen the other site, but didn't realise it was changing to be the main version. I'll look into making a translator for that as well. Thanks for the heads up!

zuphilip:
Thank you @GuyAglionby very much! I did some final changes and the PR is now merged :tada: .

@adam3smith Did you start to work on the new site of ACL? Or can @GuyAglionby start from scratch?

PR #1578, opened 2018-03-07
[Zenodo] Prevent mis-importing software as report

katrinleinweber:
As [discussed in the forum](https://forums.zotero.org/discussion/comment/301116/) this ensures that `Software` items on Zenodo are imported as `computerProgram`, even though along the way, Zenodo is offering a non-CSL item type

### Potential/Remaining isues to discuss

- [x] Scaffold imports the full author names as `lastName`. This could occur in https://github.com/zotero/translators/blame/master/Zenodo.js#L172-L180, but I'm not sure whether or how to adjust that. Or, should I adjust the new test case to expect `lastName` and `firstName` normally?
- [x] Would it be helpful, if Zenodo stopped putting a clearly mis-matching CSL type in the JSON that Zotero parses, and instead send none? If yes, could we then omit both the ["fix"](https://github.com/zotero/translators/blame/master/Zenodo.js#L190-L197) and this "undo"?
- [x] Scaffold added `-` and `+` in the `doWeb` test output (near the bottom in the diff). [Not sure](https://www.mediawiki.org/wiki/Citoid/Creating_Zotero_translators#Write_test_cases) what those mean, but I guess both symbols need to be removed from the JSON block.

### DIFF
--- a/Zenodo.js
+++ b/Zenodo.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2018-01-22 21:59:40"
+	"lastUpdated": "2018-03-07 14:55:24"
 }
 
 /*
@@ -195,6 +195,11 @@ function scrape(doc, url) {
 
 			if (item.itemType == "document" && zoteroType[type]) {
 				item.itemType = zoteroType[type];
+				
+				//undo the fix for non-existent CSL types, if Zotero supports the original itemType
+				if (detectWeb(doc, url) == "computerProgram") {
+					item.itemType = "computerProgram"
+				}
 			}
 			item.itemID = "";
 			item.complete();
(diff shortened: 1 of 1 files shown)


adam3smith:
How are you generating the tests in Scaffold? This is fine now, but to save you the work going forward, if you actually generate the test using the Scaffold testing panel, you won't get the + and - (which are from a diff display Scaffold shows).

Zenodo.js
```diff
@@ -195,6 +195,11 @@ function scrape(doc, url) {
 
 			if (item.itemType == "document" && zoteroType[type]) {
 				item.itemType = zoteroType[type];
+
```
adam3smith: what would happen if we replace this whole thing, including l. 197 by `item.itemType = detectWeb(doc, url)`
That seems like a cleaner solution, provided our detect function is good.

katrinleinweber: Could the `var zoteroType` block be removed as well, then?

adam3smith:
The Zenodo item types in the CSL are fine -- they are as good as it gets for e.g. automated citation generation directly from the CSL and Zotero can customize at will in the import.

For the names, we can't really parse these correctly without breaking institutional names since they're entered incorrectly in Zenodo, but we _can_ switch fieldMode to true along the lines for 
```
for (let i =0; i< item.creators.length; i++) {
  if (!item.creators[i].firstName) {
    item.creators[i].fieldMode = true;
  }
}
```

This makes them easier to fix manually.

Zenodo.js
adam3smith: exactly, yes

Zenodo.js
```diff
@@ -175,6 +175,7 @@ function scrape(doc, url) {
 				if (!item.creators[i].firstName && item.creators[i].lastName.indexOf(",")!=-1) {
 					item.creators[i].firstName = item.creators[i].lastName.replace(/.+?,\s*/, "");
 					item.creators[i].lastName = item.creators[i].lastName.replace(/,.+/, "");
+					item.creators[i].fieldMode = true;
 				}
```
zuphilip: Does this work? No changes in the test cases??

I think what we want is rather want is:
```js
if (!item.creators[i].firstName) {
	if (item.creators[i].lastName.indexOf(",")!=-1) {
		item.creators[i].firstName = item.creators[i].lastName.replace(/.+?,\s*/, "");
		item.creators[i].lastName = item.creators[i].lastName.replace(/,.+/, "");
	} else {
		item.creators[i].fieldMode = true;
	}
}
```

Zenodo.js
```diff
@@ -187,15 +188,7 @@ function scrape(doc, url) {
...
 
-			//Zenodo uses some non-existent CSL types we're fixing.
-			var zoteroType = {
-				"figure": "artwork",
-	 			"article": "report"
-			};
-
-			if (item.itemType == "document" && zoteroType[type]) {
-				item.itemType = zoteroType[type];
-			}
```
zuphilip: Can we delete the first introduction of the variable `type` as well?

Zenodo.js
```diff
@@ -187,15 +188,7 @@ function scrape(doc, url) {
...
+			item.itemType = detectWeb(doc, url)
```
zuphilip: Add `;` at the end.

Zenodo.js
```diff
@@ -469,6 +462,61 @@ var testCases = [
...
+		"url": "https://zenodo.org/record/1048320",
+		"items": [
+			{
+				"itemType": "computerProgram"
+				"creators": [
+					{
+						"lastName": "Carl Boettiger"
+						"firstName": ""
+						"creatorType": "author"
+					}
```
zuphilip: Rerun the tests should yield some `fieldMode` fields.

Zenodo.js
katrinleinweber: Do you mean [L42](https://github.com/katrinleinweber/translators/blob/2b436bebeaf2af72c2c95cf20f5f21740ef2953b/Zenodo.js#L42) or [L133](https://github.com/katrinleinweber/translators/blob/2b436bebeaf2af72c2c95cf20f5f21740ef2953b/Zenodo.js#L133). The latter can indeed be removed without breaking the tests.

zuphilip: Yes, the latter one seems not necessarily anymore. We certainly still need the first one.

zuphilip:
:tada: Thank you @katrinleinweber !

katrinleinweber:
You're welcome, and thanks for merging :-)

Issue #1544, opened 2018-02-03
WHO document translator request


mvolz:
i.e. http://apps.who.int/iris/handle/10665/70863

Has relatively good metadata in it so I would think the embedded metadata translator would be a good start.

adam3smith:
yeah, definitely seems worthwhile + quick

zuphilip:
Duplicate of #791

zuphilip:
This is now implemented.

PR #1593, opened 2018-03-16
Copernicus improvements + basic WHO translator

mtrojan-ub:
Note: WHO translator was just for workshop, development will not be continued

see #1544


WHO.js
```diff
@@ -0,0 +1,192 @@
...
+			case 'URI:':
+				item.url = value;
+				break;
+			case 'Abstract:':
+				item.abstractNode = value;
+				break;
+		}
+	}
+	
+	//item.title = "dummy";
```
adam3smith: remove

WHO.js
```diff
@@ -0,0 +1,192 @@
...
+	var item = new Zotero.Item(type);
+	
+	var rows = doc.querySelectorAll("table.itemDisplayTable > tbody > tr");
+	for (let i=0;i<rows.length;i++) {
+		let key = text(rows[i], "td.metadataFieldLabel").trim();
+		let value = text(rows[i], "td.metadataFieldValue");
+		
+		Z.debug(key + " " + value);
+		
+		switch (key) {
```
adam3smith: We'll want to make this multilingual to work with all languages on the WHO page; I'm afraid creating doing that manually, i.e. looking up the lables for all 6 languages, is the only way to go.

WHO.js
```diff
@@ -0,0 +1,192 @@
...
+function detectWeb(doc, url) {
+	if (url.includes('/simple-search')) {
+		return "multiple";
+	}
+	
+	if (text(doc, "h2 > small") == "Collection home page") {
+		return "multiple";
+	}
+	
+	return "document";
```
adam3smith: This needs to be more robust (e.g. would falsely detect on the homepage). Also, `report` is probably a better option here

WHO.js
```diff
@@ -0,0 +1,192 @@
...
+
+function getSearchResults(doc, checkOnly) {
+	var items = {};
+	var found = false;
+	
+	var rows = doc.querySelectorAll('a.list-group-item');
+	for (let i=0; i<rows.length; i++) {
+		let href = rows[i].href;
+		Z.debug(href);
+		var title = text(rows[i], 'div.row > div.col-md-10 > h5');
```
adam3smith: This looks fragile to me -- no attributes we could use to make it more robust?

WHO.js
```diff
@@ -0,0 +1,192 @@
...
+*/
+
+
+// attr()/text() v2
+function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null;}
+function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null;}
+function node(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem:null;}
+
+function detectWeb(doc, url) {
+	if (url.includes('/simple-search')) {
```
adam3smith: The two conditions returning `multiple` can be combined into one `||` statement. Also, we'll want to test for `getSearchResults(doc, true)` to avoid false positives.

WHO.js
```diff
@@ -0,0 +1,192 @@
...
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+
+// attr()/text() v2
+function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null;}
+function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null;}
+function node(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem:null;}
```
adam3smith: For custom functions, please always include a brief comment with a description of what it does.

WHO.js
```diff
@@ -0,0 +1,192 @@
...
+		}
+	}
+	
+	//item.title = "dummy";
+	
+	item.complete();
+}
+
+
+function getBrValues(element) {
```
adam3smith: Same here -- please describe briefly.

WHO.js
```diff
@@ -0,0 +1,192 @@
...
+}
+
+/** BEGIN TEST CASES **/
+var testCases = [
+	{
+		"type": "web",
+		"url": "http://apps.who.int/iris/handle/10665/70863",
+		"items": [
+			{
+				"itemType": "document",
```
adam3smith: As per above, I want this as a report and then we can probably add some information under report number

WHO.js
```diff
@@ -0,0 +1,192 @@
...
+				"creators": [
+					{
+						"lastName": "World Health Organization",
+						"type": "author",
+						"fieldMode": 1
+					}
+				],
+				"language": "English",
+				"libraryCatalog": "WHO",
+				"publisher": "Geneva : World Health Organization",
```
adam3smith: That should be split up in publisher & place

WHO.js
```diff
@@ -0,0 +1,192 @@
...
+				"url": "http://www.who.int/iris/handle/10665/70863",
+				"attachments": [],
```
adam3smith: We'll want to attach the PDF here

WHO.js
zuphilip: @mtrojan-ub Did you try
```js
var title = rows[i].textContent;
```

I expect this gives the title together with the year (No?), but that is okay for the list to choose from.

WHO.js
mtrojan-ub: attr() and text() are from Scaffold code template => the comment should be added there, but i can gladly add a comment to node()

zuphilip:
I think all the points were addressed in the new commit. @adam3smith Do you have time to look at these changes?

zuphilip:
Superseeded by #1731.

PR #1600, opened 2018-03-22
WIP: expand OSTI...js to DOE Code

katrinleinweber:
This is currently just a demo to discuss whether to expand this translator, or whether to write a new one.
After 199a680, using this translator returns a minimal citation:

```json
11:35:26 Returned item:
           {
             "creators": [],
             "notes": [],
             "tags": [],
             "seeAlso": [],
             "attachments": [
               {
                 "title": "Snapshot"
               }
             ],
             "url": "https://www.osti.gov/doecode/biblio/9158",
             "title": "Doe Code: Project Metadata",
             "abstractNote": "XMOF2D [...]",
             "language": "en",
             "libraryCatalog": "www.osti.gov",
             "accessDate": "CURRENT_TIMESTAMP",
             "itemType": "webpage",
             "shortTitle": "Doe Code"
           }
```

The goal here would be to get `"itemType": "computerProgram"`, fill `"creators": []`, etc. to further https://github.com/force11/force11-sciwg/issues/9

### DIFF
--- a/OSTI Energy Citations.js
+++ b/OSTI Energy Citations.js
@@ -2,14 +2,14 @@
 	"translatorID": "0cdc6a07-38cf-4ec1-b9d5-7a3c0cc89b15",
 	"label": "OSTI Energy Citations",
 	"creator": "Michael Berkowitz",
-	"target": "^https?://www\\.osti\\.gov/(energycitations|scitech)",
+	"target": "^https?://www\\.osti\\.gov/(doecode|energycitations|scitech)",
 	"minVersion": "1.0.0b4.r5",
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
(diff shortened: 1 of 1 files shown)


zuphilip:
Extending the current translator for `DOE Code` sounds the right thing to do here. We may want to consider to slightly rename it afterwards.

However, I guess you have to handle these two cases almost independently in the translator: embedded metadata does not look useful here (the title is wrong, no creators, no year, no DOI). Thus, I would suggest to add another case using the DOI translator for the code part and improve from the result there (e.g. itemType).

katrinleinweber:
Not from my side, I'm sorry. [Without auto-testing](https://github.com/zotero/scaffold/issues/76) and only little JavaScript knowledge, I found large translator changes too cumbersome. Please feel free to continue :-) where I had to leave :-/

PR #1609, opened 2018-03-27
 Update Sveriges Radio translator to work with Wikimedia's installation

sebastian-berlin-wmse:
The version that Wikimedia is running doesn't support `String.include()`, which is used in the template version of `detectWeb()`. Changed this to `String.indexOf()`, which works in the latest version too.


dstillman:
They'll need to update to a later version of translation-server.

sebastian-berlin-wmse:
Would it be possible to have this merged anyway? There are plans to update the server, but no ETA at the moment.

As I wrote, it's still fully compatible with the latest version and contain a fix for a test as well.

dstillman:
No, sorry. `includes()` is used in many translators.

zuphilip:
It is even part of the CI here to replace unnecessary `.indexOf` with `.includes`:

https://github.com/zotero/translators/blob/c51821e26a97a1371730efdec17c80d83b0cef02/.ci/checkTranslator.sh#L93-L99

zuphilip:
Related discussion on phabricator: https://phabricator.wikimedia.org/T140539

PR #1618, opened 2018-04-12
CrossRef REST

mrtcode:


CrossRef-REST.js
```diff
@@ -0,0 +1,245 @@
...
+	// }
+	// else
+	if (item.query) {
+		query = '?query.bibliographic=' + encodeURIComponent(item.query);
+	}
+	else return;
+	
+	query += '&rows=50';
+	
+	ZU.doGet('http://api.crossref.org/works/' + query, function (responseText) {
```
dstillman: `https`

CrossRef-REST.js
```diff
@@ -0,0 +1,245 @@
...
+function doSearch(item) {
+	let query = null;
+	// if (item.DOI) {
+	// 	query = '?filter=doi:' + ZU.cleanDOI(item.DOI.toString());
+	// }
+	// else
+	if (item.query) {
+		query = '?query.bibliographic=' + encodeURIComponent(item.query);
+	}
+	else return;
```
dstillman: Remove

CrossRef-REST.js
```diff
@@ -0,0 +1,245 @@
...
+		
+		// result.URL is always http://dx.doi.org/..
+		
+		if (result.link && result.link.URL) item.url = result.link.URL;
+		
+		if (result.title && result.title[0]) {
+			item.title = result.title[0];
+			if (result.subtitle && result.subtitle[0]) {
+				// Don't duplicate subtitle if it already exists in title
+				if (item.title.toLowerCase().indexOf(result.subtitle[0].toLowerCase()) < 0) {
```
dstillman: `.includes()`

CrossRef-REST.js
```diff
@@ -0,0 +1,245 @@
...
+					item.title += ': ' + result.subtitle[0];
+				}
+			}
+			item.title = item.title.replace(/<!\[CDATA\[/g, '');
+			item.title = item.title.replace(/\]\]>/g, '');
+			item.title = item.title.replace(/<[a-zA-Z\/].*?>/g, '');
```
dstillman: What about supported/transformed markup, [as in the old CrossRef translator](https://github.com/zotero/translators/blob/c51821e26a97a1371730efdec17c80d83b0cef02/CrossRef.js#L40)? While currently we don't officially support markup in titles, citeproc-js does render it, so we don't want to throw away fields that citeproc-js supports.

adam3smith:
The reason I haven't moved ahead with this is that the old openURL API actually performs better on some item types (chapters, especially) than this one, but I think we should probably switch. That said, if we do switch, could we just do DOI content negotiation instead of separate CrossRef translators?
See https://citation.crosscite.org/docs.html and 
https://github.com/zotero/translators/pull/1135

dstillman:
@adam3smith: Note that this is currently only for text search for use in translation-server and doesn't handle DOI lookup. I assume content negotiation is only for DOI metadata lookup.

adam3smith:
ah yes, sorry, looked at this too quickly

CrossRef-REST.js
```diff
@@ -0,0 +1,245 @@
+{
+	"translatorID": "0a61e167-de9a-4f93-a68a-628b48855909",
+	"translatorType": 8,
+	"label": "CrossRef REST",
+	"creator": "Martynas Bagdonas",
+	"target": "",
+	"minVersion": "5.0.0",
```
zuphilip: I guess that all special Zotero functions which you use here are also available in version 3.0. Thus the translator is compatible, but it might be that it will not be possible to call it before.

CrossRef-REST.js
```diff
@@ -0,0 +1,245 @@
...
+	"maxVersion": null,
+	"priority": 90,
+	"inRepository": true,
+	"browserSupport": "gcsv",
```
zuphilip: Add all of them "gcsibv". Or are there any reasons for the selection?

CrossRef-REST.js
```diff
@@ -0,0 +1,245 @@
...
+	"lastUpdated": "2018-04-11 10:00:01"
+}
+
+// Based on CrossRef.js (by Simon Kornblith), which uses OpenURL API
```
zuphilip: Can we add a AGPL license for this new translator?

CrossRef-REST.js
```diff
@@ -0,0 +1,245 @@
...
+
+// This translator uses the newer REST API
+// https://github.com/CrossRef/rest-api-doc
+// https://github.com/CrossRef/rest-api-doc/blob/master/api_format.md
+// REST API documentation not always reflect the actual API i.e. and
+// some fields are undocumented.
+// All CrossRef item types can be retrieved at http://api.crossref.org/types
+
+function fixAuthorCapitalization(string) {
```
zuphilip: We should move this function to `Zotero.Utilities` because it occurs also in several other translators.

CrossRef-REST.js
```diff
@@ -0,0 +1,245 @@
...
+	if (typeof string === "string" && string.toUpperCase() === string) {
+		string = string.toLowerCase().replace(/\b[a-z]/g, function (m) {
+			return m[0].toUpperCase()
```
zuphilip: Add `;`

PR #1619, opened 2018-04-12
Add text search for GBV and LoC

mrtcode:


Library of Congress ISBN.js
```diff
@@ -25,7 +25,18 @@ function detectSearch(item) {
...
+	
+	let url;
+	
+	if (item.ISBN) {
+		url = "http://lx2.loc.gov:210/LCDB?operation=searchRetrieve&version=1.1&query=bath.ISBN=^" + ZU.cleanISBN(item.ISBN) + "&maximumRecords=1";
+	}
+	else if (item.query) {
+		url = "http://lx2.loc.gov:210/LCDB?operation=searchRetrieve&version=1.1&query=" + encodeURIComponent(item.query) + "&maximumRecords=50";
+	}
+	else return;
```
dstillman: We'd want braces here for style reasons, but the `else` can just be removed from both translators. While at the moment we're not testing for `query` in `detectSearch`, the assumption in translators is always that the page/input should clear `detect*` in order for `do*` to be called (and it'd be fine for `doGet` to throw an error on an empty `url` if that didn't happen for some reason).

zuphilip:
You need to update the timestamp of the two translators.

dstillman:
This has a conflict with the GBV translator now. Can you rebase?

Issue #1587, opened 2018-03-12
Support Codemeta format


zuphilip:
The Codemeta format is for software citations and uses JSON-LD. See here for more details: https://codemeta.github.io/

I am thinking about writing an import translator for that format. This can be used to improve the Github translator (and possibly others as well).

Maybe even an import+export translator could be interesting.

Let me know what you think.

PR #1621, opened 2018-04-13
Update GitHub API comment

katrinleinweber:
`stats` is nowhere to be found on https://api.github.com/repos/zotero/zotero, nor any users' forks like https://api.github.com/repos/zuphilip/zotero. I think GH updated the API URL as in the diff. If that's the one meant by the comment, please consider merging this.

PS: If the above is true, I'm guessing that "organization in GitHub cannot itself provide contributions" from #1296 is now outdated, and the translator could be updated in that way. If yes, shall I open an issue, or is that change trivial and can be appended here?

### DIFF
--- a/Github.js
+++ b/Github.js
@@ -93,7 +93,7 @@ function scrape(doc, url) {
 
 	item.rights = ZU.xpathText(doc, '//a[*[contains(@class, "octicon-law")]]');
 	
-	//api calls to /repos and /repos/../../stats/contribuors
+	//api calls to /repos and /repos/../../contribuors
 	var apiUrl = "https://api.github.com/";
 	ZU.doGet(apiUrl+"repos/"+repo, function(result) {
 		var json = JSON.parse(result);
--- a/Github.js
+++ b/Github.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2017-05-12 20:27:47"
+	"lastUpdated": "2018-04-14 12:00:32"
 }
 
 /**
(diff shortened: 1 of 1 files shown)


zuphilip:
I updated the translator a little bit more and updated also the test cases. Please have a look.

I don't know whether the Github API changed but these comments seem for me of little value anyways. I tried to replace them with something more meaningful.

Why do you think we should change how we save the owner information? E.g. zotero is the company but Dan would be a programmer seems still fine for me.

katrinleinweber:
Thanks for the update :-)

To elaborate on my "PS": Maybe I'm seeing a bug. Does importing an organisational repo result in _any_ ´programmers´ for you? Neither `HEAD` nor beea162 does in my case (macOS 10.13.3, Zotero 5.0.35).

Dan for example, only appears as `programmer` for me, when I trigger the translation on _his repos_. The field remains empty, whenever I translate the upstream repo, even though he obviously is a (or the main) contributor to them.

screen shot 2018-04-14 at 17 27 46

From #1296 I understood that `owner == org` back then prevented any `programmer` name from getting extracted. But from your comment above I gather that both `company` and `programmer`  are expected to get extracted, correct? I agree that would be fine, but I never observed that expected behavior so far.

zuphilip:
> Does importing an organisational repo result in any ´programmers´ for you?

No, the organization will be saved in the company field.

We also tried to look at the contributors, which need another API call, e.g. https://api.github.com/repos/zotero/zotero/contributors, but this list was too much to be meaningful in general. For doing better we might need more information to rely on, e.g. a CODEMETA file.

katrinleinweber:
Hm. I'd argue that at least adding the 0-th contributor (with most contributions) as `programmer` would be more "meaningful" than none, or a fork's owner.

zuphilip:
@katrinleinweber Thank you!

I try to look at the #1587 soon and then we possibly can revisit GitHub translator afterwards again.

katrinleinweber:
Thank you :-)

PR #1626, opened 2018-04-17
Add search by identifier capability to arXiv.org.js

mrtcode:


arXiv.org.js
```diff
@@ -7,9 +7,19 @@
...
-	"lastUpdated": "2017-03-22 21:23:25"
+	"lastUpdated": "2018-04-17 20:00:00"
+}
+
+function detectSearch(item) {
+	return !!item.arXiv;
+}
+
+function doSearch(item) {
+	var url = 'http://export.arxiv.org/oai2?verb=GetRecord&metadataPrefix=oai_dc'
```
dstillman: `https` (almost nothing should be using `http` these days)

arXiv.org.js
```diff
@@ -7,9 +7,19 @@
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
-	"translatorType": 4,
+	"translatorType": 8,
```
dstillman: Pretty sure this would disable web translation for arXiv.org — make sure you've tested that. I believe a combo web/search translator [should be `12`](https://www.zotero.org/support/dev/translators#metadata).

PR #1635, opened 2018-04-24
Increase text search results count for LoC and GBV

mrtcode:
Increases text search results count because LoC and GBV have quite poor result ranking and sometimes the best results doesn't appear in the first 50 results.


dstillman:
Let's hold off on this for the moment, since it would be a ton of data (e.g., 500KB for one request).

PR #1649, opened 2018-05-10
Translator for Japan National Diet Library periodicals

fbennett:
This is very fresh and only lightly tested so far. It should work against searches and content pages under https://ndlonline.ndl.go.jp/#!/  I abandoned page scraping when (a) the Ajax/Angular methods used to build pages made multiple grab-and-page-scrape impossible; and (b) I figured out how to pull the source JSON for content pages directly.

I built this originally to fetch periodicals from the [NDL periodical ToC feeds](http://www.ndl.go.jp/jp/data/sakuin/sakuin_index.html), and it sets all items to "journalArticle;" but I now realize that the coverage of NDL Online is much broader, and possibly covers their entire holdings. So this will need some more work to work sensibly with books and CDs and the like.

On the positive side, the translator pulls attachments for items that have hosted digital content.

You can examine the NDL metadata object in Scaffold by uncommenting a debug line in the translator. The schema is a little odd in places, and some modest post-parsing of fields is needed, but it's pretty good, considering the scale of their undertaking.

Anyway, see what you think!

### DIFF
--- a/National Diet Library Periodicals.js
+++ b/National Diet Library Periodicals.js
@@ -1,0 +1,248 @@
+{
+	"translatorID": "6c51f0b2-75b2-4f4c-aa40-1e6bbbd674e3",
+	"label": "National Diet Library Periodicals",
+	"creator": "Frank Bennett",
+	"target": "https://ndlonline.ndl.go.jp/#!/(detail|search)",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsbv",
+	"lastUpdated": "2018-05-10 03:29:28"
+}
+
+
+
+// In case needed ever
+//function getCookies(doc) {
+//	var ret = {};
+//	if (doc.cookie) {
+//		var lst = doc.cookie.split(/; */);
+//		for (var i=0,ilen=lst.length; i<ilen; i++) {
+//			var pair = lst[i].split('=');
+//			if (pair[1]) {
+//				pair[1] = pair.slice(1).join("=");
+//			}
+//			if ("object" === typeof pair && pair.length) {
+//				pair[1] = pair[1].split("|");
+//				if (pair[1].length === 1) {
+//					pair[1] = pair[1][0];
+//				} else if (!pair[1].slice(-1)[0]) {
+//					pair[1].pop();
+//				}
+//			}
+//			ret[pair[0]] = pair[1];
+//		}
+//	}
+//	return ret;
+//}
+
+function getPageUrl(key) {
+	return "https://ndlonline.ndl.go.jp/detail/" + key;
+}
+
+var headers = {
+	"referer": "https://ndlonline.ndl.go.jp/"
+}
+
+function getKeyFromUrl(url) {
+	var key = null;
+	var m = url.match(/detail\/([^\/]+)/);
+	if (m) {
+		key = m[1];
+	}
+	return key;
+}
+
+function getJsonUrl(url) {
+	var key = getKeyFromUrl(url);
+	var jsonUrlTemplate = "https://ndlonline.ndl.go.jp/risapi/search?id={ key }&lang=ja_JP&page=1&rows=20&searchPattern=ris_simple&start=0";
+	return jsonUrlTemplate.replace(/{\s+key\s+}/, key);
+}
+
+function callData(urls, headers) {
+	var jsonUrls = [];
+	for (var url of urls) {
+		jsonUrls.push(getJsonUrl(url));
+	}
+	ZU.doGet(jsonUrls, scrape, false, "utf8", headers);
+}
+
+function getCreators(details) {
+	var creatorSplit;
+	var creators = [];
+	if (!details.creator) return creators;
+	for (var cObj of details.creator) {
+		for (var i=0,ilen=cObj.length; i<ilen; i++) {
+			if (cObj[i] === "creator") {
+				var newCreator = {};
+				var creatorParts = cObj.slice(i+2);
+				for (var creatorPart of creatorParts) {
+					if (creatorPart[0] === "name") {
+						creatorSplit = creatorPart[2].split(/\s+/);
+						newCreator.lastName = creatorSplit[0];
+						newCreator.firstName = creatorSplit[1];
+						newCreator.creatorType = "author";
+					} else if (creatorPart[0] === "transcription") {
+						creatorSplit = creatorPart[2].split(/\/\//);
+						newCreator.multi = {
+							_key: {
+								"en": {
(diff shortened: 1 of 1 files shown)


zuphilip:
Can you add some test cases?

fbennett:
Will do. I'm working now to extend the coverage a bit. When the next version is done, I'll add test cases for each type.

zuphilip:
Great! Looking forward for the next version with the tests.

BTW you should also add an AGPL license and you can look at the other small issues which the CI script shows:
```
./.ci/checkTranslator.sh  National\ Diet\ Library\ Periodicals.js
```

adam3smith:
> Item type discrepancies between detectWeb() and the result in doWeb() are going to throw errors in almost all of the tests. Is that acceptable?

Yes, we have some other translators like Primo where detect web isn't accurate for some resources. 

I think I would comment out those tests, since they data is helpful but they're more than useless as automated tests

fbennett:
It should be happier now. It's been revised to (attempt to) cover books, journal articles, standards, PhD theses, video recordings, and audio recordings, and maps.

fbennett:
I fixed a few small issues in a further checkin. Not major problems, just a bit of fine-tuning.

fbennett:
NDL reference numbers are available in most CiNii pages. CiNii metadata is pretty crappy, and the RDF there is a pain to work with, so we might consider diverting CiNii grabs to this translator when possible, at some point.

fbennett:
Hmm. We might want to rename this file to "National Diet Library Online.js"?

zuphilip:
>  We might want to rename this file to "National Diet Library Online.js"?

Sounds good. The word "online" is part of the url and prominently on their page.

If you change the label on the first page in Scaffold and save, then also the file is renamed accordingly.

zuphilip:
@fbennett Is this ready for review?

fbennett:
`monitorDOMChanges()` did not help. It looks like this site has the same problem as HeinOnline: the URL and site content are changed programmatically, without registering the state change in page history. As a result, the site translator is not triggered (or is triggered on the wrong URL). User has to refresh the page to change the icon. The translator really shouldn't go live in this state.

I will try to contact someone inside the NDL, to see if we can get the site fixed.

fbennett:
Came across a couple of bugs in Imperial date conversion when running that code in another translator. All fixed now.

National Diet Library Online.js
```diff
@@ -0,0 +1,811 @@
...
+				"itemType": "book",
+				"title": "Choctaw tales / collected and annotated by Tom Mould",
+				"creators": [
+					{
+						"lastName": "Mould,",
+						"firstName": "Tom,",
+						"creatorType": "author",
+						"multi": {
+							"_key": {}
+						}
```
zuphilip: What is this `multi`?

National Diet Library Online.js
```diff
@@ -0,0 +1,811 @@
...
+							creatorSplit = creatorPart[2].split(/(?:\/\/|,)/);
+							newCreator.multi = {
+								_key: {
+									"en": {
+										lastName: creatorSplit[0],
+										firstName: creatorSplit[1],
+										creatorType: creatorType
+									}
+								}
+							};
```
zuphilip: I guess this is for Juris-M only?

National Diet Library Online.js
```diff
@@ -0,0 +1,811 @@
...
+			institutionalEditor = {
+				lastName: publicationMainSplit[1].replace(/\u7d38/, ""),
+				fieldMode: 1,
+				creatorType: "editor"
+			};
+		}
+		if (publicationSplit[1]) {
+			var publicationAlternativeSplit = publicationSplit[1].split(/\s+\/\s+/);
+			item.multi._keys.publicationTitle = {};
+			item.multi._keys.publicationTitle[altLang] = publicationAlternativeSplit[0];
```
zuphilip: I guess this is for Juris-M only?

National Diet Library Online.js
```diff
@@ -0,0 +1,811 @@
...
+			if (publicationAlternativeSplit[1] && institutionalEditor) {
+				institutionalEditor.multi = {_key:{}};
+				institutionalEditor.multi._key[altLang] = {
+					lastName: publicationAlternativeSplit[1],
+					fieldMode: 1,
+					creatorType: "editor"
+				};
```
zuphilip: I guess this is for Juris-M only?

National Diet Library Online.js
```diff
@@ -0,0 +1,811 @@
...
+					return elem;
+				}).map(function(elem){
+					while (elem.length < 2) {
+						elem = "0" + elem;
+					}
+					return elem;
+				}).join("-");
+		}
+	}
+	return dateStr;
```
zuphilip: Did you try to use `ZU.strToISO` for some of the date parsing here?

National Diet Library Online.js
```diff
@@ -0,0 +1,811 @@
...
+	// Uncomment to dump the data object in all its glory
+	// Zotero.debug(JSON.stringify(obj, null, 1));
+	obj = obj.response.docs[0];
+	var details = obj.view_detailParameters;
+
+	var itemType = getItemType(details);
+	var item = new Zotero.Item(itemType);
+
+	var altLang;
+	if (item.language === "jpn") {
```
zuphilip: The `item` is just defined two lines before as a new Zotero item. How can then the `item.language` field be set?

National Diet Library Online.js
```diff
@@ -0,0 +1,811 @@
...
+		}
+		item.date = dateSplit.join("-");
+	}
+
+	if (item.title) {
+		item.title = item.title.replace(/^\[(.*)\]$/, "$1");
+	}
+	item.url = getPageUrl(obj.risGrouping_parentIssToken_ssd);
+
+	item.multi._keys.title = {};
```
zuphilip: I guess this is for Juris-M only?

National Diet Library Online.js
```diff
@@ -0,0 +1,811 @@
...
+		return "multiple";
+	} else if (info.single) {
+		return "journalArticle";
+	} else {
+		var body = ZU.xpath(doc, "//body")[0];
+		Z.monitorDOMChanges(body, {childList: true, subtree: true});
+		if (info.single) {
+			return "journalArticle";
+		}
+	}
```
zuphilip: I don't think we need to put these lines in the else case, because in the other two cases we return something and will not continue here. Moreover, I think we can delete the second check for `info.single` because when the DOM changes then the function `detectWeb` will be called again.

AbeJellinek:
Alright, I took a stab at addressing some of @zuphilip's comments on this, but I can't push to the `ndl-periodicals` branch. @fbennett, could you see if you can enable maintainer contributions? I'd love to get this merged.

PR #1661, opened 2018-05-21
ScienceDirect: fixes a buggy pdf query selector

adomasven:
Report here https://forums.zotero.org/discussion/comment/307937/#Comment_307937

Safari seems to be served different content than other browsers and the conditional branch on that different content had a buggy query selector.


zuphilip:
Thank you very much @adomasven!

BTW do you also see "error parsing XML" in Scaffold when running the automatic tests? Cf. https://github.com/zotero/scaffold/issues/74

PR #1668, opened 2018-06-01
Japanese Diet Committee Transcripts

fbennett:
The Japanese National Diet Library hosts transcripts of hearings in the National Diet. This translator supports references to statements by individual speakers in the Diet records. The site is an old-school design that makes heavy use of frames, and there are a couple of quirks in the translator.

First, in Scaffold, the Regexp Test of the URL under the Metadata tab will fail on a test page, for reasons explained below.

Second, test fixtures cannot be usefully set for the translator, because access to the content pages depends on acquiring a session ID from the landing page of the site. It is not possible to set persistent links to the inner pages.

Third, in order to build the `multiple` listing (a list of speaker statements in a given session), I have had to set the target of the translator as the URL of the page, and then wrap the call to `Zotero.selectItems()` in a call to `ZU.processDocuments()` that fetches the frame with the list data that we need. This seems to work just fine, but it has two weird side-effects.

The first side-effect is that the Regexp Test button always fails in Scaffold, because Scaffold sees only the URLs of the frames, and not the URL of the outer page. I tried hard to get `detectWeb()` to locate the frame with the list data, but I gave up in the end.

The second side-effect is a little more serious. The item picker from `Zotero.selectItems()` comes up just fine, and selected items download correctly, but there is no feedback in the UI during the download. No popup to show the target collection, no progress displays of the items being added, nothing. I think this may be due to having `Zotero.selectItems()` wrapped in `ZU.processDocuments()`, but I don't see a way around that on this site.

### DIFF
--- a/Japanese Diet Committee Transcripts.js
+++ b/Japanese Diet Committee Transcripts.js
@@ -1,0 +1,449 @@
+{
+	"translatorID": "ad3a50fa-4f2f-4ca8-9fdf-eabc03f3fc57",
+	"translatorType": 4,
+	"label": "Japanese Diet Committee Transcripts",
+	"creator": "Frank Bennett",
+	"target": "http://kokkai.ndl.go.jp/cgi-bin/KENSAKU/swk_dispdoc.cgi",
+	"minVersion": "1.0.0b3.r1",
+	"maxVersion": null,
+	"priority": 100,
+	"inRepository": true,
+	"browserSupport": "gcsv",
+	"lastUpdated": "2018-06-01 18:21:18"
+}
+
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2018 Frank Bennett
+
+	This file is part of Zotero.
+
+	Zotero is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


fbennett:
Thanks again - I'll fix this up and push an update.

fbennett:
I've replaced the unneeded `processDocuments()` call with a couple of DOM operations as suggested by Dan. Unfortunately it still isn't showing anything in the UI when downloads are in progress. Everything works otherwise, though.

fbennett:
Added small bits of important information about speakers -- phonetic form of name, Diet faction (for Diet members), title (for Ministry officials), and committee role, each if available. Also added the index number of the statement captured, as `archiveLocation`.

adam3smith:
@fbennett did some clean-up but otherwise looks good -- is this ready to go then? Could we get a couple of tests? If tests don't work, at least include links to test with?

AbeJellinek:
This is now out of date - the site has moved to kokkai.ndl.go.jp and has a new structure. I don't speak Japanese and wouldn't even know where to start with updating it. If @fbennett wants to write a translator for the current site, that would be excellent, but I'm going to close for lack of activity for now.

PR #1678, opened 2018-06-21
[the Lens] New translator

eurruby:
This is a beta translator for https://www.lens.org/

### DIFF
--- a/the Lens.js
+++ b/the Lens.js
@@ -1,0 +1,215 @@
+{
+	"translatorID": "b5f3b896-df57-4c68-8841-ea2171ee7419",
+	"label": "the Lens",
+	"creator": "Laurent Lhuillier",
+	"target": "^https?://www\\.lens\\.org",
+	"minVersion": "4.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2018-06-21 09:15:02"
+}
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+	
+	the Lens translator
+	
+	This file is part of Zotero.
+	
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+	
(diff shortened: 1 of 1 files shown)


the Lens.js
```diff
@@ -0,0 +1,215 @@
...
+	newItem.filingDate = filingDate.toISOString().substr(0,10);
+
+	var ul = doc.getElementById("inventorFilter");
+	var liNodes = [];
+	for (var i = 0; i < ul.childNodes.length; i++) {
+		if (ul.childNodes[i].nodeName == "LI") {
+			liNodes.push(ul.childNodes[i]);
+		}
+	}
+	iAu = liNodes.length;
```
zuphilip: It seems for me that you don't need much of these...

I suggest to do something like
```js
var inventors = ZU.xpath(doc, '//*[@id="inventorFilter"]/li/a');
for (let inventor of inventors) {
   newItem.creators.push(ZU.cleanAuthor(inventor, "inventor", false);
}
```
(not tested)

the Lens.js
```diff
@@ -0,0 +1,215 @@
...
+	***** END LICENSE BLOCK *****
+*/
+
+function detectWeb(doc, url) {
+	if(url.indexOf("search?") !== -1) {
+		return "multiple";
+	}
+		
+	if (url.indexOf("patent") !== -1 && getTitle(doc)) {
+		return "patent";
```
zuphilip: What about some of the other subpages like https://www.lens.org/lens/patent/109-941-370-099-325/citations  ?

the Lens.js
```diff
@@ -0,0 +1,215 @@
...
+
+	newItem.creators.push(Zotero.Utilities.cleanAuthor(authors[iAu], "inventor", true));
+	}
+
+	var pdfurl = ZU.xpathText(doc, '//*[@id="patentFrontPage"]/div[2]/a/@href');
+	newItem.attachments.push({
+		title: 'Full Text PDF',
+		url: pdfurl,
+		mimeType: 'application/pdf'
+	});
```
zuphilip: We should test here that `pdfurl` has some contents and only then do these lines here.

eurruby:
no sorry I can't because I changed job

AbeJellinek:
The Lens now has COinS metadata that Zotero scrapes quite well. Closing this.

PR #1684, opened 2018-07-01
Add PC Gamer.js

owcz:
https://www.pcgamer.com

* based on [EM](https://github.com/zotero/translators/blob/master/Embedded%20Metadata.js) with fixes for bad metadata
* handles articles, search, categories, multiple authors

this outlet has a magazine too but I think the type is better set as "blog" since one doesn't appear to mirror the other

### DIFF
--- a/PC Gamer.js
+++ b/PC Gamer.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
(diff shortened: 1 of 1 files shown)


owcz:
this outlet has a magazine too but I think the type is better set as "blog" since one doesn't appear to mirror the other

adam3smith:
Thanks!

PR #1685, opened 2018-07-02
EBSCOhost fixes: handle full text HTML, tags

owcz:
major translator hasn't been saving full text in recent memory

this patch handles tags with regex from the RIS's N1 and saves full text to a note, since I can't find a way to save a snapshot with the convoluted session ID system

citation will drop the url if there is no full text (HTML or PDF) to link

this patch also resolves random errors i had in the multi, likely due to attachment failures

### DIFF
--- a/EBSCOhost.js
+++ b/EBSCOhost.js
@@ -12,10 +12,6 @@
 	"lastUpdated": "2018-07-02 02:12:50"
 }
 
-// attr()/text() v2 per https://github.com/zotero/translators/issues/1277
-function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null;}
-function textr(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null;}
-
 
 function detectWeb(doc, url) {
 	// See if this is a search results or folder results page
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks for working on this!
I'm not happy with having the full text in a note, though -- that's going to be confusing for users and for long texts may break Zotero sync. Haven't tested, but couldn't we build an html doc that we then attach?

owcz:
@adam3smith if you have an example of a translator doing that, I can use it as a model

adam3smith:
I don't think we do this anywhere, so there's no template.

owcz:
Is there something in [here](https://github.com/zotero/translators/blob/f24f9494133c2f5299ce310725f3dda18429785f/Google%20Scholar.js#L527-L546) to reuse? Looks promising but I can't find documentation for that `composeDoc` function

adam3smith:
I agree it looks promising, but I don't know either. 
@dstillman -- where is `ZU.composeDoc`? I can't even find it in the source code, but it clearly exists.

dstillman:
That's apparently MLZ stuff [that Frank added](https://github.com/zotero/translators/commit/b37c487648e5458256aecab7cab3d793a1600c68) (which I'm not crazy about having in the main translator). Being able to build an HTML file could be useful in general, though.

adam3smith:
@dstillman I'd like your input on this please -- not getting EBSCO full texts isn't good; currently they're broken Snapshots, which is even worse. I don't like putting this into a note, so the ideal solution would be a function that's equivalent to whatever Frank's `composeDoc` does. Is that in the cards?

adam3smith:
ping @dstillman

zuphilip:
Did you try out DOMParser with `parseFromString` function? Can the output of that function saved as an attachment in Zotero?

dstillman:
DOMParser won't work in the client currently, but we're hoping to fix that with a new, less restrictive sandboxing method.

The larger problem is that the connector [doesn't actually pass the HTML document to the client](https://github.com/zotero/zotero/blob/01e2f32d03bdb758244e0a984a81363877771d13/chrome/content/zotero/xpcom/translation/translate.js#L155), only the URL, since there's not really a good way for it to do so with auxiliary files (unless we find a snapshot replacement that can run in the connector). But simple fix here would be a `parseHTMLDocument()` function that would use DOMParser and also set a special property in the object in the `attachments` array that caused it to be sent to the client as raw HTML source.

PR #1687, opened 2018-07-05
Add MCV.js

owcz:
Translator for https://www.mcvuk.com

### DIFF
--- a/MCV.js
+++ b/MCV.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2018-07-08 02:49:52"
+	"lastUpdated": "2018-07-08 02:58:59"
 }
 
 /*
@@ -56,8 +56,9 @@ function scrape(doc, url) {
 	translator.setDocument(doc);
 	
 	translator.setHandler('itemDone', function (obj, item) {
-		item.itemType = "blogPost";
+		item.itemType = "magazineArticle";
 		item.publicationTitle = "MCV";
+		item.ISSN = "1469-4832";
 		item.date = attr(doc,'meta[name="published"]','content');
 		item.title = item.title.replace(/- MCV$/,'');
 		item.language = item.language.replace('us','US');
--- a/MCV.js
+++ b/MCV.js
@@ -1,0 +1,206 @@
+{
+	"translatorID": "b51ac026-ed35-4c68-89bb-b42b1e1ce8f2",
+	"label": "MCV",
+	"creator": "czar",
+	"target": "^https?://(www\\.)?mcvuk\\.com",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2018-07-05 01:11:13"
+}
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2018 czar
+	http://en.wikipedia.org/wiki/User_talk:Czar
+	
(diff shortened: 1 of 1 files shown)


MCV.js
```diff
@@ -0,0 +1,206 @@
...
+
+
+// attr()/text() v2 per https://github.com/zotero/translators/issues/1277
+function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null;}function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null;}
+
+
+function detectWeb(doc, url) {
+	if (/\/(business|development|esports|influence)\//.test(url)) {
+		return "blogPost";
+	} else if (url.indexOf("/search?query=") != -1) {
```
adam3smith: use `string.includes()` for string testing & take advantage of `getSearchResults`

MCV.js
```diff
@@ -0,0 +1,206 @@
...
+		return "multiple";
+	} else return null;
```
adam3smith: no need for the `else`; that's implicit.

MCV.js
```diff
@@ -0,0 +1,206 @@
...
+	var translator = Zotero.loadTranslator('web');
+	translator.setTranslator('951c027d-74ac-47d4-a107-9c3069ab7b48'); // embedded metadata
+	translator.setDocument(doc);
+	
+	translator.setHandler('itemDone', function (obj, item) {
+		item.itemType = "blogPost";
+		item.publicationTitle = "MCV";
+		item.date = attr(doc,'meta[name="published"]','content');
+		item.title = item.title.replace(/- MCV$/,'');
+		if (item.creators[0].lastName == "Editors") {
```
adam3smith: just to make sure -- this would always be the first author? No need to cycle through creatores?

owcz: yep this is just so the generic "MCV Editors" byline doesn't get digested into "Editors, MCV"

don't believe I've seen two+ authors on this site, but the generic byline wouldn't be used with a second author, fwiw

owcz:
good point re: type
I didn't know if all or just some of their content was also in their magazine
but changed the type anyway

MCV.js
```diff
@@ -0,0 +1,208 @@
...
+*/
+
+
+// attr()/text() v2 per https://github.com/zotero/translators/issues/1277
+function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null;}function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null;}
+
+
+function detectWeb(doc, url) {
+	if (/\/(business|development|esports|influence)\//.test(url)) {
+		return "blogPost";
```
adam3smith: update the item type here and in `doWeb` and update tests

MCV.js
```diff
@@ -0,0 +1,208 @@
...
+	translator.setTranslator('951c027d-74ac-47d4-a107-9c3069ab7b48'); // embedded metadata
+	translator.setDocument(doc);
+	
+	translator.setHandler('itemDone', function (obj, item) {
+		item.itemType = "magazineArticle";
+		item.publicationTitle = "MCV";
+		item.ISSN = "1469-4832";
+		item.date = attr(doc,'meta[name="published"]','content');
+		item.title = item.title.replace(/- MCV$/,'');
+		item.language = item.language.replace('us','US');
```
adam3smith: just in case let's wrap this into an `if (item.language)` so it doesn't break things if they remove language or it's missing.

PR #1691, opened 2018-07-06
Add translator for dn.se

sebastian-berlin-wmse:
This is the website for the Swedish newspaper Dagens Nyheter.

### DIFF
--- a/Dagens Nyheter.js
+++ b/Dagens Nyheter.js
@@ -131,11 +131,7 @@ function scrape(doc, url) {
 			item.abstractNote = abstractString.replace(/&nbsp;/g, " ");
 		}
 
-		var sectionString =
-			attr(doc, 'meta[property="article:section"]', "content");
-		if(sectionString) {
-			item.section = sectionString;
-		}
+		item.section = attr(doc, 'meta[property="article:section"]', "content");
 
 		var timeString =
 			attr(doc, 'meta[property="article:published_time"]', "content");
--- a/Dagens Nyheter.js
+++ b/Dagens Nyheter.js
@@ -1,0 +1,373 @@
+{
+	"translatorID": "192a29de-3d6e-4850-984f-943764126429",
+	"label": "Dagens Nyheter",
+	"creator": "Sebastian Berlin",
+	"target": "^https?://www\\.dn\\.se/(nyheter|ekonomi|kultur-noje|sport|sok)/",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2018-07-19 12:35:38"
+}
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2018 Sebastian Berlin
+	
+	This file is part of Zotero.
+
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+
+// attr()/text() v2
+function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null;}function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null;}
+
+
+function detectWeb(doc, url) {
+	if (url.includes('/sok/')) {
+		if (getSearchResults(doc, true)) {
+			return "multiple";
+		}
+	} else {
+		return "newspaperArticle";
+	}
+}
+
+
+function getSearchResults(doc, checkOnly) {
+	var items = {};
+	var found = false;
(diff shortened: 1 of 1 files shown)


Dagens Nyheter.js
```diff
@@ -0,0 +1,230 @@
+{
+	"translatorID": "192a29de-3d6e-4850-984f-943764126429",
+	"label": "Dagens Nyheter",
+	"creator": "Sebastian Berlin",
+	"target": "https://www.dn.se/(nyheter|sok)/",
```
adam3smith: The Target is a regular expression, i.e. special characters should be escaped and for efficiency you should mark the string beginning, i.e. here:
`^https?://www\.dn\.se/(nyheter|sok)/' (this is when you do it in Scaffold. If you write straight to the translator, you need double backslashes -- www\\.dn\\.se)

adam3smith: Also, this looks like it might be too restrictive? e.g. 
https://www.dn.se/sport/fotboll/sverige-ute-ur-fotbolls-vm-forlust-mot-england-i-kvarten/
https://www.dn.se/ekonomi/ostkriget-skrammer-skanska-bonder/

Dagens Nyheter.js
```diff
@@ -0,0 +1,230 @@
...
+	***** END LICENSE BLOCK *****
+*/
+
+
+// attr()/text() v2
+function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null;}function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null;}
+
+
+function detectWeb(doc, url) {
+	if (url.includes('/nyheter/')) {
```
adam3smith: If I'm right about the regex above, this is too restrictive as well.

Dagens Nyheter.js
```diff
@@ -0,0 +1,230 @@
...
+	translator.setTranslator('951c027d-74ac-47d4-a107-9c3069ab7b48');
+
+	translator.setHandler('itemDone', function (obj, item) {
+		item.publicationTitle = "Dagens Nyheter";
+
+		var nameString = attr(doc, 'div[class="js-article"]', "data-authors");
+		var author = ZU.cleanAuthor(nameString, "author");
+		if(author.firstName === "") {
+			// If there's only one name, the author is not a person,
+			// e.g. "TT".
```
adam3smith: Is this actually helpful information? I'm happy to go either way if you have a strong opinion, but would we really want a citation to start with "TT" -- my assumption is that mostly not. So maybe just remove these sorts of authors entirely?

Dagens Nyheter.js
```diff
@@ -0,0 +1,230 @@
...
+			author.firstName = undefined;
+			author.fieldMode = true;
+		}
+		item.creators = [author];
```
adam3smith: Are there really no articles with multiple authors?

Dagens Nyheter.js
```diff
@@ -0,0 +1,230 @@
...
+
+		var abstractString =
+			attr(doc, 'div[class="js-article"]', "data-article-description");
+		item.abstractNote = abstractString.replace("&nbsp;", " ");
```
adam3smith: This only replaces the first nbsp -- is that on purpose? If you want to replace all, you need to use a regex with the g flag

adam3smith: Also, both here and for section below, please test for the scraped string before processing it, since a missing string will otherwise cause an error, i.e. simply `if (abstractString) item.abstractNote .= ..`

Dagens Nyheter.js
```diff
@@ -0,0 +1,230 @@
...
+	}
+}
+
+
+function scrape(doc, url) {
+	var translator = Zotero.loadTranslator('web');
+	// Embedded Metadata
+	translator.setTranslator('951c027d-74ac-47d4-a107-9c3069ab7b48');
+
+	translator.setHandler('itemDone', function (obj, item) {
```
adam3smith: it doesn't look like you're using the EM translator results here at all, are you? In that case you might as well just create a new item and scrape, i.e. `var item = new Zotero.Item("newspaperArticle")`. While in general we push towards using EM where it helps, this isn't a blanket rule

Dagens Nyheter.js
```diff
@@ -0,0 +1,230 @@
...
+		var abstractString =
+			attr(doc, 'div[class="js-article"]', "data-article-description");
+		item.abstractNote = abstractString.replace("&nbsp;", " ");
+
+		item.section = attr(doc, 'meta[property="article:section"]', "content");
+
+		var timeString =
+			attr(doc, 'meta[property="article:published_time"]', "content");
+		item.date = timeString.split("T")[0];
+
```
adam3smith: Let's add the ISSN for the publication as `item.ISSN` please. I assume that's going to be hardcoded.

Dagens Nyheter.js
sebastian-berlin-wmse: Since the paper itself cites TT as a source, I'd say it's relevant. I don't know if Zotero has some other criteria as to what counts as an author, though. If you want to know more about TT: https://en.wikipedia.org/wiki/Tidningarnas_Telegrambyr%C3%A5 and their own "about us": https://tt.se/en/about-us/.

Dagens Nyheter.js
sebastian-berlin-wmse: Actually, yes. When I added tests for more articles, I found some that did. Added logic to handle that.

Dagens Nyheter.js
sebastian-berlin-wmse: Huh, I didn't know that `replace()` was implemented differently depending on if you used string or regex. No, that wasn't on purpose; changed this.

Dagens Nyheter.js
sebastian-berlin-wmse: It seems that at least `item.title` is set by that translator. I get this if remove it:
```
Error: No title specified for item
```

adam3smith:
Thanks, looks good. For future reference, please don't squash commits after updating a PR -- it's much easier if I can see what you changed from the PR I already reviewed.

sebastian-berlin-wmse:
Thank you and sorry about the mess in the PR. I don't usually work with Github, so I didn't quite remember how you're supposed to do it and when I realized I'd messed up, it was already committed. I'll do it properly next time.

Issue #1682, opened 2018-06-27
[Datacite]: Import fields not included in item type to Extra


adam3smith:
E.g.
`curl -LH "Accept: application/vnd.citationstyles.csl+json" https://doi.org/10.17171/2-3-1`

Has a `container-title` field that makes a lot of sense for them and does show up e.g. in APA citations, but doesn't import into Zotero. We should include it in Extra as `container-title: xxxx` just as we do with the DOI. While we're at it, we should also make datasets produce `type: dataset` (which I thought we did already)

adam3smith:
Having looked at this some, I think the only way to reasonably do this is on a case-by-case basis, so I'll close this from the one open PR

PR #1692, opened 2018-07-08
Add type & container-title to Datasets

adam3smith:
Make sure DOI doesn't overwrite extra.
Partially addresses https://github.com/zotero/translators/issues/1682
I'm not sure there's an easy solution that does this across the board, but ideas welcome.

### DIFF
--- a/DataCite.js
+++ b/DataCite.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 8,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2018-07-08 11:41:00"
+	"lastUpdated": "2018-07-09 07:15:00"
 }
 
 function detectSearch(items) {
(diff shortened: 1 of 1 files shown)


DataCite.js
```diff
@@ -54,9 +54,14 @@ function doSearch(items) {
 function fixJSON(text) {
 	try {
 		var item = JSON.parse(text);
-		
-		if(item.type == 'misc') item.type = 'article-journal';
-		
+	  //Z.debug(item)
+		if (item.type == 'misc') item.type = 'article-journal';
+		if (item.type == 'dataset') {
+			item.note = "type: dataset";
```
zuphilip: Shouldn't we also add a line
```
item.type = 'article-journal';
```
as well here? Because dataset is still imported as document.

adam3smith: no, we switched that recommendation a while ago to suggest `Document` with `type: dataset` in Extra ; since we need to use the Extra field anyway and Document has a publisher, that seemed to make more sense (I think you may even have been part of that conversation), see https://www.zotero.org/support/dev/translators/datasets

adam3smith:
Is there a reasonably easy way to add a test? I wouldn't really know how to -- Z.debug(item) right before item.complete and then copy the output from the debug? Or is there an easier option?

DataCite.js
zuphilip: Okay, thanks for the info.

zuphilip:
Debugging the item should work. I also have put together a very simple export translator for that https://github.com/UB-Mannheim/zotkat/blob/master/Zotero%20JSON.js . If everything is complicated then also just the url and a comment is helpful.

adam3smith:
Thanks! Having the JSON export is convenient.

PR #1696, opened 2018-07-14
Update Eurogamer for March 2018 redesign

owcz:
- updated translator to work post-redesign
- combined similar Eurogamer and USgamer translators (same owner, slightly different code)
- expanded detect to Eurogamer in other languages
- removed Framework #1277, now starts from EM


owcz:
`e570f517-83f2-4735-97d2-44499aee0b21` was renamed, not deleted, so false alarm on Travis CI, I think

adam3smith:
@zuphilip if you have a chance to look at the false positive test above at some point, that'd be great.

adam3smith:
Thanks!

zuphilip:
@adam3smith Renames are difficult to detect for the script. This works only reliable when you use `git mv` for that. See also https://github.com/zotero/translators/pull/1357

adam3smith:
Ah yes, thanks. OK, we'll just know to double-check & ignore the test in those cases then

PR #1701, opened 2018-07-30
Revised SROWA translator

ontomoc:
Recoded to eliminate the use of FW, as requested. Could you please check whether the revision meets Zotero requirements.

### DIFF
--- a/State Records Office of Western Australia.js
+++ b/State Records Office of Western Australia.js
@@ -1,3 +1,16 @@
+{
+	"translatorID": "2b62db1e-1d49-4e2b-a409-a3f9f1870a41",
+	"label": "State Records Office of Western Australia",
+	"creator": "Mike O'Connor",
+	"target": "^https://archive\\.sro\\.wa\\.gov\\.au/",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2018-10-05 03:30:23"
+}
 /*
 	***** BEGIN LICENSE BLOCK *****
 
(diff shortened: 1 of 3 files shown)


State Records Office of Western Australia.js
```diff
@@ -0,0 +1,372 @@
...
+			}
+		]
+	},
+	{
+		"type": "web",
+		"url": "https://archive.sro.wa.gov.au/index.php/cancelled-public-plans-districts-post-standard-series-s980",
+		"items": [
+			{
+				"itemType": "manuscript",
+				"title": "CANCELLED PUBLIC PLANS (DISTRICTS) - POST STANDARD SERIES",
```
adam3smith: question: we can try to convert these all caps strings to something more sensible -- would that make sense or is it important that they stay in all caps for some reason?

ontomoc: Thank you for improving my javascript. I will:
1. Review that it still works as expected.
2. Add some code to convert the caps to Title case or similar (SROWA have OK'ed that).
3. Add the tests.

Also, is ZU.Xpath still acceptable in translators?  I went with DOM methods eg. querySelectorAll(), but this may have added

adam3smith: We prefer using CSS selectors over xpaths for new translators if at all possible, yes, so that was the right call.

ontomoc:
Revised translator ready for review.

ontomoc:
BTW, the Zotero icon doesn't show in the browser's address bar when Zotero makes available this translator.

adam3smith:
(I'm still making changes to this, so please leave it for the time being)

adam3smith:
OK, take a look at my changes -- this is ready to merge from my end now.

ontomoc:
Thank you for the checks and improvements. Have worked through changes, and run some tests. Results as expected. Happy for it to be merged now.

adam3smith:
Great, thanks for checking!

PR #1705, opened 2018-08-04
TEI.js fix unvalid generated xml-ids  and imprint

oriflamms:
TEI.js fix invalid generated xml-ids by replacing invalid characters by underscore and creates empty <date/>, <publisher/> and <pubPlace/> elements if there is no info

### DIFF
--- a/TEI.js
+++ b/TEI.js
@@ -359,19 +359,20 @@ function generateItem(item, teiDoc) {
 			curCreator.appendChild(forename);
 		}
 		if (item.creators[i].lastName) {
-			var surname = null;
-			if (item.creators[i].firstName) {
-				surname = teiDoc.createElementNS(ns.tei, "surname");
-			}
-		else {
-			surname = teiDoc.createElementNS(ns.tei, "name");
-			}
-			
-			surname.appendChild(teiDoc.createTextNode(item.creators[i].lastName));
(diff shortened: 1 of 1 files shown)


TEI.js
```diff
@@ -446,10 +451,14 @@ function generateItem(item, teiDoc) {
 		imprint.appendChild(imprintDate);
 	}
 
-	// flag unpublished if there is no date | publisher | place
+	// flag unpublished if there is no date | publisher | place by creating empty elements date | publisher | pubPlace
```
adam3smith: this doesn't work as is: it creates an empty node for all three if any one of them doesn't exist. Why not instead just tag this on as `else` conditions individually for `if (item.date)` etc.?

oriflamms: As a matter of fact, it did not work that way and in the previous code neither (while the pipes should have said OR, it was correct for all items with at least one of the field). Following your proposition and given that only one field is needed, I've added "!(item.date)" => empty date element.
And I've added some lines to keep the website title (webpage as analytic

adam3smith:
Thanks -- and sorry about the "or" mixup. `!(a||b)` is of course `!a && !b`. Just checking for date should  work nicely, though. If people want the empty nodes and/or the comment back we can do so later.

Issue #1702, opened 2018-07-30
New ACL Anthology


nschneid:
http://aclanthology.info/ a.k.a. http://aclanthology.coli.uni-saarland.de/ has replaced http://www.aclweb.org/anthology/, which will no longer index new publications.

The Embedded Metadata translator partially works for the new site, but it would be better to have a translator based on BibTeX, e.g. to distinguish conference and journal papers.

@GuyAglionby, thanks for #1577—would it be straightforward to adapt the ACLWeb translator to the new anthology?

GuyAglionby:
Hey there. I suspect the new translator will be pretty straightforward to write. A lot of the complexity in the other translator handles cases where BibTeX is not available (e.g. [here](http://www.aclweb.org/anthology/Q/Q14/)), but this doesn't appear to be an issue on the new website (e.g. [corresponding](https://aclanthology.coli.uni-saarland.de/events/tacl-2014)). As such I'd probably just start from scratch -- shouldn't take long. Happy to do this but will have to be late next week. Thanks!

Evpok:
<http://aclweb.org/anthology> is now claiming it is the latest version of ACL anthology and the exporter is not doing well there (e.g. failing to pick up the venue). Are there any plans to support this version?

nschneid:
In light of https://github.com/acl-org/acl-anthology/issues/170#issuecomment-475883559, https://github.com/zotero/translators/blob/master/ACLWeb.js needs to be removed and https://github.com/zotero/translators/blob/master/ACLAnthology.js updated to work with the newest version of the site.

We are also waiting for #1797 to be merged.

mjpost:
Thanks for working on this! If there are things the Anthology can do to facilitate integration please let us know. 

I’ll add that it might look like there’s a lot of churn with Anthology formats, but the recent static rewrite is the third version of the website (under http://www.aclweb.org/anthology) and should be stable for many years to come.

Evpok:
Side note (and I'll open a separate feature request if suitable) : it would also be nice to be able to download the supplementary materials (appendix, posters, slides…). Is it possible from the translator side or should it be done in a full-fledged add-on?

mjpost:
What would you need for this to happen? Additional fields in the BibTeX? Or a BibLaTeX file?

nschneid:
I think the translator could retrieve these without modification to the Anthology .bib files, but @GuyAglionby would know better.

Evpok:
[...]

PR #1707, opened 2018-08-09
Add translator for new ACL Anthology site

GuyAglionby:
New translator for https://aclanthology.coli.uni-saarland.de/ (cf #1702)

One small problem with the test case https://aclanthology.coli.uni-saarland.de/volumes/computational-linguistics-volume-44-issue-1-april-2018 . Scaffold happily imports it automatically, but when actually running it the `xpath` on line 74 erroneously returns nothing, causing `items` on line 88 to be empty and `Zotero.selectItems` on line 49 to error out with `Error: Translator called select items with no items`. This happens despite the test case working in manual testing. Not too sure what might cause this; any help appreciated!

Also worth noting that the PDFs of 2018 papers are still stored on the old domain (e.g. [paper overview](https://aclanthology.coli.uni-saarland.de/papers/J18-1001/j18-1001) with [corresponding PDF](http://aclweb.org/anthology/J18-1001)), so trying to save from the PDF is dealt with by the old translator (ACLWeb.js). This translator fails on the '18 papers, as they are not actually indexed on the old site. As a result, I suggest a small change to ACLWeb.js to pull BibTeX information from the new site, rather than scraping it from the old. Happy to do this if it sounds reasonable to you.

Thanks!

### DIFF
--- a/ACLAnthology.js
+++ b/ACLAnthology.js
@@ -1,0 +1,168 @@
+{
+	"translatorID": "9b2f2d29-02e3-4693-b728-4cd435650669",
+	"label": "ACLAnthology",
+	"creator": "Guy Aglionby",
+	"target": "^https://aclanthology\\.coli\\.uni\\-saarland\\.de/(events|papers|volumes)/[^#]+",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2018-08-09 10:23:41"
+}
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2018 Guy Aglionby
+	This file is part of Zotero.
+
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero.  If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+function detectWeb(doc, url) {
+	if (url.includes('/events/') || url.includes('/volumes/')) {
+		return 'multiple';
+	} else if (url.includes('/papers/')) {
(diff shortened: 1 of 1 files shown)


nschneid:
Sounds good to me. Thanks!

ACLAnthology.js
```diff
@@ -0,0 +1,168 @@
...
+		});
+	}
+}
+
+function extractPapersFromListing(doc) {
+	let listing = ZU.xpath(doc, '//div[@id="content"]//p');
+	let items = {};
+	for (let i = 0; i < listing.length; i++) {
+		let title = ZU.xpath(listing[i], 'strong')[0].textContent;
+		let authors = ZU.xpath(listing[i], 'br/following-sibling::a');
```
adam3smith: this seems quite fragile & will throw an error if the xpath doesn't  detect. We typically just display the title for select, why not stick to that?

ACLAnthology.js
```diff
@@ -0,0 +1,168 @@
...
+				let bibtexUrl = WEBSITE_STUB + partialBibtexUrl;
+				ZU.doGet(bibtexUrl, function(responseString, responseObj) {
+					scrapeBibtex(responseString);
+				});
+			});
+		});
+	} else if(url.endsWith('.bib')) {
+		scrapeBibtex(ZU.xpath(doc, '//pre')[0].innerHTML);
+	} else {
+		let partialBibtexUrl = ZU.xpath(doc, '//a[contains(., "BibTeX")]/@href')[0].value;
```
adam3smith: I think
`ZU.xpath(doc, '//a[contains(., "BibTeX")]')[0].href` should work & is cleaner

ACLAnthology.js
```diff
@@ -0,0 +1,168 @@
...
+		let bibtexUrl = WEBSITE_STUB + partialBibtexUrl;
+		ZU.doGet(bibtexUrl, function(responseString, responseObj) {
+			scrapeBibtex(responseString);
+		});
+	}
+}
+
+function extractPapersFromListing(doc) {
```
adam3smith: This is fine (except for the author comment below), but would you mind renaming this & following the structure of the `getSearchResults` function we use in templates? (see e.g. https://github.com/zotero/translators/blob/master/Denik%20CZ.js#L53 ) same name&structure for same function makes maintenance so much easier.

GuyAglionby:
Thanks for the review! I display the authors + the paper ID when selecting multiple as this is what we do with the old translator. Happy to change to paper title exclusively for consistency with the rest of the translators. 

As for the failing test, I can go ahead and remove it and instead add the link as a comment in the relevant part of the code, if this sounds reasonable?

adam3smith:
Thanks -- everything looks good on your end, but I want to see if I can't get the test to work.

adam3smith:
OK, I removed the test and added as comment for now. The discussion on zotero-dev for the functionality that I think would get the test to work is here: https://groups.google.com/forum/#!topic/zotero-dev/Dy01PtaZ53E
Thanks for working on this!

PR #1709, opened 2018-08-12
[CrossRef] Fix conference publication title, place, and conference name lookup

bmoenecl:
This fixes the lookup of publication title, place and conference name lookup for conference publications using CrossRef.

Perhaps adding a test case for such `conferencePublication` types is in order, but I can't get Scaffold to run the tests for the Crossref translator. The relevant XML structure was derived from [this example](http://www.crossref.org/openurl/?pid=zter:zter321&url_ver=Z39.88-2004&&rft_id=info:doi/10.1109/ISSCC.2017.7870285&noredirect=true&format=unixref).

### DIFF
--- a/CrossRef.js
+++ b/CrossRef.js
@@ -240,9 +240,9 @@ function processCrossRef(xmlOutput) {
 		metadataXML = ZU.xpath(itemXML, 'c:proceedings_metadata', ns);
 		seriesXML = ZU.xpath(metadataXML, 'c:proceedings_metadata', ns);
 
-		item.publicationTitle = ZU.xpathText(metadataXML, 'c:publisher/c:proceedings_title', ns);
-		item.place = ZU.xpathText(metadataXML, 'c:event_metadata/c:conference_location', ns);
-		item.conferenceName = ZU.xpathText(metadataXML, 'c:event_metadata/c:conference_name', ns);
+		item.publicationTitle = ZU.xpathText(metadataXML, 'c:proceedings_title', ns);
+		item.place = ZU.xpathText(itemXML, 'c:event_metadata/c:conference_location', ns);
+		item.conferenceName = ZU.xpathText(itemXML, 'c:event_metadata/c:conference_name', ns);
 	}
 
 	else if((itemXML = ZU.xpath(doiRecord, 'c:crossref/c:database', ns)).length) {
--- a/CrossRef.js
+++ b/CrossRef.js
@@ -9,7 +9,7 @@
 	"priority": 90,
 	"inRepository": true,
 	"browserSupport": "gcsv",
-	"lastUpdated": "2018-04-19 05:16:03"
+	"lastUpdated": "2018-08-13 12:17:00"
 }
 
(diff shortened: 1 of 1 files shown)


adam3smith:
Looks good, thanks! Tests can only be added manually for search translators, so we don´t expect that here, but I would like the URL used for this example in the translator as they are for other item types.

Given that I imagine this was originally based on a real example, it would be good to check a couple of other conference paper examples from other publishers (Springer and ACM come to mind) just to make sure they all follow the same pattern -- I have not found this to necessarily be the case in the past.

bmoenecl:
Thanks for your quick reply.

In this last commit I have manually added a test ([doi 10.1109/ISSCC.2017.7870285](http://www.crossref.org/openurl/?pid=zter:zter321&url_ver=Z39.88-2004&&rft_id=info:doi/10.1109/ISSCC.2017.7870285&noredirect=true&format=unixref))  for the `conferencePaper` item type and updated the `lastUpdated` field.

As per your suggestion, I've also checked a conference paper from ACM ([doi 10.1145/3209415.3209476](http://www.crossref.org/openurl/?pid=zter:zter321&url_ver=Z39.88-2004&&rft_id=info:doi/10.1145/3209415.3209476&noredirect=true&format=unixref)): it has a similar XML structure and the fields `publicationTitle`, `place`, and `conferenceName` are populated.
It seems that all conference papers from Springer are also book chapters, and are thus resolved as a `bookSection`.


The result of the test before and after the fix is compared below.

```diff
--- /tmp/crossref/a_before.json	2018-08-13 05:46:55.000000000 +0200
+++ /tmp/crossref/a_after.json	2018-08-13 05:47:20.000000000 +0200
@@ -1,44 +1,44 @@
{
     "itemType": "conferencePaper",
     "creators": [
         {
             "creatorType": "author",
             "firstName": "Pen-Jui",
             "lastName": "Peng"
         },
         {
             "creatorType": "author",
             "firstName": "Jeng-Feng",
             "lastName": "Li"
         },
         {
             "creatorType": "author",
             "firstName": "Li-Yang",
             "lastName": "Chen"
         },
         {
             "creatorType": "author",
             "firstName": "Jri",
             "lastName": "Lee"
         }
     ],
     "notes": [],
     "tags": [],
     "seeAlso": [],
     "attachments": [],
-    "publicationTitle": null,
-    "place": null,
-    "conferenceName": null,
+    "publicationTitle": "2017 IEEE International Solid-State Circuits Conference (ISSCC)",
+    "place": "San Francisco, CA, USA",
+    "conferenceName": "2017 IEEE International Solid- State Circuits Conference - (ISSCC)",
     "abstractNote": null,
     "language": null,
     "ISBN": "978-1-5090-3758-2",
     "ISSN": null,
     "publisher": "IEEE",
     "edition": null,
     "volume": null,
     "date": "2/2017",
     "pages": "110-111",
     "DOI": "10.1109/ISSCC.2017.7870285",
     "url": "http://ieeexplore.ieee.org/document/7870285/",
     "title": "6.1 A 56Gb/s PAM-4/NRZ transceiver in 40nm CMOS"
 }
```

adam3smith:
great, thank you!

PR #1711, opened 2018-08-15
Update Springer Link

adam3smith:
only relies on RIS now with some supplementing. Tests look very good.

### DIFF
--- a/Springer Link.js
+++ b/Springer Link.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsbv",
-	"lastUpdated": "2018-02-13 19:20:12"
+	"lastUpdated": "2018-08-15 11:38:22"
 }
 
 function detectWeb(doc, url) {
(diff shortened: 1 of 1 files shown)


adam3smith:
@zuphilip -- let me know if you'll be able to review this

adam3smith:
https://forums.zotero.org/discussion/73189/zotero-chrome-plug-in-overriding-editors-in-book-sections-chapters#latest

Springer Link.js
```diff
@@ -89,13 +89,19 @@ function complementItem(doc, item) {
 					or @id="abstract-about-doi"][1]'
 		);
 	}
+	if (!item.language) {
+		item.language = ZU.xpathText(doc, '//meta[@name="citation_language"]/@content');
+	}	
 	if(!item.publisher) {
```
zuphilip: I prefer to add a space after the `if` as you did here (but there is quite a mismatch in that file). Can we agree that we generally want that? See also https://eslint.org/docs/rules/keyword-spacing

adam3smith: I believe the space is standard Zotero style, too, so yes, let's do that.

adam3smith:
Thanks!
Do you want to add the eslint rules to the built-in tests?

Springer Link.js
zuphilip: Great!

> Unsure whether to address that retrospectively, though.

Should be possible to replace this by simple search and replace. I will try to look into this...

zuphilip:
We are using currently JSHint. Thus, we would need to change to ESLint before we can add the rule there. I can see whether a PR is easy or maybe a first open a issue.

PR #1715, opened 2018-08-17
New translator: Mikado

shadeMe:
No usable test cases as the website is built as a single-page webapp and therefore has no unique URLs for individual pages.

Website: http://www.mikado-ac.info


shadeMe:
I've pushed a new version of the translator that's less rubbish. It ought to addresses most, if not all, of your concerns.

PR #1720, opened 2018-08-21
Sciencedirect book main pages

mbwhitfield:
Changed target to make "science/" in URL optional, to hopefully accomodate old and new sciencedirect non-article pages.

Also added code to ZU.xpath to identify chapters on book main pages. (This also fixed detectweb on book main pages, which requires getArticleList to not be empty in order to return "multiple".)

### DIFF
--- a/ScienceDirect.js
+++ b/ScienceDirect.js
@@ -474,7 +474,7 @@ function scrape(doc, url) {
 	// On newer pages, there is an GET formular which is only there if
 	// the user click on the export button, but we know how the url
 	// in the end will be built.
-	form = ZU.xpath(doc, '//div[@class="ExportCitation"]//button[contains(@class, "ExportCitationButton")]')[0];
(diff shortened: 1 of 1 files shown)


adam3smith:
thanks! could you post a url?

mbwhitfield:
Sure, this is what I was aiming at: https://www.sciencedirect.com/book/9781893997981/industrial-oil-crops

mbwhitfield:
By coincidence, I found a problem with the code to identify whether there is an RIS export button on the page (probably b/c of changes in the sciencedirect website). I made the edit to my fork, not realizing it would automatically add to the pull request (github newbie). Hope that's not a problem. 
Example URL: https://www.sciencedirect.com/science/article/pii/S0921042300800180
I simply removed the class from the button element on line 477. Since it's underneath `div[@class="ExportCitation"]`, I don't _think_ that will break anything.

adam3smith:
Thanks!

PR #1721, opened 2018-08-23
Fix if condition in Code4Lib Journal translator

bencomp:
Due to a missing `)` in a condition statement, the Code4Lib Journal translator could not be injected.


adam3smith:
Thanks! I'm surprised our tests didn't catch that.

PR #1733, opened 2018-09-04
adjust export button xpath

adam3smith:
remove/change some tests

### DIFF
--- a/ScienceDirect.js
+++ b/ScienceDirect.js
@@ -2,14 +2,14 @@
 	"translatorID": "b6d0a7a-d076-48ae-b2f0-b6de28b194e",
 	"label": "ScienceDirect",
 	"creator": "Michael Berkowitz and Aurimas Vinckevicius",
-	"target": "^https?://[^/]*science-?direct\\.com[^/]*/((science/)?(article/|(journal|bookseries|book|handbooks|referenceworks)/\\d)|search\\?|journal/[^/]+/vol)",
+	"target": "^https?://[^/]*science-?direct\\.com[^/]*/((science/)?(article/|(journal|bookseries|book|handbook)/\\d)|search\\?|journal/[^/]+/vol)",
 	"minVersion": "3.0",
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2018-09-02 01:19:42"
+	"lastUpdated": "2018-09-05 01:23:19"
 }
 
 // attr()/text() v2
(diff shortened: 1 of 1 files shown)


adam3smith:
well, still in my master branch, but oh well.

ScienceDirect.js
```diff
@@ -884,23 +884,6 @@ var testCases = [
...
-	},
-	{
-		"type": "web",
-		"url": "http://www.sciencedirect.com/science/bookseries/00652458",
-		"defer": true,
-		"items": "multiple"
-	},
 	{
 		"type": "web",
 		"url": "http://www.sciencedirect.com/science/article/pii/0584853976801316",
```
zuphilip: Switch to `https` for all tests.

ScienceDirect.js
```diff
@@ -28,7 +28,7 @@ function detectWeb(doc, url) {
 	if ((url.includes("pdf") &&
 			!url.includes("_ob=ArticleURL") &&
 			!url.includes("/article/")) ||
-		url.search(/\/(?:journal|bookseries|book|handbooks|referenceworks)\//) !== -1) {
+		url.search(/\/(?:journal|bookseries|book|handbook|referenceworks)\//) !== -1) {
```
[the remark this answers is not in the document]
zuphilip: And shouldn't you adjust the changes also in the target?

ScienceDirect.js
```diff
@@ -57,7 +57,7 @@ function detectWeb(doc, url) {
 
 function getPDFLink(doc, onDone) {
 	// No PDF access ("Get Full Text Elsewhere" or "Check for this article elsewhere")
-	if (doc.querySelector('.accessContent') || doc.querySelector('.access-options-link-text')) {
+	if (doc.querySelector('.accessContent') || doc.querySelector('.access-options-link-text') || doc.querySelector('#check-access-popover')) {
```
zuphilip: Do we have an example where this is needed? Is it already part of the test cases?

adam3smith: it's the standard PDF link behavior I now get when I'm not logged in.

ScienceDirect.js
adam3smith: the site still runs on http, too; also, a number of tests don't run correctly because of https://github.com/zotero/zotero/issues/1502, so can't really update those.

adam3smith:
I think I addressed everything. Please take another look when you get a chance

ScienceDirect.js
zuphilip: It is reachable through http, but you get redirected to https. True, about the problem with updating, thus let us postpone this.

zuphilip:
Thank you @adam3smith ! You might want to clean your master branch now after the squash.

Issue #1735, opened 2018-09-05
ACL Anthology translator missing .info URLs


nschneid:
#1707 correctly recognizes pages served under https://aclanthology.coli.uni-saarland.de/, but fails to work for https://aclanthology.info/, which is an alias.

nschneid:
Working. Thanks!

PR #1736, opened 2018-09-07
Fix ACLAnthology target regex

GuyAglionby:
Add alternate domain to the target regex per #1735


adam3smith:
Thanks for taking care of that!

Issue #1716, opened 2018-08-19
CAOD/China Asia on Demand


adam3smith:
Requested: https://forums.zotero.org/discussion/73241/omniprobe-china-asia-on-demand-author-name-import-problem#latest

Examples: http://caod.oriprobe.com/articles/53195387/Thinking_on_Spatial_Planning_Reform.htm
http://caod.oriprobe.com/articles/52762114/Preadolescents__refusal_skills_and_responses_emoti.htm

Note the consistent pattern of LASTNAME Firstname in the EM metadata.

PR #1737, opened 2018-09-08
Add China/Asia On Demand translator

GuyAglionby:
Addresses #1716.

The second test case has the university names as authors, as this is what is indicated on the website. Unfortunately I can't think of a good way of detecting these and removing them, given the range of things (especially wrt. spacing) seen in the author field (e.g. [one](http://caod.oriprobe.com/articles/50222983/Intensive_outpatient_comprehensive_behavioral_intervention_for_tics__A.htm), [two](http://caod.oriprobe.com/articles/26107512/Study_of_Perioral_Force_Exerted_on_the_Teeth_in_Preadolescent_Children.htm)).

Thanks!

### DIFF
--- a/CAOD.js
+++ b/CAOD.js
@@ -1,0 +1,250 @@
+{
+	"translatorID": "d2a9e388-5b79-403a-b4ec-e7099ca1bb7f",
+	"label": "CAOD",
+	"creator": "Guy Aglionby",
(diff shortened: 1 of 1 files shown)


GuyAglionby:
Thanks for the feedback and the heads up re. `getSearchResults` -- made both changes :)

adam3smith:
Thanks!

Issue #1727, opened 2018-08-29
mistakes in cnki.js


fredericky123:
when I save items from cnki, it will be like this:
![image](https://user-images.githubusercontent.com/13982749/44772200-ca7adf00-ab9f-11e8-9d17-691f9a10936f.png)
however, it should be **Volume:16 Issue:03**
The mistakes exists for all items saved from cnki
http://www.cnki.net/

zuphilip:
Fixed in #1740.

PR #1740, opened 2018-09-08
Fix volume and issue in CNKI

GuyAglionby:
Addresses #1727 -- problem was that the RefWorks citation exported by the site has the volume line indicated by a lowercase `vo`, causing the RefWorks translator to think it's a run-on from the issue line. The metadata order was consistent in all cases I saw, with issue always preceding volume, so I included this in the regex to avoid capitalising other instances. 

Also added a test as there were none.

Thanks!

### DIFF
--- a/CNKI.js
+++ b/CNKI.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcs",
-	"lastUpdated": "2017-09-02 11:17:43"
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks -- this certainly seems like a very low-risk fix. If it turns out there are non-numeric issues (e.g. roman numerals) we may need to relax the regex a bit, but for now let's go with this.

PR #1741, opened 2018-09-09
Add Google AI translator

GuyAglionby:
Add translator for https://ai.google/research . Supersedes #1503 as they now redirect https://research.google.com to this URL. Put URLs in the code for testing the search page and for personal pages, as they use Angular to populate data and thus `defer: true` is needed for automated testing, which doesn't work ([zotero-dev](https://groups.google.com/forum/#!topic/zotero-dev/Dy01PtaZ53E)).

Thanks :)

### DIFF
--- a/Google AI.js
+++ b/Google AI.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2018-09-09 17:55:01"
+	"lastUpdated": "2018-09-15 16:07:25"
 }
 
 /*
@@ -53,8 +53,7 @@ function detectWeb(doc, url) {
 }
 
 function doWeb(doc, url) {
-	let detectType = detectWeb(doc, url);
-	if (detectType === 'multiple') {
+	if (detectWeb(doc, url) === 'multiple') {
 		// Test cases: (defer: true is broken so cannot currently use automated tests)
 		// https://ai.google/research/people/105197
 		// https://ai.google/research/pubs/
@@ -64,7 +63,7 @@ function doWeb(doc, url) {
 			}
 			ZU.processDocuments(Object.keys(selected), scrape);
 		});
-	} else if (detectType) {
+	} else {
 		scrape(doc, url);
 	}
 }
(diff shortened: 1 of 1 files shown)


Google AI.js
```diff
@@ -0,0 +1,188 @@
...
+	"article"      : "journalArticle",
+	"techreport"   : "report",
+	"patent"       : "patent"
+};
+
+function detectWeb(doc, url) {
+	if (url.includes('/pubs/pub')) {
+		let bibtex = extractBibtex(doc);
+		let doctype = bibtex.split('{')[0].replace('@', '');
+		return bibtex2zoteroTypeMap[doctype];
```
adam3smith: I wonder if it's worth putting in a fallback here? Just in case they add an unforeseen item type -- we'd detect the wrong type, but at least we'd show a save icon and the actual import would likely be correct.

Google AI.js
```diff
@@ -0,0 +1,188 @@
...
+		// Test cases: (defer: true is broken so cannot currently use automated tests)
+		// https://ai.google/research/people/105197
+		// https://ai.google/research/pubs/
+		Zotero.selectItems(getSearchResults(doc), function (selected) {
+			if (!selected) {
+				return true;
+			}
+			ZU.processDocuments(Object.keys(selected), scrape);
+		});
+	} else if (detectType) {
```
adam3smith: You can just do else here. It's impossibel to call `doWeb` if `detectWeb` is false.

Google AI.js
```diff
@@ -0,0 +1,188 @@
...
+		}
+		delete item.itemID;
+		item.complete();
+	});
+	translator.translate();	
+}
+
+function extractBibtex(doc) { 
+	let bibtexElement = ZU.xpath(doc, '//a[span[contains(@class, "icon--copy")]]')[0];
+	let escapedBibtex = bibtexElement.getAttribute('ng-click')
```
adam3smith: any reason you're not just using `ZU.xpathText(doc, '//a[span[contains(@class, "icon--copy")]]/@ng-click')`? or the `attr` function with a CSS selector? Same above for getting the PDF

Google AI.js
GuyAglionby: Could it be worth copying in the whole list from [BibTeX.js](https://github.com/zotero/translators/blob/2d3e4bdbaa217b31eee10591e84d6afb0cd2efb2/BibTeX.js#L200)? Or is there a cleaner way of doing the mapping? I only added types that I saw, but it seems like they put a variety of content on there so adding this and a fallback seems a good idea.

GuyAglionby:
Thanks for the hints, made the changes as suggested :)

Google AI.js
adam3smith: Good point. I don't see any real cost of adding the whole list. It's pretty short. I'd still do a fallback, likely to journal article, just in case because bibtex is so poorly standardized. (and sorry for the delay, I could have written those 2 sentences faster ;)

GuyAglionby: Didn't know it was so poorly standardised -- good to know. Amended as suggested, and no problem about the delay

adam3smith:
Great, thanks!

PR #1744, opened 2018-09-11
Fix semantic scholar translator

GuyAglionby:
The semantic scholar website has changed a bit, breaking the xpaths in the translator. Summary of changes:

* Fix BibTeX xPath
* They now call "tags" "entities", so amended for this
* Updated tests accordingly (some of the other data on the website also changed, which the new tests reflect)
* Removed test which had incorrectly parsed data on their side ([link](https://www.semanticscholar.org/paper/The-German-hospital-malnutrition-study.-Pirlich-Schuetz/b59a79b2194f5f6d82b06593c23f25f67fbef512) to paper, see authors `Jürgen Weiner. Benno Stein` conflated)
* Removed fix for BibTeX export having a bracket mismatch -- this is fixed on their end now.

Thanks!

### DIFF
--- a/Semantic Scholar.js
+++ b/Semantic Scholar.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2018-03-05 00:00:22"
+	"lastUpdated": "2018-09-11 19:58:38"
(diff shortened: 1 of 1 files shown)


adam3smith:
Perfect. Too bad they're no longer putting full first names in the author data.

PR #1747, opened 2018-09-17
Update neural information processing systems

GuyAglionby:
* Amended target regex -- books.nips.cc doesn't seem to be used
* Support saving from author pages
* Support saving from search results
* Support saving from pdf 

Thanks!

### DIFF
--- a/Neural Information Processing Systems.js
+++ b/Neural Information Processing Systems.js
@@ -20,15 +20,22 @@ function detectWeb(doc, url) {
 }
 
 function scrape(doc, url) {
-	let pdfurl = url + ".pdf";
-	let bibtexurl = url+ "/bibtex";
+	//work on PDF pages
+	let baseurl = url.replace(/\.pdf$/, "");
+	let pdfurl = baseurl + ".pdf";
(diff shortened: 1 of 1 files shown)


adam3smith:
just to explain these changes: adding attachments by using the current document doesn't check for mimeType, so I think this would have imported two PDFs on the PDF page (and it definitely wouldn't have gotten the Snapshot of the item page, producing inconsistent behavior). Instead I'm distinguishing between the two cases in the attachment behavior.

adam3smith:
Thanks!

GuyAglionby:
Makes sense -- thanks for the catch :)

zepinglee:
The hostname of NeurIPS proceedings has changed to proceedings.neurips.cc . The url in the test case http://papers.nips.cc/paper/3689-information-theoretic-lower-bounds-on-the-oracle-complexity-of-convex-optimization is now redirected to https://proceedings.neurips.cc/paper/2009/hash/2387337ba1e0b0249ba90f55b2ba2521-Abstract.html . Besides the entry type should be “conference” paper rather than “book section” because “inproceedings” is used in the bibtex of each paper.

PR #1748, opened 2018-09-21
Bug fixes (especially issue with locale identification) and minor improvements

pjalet:
Locale identification failed because the URL may end with an # -> fixed with new regex on line 93.
Using secondaryInventors  and secondaryApplicants led to duplicates -> fixed by using inventors and applicants instead (l. 156, 165)
Names from inventors and applicants include country code in brackets -> fixed with modified regex on line 129.
Case correction did not work when some names where upper case and some not -> fixde by eliminating test on line 131 and going first to lower case on line 132.
Catching the application number through default case led to inclusion of parasitic text ("global dossier") -> fixed with specific case and XPath on line 190
Current patent number is directly accessible through navigation block on left of the page, so that text cleanup is not necessary -> new assignement on line 204.

### DIFF
--- a/ESpacenet.js
+++ b/ESpacenet.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2016-10-31 16:26:28"
+	"lastUpdated": "2018-09-21 14:23:22"
 }
 
 /*
@@ -37,6 +37,7 @@
 
 function detectWeb(doc, url) {
 	if (url.indexOf("searchResults?") !== -1
+	if(url.indexOf("searchResults?") !== -1
 		&& getSearchResults(doc).length) {
 			return "multiple";
 	}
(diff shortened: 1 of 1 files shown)


pjalet:
Good point regarding the tests. The changes indeed have an impact on the expected patent number; this is changed in the latest commit. However the tests consistently return an error ("TranslatorTester: ESpacenet Test 1: failed (Detection failed)") in Scaffold but on this code and the original one in zotero:master.
Change notes in the code are removed.

zuphilip:
> However the tests consistently return an error ("TranslatorTester: ESpacenet Test 1: failed (Detection failed)") in Scaffold but on this code and the original one in zotero:master.

I guess this is because the detection uses `Z.monitorDOMChanges` function which the automatic testing is IMO ignoring. You can temporarily output the expected type in `detectWeb` always, and then run the tests again. (Maybe even for the test cases save the types hard coded, i.e. if url = ... then type = ...)

ESpacenet.js
```diff
@@ -187,6 +181,9 @@ function scrape(doc) {
 			case "Page bookmark":
 				applyValue(newItem, label, value.firstElementChild.href)
 			break;
+			case "Application number:":
+				applyValue(newItem, label, ZU.xpathText(value,'./node()[following-sibling::a]'));
```
zuphilip: There is not always such a "Global Dossier" button, see e.g. https://worldwide.espacenet.com/data/publicationDetails/biblio?locale=en_EP&II=9&FT=D&CC=AU&DB=EPODOC&NR=2814389A&date=19890601&ND=3&KC=A&adjacent=true

adam3smith:
Thanks and sorry for the delay; I missed the review request.

PR #1753, opened 2018-09-27
Update Engineering Village

adam3smith:
after some site changes. Thankfully small changes -- the post request happens in a window,that only pops up briefly & then closes. This wouldn't have been fun to re-create...

### DIFF
--- a/Engineering Village.js
+++ b/Engineering Village.js
@@ -8,22 +8,23 @@
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
-	"browserSupport": "gcsb",
-	"lastUpdated": "2018-09-27 01:43:46"
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2018-10-07 10:02:04"
 }
 
 function detectWeb(doc, url) {
 	Z.monitorDOMChanges(doc.getElementById("ev-application"), {childList: true});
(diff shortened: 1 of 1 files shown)


Engineering Village.js
```diff
@@ -8,51 +8,51 @@
...
 		if (!title) continue;
 		
 		if (checkOnly) return true;
 		found = true;
 		
 		items[docid] = {
 			title: ZU.trimInternal(title.textContent),
-			checked: checkbox.checked
-		}
+			checked: checkbox.checked // TODO is this still needed?
```
zuphilip: It looks for me that this is  not anymore needed and therefore we can also switch to our default template for the function `getSearchResults`. Can you try that out @adam3smith ?

Engineering Village.js
```diff
@@ -8,51 +8,51 @@
...
 	var m = url.match(/\bdocidlist=([^&#]+)/);
 	if (!m) return false;
 	
 	return decodeURIComponent(m[1]).split(',');
 }
 
+
 function getSearchResults(doc, checkOnly) {
-	var rows = doc.getElementsByClassName('result'),
+	var rows = doc.querySelectorAll('div[class*=result-row]'),
```
zuphilip: FYI A comma (not semi-colon) because several variables are defined here.

Issue #1719, opened 2018-08-20
elife translator issues


tmchartrand:
The site layout for eLife seems to have changed at some point, and the regex no longer matches.
For an example url, see https://elifesciences.org/articles/16553
Simply adding "|articles" to the regex fixed the issue for me.
Full target regex:
`^https?://(elife\.)?elifesciences\.org/(content/|search|category/|browse|archive|articles)`

Additionally, there's some code that apparently used to clean up extra authors, but now removes all:
```
//clean out the authors if they're not in citation_author, so we don't import junk from elsewhere
// as on http://elifesciences.org/content/2/E00565V1
if (!ZU.xpathText(doc, '//meta[@name="citation_author"]/@content')) {
   item.creators = [];
}
```
I'm unable to find an example of the old scheme where this was necessary, but commenting it out fixes the import for me.

Finally, the pdf attachments are no longer found. This currently relies on the embedded metadata translator, and it seems some custom code is needed for the new layout.

PR #1754, opened 2018-09-27
eLife rewrite after site redesign

adam3smith:
Closes https://github.com/zotero/translators/issues/1719


adam3smith:
The PMID is nowhere to be found anymore, so we can't include that.

PR #1755, opened 2018-09-28
Update Google Patents.js

hpxing:
fix pdf url


adam3smith:
do you have a link where that's needed? I don't have an example for the relative link, but typically they'd include a beginning slash?

hpxing:
just the test case   https://patents.google.com/?q=book&oq=book
without the / the pdf will not attached correctly

adam3smith:
perfect, thanks, that's what I needed

PR #1757, opened 2018-10-04
Update translator for Open Journal Systems

jrwiebe:
This update brings compatibility with recent versions of Open Journal Systems. 

Two small changes were made to the actual OJS translator:
* Only one `<script>` link containing `/lib/pkp/js` is required for detectWeb(). One is all I had in my live example. (Previously it expected >= 10.)
* The new code looks for an `<a>` link to the PDF view if a PDF attachment is not linked in the header.

Most of the changes are located in RDF.js, adding Dublin Core elements and values that are used by OJS. Some of these might be used by other systems, but at least one -- using "Text.Serial.Journal" as the default value for DC.Type -- appears unique to OJS. DC.Creator.PersonalName is also not commonly used elsewhere. The need for these changes seems to date to OJS moving metadata to plugins two years ago (pkp/pkp-lib#1815). It was probably not noticed by many because most OJS installations have citation_* metadata.

### DIFF
--- a/Open Journal Systems.js
+++ b/Open Journal Systems.js
@@ -74,7 +74,6 @@ function doWeb(doc, url) {
 		var pdfUrl = doc.querySelector("a.obj_galley_link.pdf");
 		//add linked PDF if there isn't one listed in the header
 		if (!pdfAttachment && pdfUrl) { 
-			Z.debug("hereherehere");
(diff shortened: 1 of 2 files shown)


adam3smith:
Thanks, looks good to me!

PR #1760, opened 2018-10-08
Update Old Bailey Online.js

sharonhoward:
Fixed function

### DIFF
--- a/Old Bailey Online.js
+++ b/Old Bailey Online.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2018-10-08 14:06:13"
+	"lastUpdated": "2018-10-08 14:15:17"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


Old Bailey Online.js
```diff
@@ -296,6 +297,40 @@ var testCases = [
...
 			}
 		]
+	},
+	{
+		"type": "web",
+		"url": "https://www.oldbaileyonline.org/browse.jsp?id=t16780828-12&div=t16780828-12&terms=hog#highlight",
+		"items": [
+			{
+				"itemType": "case",
+				"caseName": ".",
```
adam3smith: That doesn't seem right/helpful as data import. I think what I would do is strip all trailing periods and commas, and then set `newItem.title = [no title]` when there's nothing left -- does that sound right?

adam3smith:
@sharonhoward -- could you take a look what you think about my changes? Looks good to me, but I don't use the database.

zuphilip:
Thank you both for the work here!

PR #1765, opened 2018-10-11
Update newspapers.com.js to handle new markup

pbinkley:
The html on newspapers.com has changed slightly, so that the translator was no longer capturing the date, publication, edition or page.


adam3smith:
Thanks!

Issue #1763, opened 2018-10-11
PDF download missing for Nature Scientific Data


quazgar:
When trying to save this (pretty famous) OpenAccess article on Nature Scientific Data, only a snapshot for the web page is saved (as item type `web page`), not the PDF version of the article: https://www.nature.com/articles/sdata201618

Downloading articles from other sites, e.g. Phys. Rev. E, works for me.

This is Zotero Connector 5.0.41 (Firefox) with Standalone 5.0.55 on Kubuntu 18.04 LTS.

quazgar:
Or is this expected behaviour as per #1599?
@adomasven @zuphilip

quazgar:
Screenshot to document the missing *Save* options:

![screenshot_20181011_134439](https://user-images.githubusercontent.com/302587/46802036-7f3a1b80-cd5c-11e8-8f9d-d5bff70e0761.png)

adam3smith:
No, that's not right, thanks for reporting

zuphilip:
Fixed in #1766 by @adam3smith.

PR #1766, opened 2018-10-12
Fix recognition of articles and getting PDFs in Nature

adam3smith:
(this was reported on the Zotero forums, but can't find it (?)

### DIFF
--- a/Nature Publishing Group.js
+++ b/Nature Publishing Group.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2018-03-26 20:45:23"
+	"lastUpdated": "2018-10-12 15:27:55"
 }
 
 /**
@@ -269,9 +269,12 @@ function getKeywords(doc) {
 }
 
 //get PDF url
-function getPdfUrl(url) {
+function getPdfUrl(doc, url) {
 	var m = url.match(/(^[^#?]+\/)(?:full|abs)(\/[^#?]+?\.)[a-zA-Z]+(?=$|\?|#)/);
 	if (m && m.length) return m[1] + 'pdf' + m[2] + 'pdf';
+	else if (attr(doc, 'a[data-track-action="download pdf"]', 'href')) {
+		return attr(doc, 'a[data-track-action="download pdf"]', 'href');
+	}
 }
 
 //add using embedded metadata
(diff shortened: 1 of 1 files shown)


Nature Publishing Group.js
```diff
@@ -651,14 +658,20 @@ function scrape(doc, url) {
 			item.publicationTitle = 'Nature';// old articles mess this up
 		}
 		delete item.journalAbbreviation;
-		
-		if (!item.attachments) {
+		var hasPDF = false;
```
adam3smith: this is potentially the most problematic bit

zuphilip:
I guess that it fixes #1763 and maybe this is the report you remembered?

Nature Publishing Group.js
```diff
@@ -269,9 +269,12 @@ function getKeywords(doc) {
 }
 
 //get PDF url
-function getPdfUrl(url) {
+function getPdfUrl(doc, url) {
 	var m = url.match(/(^[^#?]+\/)(?:full|abs)(\/[^#?]+?\.)[a-zA-Z]+(?=$|\?|#)/);
 	if (m && m.length) return m[1] + 'pdf' + m[2] + 'pdf';
+	else if (attr(doc, 'a[data-track-action="download pdf"]', 'href')) {
+		return attr(doc, 'a[data-track-action="download pdf"]', 'href');
```
zuphilip: Is the check before here needed? `attr` should return `null` if the element is not found, which should be okay here.

Nature Publishing Group.js
```diff
@@ -354,7 +361,7 @@ function scrapeRIS(doc, url, next) {
 	if (!risURL) risURL = doc.evaluate('//li[@class="download-citation"]/a', doc, null, XPathResult.ANY_TYPE, null).iterateNext();
 	if (!risURL) risURL = doc.evaluate('//a[normalize-space(text())="Export citation" and not(@href="#")]', doc, null, XPathResult.ANY_TYPE, null).iterateNext();
 	if (!risURL) risURL = ZU.xpath(doc, '//ul[@data-component="article-info-list"]//a[@data-track-source="citation-download"]')[0];
-
+	if (!risURL) risURL = doc.querySelectorAll('a[data-track-action="download article citation"]')[0];
```
zuphilip: For the first element you can also use `.querySelector` cf. https://developer.mozilla.org/de/docs/Web/API/Document/querySelector

adam3smith:
OK, all done

zuphilip:
Thank you very much!

PR #1767, opened 2018-10-12
Open Journal Systems: Get metadata for articles in galley view (need for OJS 3)

jrwiebe:
In Open Journal Systems an article will have a landing page, which might link to galleys of the article in one or more formats (e.g., PDF, HTML). In OJS 2 the galley-view page included full metadata about the article; in OJS 3 it does not. I believe this is an inadvertent omission, although I haven't delved into the OJS version history to verify this. 

I updated the OJS translator to fetch the metadata from the article landing page. It does this regardless of the OJS version, since a) I didn't initially realize that this was specifically an OJS 3 issue, and b) it makes for simpler code. (One could load the 'web' translator in doWeb(), translate, and test for an element like item.volume to see if fetching the landing page is necessary.) 

I have also opened an issue about this with OJS (https://github.com/pkp/pkp-lib/issues/4141). If it gets addressed on that end, and this pull request is accepted, we could be in a position where the OJS translator is doing extra work to accommodate for a small percentage of OJS installs. (It seems like the majority "in the wild" are still on OJS 2.)

### DIFF
--- a/Open Journal Systems.js
+++ b/Open Journal Systems.js
@@ -23,7 +23,7 @@ function detectWeb(doc, url) {
 function doWeb(doc, url) {
 	// In OJS 3, up to at least version 3.1.1-2, the PDF view does not
 	// include metadata, so we must get it from the article landing page.
-	var urlParts = doc.location.href.match(/(.+\/article\/view\/)([^/]+)\/[^/]+/);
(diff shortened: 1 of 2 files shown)


jrwiebe:
Will do.

PR #1775, opened 2018-11-01
Translator for Deutsches Ärtzeblatt

sebastian-de:
Tested with quite a few different articles, all work fine. So far this translator is only for single articles. Maybe I'll add support for multiple articles later (for example from search results).


sebastian-de:
ignore this PR, adding new one without typo

adam3smith:
for future reference, makes our life easier if you just update existing pull requests -- you can fix typos etc. easily.

PR #1776, opened 2018-11-01
Add translator for Deutsches Ärzteblatt

sebastian-de:
Tested with quite a few different articles, all work fine. So far this translator is only for single articles. Maybe I'll add support for multiple articles later (for example from search results).

### DIFF
--- a/Deutsches Arzteblatt.js
+++ b/Deutsches Arzteblatt.js
@@ -1,0 +1,286 @@
+{
+	"translatorID": "72543881-a05b-4a40-b891-de8726e29d77",
+	"label": "Deutsches Ärzteblatt",
+	"creator": "Sebastian Fleer",
+	"target": "^https?://www\\.aerzteblatt\\.de/archiv/",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2018-11-01 15:42:33"
+}
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2018 Sebastian Fleer
+
+	This file is part of Zotero.
+
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+function detectWeb(doc, url) {
+	if (url.match(/archiv\/\d+/) || url.match(/aid\=\d+/)) {
+		return "journalArticle";
+	}
+}
+
+function doWeb(doc, url) {
+	var type = detectWeb(doc, url);
+	if (type == "journalArticle") {
+		scrape(doc, url);
+	}
+}
+
+function scrape(doc, url) {
+	var newItem = new Zotero.Item("journalArticle");
+
+	// detect article ID
+	var tmp = "null";
+	if (url.match(/archiv\/\d+/)) { // normal article page
+		tmp = url.match(/archiv\/\d+/);
+	} else if (url.match(/aid\=\d+/)) { // article page in search results
+			tmp = url.match(/aid\=\d+/);
+	}
+	aid = String(tmp).match(/\d+/);
+	//Z.debug("aid: " + aid);
+
+	// use BibTeX Translator
+	var translator = Zotero.loadTranslator("import");
+	translator.setTranslator("9cb70025-a888-4a29-a210-93ec52da40d4");
+
+	translator.setHandler("itemDone", function(obj, item) {
+		item.attachments.push({
+			title:"Snapshot",
+			document:doc
+		});
+
+		// PDFs are stored under https://www.aerzteblatt.de/pdf/<volume>/<issue>/pdfPageNum.pdf
+		// pdfPageNum is in the form a<firstpage> or m<firstpage>, depending on the journal part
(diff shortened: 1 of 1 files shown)


sebastian-de:
I'd like to clean up the process of generating the URL to the PDF file. Since the URL can't be found on the article page itself, it can be generated the way I implemented, or by fetching a "PDF viewer" page. The only way I I'm aware of doing this in a translator is using Zotero.Utilities.HTTP.doGet(). The following code produces the correct URL, but how can I access it outside of the callback function?
```
var pdfPageURL = "https://www.aerzteblatt.de/pdf.asp?id=202039";
Zotero.Utilities.HTTP.doGet(pdfPageURL, function(html) {
	var parser = new DOMParser();
	var pdfDoc = parser.parseFromString(html, "text/html");
	var pdfFileURL = ZU.xpathText(pdfDoc, '//div[contains(@class, "save")]/a/@href');
	Z.debug("pdfFileURL :" + pdfFileURL);
});
```

zuphilip:
> The following code produces the correct URL, but how can I access it outside of the callback function?

That might not be possible. If I understand you correctly, then you rather need to continue in the callback function. Maybe pack these steps into a function `finalize()`, something like
```js
var pdfPageURL = "https://www.aerzteblatt.de/pdf.asp?id=202039";
if (pdfPageURL) {
	Zotero.Utilities.HTTP.doGet(pdfPageURL, function(html) {
		var parser = new DOMParser();
		var pdfDoc = parser.parseFromString(html, "text/html");
		var pdfFileURL = ZU.xpathText(pdfDoc, '//div[contains(@class, "save")]/a/@href');
		finalize();
	});
} else {
	finalize();
}
```

A similar example can be found here https://github.com/zotero/translators/blob/master/Cell%20Press.js . Does this helps?

zuphilip:
Let us know when this is ready for review.

Deutsches Arzteblatt.js
```diff
@@ -0,0 +1,255 @@
...
+		}
+		var articleID = String(tmp).match(/\d+/);
+
+		// use BibTeX Translator
+		var translator = Zotero.loadTranslator("import");
+		translator.setTranslator("9cb70025-a888-4a29-a210-93ec52da40d4");
+	
+		translator.setHandler("itemDone", function(obj, item) {
+			item.attachments.push({
+				title:"Snapshot",
```
AbeJellinek, line 70: Rebase on master and lint to fix formatting errors

Deutsches Arzteblatt.js
```diff
@@ -0,0 +1,255 @@
...
+
+function detectWeb(doc, url) {
+	if (url.match(/archiv\/\d+/) || url.match(/aid\=\d+/)) {
+		return "journalArticle";
+	}
+}
+
+function doWeb(doc, url) {
+	var type = detectWeb(doc, url);
+	if (type == "journalArticle") {
```
AbeJellinek, line 46: \The connector calls `detectWeb` and then `doWeb`, so this translator is running `detectWeb` three times: once in the connector, once in `doWeb`, and once in `scrape`. All calls but the first are redundant: `doWeb` will never be called if `detectWeb` doesn't return a type, and the only type it can return is `journalArticle`.

Deutsches Arzteblatt.js
```diff
@@ -0,0 +1,255 @@
...
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+function detectWeb(doc, url) {
+	if (url.match(/archiv\/\d+/) || url.match(/aid\=\d+/)) {
+		return "journalArticle";
+	}
```
AbeJellinek, line 41: Add `return false` when the URL doesn't match.

Deutsches Arzteblatt.js
```diff
@@ -0,0 +1,255 @@
...
+						"firstName": "Christian",
+						"lastName": "Thielscher",
+						"creatorType": "author"
+					}
+				],
+				"date": "2018",
+				"issue": "43",
+				"itemID": "DAE202039",
+				"libraryCatalog": "Deutsches Ärzteblatt",
+				"pages": "A-1946-",
```
AbeJellinek, line 109: Can you fix up the pages? This feels wrong.

Deutsches Arzteblatt.js
```diff
@@ -0,0 +1,255 @@
...
+						"firstName": "Kathrin",
+						"lastName": "Gießelmann",
+						"creatorType": "author"
+					}
+				],
+				"date": "2018",
+				"issue": "42",
+				"itemID": "DAE201948",
+				"libraryCatalog": "Deutsches Ärzteblatt",
+				"pages": "A-1840-A-1840",
```
AbeJellinek, line 238: As does this.

Deutsches Arzteblatt.js
```diff
@@ -0,0 +1,255 @@
...
+				"publicationTitle": "Dtsch Arztebl International",
+				"shortTitle": "Bluttests",
+				"url": "http://www.aerzteblatt.de/int/article.asp?id=201948",
```
AbeJellinek, line 241: URLs may have changed; update tests just in case.

Deutsches Arzteblatt.js
```diff
@@ -0,0 +1,255 @@
...
+						"lastName": "Thielscher",
+						"creatorType": "author"
+					}
+				],
+				"date": "2018",
+				"issue": "43",
+				"itemID": "DAE202039",
+				"libraryCatalog": "Deutsches Ärzteblatt",
+				"pages": "A-1946-",
+				"publicationTitle": "Dtsch Arztebl International",
```
AbeJellinek, line 110: Maybe this belongs in `journalAbbreviation` (although I believe `International` should be `int.`), but not `publicationTitle`, where it should be spelled out completely.

AbeJellinek (changes requested):
Thanks for contributing this, and sorry it took me a while to get to! Working through a big backlog, but it's coming along. I've requested a few changes. Could you take a look?

AbeJellinek:
PDF downloads should be pretty easy to get working once the rest is done. `processDocuments` to grab the page where the PDF is embedded, then just save its URL as an attachment.

sebastian-de:
@AbeJellinek thank you very much for taking your time and doing a review.
Currently I'm not using Zotero unfortunately. If I find the time I'll setup a dev-environment in the coming days and have a look at this PR again.

AbeJellinek:
Thanks, that would be great!

PR #1787, opened 2018-11-26
Fix ResearchGate translator

GuyAglionby:
Some changes to the ResearchGate HTML broke the detection code, so I've quickly fixed that. Some other minor changes are:

* The way articles are listed on author pages and search results are not (any longer) the same; added a new xpath for this
* Updated URL for downloading the abstract with the citation
* Added support for downloading from authors who haven't registered for a RG profile (and a test for this)

Thanks!


adam3smith:
Looks great, thanks!

Issue #1734, opened 2018-09-05
ACL Anthology translator enhancement


nschneid:
@GuyAglionby, thanks for your work on #1707—the new BibTeX-based translator is working great!

I'm wondering if it would be possible to add an enhancement, namely populating the Conference Name field with the abbreviated name of the conference—"ACL 2018", "COLING 2018", "NAACL-HLT 2018", etc.—for main conference papers only (excluding regular and student research workshops, demos, and tutorials). Currently I fill this field in manually so my medium-form and short-form [custom exporters](https://gist.github.com/nschneid/2257875) will display that instead of the spelled-out name.

I assume this would require the code to contain a list of the standard conference abbreviations (as seen on the Anthology [homepage](https://aclanthology.coli.uni-saarland.de/)), and correctly discern hyphenated conferences like "ACL-IJCNLP" from the proceedings title. As the proceedings titles are formulaic, I think this could be accomplished with a few regexes.

GuyAglionby:
This seems like a good addition to me. Helpfully, each event has a unique letter code as seen in the URLs and PDF titles for each paper (e.g. ACL is P, NAACL is N, COLING is C) which have been stable over the years and so could be hardcoded in. Otherwise as you say we could regex on BibTeX `booktitle`, but this seems a little more complicated. Keen for feedback before I go ahead and implement this. Thanks!

nschneid:
The letter code may help, but note that tutorials, demos, and student research workshop papers are typically listed with the same letter as the main conference papers. The letter doesn't indicate which years are plain ACL and which are ACL-IJCNLP (for example), and *SEM and SemEval fall under the same letter but are cited differently. So I think some processing of the booktitle is necessary.

GuyAglionby:
[...]

PR #1797, opened 2018-12-14
Add conference name for ACLAnthology translator

GuyAglionby:
Addresses #1734 . This adds an abbreviated conference name to ACL imports, except those that:

* Have `WS` as a venue, indicating that the paper is from a workshop
* Have `Student`, `Demonstration`, or `Tutorial` in the conference title, as these are not part of the main conference.

@nschneid can you please confirm if this seems reasonable? Thanks!

### DIFF
--- a/ACLAnthology.js
+++ b/ACLAnthology.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2018-12-14 15:12:34"
+	"lastUpdated": "2018-12-14 15:23:38"
 }
 
 /*
@@ -100,7 +100,7 @@ function getSearchResults(doc) {
 		let title = ZU.xpath(listing[i], 'strong')[0].textContent;
 		let bibtexLink = ZU.xpath(listing[i], 'a[img[@alt = "Export"]]/@href')[0].value;
 		let id = bibtexLink.split('/')[2];
-		items[id] = title;
+		items[id] = title.trim().replace(/\n\s*/g, ' ');
 	}
 	return items;
 }
--- a/ACLAnthology.js
+++ b/ACLAnthology.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2018-09-07 18:10:44"
+	"lastUpdated": "2018-12-14 15:12:34"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


nschneid:
Awesome! The only other thing I would suggest is special-casing *SEM/SemEval, which right now are listed as [*SEMEVAL](https://aclanthology.coli.uni-saarland.de/venues/semeval)—note the variation in titles over the years.

Suggested heuristic based on the proceedings title:

1. If it contains "SENSEVAL", return "SENSEVAL".
2. Else if it contains "Evaluation" and not "Joint", return "SemEval".
3. Else if it contains "Joint" and not "Evaluation", return "*SEM".
4. Else if it contains "Volume 1" and "Volume 2", return "*SEM/SemEval". (special case for 2012, where the numeric prefixes on papers do not distinguish the two volumes)
5. Else if it contains "SemEval", return "SemEval".
6. Else return "*SEM".

nschneid:
The *SEMEVAL heuristics should be applied before filtering out WS venues, as SENSEVAL in 2004 is also listed as WS.

GuyAglionby:
Thanks for supplying those heuristics @nschneid . The solution now looks something like:

* If the proceedings name includes 'Student', 'Demonstration', or 'Tutorial', assume there is no conference name.
* If the proceedings name contains '*SEMEVAL', apply the heuristics you provide to get the name.
* Otherwise, scrape the name from the 'Venues' table row on the website, unless this contains 'WS', in which case do not return anything.

nschneid:
Looks great, thanks!

nschneid:
@adam3smith is this ready to merge?

Evpok:
If I understand correctly, this will use the initialized name (ie ACL 2019 instead of xth Annual Meeting of the Association for Computational Linguistics). Is it the usual way to do this in Zotero? E.g. in BibLaTeX, the full name is the event name and the abbreviation is an event title addon (https://github.com/plk/biblatex/issues/122).

mjpost:
Also, isn't this referencing the wrong version of the Anthology? I see references to the Saarland server in the diff.

nschneid:
Zotero has two fields for conference papers, Proceedings Title and Conference Name. This will put the abbreviation in Conference Name. That way the user can choose whether the exported bibliography/BibTeX has the short name or the full proceedings title.

Evpok:
@nschneid I understand, that is certainly nitpicking on my part, but I thought that the two fields existed in case the proceedings had a title other than simply “Proceeding of <conference name>”

nschneid:
Well, for the original intent you'd have to ask those who decided on the fields, but Proceedings Title is already populated based on booktitle, so Conference Name is a convenient place to store the abbreviation.

GuyAglionby:
> Also, isn't this referencing the wrong version of the Anthology? I see references to the Saarland server in the diff.

Yes, this targets the old version. This change(/translator) will be superseded once aclanthology.info redirects to aclweb.org/anthology, the translator for which will have to be re-written anyway but can include this facility.

GuyAglionby:
https://aclanthology.info has been shut down; #1911 supersedes this

PR #1799, opened 2018-12-14
Extract DOI from the current web page URL

mrtcode:
From some URLs DOI can't be correctly extracted, i.e. `http://onlinelibrary.wiley.com/journal/10.1111/%28ISSN%291470-9856/issues` would result to `10.1111/%28ISSN%291470-9856/issues` instead of `10.1111/%28ISSN%291470-9856`.

Some URLs can also have multiple DOIs i.e. `http://api.crossref.org/works/?filter=doi:10.1117/3.1002595.ch10,doi:10.3403/00522251u,doi:10.3403/00522251,doi:10.3403/30217493,doi:10.3403/30289582,doi:10.1117/12.939903,doi:10.3403/02454346u,doi:10.1364/ofc.1979.thf1,doi:10.5772/7558,doi:10.3758/BF03202760,doi:10.3758/bf03195760,doi:10.1006/jmla.1997.2532,doi:10.1037/h0082866`.


dstillman:
Shouldn't we return `document` instead of `multiple` when there's a single DOI? Even if it's invalid or not found, there's still no reason to show the Select Items dialog. (That would only make sense if it could actually result in a different item from the one you were expecting.)

dstillman:
No. In the page we don't know if it describes the main item for the page. In the URL we do.

mrtcode:
If a random string, which looks like a DOI, is extracted from a URL, it would prevent the further DOIs from document extraction. Which means if we want to return a single item we have to firstly resolve DOI metadata to make sure it's valid. And if not then extract and resolve items from a document. The current commit just puts all DOIs into one list, where all items are resolved and user just needs to select which one he thinks is correct. To be able to separately resolve DOI extracted from URL and from document, we would probably need serious modifications of the translator.

DOI.js
```diff
@@ -20,7 +20,17 @@ var selectArray = {};
 
 
 // builds a list of DOIs
-function getDOIs(doc) {
+function getDOIs(doc, url) {
+	var dois = [], m;
+	
+	// Extract DOIs from the current URL
+	var rx = /10.[0-9]{4,}?\/[^\s&"'?#,]*[^\s&"'?#\/.,]/g;
```
dstillman: We usually use `re` for this, not `rx`.

Also, seems like we don't need to exclude quotes in this context.

And maybe we do need to URL decode? I'm not sure whether a URL that passes through here is necessarily decoded.

DOI.js
```diff
@@ -20,7 +20,17 @@ var selectArray = {};
...
+	while (m = rx.exec(url)) {
+		if (dois.indexOf(m[0]) === -1) {
```
dstillman: `includes` instead of `indexOf`

mrtcode:
So as I said previously, it's not that rare to encounter web page URLs that contain a DOI, but we can't extract it reliably. For example:

URL: http://iopscience.iop.org/article/10.1088/0004-637X/768/1/87
Extracted DOI: 10.1088/0004-637X/768/1/87
Actual DOI: 10.1088/0004-637X/768/1/87

URL: http://journal.frontiersin.org/article/10.3389/fmicb.2014.00402/full
Extracted DOI: 10.3389/fmicb.2014.00402/full
Actual DOI: 10.3389/fmicb.2014.00402

And we can't do anything about this.

DOI from URL can be extracted incorrectly but it applies not only for the web page URL, but also for URLs found in the body.

For DOI(s) extracted from a web page URL there are a few possible outcomes:
1. It's correct and results to correct metadata
2. It's incorrect and results to no metadata
3. It's correct but results to no metadata (DOI RAs don't have it, i.e. JSTOR)
4. It's incorrect and results to incorrect metadata - is not possible, I would say. Except maybe in same rare cases it can result to the actual journal instead of the article

So the translator should work like this, depending on where and how many DOIs where found:
- one in URL - single
- one in URL, one in body and they are equal - single
- one in body - multiple
- in all other cases - multiple

dstillman:
> URL: http://iopscience.iop.org/article/10.1088/0004-637X/768/1/87
> Extracted DOI: 10.1088/0004-637X/768/1/87
> Actual DOI: 10.1088/0004-637X/768/1/87

Is this what you meant? These are the same.

> URL: http://journal.frontiersin.org/article/10.3389/fmicb.2014.00402/full
> Extracted DOI: 10.3389/fmicb.2014.00402/full
> Actual DOI: 10.3389/fmicb.2014.00402
> 
> And we can't do anything about this.

We could try stripping some likely suffixes, like `/full` and `/pdf`.

adam3smith:
> We could try stripping some likely suffixes, like /full and /pdf

given the most common academic CMS's, `/full$` `/pdf$`  
(as per the above examples)
`/abstract$` and `/abs$`  (speculating about these -- I've mainly seen them before the DOI where they're no problem)

`#.*`
https://link.springer.com/article/10.1007%2Fs11104-018-3787-2#enumeration

`\?.+`
https://www.tandfonline.com/doi/ref/10.1080/10301763.1988.10669055?scroll=top
 should cover the vast majority of cases

zuphilip:
> one in URL, one in body and they are equal - single

I think you can here also test wether the one in the body is the beginning part of the one in the URL. This would be IMO more stable and do (at least for all the examples here) the same as deleting `/full` or similar postfixes. If a publisher is making the URLs according to the DOIs then I expect also that he is mentioning this DOI in the body text. Therefore I would suggest to make this instead and not additional to deleting anything at the end.

> in all other cases - multiple

How about the case: There are multiple DOIs in the body, one in the URL which matches the first one in the body. Should we then just go for this one, i.e. a single case, e.g. https://olh.openlibhums.org/article/10.16995/olh.46/ (there is another DOI in the reference section at the bottom). The same would be true for the Frontiers paper. There is the drawback that then it is not possible to save all references with DOI instead of the article from pages like https://www.frontiersin.org/articles/10.3389/fmicb.2014.00402/full#h12 . But is this really a "feature" we need? (Maybe a hidden preference to toggle it on/off would be enough.) CC @adam3smith

DOI.js
```diff
@@ -191,6 +199,11 @@ var testCases = [
 		"type": "web",
 		"url": "https://en.wikipedia.org/wiki/Template_talk:Doi",
 		"items": "multiple"
+	},
+	{
+		"type": "web",
+		"url": "https://zotero.org/?d=10.7208/chicago/9780226924632.001.0001",
+		"items": "multiple"
 	}
```
zuphilip: Can you also add https://onlinelibrary.wiley.com/doi/abs/10.1002/(SICI)1099-1360(199711)6:6%3C320::AID-MCDA164%3E3.0.CO;2-2 as a test case?

PR #1803, opened 2018-12-22
Add Le Maitron

owcz:
This is a major French set of encyclopedias with open online access. No metadata => all manual scrape.

### DIFF
--- a/Le Maitron.js
+++ b/Le Maitron.js
@@ -1,0 +1,281 @@
+{
+	"translatorID": "ee016722-5f02-4362-8ffe-c96e06872b3e",
+	"label": "Le Maitron",
+	"creator": "czar",
+	"target": "^https?://maitron-en-ligne\\.univ-paris1\\.fr",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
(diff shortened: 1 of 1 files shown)


owcz:
Also noticed that `ZU.strToISO` struggles with certain French dates. It only retains the year from `14 avril 2014` and `16 mai 2014`

`30 décembre 2010` doesn't work without removing the diacritic with `.normalize('NFD').replace(/[\u0300-\u036f]/g, "")`

Not a big deal for this translator in particular, but worth flagging in general

dstillman:
> Also noticed that ZU.strToISO struggles with certain French dates. It only retains the year from 14 avril 2014 and 16 mai 2014

That's a little tricky. Zotero handles all of those dates, but only if the current locale is French.

The options here are basically 1) supporting an optional locale that could be passed to the `strTo*` functions or 2) trying to parse dates from all locales, regardless of the current locale.

retorquere:
If you know the dates on the website are always french, you could use string replace to change the french months into their english equivalents before handing it to strToISO.

AbeJellinek:
Thanks, this looks good. I rebased, linted, added a simple date translator, and updated the tests. Merging.

owcz:
The translator is failing on a few articles. Noting here for me to review later:
* https://maitron.fr/spip.php?article153818

PR #1808, opened 2019-01-15
Update the way to retrieve date to match the new format

sebastian-berlin-wmse:
Fixed some tests that was outdated. Fixed multiple matches; search page had changed.


adam3smith:
Thanks -- I added a test for `dateParts` so that this won't break if the publication data metadata changes.

PR #1810, opened 2019-01-21
TEI: Use Better BibTeX citation key if available

adunning:
This will use the citation key generated by Better BibTeX for Zotero as the `xml:id` if it is available. If that add-on is not enabled, this change has no effect. Solution from @retorquere in <https://github.com/retorquere/zotero-better-bibtex/issues/1124>.

### DIFF
--- a/TEI.js
+++ b/TEI.js
@@ -19,7 +19,7 @@
 		"Full TEI Document": false,
 		"Export Collections": false
 	},
-	"lastUpdated": "2017-11-18 10:45:00"
+	"lastUpdated": "2019-01-22 09:47:00"
 }
 
 // ********************************************************************
@@ -182,7 +182,7 @@ function generateItem(item, teiDoc) {
 
 	if (Zotero.getOption("Generate XML IDs")) {
 		if (!generatedItems[item.uri]) {
-			var xmlid = item.citekey || genXMLId(item); // use Better BibTeX for Zotero citation key if available
+			var xmlid = item.citationKey || genXMLId(item); // use Better BibTeX for Zotero citation key if available
 			bibl.setAttributeNS(ns.xml, "xml:id", xmlid);
 			exportedXMLIds[xmlid] = bibl;
 		} else {
--- a/TEI.js
+++ b/TEI.js
@@ -182,7 +182,7 @@ function generateItem(item, teiDoc) {
 
 	if (Zotero.getOption("Generate XML IDs")) {
 		if (!generatedItems[item.uri]) {
-			var xmlid = genXMLId(item);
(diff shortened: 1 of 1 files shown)


TEI.js
```diff
@@ -182,7 +182,7 @@ function generateItem(item, teiDoc) {
 
 	if (Zotero.getOption("Generate XML IDs")) {
 		if (!generatedItems[item.uri]) {
-			var xmlid = genXMLId(item);
+			var xmlid = item.citekey || genXMLId(item); // use Better BibTeX for Zotero citation key if available
```
adam3smith: @dstillman what's your view on this?

retorquere: I fully understand Zotero wouldn't want to do stuff that specifically targets BBT. Are there prospective thoughts on the name for the citation key field when it comes to Zotero natively? I wouldn't mind renaming my field to that ahead of time.

adunning: I should perhaps explain that I'm proposing this because there is a group of us using the TEI translator for a few projects, and we need something more reliable and configurable than the `xml:id` generation in the TEI translator. This is the best approach to improving it that I can think of without breaking existing workflows.

dstillman:
> Are there prospective thoughts on the name for the citation key field when it comes to Zotero natively?

`citationKey`, I would imagine?

I'm going to try to get to the new fields soon, so I'm OK with looking for `citationKey` now.

(Also, this is equivalent to `citation-label` in CSL, right?)

retorquere:
citationKey works for me, but I'd likely have to write migrations for it to work and I'd rather not stack migration on migration. But if we can pin down (so to speak) `citationKey`, I'd be happy to start doing that.

Looking at the [CSL docs](http://docs.citationstyles.org/en/stable/specification.html), `citation-label` is the "label identifying the item in in-text citations of label styles" -- that sounds to me like something that's exposed to the reader.

dstillman:
> Looking at the CSL docs, citation-label is the "label identifying the item in in-text citations of label styles" -- that sounds to me like something that's exposed to the reader

Ah, OK, right. Disregard.

So unless anyone has a reason why the citekey field wouldn't be called "Citation Key" in the interface, we can go with `citationKey` for the field name.

adunning:
I've updated it to use `citationKey` (works with the test build of Better BibTeX at <https://github.com/retorquere/zotero-better-bibtex/issues/1124#issuecomment-456224353>).

adam3smith:
Thanks, I'll take this as soon as the BBT change is fully released; please just ping me here.

(On a side note, we'll almost certainly want to add a variable for citekey to CSL -- it'd do a ton to facilitate CSL/BibTeX interoperability cc @rmzelle  )

retorquere:
Done.

adunning:
Thank *you*!

Issue #1310, opened 2017-05-28
Automatic translator testing as part of the continuous integration


zuphilip:
It would be really good to have automatic translator testing as part of the continuous integration. This doesn't have to be a strict test, but the information could be just shown for all participants of a PR.

I imagine something like this:
1. There is a testing instance of the translation server running online somewhere, e.g.
    a. The same instance we are using now for the daily testing.
    b. The same machine but different instance we are using now for the daily testing.
    c. Some service like heroku should be easy to set up with the [Dockerfile](https://github.com/zotero/translation-server/blob/master/Dockerfile)
2. Some script is called by some CI
   a. Travis CI which we already use
   b. CircleCI or anything else
3. The script will on the testing instance checkout the suggested code in the PR
4. The [tests are run](https://github.com/zotero/translation-server#tests) (possibly only the ones from the changed translators) and compared to the test results from the master branch.
5. Some diff feedback is given.

dstillman:
That's what this ticket is for, yes. As I mentioned in the other thread, we already have [daily translator tests](https://translator-tests.zotero.org/) using the same test runner used in Scaffold. For CI tests, we'd test just the modified translator. We could consider also testing translators that depend on the modified one, but that could mean hundreds of translators for something like RIS, so it probably makes more sense just to make sure those translators have appropriate coverage in their own tests.

retorquere:
I'm still seeing some unexpected things assuming the existing test already run daily. I'm not sure about any of this though, please be gentle -- these are what I found, but the way I found it could be way off base.

testcase 2 (and a similar issue in testcase 6) in RIS.js: the input has `Y2  - 1986/6/23` but the `items` array in that test case doesn't have an `accessDate` field, so the test should fail (I think?), and in Zotero, the date is imported as `1986-06-22`

maybe more a question than anything else, but in testcase 3 and 5 of RIS.js, the `items` array has a `version` field but I've been going over the code and I don't see where it is set from the `IS` input field

in testcase 7 of RIS.js there's a `"volume": "7"` field in the `items` array that I think should be `edition` (if I'm interpreting line 6403 of RIS.js correctly)

BibTeX.js test case 8 misses an `"accessDate": "2012-07-27"`

Frieze.js has a syntax error (https://github.com/zotero/translators/pull/1820)

dstillman:
> Maybe web, but straight imports should just believe the accessDate being passed right?

Yes. I noted that on the linked ticket.

> I meant line 867, not 6403, sorry.

[...]

PR #1818, opened 2019-02-01
Re-introduce changes from #1814

adam3smith:
But this time without breaking stuff...

### DIFF
--- a/BibTeX.js
+++ b/BibTeX.js
@@ -19,7 +19,7 @@
 	"inRepository": true,
 	"translatorType": 3,
 	"browserSupport": "gcsv",
-	"lastUpdated": "2019-01-31 13:10:16"
+	"lastUpdated": "2019-01-31 20:37:00"
 }
 
 function detectImport() {
@@ -420,8 +420,8 @@ function processField(item, field, value, rawValue) {
 			}
 		}
 	
-	} 
-	//accept lastchecked or urldate for access date. These should never both occur. 
+	}
+	//accept lastchecked or urldate for access date. These should never both occur.
 	//If they do we don't know which is better so we might as well just take the second one
 	else if (field == "lastchecked"|| field == "urldate"){
 		item.accessDate = value;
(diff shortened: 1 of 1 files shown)


retorquere:
(not that it's any kind of excuse for my error, but why didn't the test suite pick this up?)

retorquere:
I see now, travis only does a lint check. This looks good to me.

adam3smith:
Yeah, more advanced CI for translators would be nice, especially since most translators already have built in unit tests (wouldn't have helped here because export translators don't). @zuphilip and @dstillman talked about this briefly when they implemented the current CI but I forgot where that was left. Obviously the overhead to run the tests is pretty significant, so this isn't a trivial problem.

retorquere:
Already working on both. Current tests run in a second or two.

retorquere:
(and I've already found a few errors in existing translators)

zuphilip:
There is a ticket for supporting export translators in Scaffold https://github.com/zotero/scaffold/issues/103 and I would appreciate also automatic tests for export translators.

I guess that now with the new [translation-server](https://github.com/zotero/translation-server) it is more easily possible to include to run also automatically the test cases for each pull request or commit. As most PR are changing only a single translator, it is enough to test these test cases and that shouldn't take that long.

dstillman:
> Already working on both. Current tests run in a second or two.

Not sure what you're referring to here, but we have a [ticket](https://github.com/zotero/translators/issues/1310) for CI-based testing of translator commits. This would be separate from the [existing daily translator tests](https://translator-tests.zotero.org/), though it would use the [same Node-based testing functionality](https://github.com/zotero/provo/blob/94c6fa691005e5935f58209aaf557723bbd2a6b9/test_v2.sh#L30). (Web translators can easily fail when run that way because JS isn't run, but import/export tests should never fail.)

If you're working on this, it might be worth posting to that ticket to make sure you're not doing unnecessary work.

PR #1820, opened 2019-02-01
Missing closing parenthesis

retorquere:


dstillman:
Thanks! (Not sure how this made it through.)

retorquere:
I have no right to remark on that one given my resent slip-up with BibTeX.js (for which, again, apologies)

PR #1829, opened 2019-02-17
WIP: Add Veridian.js

pbinkley:
This translator handles Veridian newspapers. Veridian markets a digitization platform often used by college newspapers, e.g. [The Stanford Daily](https://stanforddailyarchive.com/). It has changed a good deal over time, and different instances vary a lot in the html patterns that a translator has to track. I've worked with two instances which I think cover the range of options.

Before this can be merged there's a problem I'd love to get help with: the Scaffold tests don't work, I think because the element that contains the article-level information is populated by an AJAX call and that call doesn't run or doesn't complete in time. If someone could point me to an example of how to handle this in a translator, I'd be very grateful.

### DIFF
--- a/Veridian.js
+++ b/Veridian.js
@@ -1,0 +1,211 @@
+{
+	"translatorID": "fb6773ac-74b1-406f-b6b0-9343fed56bb5",
+	"label": "Veridian",
+	"creator": "Peter Binkley",
+	"target": "^.*\\?a=d&d=.*",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2019-02-12 03:52:30"
+}
+
+function detectWeb(doc, url) {
+  // TODO handle other item types besides newspaperArticle, since Veridian is 
+  // also used for general digitization projects
+  if (ZU.xpath(doc, '/html/head/meta[@name="veridian-license"]').length > 0) {
+  return 'newspaperArticle';
+  }
+}
+
+function doWeb(doc, url) {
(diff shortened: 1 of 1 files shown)


GuyAglionby:
The `defer` property in tests should get around the problem you describe of things being populated with AJAX, but it appears only to work on test update and not on test execution (see https://github.com/zotero/scaffold/issues/96). I've seen this addressed by leaving the appropriate test cases as comments in the code, but will leave it to the maintainers to suggest the appropriateness of this.

In this case, it looks like this might be able to be circumvented by doing a call directly to the API, e.g. for test instance https://stanforddailyarchive.com/cgi-bin/stanford?a=d&d=stanford19640423-01.2.15&srpos=2&e=-------en-20--1--txt-txIN-microfilm------, adding `&f=XML` to get [this](https://stanforddailyarchive.com/cgi-bin/stanford?a=d&d=stanford19640423-01.2.15&srpos=2&e=-------en-20--1--txt-txIN-microfilm------&f=XML), which also appears to include the necessary metadata.

AbeJellinek:
This is really cool work, and it would cover a lot of archives that we could really use support for. The Stanford Daily (one of your test cases) has switched to something of their own creation, but the [California Digital Newspaper Collection](https://cdnc.ucr.edu/) and others are still using Veridian installs that look compatible with this translator. Would you be able to rebase on `master`, implement @GuyAglionby's suggestion, and update the tests? I'd love to get this merged.

pbinkley:
Ugh, I had forgotten about this - but a couple of weeks ago I saved some stuff from a Veridian site and noticed some problems, and thought about posting an issue ... I guess it only worked at all because my dev translator is still installed in Firefox. I'll try to finish this up.

AbeJellinek:
Excellent, thank you!

PR #1830, opened 2019-02-18
eslint with plugin

retorquere:
Translator linter in javascript, reports line numbers for errors.


dstillman:
Thanks!

I'll look at this more, but can you convert .eslintplugin.js to tabs to start?

Also, if this is replacing the existing checks, we can remove jshint and the other scripts, right?

retorquere:
spaces -> tabs, done, checkTranslators.sh removed, added two checks I forgot. I don't know what `checkDeleted.sh` does so I've not removed that.

.ci/package.json
```diff
@@ -4,15 +4,18 @@
   "description": "Continuous integration of Zotero translators",
   "main": "index.js",
   "scripts": {
-    "test": "find .. -maxdepth 1 -name '*.js' -print0|xargs -0 -n1 -P8 ./checkTranslator.sh --skip-warn",
-    "start": "find .. -maxdepth 1 -name '*.js' -print0|xargs -0 -n1 -P1 ./checkTranslator.sh"
+    "errors": "find .. -maxdepth 1 -name '*.js' -print0|xargs -0 eslint -c eslintrc --quiet",
```
dstillman: 1) Add spaces around `|`, here and below
2) Why not move `eslintrc` to `.eslintrc` and have it be implicit?

.ci/package.json
```diff
@@ -4,15 +4,18 @@
...
+    "test": "find .. -maxdepth 1 -name '*.js' -print0|xargs -0 eslint -c eslintrc",
+    "start": "find .. -maxdepth 1 -name '*.js' -print0|xargs -0 eslint -c eslintrc --max-warnings 0"
```
dstillman: Do we need the `start` line? Seems like `errors` and `test` are sufficient, and it's less clear what `start` should do anyway, so I'd remove this.

.ci/.eslintplugin.js
```diff
@@ -0,0 +1,167 @@
...
+
+const deleted = new Set(
+	fs.readFileSync('../deleted.txt', 'utf-8')
+	.split('\n')
+	.map(line => line.split(' ')[0])
+	.filter(id => id && id.indexOf('-') > 0)
+)
+const headers = {}
+
+const translators = new Set([])
```
dstillman: Don't need `[]`

.ci/.eslintplugin.js
```diff
@@ -0,0 +1,167 @@
+const fs = require('fs')
```
dstillman: Add semicolons throughout

.ci/package.json
retorquere: 1. Done, but I took that from `checkTranslators.sh`
2. eslint looks in the directory where `.git` lives so it won't find .eslintrc there

.ci/package.json
retorquere: I don't know -- checkTranslators had several check modes and I didn't know where they were used.

"start" reports warning and errors but will exit with a non-0 exit status if it finds warnings, "test" will report warnings and errors but only exits with non-0 when any errors are found

I added "errors" because there were so many warnings it was hard to see the errors...

.ci/.eslintplugin.js
retorquere: Removed

.ci/.eslintplugin.js
retorquere: Done, and a check added on the linter itself for this.

dstillman:
I'm not really sure what's going on at this point, but it seems like it's becoming way too verbose and complicated. This was fine as of my last review, and it simply required running `npm i` — there shouldn't be any need to install packages manually, to use `--save-dev` (which is for modifying package.json), or to specify rules, and there was nothing wrong with the eslint-plugin-local dependency.  Unless there's some clear problem, can we just go back to that?

dstillman:
If anything, I think the change here should be to move .eslintrc into the root so that it's available to someone using an editor that supports ESLint without having to run commands manually (even if there are additional checks run by CI).

retorquere:
No problems going back. But `npm i` should still just work? It does for me, no need to install anything manually. For this setup, save-dev and save do the same thing really as the translator repo isn't likely to be published as an npm module. Maybe I'm missing something, maybe my dev env is polluted in some way, but I can `rm -rf node_modules && npm i && npm test` and that works here.

Oh wait -- maybe you mean the devDeps in `eslint-plugin-zotero-translator`? I hadn't noticed those but they can go, they were generated by the yeoman generator but you're not actually expected to install them. I don't run `npm i` there. The `package.json` is only there so that the `package.json` in the `.ci` directory resolves `file:eslint-plugin-zotero-translator` as a proper package.

I'm totally fine moving the `.eslintrc` file up, I just didn't want to touch that directory wantonly, I don't know what the expectations are for files that can live there by programs consuming that directory.

The ideas behind the later changes were:

1. `eslint-plugin-local` was only added because eslint can only use requirable packages as plugins, but it turns out that there's a way to load packages directly rather than using `eslint-plugin-local` as a trampoline, so `eslint-plugin-local` was replaced by `eslint-plugin-zotero-translator`; that's one less external dependency as `eslint-plugin-zotero-translator` is just a directory in the `.ci` directory. Now that npm packages can be loaded this way it seems to me `eslint-plugin-local` is obsolete.
2. The single-file `eslint-plugin-local` setup worked but split configuration of what errors/warnings one could expect between hardcoded in the plugin code and configured in eslintrc. That seemed bad from a discoverability POV to me. The later changes just took the existing code in the `eslint-plugin-local`, split it to rules, and let eslintrc do the config. There's not actually (much) more code, the errors/warnings that are checked for have just been moved to separate files. It'd be trivial to move everything back into one file, but this did help for readability for me at least.
2. `node-exec-promise`, `simple-git` and terminal-kit` were only added to replace checkDeleted.sh by checkDeleted.js. That can simply be reverted wholesale.

dstillman:
Great, thanks — we can make some further tweaks, but this is very helpful.

I don't really have an opinion on checkDeleted. We can certainly switch to the JS version if it's solving a particular problem or if people think it'll be easier to maintain.

retorquere:
It's more maintainable IMO (bash is great for availability but not for readability), but then checkdeleted has not seen much management, so maybe it's just done, and then maintainability doesn't mean much.

retorquere:
A lot of the indent warnings come down to differences in indentations from switch statements. There's two major styles in the translators:

```
switch (field) {
case 'key':
case 'version':
  continue;

case 'creators':
case 'tags':
  break;
}
```

and 

```
switch (field) {
  case 'key':
  case 'version':
    continue;

  case 'creators':
  case 'tags':
    break;
}
```

eslint can accommodate either. I know which I prefer, but then I also prefer spaces over tabs, so our aesthetic preferences may not align :)

dstillman:
Zotero style is the former, which is also the default in ESLint.

Note that we now have an [ESLint config](https://raw.githubusercontent.com/zotero/zotero/master/.eslintrc) for the client, which should also apply to most Zotero JS code. A lot of existing code doesn't match that, so we're not enforcing it anywhere yet, and I want to close some open pull requests/branches before using `--fix` or making other changes, but it's aspirational.

Translators should use the same rules. I'm generally not a fan of mass-edits for style reasons, though, so I'm inclined to say that translator authors should just make style fixes as separate commits before making other changes to the translator.

PR #1831, opened 2019-02-20
Support inverted creator names for schema.org

mrtcode:
In more than a half cases author names are provided inverted when using schema.org ontology and JSON-LD.


zuphilip:
@mrtcode Okay, that is fine. Do you happen to have an example in RDF which we could save as a test case? If yes, this could be interesting to save, otherwise we can also merge that without.

PR #1833, opened 2019-02-23
BBC language fix

leoncastro:
Language code `en-GB` for the BBC must not be fixed ([as done here](https://github.com/zotero/translators/blob/8203550b89c3288e3e0b5171c6d6db2f937ef72a/BBC.js#L177)), because it uses different languages (examples: [fr](https://www.bbc.com/afrique), [es](https://www.bbc.com/mundo), [pt](https://www.bbc.com/portuguese/), [ru](https://www.bbc.com/russian), ...). Language can be found in the HTML `lang` tag, using the preset only as fallback.

Maybe it works if you replace [that line](https://github.com/zotero/translators/blob/8203550b89c3288e3e0b5171c6d6db2f937ef72a/BBC.js#L177) with:
`
item.language = ZU.xpathText(doc, '//*/@lang/') or "en-GB";
`

### DIFF
--- a/BBC.js
+++ b/BBC.js
@@ -174,7 +174,7 @@ function scrape(doc, url) {
 		for (var i in item.tags)
 			item.tags[i] = item.tags[i].charAt(0).toUpperCase()+item.tags[i].substring(1);
 
-		item.language = "en-GB";
+		item.language = ZU.xpathText(doc, '//*/@lang/') or "en-GB";
 
 		if (url.substr(-4)==".stm") {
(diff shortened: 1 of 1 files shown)


zuphilip:
Thanks @leoncastro for the try, but that does not work and is not valid JS.

Anyway, I see the issue and can fix it, which might be easier.

leoncastro:
My bad. Replace "or" with "||" and you will get an easier, working and valid JS.
Anyway, this pull can be closed if any other effective solution is applied.

zuphilip:
Superseeded by #1956.

PR #1835, opened 2019-02-24
upgrade to eslint:notice makes 'template' unnecesary.

retorquere:


dstillman:
You'll need to rebase and force-push this to avoid the duplicate commits.

retorquere:
I was wondering about that... I know you've explained before, but what are the steps to do this? Rebasing is the one git concept that keeps eluding me.

dstillman:
Depends how you have things set up and the current state, but something like this should do it:

```
git fetch upstream master
git rebase -i origin/master
```

(This is assuming you have `upstream` set as the remote for `zotero/translators` (as shown in `git remove -v`).)

If all goes well, you should see an editor window with only the relevant commit. You can then run `git push -f` to force-push it to this branch.

dstillman:
Oh, and you'll need to delete the extra commits in the editor window during the interactive rebase, since those are commits that don't exist upstream (since I squashed them into a single commit in the other PR). So you just want to leave the last line, which is the relevant commit.

(In the future, the way to avoid this is to start a new branch based on the latest upstream master rather than anything you have locally.)

retorquere:
How do I do that? I merged upstream/master into my master and had hoped to avoid this by doing that. Can I checkout upstream/master and branch that?

dstillman:
How do you do what?

dstillman:
If you mean how do you delete commits, you just delete the lines in the editor during the interactive rebase. It gives instructions in the window.

This happened because, somehow or another, you based this branch on your PR rather than on the merged master.

retorquere:
I really thought I did base the branch on the merged master. I'll figure this out and resubmit.

retorquere:
I meant how do I start a new branch based on the latest upstream master rather than anything you have locally, because I did just that.

dstillman:
(There's pretty much never a need to delete a pull request. You can always fix the branch and force-push, and it's almost always better to do that.)

retorquere:
Wait, that's what's gone wrong then. I have a local branch named `eslint-notice-0.7.8`, and this pull request is meant to come from that. I did delete the `eslint` branch but that is what was merged earlier.

PR #1836, opened 2019-02-24
upgrade to eslint:notice makes template unnecesary

retorquere:


retorquere:
So I can delete my own branches of this now right? And fetch upstream/master?

PR #1847, opened 2019-02-25
WorldCat Discovery Service.js -- fix eslint warnings

retorquere:


adam3smith:
You're a hero slogging through all of these!

retorquere:
I needed something to take my mind off. This is about as Zen as it gets :smile: I have some scripts set up to create branches of course, but cautious checking with `git diff -w` remains necessary. I''l do a few more tonight but must stop before I get change-blind.

dstillman:
Thanks a lot for doing this, but I would actually hold off before going through more of these. As [I noted](https://github.com/zotero/translators/pull/1830#issuecomment-466818968), we have a much larger ESLint config that we might want to apply to this repo (and most of our repos). I just added that config recently, so there may be some additional tweaks to make (and the warn/error distinction is somewhat arbitrary at this point), but it should mostly describe the bulk of existing Zotero code. Whether that matters for translator is up for debate, but it could certainly address things like [this](https://github.com/zotero/translators/pull/1848)

(I also suggested that we shouldn't do mass edits for style changes and translator authors should just fix them in separate commits when they update styles. Conflicts are much less of an issue here, though, so if @adam3smith and @zuphilip don't mind the extra churn, I'm OK with going through these now. But it doesn't make sense to do it more than once.)

dstillman:
(And one reason to do them along with other translator changes, after expanding the config, is that more intensive checking is likely to reveal actual bugs that need to be investigated and tested against the site.)

retorquere:
I'm OK holding off, it's mostly a papercuts issue for me. I reckon though that whatever edits I make now are going to have to be done in any case. One argument I could offer for doing it anyhow is that most translators seem to be copy-and-change jobs, so whatever code problems exist are more likely to proliferate.

But I'll stop sending in pull requests for now.

retorquere:
Those bugs would be unearthed by heavier linting in any case -- if you look at the `git diff -w`, I'm mostly just changing indents and indexOfs.

But as said, I'll stop for now. Feel free to holler when the new lint config is in place -- code tidying is relaxing work for me right now.

PR #1869, opened 2019-02-25
UpToDate References.js -- fix eslint warnings

retorquere:


retorquere:

UpToDate References.js
```diff
@@ -36,48 +36,49 @@
 	***** END LICENSE BLOCK *****
 */
 
-function detectWeb(doc, url) {
+function detectWeb(doc) {
```
zuphilip: I would prefer to use here the `detectWeb(doc, _url)` variable, because it may be that in some later changes we want to start using that variable.

UpToDate References.js
```diff
@@ -36,48 +36,49 @@
...
 	if (ZU.xpathText(doc, '//ol[@id="reference"]//a')) return "multiple";
 	else if (ZU.xpathText(doc, '//div[@class="abstractRow"]/div[@class="label" and contains(text(), "TI")]')) return "journalArticle";
+	return false;
 }
 
 function doWeb(doc, url) {
 	if (detectWeb(doc, url) == "multiple") {
-		items = {}
+		const items = {};
```
zuphilip: Why `const` for something that is changing?

UpToDate References.js
```diff
@@ -197,5 +199,5 @@ var testCases = [
 		"url": "http://www.uptodate.com/contents/approach-to-the-diagnosis-and-evaluation-of-low-back-pain-in-adults?source=search_result&search=back+pain&selectedTitle=1%7E150",
 		"items": "multiple"
 	}
-]
-/** END TEST CASES **/
\ No newline at end of file
+];
```
zuphilip: no

UpToDate References.js
retorquere: The `items` variable is constant (it always points to the same object), even if it contents changes; scalars can't have this behavior, but Objects, Sets, arrays, etc can.

zuphilip: I see. But we currently don't use `const` often/at all (?) in translator coding. We have in our template in scaffold also a `var`, cf. https://github.com/zotero/scaffold/blob/master/src/chrome/content/scaffold/templates/newWeb.js#L40 . Thus, I would argue, that it is more consistent to use here also `var`.

retorquere: That's your call -- personally, I find the rules for `var`-declared variables pretty confusing, and hoisting sometimes creates strange situations, so for my own coding, I've settled on using `const` by default, `let` when required, and `var` just never.

I'm inclined to think `var` is a relic from pre-ES2015 habits (and we have shifted from pre-ES2015 `indexOf(...) >= 0` to `includes(...)` too, so..), but again -- your call, I have no deep vested interest here.

retorquere: I mean I'd sooner argue that the scaffold template should be updated -- I would likewise suggest that https://github.com/zotero/scaffold/blob/master/src/chrome/content/scaffold/templates/newWeb.js#L44 be replaced with

```
for (const row of rows) {
```

for better readability

retorquere: https://github.com/zotero/scaffold/blob/master/src/chrome/content/scaffold/templates/newWeb.js#L61 should also be updated, because it should just `return`.

retorquere: Wouldn't https://github.com/zotero/scaffold/blob/master/src/chrome/content/scaffold/templates/newWeb.js#L64 - 68 be the same as 

```
ZU.processDocuments(Object.keys(items), scrape);
```

?

I could put together a pull request if so desired.

dstillman: FWIW, we don't really use `const` at all in Zotero code. We always use `let` for block scope in new code (and while we generally use `var` outside of block scope, I don't have strong objections to using `let` everywhere). Personally, I find `const` mostly just distracting with little benefit over `let` (and plenty of the sort of confusion that @zuphilip shared). While I haven't asked for changes when you've used it, I think it'd be better to avoid it.

retorquere: This isn't really "my" code, so I'll just follow you guys' recommendations. In this case, either var or let will do.

zuphilip: Okay, will you do the change here then?

A PR for the template code in Scaffold would be great. The other suggestions you mentioned here look fine.

retorquere: Yep, sure. `var` or `let`?

dstillman: Always `let` within a block

retorquere: Done.

Wouldn't that mean almost everywhere though? Aren't function bodies blocks?

In the lint-fixes I don't change all `var`s to `let`s BTW, nor have I replaced `for(; ;)` with `for ( of )`; I take it it has your preference such lint-fixes introduce the minimum amount of changes to make it pass the linter. I haven't changed the `var titles` below it for example.

retorquere: I'm OK with ebing asked for changes where I've used it as far as I'm concerned. You have a better view of what will resonate with "the average" translator writer (if there is such an average) than me. I don't mind.

dstillman: > Wouldn't that mean almost everywhere though? Aren't function bodies blocks?

I mean other than function blocks — i.e., where there's an important scope difference (and not just the more subtle differences of hoisting behavior and redeclaration).

> In the lint-fixes I don't change all vars to lets BTW, nor have I replaced for(; ;) with for ( of )

Yes, that's fine. These things can get updated as they're changed for other reasons.

retorquere: Fine as in I don't have to do it, or as in you prefer I don't do it. I'm OK either way, just would be good for me to know.

zuphilip:
Thank you!

PR #1868, opened 2019-02-25
Rewrite VoxEU.js

retorquere:

PR #1871, opened 2019-02-25
Vanity Fair: Remove FW, update for current site

retorquere:


retorquere:

zuphilip:
That is a translator using the FW as well. Thus, I guess it makes more sense to do a rewrite before fixing any linting errors.

retorquere:
I'm looking at the FW now -- there's actually quite a bit of code hidden in that single line. I know you recently replaced the FW in a translator, which was that again? I can take a look to see what you actually changed for guidance.

zuphilip:
See https://github.com/zotero/translators/pull/1868

retorquere:
I see a checkbox in scaffold that says "use external editor", how does that work? And can I point Zotero to another translators directory so that I can have it use my repo clone?

retorquere:
OK so if I'm getting this right, `external` means it will reload the translator just before it's ran... but all of the Vanity Fair tests fail currently with the TFW. Looking into it.

retorquere:
It looks to me the Vanity Fair site has changed substantially. I think I have the rewrite mostly tackled, but there's two things that still stump me:

1. In the test for http://www.vanityfair.com/society/features/2011/05/top-one-percent-201105, I can't seem to get the `abstractNote` captured, even if I can see that it's in the page (it shows up in the log using `Zotero.debug`)
2. In the `multiple` test I can see that I'm getting urls and titles but the test flames out with a `failed (Translation failed: no items returned)`

retorquere:
Wait, I have to pass `scrape` to `processDocuments`. But how would `scrape` know which item type to create?

retorquere:
Oh wait I can just run detectWeb again

retorquere:
OK so point 2 is updated now but I'd still appreciate an assist.

retorquere:
OK, so strange... the `multiple` test case gives me https://www.vanityfair.com/style/2019/06/royal-ascot-queen-tradition, and when that's scraped I get a 402, but when I scrape that URL directly, it just works. I must be doing something wrong in `processDocuments`.

zuphilip:
Sorry for not reacting earlier. Is this PR ready for a review or do you still struggle with multiples?

retorquere:
The Vanity Fair website has changed in the interim, and none of the tests pass now. I'll see if I can get to it next weekend -- this week is packed.

retorquere:
I don't know what distinguishes a blog post from a magazine article on Vanity Fair, so I don't think I can fix this. The HTML has changed substantially on the site.

retorquere:
I'm afraid this PR will have to be closed. While it won't be hard to fix from a technical POV, I don't know what distinguishes a blog post from a magazine article, conceptually, on Vanity Fair, and I don't immediately see any distinguishing HTML in the source between the test pages.

zuphilip:
I think that magazineArticle and blogPost are quite similar and wouldn't take this as a show-stopper if we cannot always properly distinguish between the two of them.

However, here in the test cases it seems that for magazinArticles there is some kind of breadcrumb list at the beginning, which you can check for presences, i.e. `.content-header__rubric--issue-date`.

Okay?

retorquere:
The single-item tests pass now, the multiple fails. The error occurs because `text('.byline__name')` return `null` in the sub-scrape, but if I look in the page it's trying to scrape (https://www.vanityfair.com/style/2019/11/alison-roman-well-then-wellness-diary-nothing-fancy-cookbook), `div.byline__name` is present.

zuphilip:
It looks like they are loading some content later. Can you just check whether `text('.byline__name')` is not null and non-empty before giving it in another function? Possibly also with some of the other variables. I am okay, that we don't have the best data for multiple but something with works and then we can finally review and merge this PR here.

retorquere:
I'm not giving it to another function, that's just how the multi scraper works right? It gets a bunch of urls, scrapes each. When I scrape each of those urls individually, it just works. Maybe it's a timing issue.

zuphilip:
Well, you are giving the result of this byline to the `ZU.cleanAuthor` function and this will give you the error in the end, no? Thus I would suggest to do something like
```js
var byline= text('.byline__name');
if (byline) {
   item.creators = [ZU.cleanAuthor(byline, 'author')];
}
```
(not tested)

The same error can happen in the future, if they switch to a different structure. Usually, I think that then it is better to still provide some data than to fail immediately.

retorquere:
The thing is that if I'm visiting the page it's scraping, that element does exist. It's just when the multiple scraper does it, it seems to not exists.

zuphilip:
Sorry, I don't understand the problem. If the element does not exist, then this will be filtered out in the suggested code above. If the element exists, then you pass the if-clause and can do further steps.

Does this code give you still an error?

retorquere:
The element *does* exist. I can load the page, and I can see that it's there. If I scrape that page, I get the element back. It is *only* when `scrape` is called through `ZU.processDocuments` that it seems to not be picked up.

zuphilip:
Well, when I load the website in my browser then I see that first a page with some information missing (e.g. the author information is missing) is loaded and just after some moments the remaining information will be loaded and shown on the website. Therefore it is possible that in multiples we only grab the information of the first page shown.

retorquere:
I'm getting a different error on multi-scrape now. It looks like the first URL it tries to get data on is https://www.vanityfair.com/style/2019/11/prince-charles-art-fraud. If I load that into my own browser, no issue, if I load that into the scaffold browser tab, no issue, but during the scrap I get 

```
Error: HTTP GET https://www.vanityfair.com/style/2019/11/prince-charles-art-fraud failed with status code 402
```

zuphilip:
HTTP status code 402? Interesting. This could be temporary or somehow they are disallowing bots/headless browsers.

I am fine to drop support for multiples completely for this translator. I don't think it is worth the trouble...

retorquere:
Done.

PR #1873, opened 2019-02-25
Vice.js -- fix eslint warnings

retorquere:


zuphilip:
Let me know when you updated this one.

retorquere:
I've now updated it.

retorquere:
I'm not sure why the linter fails? I don't see tabs in the files at those spots. Looking into it, but in the Travis logs, and locally, there are tabs there.

retorquere:
Ah, new linter rules that I needed to install.

Vice.js
```diff
@@ -38,12 +38,22 @@
 
 
 // attr()/text() v2 per https://github.com/zotero/translators/issues/1277
-function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null;}function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null;}
+function attr(docOrElem, selector, attr, index) {
```
zuphilip: These lines can be deleted completely, as text/attr are now global functions from Zotero itself.

retorquere: I can add a linter rule for that if so desired. Would be an easy rule to do.

zuphilip: That would be great! Here is the corresponding issue #1913.

retorquere: https://github.com/zotero/translators/pull/1961

Issue #951, opened 2015-09-29
Licence all possible translators under AGPL?


zuphilip:
It was mentioned in <a href="https://github.com/zotero/translators/pull/947#discussion_r40689273">this thread</a> to re-licence the ACM translator to <a href="https://en.wikipedia.org/wiki/Affero_General_Public_License">AGPL</a>. I found a total of 111 result with a General GPL licence. Should we try to change all translators to AGPL? This is possible if the creator(s) accept to do so.

Here is an uncomplete list:
 * @adam3smith : <a href="https://github.com/zotero/translators/search?utf8=%E2%9C%93&q=%22copy+of+the+GNU+General+Public+License%22+karcher&type=Code">49 results</a>
 * @avram : <a href="https://github.com/zotero/translators/search?utf8=%E2%9C%93&q=%22copy+of+the+GNU+General+Public+License%22+Avram+&type=Code">18 results</a>
 * @aurimasv : <a href="https://github.com/zotero/translators/search?utf8=%E2%9C%93&q=%22copy+of+the+GNU+General+Public+License%22+aurimas&type=Code">7 results</a>

Note, there might be more than one contributors mentioned.

adam3smith:
I win!
Obviously no objections to relicense "mine".

On Tue, Sep 29, 2015 at 11:20 AM, Philipp Zumstein <notifications@github.com
> wrote:

> It was mentioned in this thread
> <https://github.com/zotero/translators/pull/947#discussion_r40689273> to
> re-licence the ACM translator to AGPL
> <https://en.wikipedia.org/wiki/Affero_General_Public_License>. I found a
> total of 111 result with a General GPL licence. Should we try to change all
> translators to AGPL? This is possible if the creator(s) accept to do so.
>
> Here is an uncomplete list:
>
>    - @adam3smith <https://github.com/adam3smith> : 49 results
>    <https://github.com/zotero/translators/search?utf8=%E2%9C%93&q=%22copy+of+the+GNU+General+Public+License%22+karcher&type=Code>
>    - @avram <https://github.com/avram> : 18 results
>    <https://github.com/zotero/translators/search?utf8=%E2%9C%93&q=%22copy+of+the+GNU+General+Public+License%22+Avram+&type=Code>
>    - @aurimasv <https://github.com/aurimasv> : 7 results
>    <https://github.com/zotero/translators/search?utf8=%E2%9C%93&q=%22copy+of+the+GNU+General+Public+License%22+aurimas&type=Code>
>
[...]

PR #1876, opened 2019-02-25
Unqualified Dublin Core RDF.js -- fix MOST eslint warnings

retorquere:
I don't know where dcterms is supposed to come from

### DIFF
--- a/Unqualified Dublin Core RDF.js
+++ b/Unqualified Dublin Core RDF.js
@@ -11,25 +11,50 @@
 		"dataMode": "rdf/xml"
 	},
 	"inRepository": true,
-	"lastUpdated": "2018-12-17 03:16:21"
+	"lastUpdated": "2019-06-11 13:33:25"
 }
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2008 Simon Kornblith
+
+	This file is part of Zotero.
+
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
 
 function doExport() {
 	var dc = "http://purl.org/dc/elements/1.1/";
 	Zotero.RDF.addNamespace("dc", dc);
 	
 	var item;
-	while (item = Zotero.nextItem()) {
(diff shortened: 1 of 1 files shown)


retorquere:

Unqualified Dublin Core RDF.js
```diff
@@ -82,42 +108,44 @@ function doExport() {
...
 		// archiveLocation
 		if (item.archiveLocation) {
-			Zotero.RDF.addStatement(resource, dc+"coverage", item.archiveLocation, true);
+			Zotero.RDF.addStatement(resource, dc + "coverage", item.archiveLocation, true);
 		}
 		
 		// medium
 		if (item.medium) {
-			Zotero.RDF.addStatement(resource, dcterms+"medium", item.medium, true);
+			Zotero.RDF.addStatement(resource, dc + "medium", item.medium, true);
```
zuphilip: There is no `medium` property in DC elements http://purl.org/dc/elements/1.1/ but in DC terms http://purl.org/dc/terms . However, this translator is named "Unqualified Dublin Core RDF" which would be just DC elements. Hm... we have to think this further.

Unqualified Dublin Core RDF.js
```diff
@@ -11,25 +11,50 @@
...
 	},
 	"inRepository": true,
-	"lastUpdated": "2018-12-17 03:16:21"
+	"lastUpdated": "2019-06-11 13:33:25"
 }
 
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2018-2019 Simon Kornblith
```
zuphilip: Did you speak with him? Any reason for the time spans here where Simon was not actively working on this?

Unqualified Dublin Core RDF.js
zuphilip: Actually, I think we can use `dc:format` here, which stands for "The file format, physical medium, or dimensions of the resource." http://www.dublincore.org/specifications/dublin-core/dcmi-terms/#elements-format . But this then should go into a separate commit.

Unqualified Dublin Core RDF.js
retorquere: I haven't, no, but I can.

zuphilip: This might not be needed, cf. https://github.com/zotero/translators/issues/951#issuecomment-145275715

zuphilip: Then I simply suggest to change the date to the original creation date to indicate this is a older translator. Maybe we also should put the relicensing commit into a separate commit.

Unqualified Dublin Core RDF.js
retorquere: OK, so what should I do in the interim?

Unqualified Dublin Core RDF.js
retorquere: Which date? The date in the license block? So just have 2018 in the license block?

zuphilip: 2008 should be okay

retorquere: done

Unqualified Dublin Core RDF.js
zuphilip: I will handle that...

zuphilip:
Thank you!

PR #1878, opened 2019-02-25
Wanfang Data.js -- fix MOST eslint warnings

retorquere:
Is GPL-licensed, not AGPL.

There's two functions that don't return a value, but I don't know what these are expected to return.


retorquere:
updated for @zotero/eslint-config@1.0.2 (license warning remains)

zuphilip:
Thank you!

PR #1879, opened 2019-02-25
University of Wisconsin-Madison Libraries Catalog.js -- fix eslint wa…

retorquere:
…rnings


retorquere:

University of Wisconsin-Madison Libraries Catalog.js
```diff
@@ -59,20 +60,21 @@ function detectWeb(doc, url) {
 				return "computerProgramm";
 			case "Photos, Drawings, Prints":
 				return "artwork";
-			//case "Journals, Magazines, Newspapers":
-			//there is no such itemType yet
+				// case "Journals, Magazines, Newspapers":
+			// there is no such itemType yet
```
zuphilip: No, this intending is not what was mean.

I suggest to move this case just before the `default` case as
```js
case "Journals, Magazines, Newspapers": // there is no such itemType yet
```

retorquere: Done. But if there's a `default` anyway, why have the two fall-through cases above it?

zuphilip: If in the future other itemTypes for these cases will exists, then we can change the behavior here easily.

zuphilip:
Thank you!

PR #1880, opened 2019-02-25
unAPI.js -- fix eslint warnings

retorquere:


retorquere:

zuphilip:
I tried to test this changes in Scaffold for the test cases and get the following error:
```
Translation using unAPI failed: 
SyntaxError: redeclaration of let FORMAT_GUIDS
         
@unAPI:1:1
         
url => https://search.library.utoronto.ca/search?N=0&Ntk=Anywhere&Ntt=nimni+challenge+of+post-zionism&Ntx=mode%252Bmatchallpartial&Nu=p_work_normalized&Np=1&formName=search_form_simple
downloadAssociatedFiles => true
automaticSnapshots => true
```
This is something new as I compared the result with the previous commit. I don't know yet what exactly causes this error...

unAPI.js
```diff
@@ -9,22 +9,54 @@
...
-	"unimarc":"a6ee60df-1ddc-4aae-bb25-45e0537be973",
-	"marcxml":"edd87d07-9194-42f8-b2ad-997c4c7deefd",
-	"ris":"32d59d2d-b65a-4da4-b0a3-bdd3cfb979e7",
-	"refer":"881f60f2-0802-411a-9228-ce5f47b64c7d",
-	"bibtex":"9cb70025-a888-4a29-a210-93ec52da40d4",
-	"rdf_dc":"5e3ad958-ac79-463d-812b-a86a9235c28f"
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2018-2019 Simon Kornblith
```
zuphilip: Let us adjust the time span to 2007-2012 here.

unAPI.js
```diff
@@ -285,52 +329,56 @@ function determineDetectItemType(ids, supportedId) {
...
+	}
+	else {
 		// Several items. We will need to call determineDetectItemType
 		determineDetectItemType(ids);
 	}
+	return false;
 }
 
-function doWeb(doc, url) {
+function doWeb(doc) {
```
zuphilip: `_url`

unAPI.js
```diff
@@ -109,89 +147,92 @@ UnAPIFormat = function(aXML) {
...
 	 * @param {Function} callback Callback to be passed items when they have been retrieved
 	 */
-	"getItems":function(callback) {
+	getItems: function (callback) {
 		if (this.items) {
 			callback(this.items);
 			return;
 		}
 		
-		var me = this;
```
zuphilip: Why did you delete this variable and what was it doing in the first place?

retorquere:
There's something strange going on in Scaffold for me. Previously, when I had "use external editor" checked and I clicked the "detectWeb" button, the code would update, but it doesn't now.

unAPI.js
retorquere: It brought a copy of `this` into the function passed to `isSupported` (otherwise `this` inside that function would be `null`), but arrow-functions retain the outer `this` so they don't need this trick.

retorquere:
I'm seeing the same "redeclaration of let FORMAT_GUIDS" but that makes no sense whatsoever. I've stuck a

```
Zotero.debug('MY_FORMAT_GUIDS: ' + typeof(MY_FORMAT_GUIDS))
```

just before the 

```
var MY_FORMAT_GUIDS = {
```

and now I'm seeing this in the log

```
15:26:41 TranslatorTester: Running 1 test for unAPI
15:26:41 TranslatorTester: Running unAPI Test 1
15:26:41 TranslatorTester: Translating https://search.library.utoronto.ca/search?N=0&Ntk=Anywhere&Ntt=nimni+challenge+of+post-zionism&Ntx=mode%252Bmatchallpartial&Nu=p_work_normalized&Np=1&formName=search_form_simple
15:26:41 MY_FORMAT_GUIDS: undefined
15:26:41 MY_FORMAT_GUIDS: object
15:26:41 Translation successful
15:26:41 TranslatorTester: unAPI Test 1: succeeded (Test succeeded)
```

This seems to me like either Zotero or Scaffold is sourcing the whole script twice (perhaps for `doWeb` and `detectWeb`?) while retaining the runtime environment, rather than running the script twice with a clean environment or running `doWeb` and `detectWeb` withing the same single runtime environment.

I can't be sure but I don't think it was previously this way. I saw a new Zotero version rolling in yesterday, could that have brought this change? Is the new `eval` sandbox live?

retorquere:
Nope, just tested on 5.0.69, same behavior. Well, what to do about this? I find this behavior pretty strange TBH.

zuphilip:
> I can't be sure but I don't think it was previously this way. I saw a new Zotero version rolling in yesterday, could that have brought this change? Is the new eval sandbox live?

No, when I try the same tests for unAPI translator in Scaffold on the master branch, then they work fine. It has to be something you changed in this PR.

@dstillman Let us know if you have any idea why we see here `redeclaration of let FORMAT_GUIDS` error now.

retorquere:
It's because I changed `var` to `let`; `var` allows a variable to be redeclared.

retorquere:
I've changed it back to var, which should work, but I find the way this runs strange (and I'll have to go back to my BBT translators because I was under the assumption that per `detectX`/`doX` combo, the top-level code would only be ran once, which from the looks of it is not true).

zuphilip:
The problem occurs only if you pressed on the run test button in Scaffold. If you load a website in Scaffold and then either run detectWeb or doWeb; this worked fine.

zuphilip:
Thank you!

PR #1893, opened 2019-02-25
Theory of Computing.js -- fix eslint warnings

retorquere:


retorquere:

Theory of Computing.js
```diff
@@ -9,10 +9,10 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcv",
-	"lastUpdated": "2017-07-03 10:55:03"
+	"lastUpdated": "2019-06-10 22:52:50"
 }
 
-/*****
+/** ***
```
zuphilip: Let us use here simply
```js
/*
...
*/
```
and delete the other asterisks in the beginning and the end.

Theory of Computing.js
```diff
@@ -31,10 +31,8 @@
 *****/
 
 
-
-
-/*****
-  Theory of Computing translator 
+/** ***
```
zuphilip: dito

Theory of Computing.js
```diff
@@ -67,127 +65,119 @@
...
+	// We now start constructing the Zoetro item
 	var newItem = new Zotero.Item(type);
 
-	//Store the snapshot and the PDF file
-	newItem.attachments.push({
-		title: "Theory of Computing Snapshot",
-		document: doc});
+	// Store the snapshot and the PDF file
+	newItem.attachments.push({ title: "Theory of Computing Snapshot",
+		document: doc });
```
zuphilip: Which rule is this? This looks unfamiliar for me...

retorquere: There's a rule that allows such indenting, but not one that enforces it. If I put it on one line, it stays there. Alternate is to only indent the keys (which I think looks best, and which I've just submitted).

zuphilip: Agree, I like this option also better.

zuphilip:
Thank you!

PR #1894, opened 2019-02-26
Add ESLint rules from client codebase

dstillman:
So this is what the config would be if we added all rules from the client. See https://github.com/zotero/translators/pull/1847#issuecomment-467231168 for context. I don't know whether we care about having this level of conformity in translators.

The warning/error distinction is also fairly arbitrary right now. Some things, particularly spacing rules, are warnings because they appear all throughout the existing client codebase, but conflicts are less of an issue here, so we could just make those errors and fix them, hopefully mostly with `--fix`. Or we might want to reserve errors for things that are more likely to cause bugs, but I'm not really sure what the point of style rules are if we're not going to enforce them consistently.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -135,7 +135,12 @@
 			"always-multiline"
 		],
 		"new-parens": "off",
-		"newline-per-chained-call": "error",
+		"newline-per-chained-call": [
+			"error",
+			{
+				"ignoreChainWithDepth": 3
+			}
+		],
 		"no-alert": "off",
 		"no-array-constructor": "error",
 		"no-await-in-loop": "warn",
--- a/.eslintrc
+++ b/.eslintrc
@@ -14,20 +14,342 @@
 		"XPathResult"
 	},
 	"rules": {
-		"semi": ["warn", "always"],
-		"indent": ["warn", "tab"],
-		"linebreak-style": ["error", "unix"],
-		"no-multi-str": "warn",
+		"accessor-pairs": "error",
+		"array-bracket-newline": [
+			"error",
+			"consistent"
+		],
+		"array-bracket-spacing": "error",
+		"array-callback-return": "error",
+		"array-element-newline": [
+			"error",
+			"consistent"
+		],
+		"arrow-body-style": "off",
+		"arrow-parens": [
+			"warn",
+			"as-needed",
+			{
+				"requireForBlockBody": true
+			}
+		],
+		"arrow-spacing": [
+			"error",
+			{
+				"after": true,
+				"before": true
+			}
+		],
+		"block-scoped-var": "off",
+		"block-spacing": [
+			"error",
+			"always"
+		],
+		"brace-style": [
+			"warn",
+			"stroustrup"
+		],
+		"callback-return": "off",
+		"camelcase": "error",
+		"capitalized-comments": "off",
+		"class-methods-use-this": "error",
+		"comma-dangle": "off",
+		"comma-spacing": [
+			"warn",
+			{
+				"after": true,
+				"before": false
+			}
+		],
+		"comma-style": [
+			"error",
+			"last"
+		],
+		"complexity": "off",
+		"computed-property-spacing": [
+			"error",
+			"never"
+		],
+		"consistent-return": "error",
+		"consistent-this": "error",
+		"curly": "off",
+		"default-case": "off",
+		"dot-location": [
+			"error",
+			"property"
+		],
+		"dot-notation": "error",
+		"eol-last": "error",
+		"eqeqeq": "off",
+		"func-call-spacing": "error",
+		"func-name-matching": "error",
+		"func-names": [
+			"error",
+			"never"
+		],
+		"func-style": "off",
+		"function-paren-newline": "off",
+		"generator-star-spacing": [
+			"error",
+			{
+				"before": false,
+				"after": true
+			}
+		],
+		"global-require": "off",
+		"guard-for-in": "off",
+		"handle-callback-err": "error",
+		"id-blacklist": "error",
+		"id-length": "off",
+		"id-match": "error",
+		"implicit-arrow-linebreak": [
+			"error",
+			"beside"
+		],
+		"indent": [
+			"error",
+			"tab"
+		],
+		"init-declarations": "off",
+		"jsx-quotes": "error",
+		"key-spacing": "warn",
+		"keyword-spacing": "warn",
+		"line-comment-position": "off",
+		"linebreak-style": [
+			"error",
+			"unix"
+		],
+		"lines-around-comment": "error",
+		"lines-around-directive": "error",
+		"lines-between-class-members": "error",
+		"max-depth": "off",
+		"max-len": "off",
+		"max-lines": "off",
+		"max-nested-callbacks": "error",
+		"max-params": "off",
+		"max-statements": "off",
+		"max-statements-per-line": "off",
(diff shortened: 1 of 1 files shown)


retorquere:
Just a side-note -- `--fix` doesn't work for the translators because of https://github.com/eslint/eslint/issues/5121

retorquere:
Oh and personally I'm all for a high level of encouraged consistency in the translators. Given that most new translators seem to be copy-and-change jobs from existing translators, enforcing rules would be low-impact (since they'd be copying fairly clean code after the first major sweep) and it'd not further proliferate bad code styles.

.eslintrc
```diff
@@ -14,20 +14,342 @@
...
+		"no-restricted-imports": "error",
+		"no-restricted-modules": "error",
+		"no-restricted-properties": "error",
+		"no-restricted-syntax": "error",
+		"no-return-assign": "off",
+		"no-return-await": "error",
+		"no-script-url": "error",
+		"no-self-compare": "error",
+		"no-sequences": "error",
 		"no-shadow": "off",
```
retorquere: I can't assess how impactful/useful this would be for Zotero, but this one has saved me on occasion from some confusion.

dstillman:
> Just a side-note -- `--fix` doesn't work for the translators because of [eslint/eslint#5121](https://github.com/eslint/eslint/issues/5121).

Oh, that's annoying. I guess not making these proper JS files was shortsighted, but we're at least a decade out from that decision, so we're probably stuck with it.

> of course a gnarly workaround can be made by physically splitting the translators into header and code files, linting them separately and recombining them if `--fix` is ran, but that's a little uglier than I'm comfortable with.

I don't know — it seems fairly trivial to just have a script that splits them, runs `--fix` on a temp file with the code part, and merges back to the original file.

retorquere:
It is annoying, and we're also stuck with it. I dunno -- it wouldn't be hard to make a script that just does what the preprocessor does and undoes it after. But it'd be changing files on disk, inside the repo. If it bombs out, it would leave changed files which could accidentally be checked in.

dstillman:
I'm not really concerned with accidental commits — anyone committing should be reviewing the diff first, and `--fix` itself has to be reviewed anyway — but it should be easy to make the script fairly safe.

retorquere:
I can have a stab at that.

retorquere:
I'll just do the split to a separate directory and stick that in .gitignore, that should minimize the risk.

retorquere:
I'll wait for https://github.com/zotero/translators/pull/1895 to clear though, it would build on some of those changes.

dstillman:
> I'll just do the split to a separate directory and stick that in .gitignore, that should minimize the risk.

That's not necessary. The header can just stay in memory, and the code can go in a temp file, and then they can be recombined back into the original file. The temp file can be cleaned up in an exit trap in case something goes wrong.

retorquere:
You mean by running eslint from the script itself? Wouldn't this split off the messages about the header from the rest of the eslint output? Just adding the temp dir to .gitignore would achieve much the same and we could just run eslint as usual.

dstillman:
Yes, we'd just have an eslint-fix script in .ci that took a filename. This is just for running `--fix`, so the header checks wouldn't be relevant.

retorquere:
I don't completely follow, but I think we agree on what we want to achieve and roughly how to do it.

retorquere:
> Oh, that's annoying. I guess not making these proper JS files was shortsighted, but we're at least a decade out from that decision, so we're probably stuck with it.

Are we though? What would be the harm from optionally allowing a translator to start with `const ZOTERO_TRANSLATOR_INFO = ` and stripping this exact string from the start of translators when they are loaded into Zotero? Old translators would continue to work, and an ESLint rule to detect that it's missing would be easy enough to do.

dstillman:
I don't think it's worth it. We'd have to make changes in many places in many repos, we'd have to either cut off translator updates or send backported translators to older versions, and we'd risk bugs in critical infrastructure, all with very little to be gained. We already have your postprocessor to get ESLint checks working, and making a `--fix` wrapper would take about 10 minutes.

retorquere:
I have a wrapperless solution that I'm testing now.

retorquere:
Alright, the results are at https://github.com/retorquere/translators/tree/eslint-split-merge

It looks a little baroque, I know, but the current implementation achieves a few goals I think are interesting:

* First of all, it allows `--fix`
* It does this while retaining the processor. Retaining the processor means that vscode et al can load the eslint config -- otherwise they would error out on the header.
* No temporary files
* Monkey-patches `notice/notice` so it can fix a missing license comment

Fix inside editors is not (currently) possible because that they must load the plugin processor.

retorquere:
> Are we though?

I had forgotten that there are more environments than just the Zotero client that load the translators.

retorquere:
Are the test cases at the end required to be valid JSON or just valid JS?

dstillman:
> Are the test cases at the end required to be valid JSON or just valid JS?

JSON, [from the looks of it](https://github.com/zotero/zotero/blob/ad27e0c5faed8eaa939ab6e40dfce95c1eefb2ad/chrome/content/zotero/tools/testTranslators/translatorTester.js#L221-L228), which doesn't really make sense, but they're only ever generated by Scaffold anyway.

retorquere:
That's ok but then the fixer workaround needs to protect them. I'll add that.

retorquere:
It's not that bad an idea -- JSON can be parsed safely using JSON.parse. Otherwise you'd have to get fancy with either sandboxes or esprima or whatnot.

I've merged the changes into https://github.com/zotero/translators/pull/1895, and added two test for testCases. Fix now works without touching the header or the testCases.

retorquere:
I have some time available this week and I'd love to plow through the translators based on this. I've done some test and after fix (which I did mostly manually so far), the linter reports quite a few genuine errors.

adam3smith:
My weekend looks (reasonably) good, so I can help speed-review & merge.

retorquere:
I take it it is still best if I do one pr per translator? Or do you want them batched at any size?

adam3smith:
Unless it's much more work, one per translator is easiest in case there are blockers/questions, so if it doesn't make a difference to you, do that. If you find it takes extra time, batches of 5 are fine, too. And thanks!

retorquere:
Nope, I have a few scripts to manage this process, one per is no problem whatsoever.

retorquere:
What about the existing PRs? Squash or submit new ones?

adam3smith:
squash if you don't mind.

retorquere:
No problem. I know what to do now.

retorquere:
Oh if I don't have to... that saves me updating my scripts.

retorquere:
So should I delete the existing PRs then?

.eslintrc
```diff
@@ -14,20 +14,347 @@
...
+		"no-empty": "off",
+		"no-empty-function": "off",
+		"no-eq-null": "error",
+		"no-eval": "error",
+		"no-extend-native": "error",
+		"no-extra-bind": "error",
+		"no-extra-label": "error",
+		"no-extra-parens": "off",
+		"no-floating-decimal": "error",
+		"no-implicit-globals": "error",
```
retorquere: Any top-level var or function (such as `doWeb`) will trigger this rule.

retorquere: Maybe we should default to sourceType `module` rather than `script`

dstillman: I've removed `no-implicit-globals`.

PR #1895, opened 2019-02-26
test lastUpdated against master

retorquere:


retorquere:
So what this does:

`git diff --name-status ${master}` will fetch the names of the files that have changed against `${master}`, where `${master}` is `master` if you're in CI, `upstream/master` otherwise. This does assume that if you're doing local development you have a remote called `upstream` and that you've fetched it.

It then does (as https://github.com/zotero/translators/pull/1821 does) a `git grep '"lastUpdated"' ${master} *.js` to get the `lastUpdated` values from the `${master}` branch. For files that are deemed changed, the lastUpdated is remembered.

Finally, in the 'last-updated' rule, the lastUpdated value of a translator is checked against this value. This extra check will thus effectively only be ran for files deemed changed.

It does not run this extra test if you're actually on `${master}` because that would always fail. I don't yet have a solution for development that happens on the `master` branch (but then that's bad form under git anyway, right?)

.ci/eslint-plugin-zotero-translator/index.js
```diff
@@ -6,68 +6,93 @@
...
 const deleted = new Set(
-	fs.readFileSync('deleted.txt', 'utf-8')
+	fs.readFileSync(path.join(repo, 'deleted.txt'), 'utf-8')
 		.split('\n')
 		.map(line => line.split(' ')[0])
 		.filter(id => id && id.indexOf('-') > 0)
 );
 
+const branch = exec('git rev-parse --abbrev-ref HEAD').trim();
+// branch to compare lastUpdated against -- assume that if you're not in CI, you have upstream/master
```
dstillman: That's not a safe assumption. A few people have write access to zotero/translators, and they might have it as origin/master. Would be better to check the actual URL from `git remote -v`.

.ci/eslint-plugin-zotero-translator/index.js
```diff
@@ -6,68 +6,93 @@
...
+		.map(changed => changed.match(/^M\t([^\/]+\.js)$/))
+		.filter(changed => changed)
+		.map(changed => changed[1]);
+
+	for (const lu of child_process.execSync(`git grep '"lastUpdated"' ${master} *.js`, { cwd: repo, encoding: 'utf8' }).split('\n')) {
+		const m = lu.match(/^[a-z\/]+:([^:]+):\s*"lastUpdated"\s*:\s*"([-0-9: ]+)"/);
+		if (m && updated.includes(m[1])) lastUpdated[m[1]] = m[2];
+	}
+}
+
```
dstillman: Can you comment this a bit so it's clearer what's happening?

.ci/eslint-plugin-zotero-translator/index.js
```diff
@@ -79,21 +104,17 @@ module.exports.rules = {
...
 					context.report(node, 'Header needs lastUpdated field');
 
-				} /* else {
-					// disabled until I figure out something smart -- git doesn't retain file modification dates, and it's too involved to hit the github API all the time
-					const stats = fs.statSync(filename)
-					if (stats.mtime > (new Date(header.parsed.lastUpdated))) {
-						context.report(node, 'lastUpdated field is older than file modification time');
-					}
+				} else if (lastUpdated[basename] && lastUpdated[basename] >= header.parsed.lastUpdated) {
+					context.report(node, `lastUpdated field must be updated to be > ${lastUpdated[basename]}`);
```
dstillman: Is this an error or a warning? Should be the latter. Let's also say `lastUpdated field must be updated to > ${lastUpdated[basename]} to push to clients` to make it clear when it's actually necessary.

.ci/eslint-plugin-zotero-translator/index.js
dstillman: Ah, I see you followed up with an explanation. Would be good to just add most of that to a comment so it's available in the code.

.ci/eslint-plugin-zotero-translator/index.js
retorquere: So would it be OK to test `git remote -v` for `upstream/master` and assume `master` if it's not present? That would test against `master` for people who have `origin/master` access.

dstillman: Yeah, I think that'd be fine.

retorquere: Done.

.ci/eslint-plugin-zotero-translator/index.js
retorquere: I've added it as a comment above that block.

.ci/eslint-plugin-zotero-translator/index.js
retorquere: This code just marks its presence, the `.eslintrc` file determines whether it's an error or a warning (or handled at all). I've added the extra text.

dstillman: Ah, OK. So we should make this a warning in .eslintrc. (Technically this is also checking whether there's a `lastUpdated` at all, which should be an error, but I'm not really concerned about that.)

retorquere: Right. I could split this into two rules easily.

dstillman: I don't think that's necessary. There's a lot that could be missing in the metadata header to break things, but it's just not an issue that comes up (since the header is almost always generated by Scaffold anyway).

retorquere:
Can someone walk me through a rebase again? I just tried but that screwed things up -- good thing there's `git rebase --abort`.

dstillman:
Did you figure out the rebase or do you still need help with that?

retorquere:
I'd really appreciate help. I tried `git rebase -i 03bf1dc` (as far as I can tell, that's latest master here), kept the top line as `pick` and set the rest to `squash`, but I got merge conflicts when I did that.

dstillman:
Yeah, you got yourself into a bit of a pickle by merging another branch, which itself was based on the root, into this branch. That prevents rebase from working properly. Keeping branches separate is the easiest way to avoid this, but if you really want to merge another branch into a topic branch, you'd want to rebase that on the topic branch first so that it's a fast-forward merge, and you can use `git merge --ff-only` to make sure you're not creating a merge commit.

In this case, the easiest way out, after aborting the rebase, is to just do `git reset 03bf1dc4e` and then `git add` or `git add -p` to create one or more new commits. That's equivalent to squashing but will get rid of all the merge-branch messiness.

dstillman:
(Both `git log` and the commits view here lie about what's really happening, but if you're curious, you can use `git log --graph` to visualize the problem.)

retorquere:
When I do that, I see two branch appearing at the point where I branched, but also a merge between the two? For some reason I can never get my head around rebasing -- TBH on many occasions I've just taken a diff between `master` and my `HEAD`, abandoned the branch I had, and re-applied the diff.

Am I doing essentially the same by doing `git add` or `git add -p`? Branch & merge I get, but bloody hell this rebase stuff...

dstillman:
This situation is very rare — I've never ended up in it myself, despite using rebasing all the time — so I wouldn't worry too much about it. If you just follow the advice I give about merging topic branches (or not), you won't run into this again.

> TBH on many occasions I've just taken a diff between master and my HEAD, abandoned the branch I had, and re-applied the diff.

That's essentially what you're doing with the `git reset` I describe. You're keeping the branch to keep the PR open, but you're throwing out all the existing commits and just keeping the diff.

> Am I doing essentially the same by doing `git add` or `git add -p`?

For that I just mean that you can use `git add -p` if you want to selectively add things to separate commits. It's not related to any of the rebase stuff — just a general super-useful git command for creating granular commits. If you just want all of this in a single commit, you can use `git add`.

retorquere:
OK, so if I want to be safe, I can do this:

1. `git checkout -b safe-keeping` of my current local branch
1. `git diff upstream/master > eslint.patch`
1. `git checkout eslint-last-updated`
1. `git reset 03bf1dc4e`
1. `patch -p0 < eslint.patch`

And that would give me my current state, ready to be added, committed, and force-pushed. Yes?

I'm still a little confused about why cross-branch merging would get me into this mess... I use cross-branch merging all the time with seemingly no ill effects.

retorquere:
Or would the reset just immediately bring me to step 4?

dstillman:
Yeah, you don't need to do any of that — the reset alone will do it. Reset without any flags just moves HEAD to the specified commit without touching the code, so it will just be like you took a diff and applied it to 03bf1dc4e without adding or committing anything.

retorquere:
I can't believe it -- that seems to have worked! I will have to remember this trick! I've added a little more explanation of what all this does in the new commit message -- we need a processor to to make our processors valid JS, but if any one plugin claims to process `.js`, all fixing is silently disabled by eslint. The current setup allows live linting by editors that can load eslint (most likely `vscode` and `atom` at this time), but allows `--fix` to do it's magic when ran from the command line.

retorquere:
(BTW current `eslint` mores strongly suggest breaking out rules and processors to separate files, one per)

dstillman:
That's fine — we can break them out again if you want. I have a bit better sense of what's going on now.

retorquere:
You want me to do that beforehand?

dstillman:
Sure.

retorquere:
Done.

dstillman:
(Sorry, there's a conflict in .eslintrc after I merged the client rules. You can keep the master changes for the main ruleset and then just reapply the zotero-translator and notice/notice rule changes from this.)

dstillman:
You wanted to rebase that, not merge it.

.ci/eslint-plugin-zotero-translator/lib/translators.js
```diff
@@ -0,0 +1,213 @@
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+const jsonParse = require('jsonify').parse;
```
dstillman: Can't we just use `JSON.parse()`?

.editorconfig
```diff
@@ -0,0 +1,8 @@
+root = true
+
+[*]
+indent_style = tab
+charset = utf-8
+trim_trailing_whitespace = true
+end_of_line = lf
+insert_final_newline = false
```
dstillman: This should be true.

.eslintrc
```diff
@@ -216,6 +216,7 @@
 		"no-self-compare": "error",
 		"no-sequences": "error",
 		"no-shadow": "off",
+		"no-undef": "off",
```
dstillman: Why did you turn this off?

.ci/eslint-plugin-zotero-translator/lib/rules/last-updated.js
```diff
@@ -0,0 +1,53 @@
...
+				if (!header.properties.lastUpdated) {
+					context.report({
+						node: header.declaration,
+						message: 'Header needs lastUpdated field',
+						fix: function (fixer) {
+							const comma = (Object.keys(header.properties).length) ? ',' : '';
+							const sourceCode = context.getSourceCode();
+							return fixer.insertTextBefore(sourceCode.getLastToken(header.body, t => t.value === '}'), `${comma}\n\t"lastUpdated": "${updated}"\n`);
+						}
+					});
```
dstillman: No need to fix this automatically. If a translator is missing a lastUpdated field, something is seriously wrong and we should just throw an error.

.ci/eslint-plugin-zotero-translator/lib/rules/no-for-each.js
```diff
@@ -0,0 +1,32 @@
...
+			category: 'Possible Errors',
+		},
+	},
+
+	create: function (context) {
+		return {
+			Program: function (node) {
+				let lineno = 0;
+				let m;
+				for (const line of context.getSourceCode().getText().split('\n')) { // eslint-disable-line newline-per-chained-call
```
dstillman: This comment may no longer be necessary after I adjusted the ignore depth for that rule.

retorquere:
Done (I think)

.ci/eslint-plugin-zotero-translator/lib/rules/test-cases.js
```diff
@@ -0,0 +1,80 @@
...
+		},
+	},
+
+	create: function (context) {
+		return {
+			Program: function (node) {
+				const translator = translators.get(context.getFilename());
+
+				const declaration = node.body.find(node => node.type === 'VariableDeclaration' && node.declarations.length === 1 && node.declarations[0].id.name === 'testCases');
+				const testCases = declaration && declaration.declarations[0].init && declaration.declarations[0].init.type === 'ArrayExpression' ? declaration.declarations[0].init.elements : [];
```
dstillman: Split up lines

e.g.,

```
const testCases = declaration
    && declaration.declarations[0].init
    && declaration.declarations[0].init.type === 'ArrayExpression'
        ? declaration.declarations[0].init.elements
        : [];
```

.ci/eslint-plugin-zotero-translator/lib/rules/test-cases.js
```diff
@@ -0,0 +1,80 @@
...
+								message: `${prefix} of type "${testCase.type}" has search term '${term}', expected one of ${expected.join(', ')}`,
+								loc,
+							});
+						}
+					}
+				}
+			}
+		};
+	},
+};
```
dstillman: Are there any existing tests that actually fail these rules?

.ci/eslint-plugin-zotero-translator/lib/rules/translator-id.js
```diff
@@ -0,0 +1,76 @@
...
+
+				if (!header.properties.translatorID) {
+					context.report({
+						node: header.declaration,
+						message: 'Header has no translator ID',
+						fix: function (fixer) {
+							const comma = (Object.keys(header.properties).length) ? ',' : '';
+							const sourceCode = context.getSourceCode();
+							return fixer.insertTextAfter(sourceCode.getFirstToken(header.body, t => t.value === '{'), `\n\t"translatorID": "${uuid()}"${comma}`);
+						}
```
dstillman: Same as `lastUpdated` — we shouldn't auto-fix this.

.ci/eslint-plugin-zotero-translator/lib/rules/translator-id.js
```diff
@@ -0,0 +1,76 @@
...
+					});
+				}
+				else if (!header.properties.translatorID.value) {
+					context.report({
+						node: header.properties.translatorID,
+						message: 'Header has empty translator ID',
+						fix: function (fixer) {
+							return fixer.replaceText(header.properties.translatorID, `"${uuid()}"`);
+						}
+					});
```
dstillman: Same — remove auto-fix.

.ci/eslint-plugin-zotero-translator/lib/rules/translator-id.js
```diff
@@ -0,0 +1,76 @@
...
+				}
+				else {
+					const conflict = translators.conflicts(filename, header.properties.translatorID.value);
+					if (conflict) {
+						const translator = translators.get(filename);
+						context.report({
+							node: header.properties.translatorID,
+							message: `Header re-uses translator ID of ${conflict.label}`,
+							fix: !translator.modified ? undefined : (fixer => fixer.replaceText(header.properties.translatorID, `"${uuid()}"`)),
```
dstillman: I guess this one we could auto-fix.

.ci/eslint-plugin-zotero-translator/lib/translators.js
```diff
@@ -0,0 +1,213 @@
...
+			// `git diff --name-status ${master}` will fetch the names of the files that have changed against `${master}`
+			for (const file of exec(`git diff --name-status ${master}`).split('\n')) {
+				const m = file.match(/^M\t([^\/]+\.js)$/); // js files that are modified but don't have a / in their path to pick out files in the root
+				if (m && this.decorated[m[1]]) this.decorated[m[1]].modified = true;
+			}
+
+			/*
+				We do a `git grep '"lastUpdated"' ${master} *.js` to get the
+				`lastUpdated` values from the `${master}` branch. For files that are
+				deemed changed, the lastUpdated is remembered (so the precense of lastUpdated implies modified).
```
dstillman: Typo: `presence`

.ci/eslint-plugin-zotero-translator/lib/translators.js
retorquere: We can, but we'd lose the position of the error.

dstillman: That's fine. I'd rather remove the dependency.

.eslintrc
retorquere: I just wanted to make it explicit -- it's off by default when missing. I'd personally prefer it to be at least a warning.

dstillman: It's not off — I [added](https://github.com/zotero/translators/commit/b3c0346183e893a88351d301a4752f9f5ec66daf) `eslint:recommended`, which enables this. The client rules assumed that was set, but I forgot to copy it over.

dstillman: (This might explain some of your confusion about the client rules.)

.ci/eslint-plugin-zotero-translator/lib/translators.js
retorquere: Done

.eslintrc
retorquere: my bad. Removed.

.ci/eslint-plugin-zotero-translator/lib/rules/no-for-each.js
retorquere: Correct

.ci/eslint-plugin-zotero-translator/lib/rules/test-cases.js
retorquere: If I indent that way I run into `Expected indentation of 5 tabs but found 6`

.ci/eslint-plugin-zotero-translator/lib/rules/test-cases.js
retorquere: Only one, and only the basic test for type.

.ci/eslint-plugin-zotero-translator/lib/rules/translator-id.js
retorquere: Done.

.ci/eslint-plugin-zotero-translator/lib/rules/translator-id.js
retorquere: Done

.ci/eslint-plugin-zotero-translator/lib/translators.js
retorquere: Done

.ci/eslint-plugin-zotero-translator/lib/rules/last-updated.js
retorquere: Done

.editorconfig
retorquere: Done.

retorquere:
> I'm getting `teslint: command not found` when I try to run this.

Can you try removing your node_modules directory and re-running `npm i`?

This is one of the downsides of using `file:<dir>` to "install" the package. It actually just creates a symlink in your node_modules and doesn't re-run the install phase when it finds the symlink present; bumping the package version of eslint-plugin-zotero-translator doesn't change this. I moved `teslint.js` into `bin`, but for that to be symlinked into `node_modules/.bin`, the install phase needs to be run. As long as no `bin` files are added/removed/moved, this will work. The only fix for this is 

1. Move `teslint.js` into `.ci` and explicitly call it as `.ci/teslint.js` (or leaving it where it is explicitly call it as `.ci/eslint-plugin-zotero-translator/bin/teslint.js`)
2. Make eslint-plugin-zotero-translator a real npm-published package.

retorquere:
I have all comments addressed except the two I left unresolved.

.ci/eslint-plugin-zotero-translator/lib/rules/test-cases.js
dstillman: We can look into that later. That style should be allowed, and is preferable to a very long line.

.ci/eslint-plugin-zotero-translator/lib/rules/test-cases.js
dstillman: That's what I figured. Tests are all generated by Scaffold, never by hand, so there's really no reason they'd be invalid. But I guess there's not much harm to having the checks other than some extra code.

.ci/eslint-plugin-zotero-translator/lib/rules/test-cases.js
retorquere: So just to make sure: leave indented as above, and leave the two lint errors in place?

dstillman:
(It's better not to force-push to a PR when you can avoid it. It's OK if you're rebasing to resolve a conflict, as before, but otherwise it's better to just to add new commits, and we can squash on merge.)

retorquere:
> (It's better not to force-push to a PR when you can avoid it. It's OK if you're rebasing to resolve a conflict, as before, but otherwise it's better to just to add new commits, and we can squash on merge.)

I much prefer this in fact, but I took

> You wanted to rebase that, not merge it.

to mean I should do the reset & force-push dance again.

dstillman:
Right, that one was fine, but then you force-pushed again after my most recent comments, no? There's only a single commit now.

retorquere:
AH, yes -- I thought that was the preferred mode, to leave only a single commit to merge. As said, I much prefer to just add commits -- I have to review what I add every time I do a reset.

dstillman:
It used to be, after a review, but GitHub added the ability to squash a while back, so now it's no longer necessary.

dstillman:
Is `eslint .` alone meant to work? I'm getting `ESLint couldn't find the plugin "eslint-plugin-zotero-translator"`. Without that working, this isn't usable from editors, right?

dstillman:
Wait, I definitely had this working previously after I moved package.json to the root. If it's working for you, I can investigate.

dstillman:
So, I keep `node_modules/.bin` in my path to allow package binaries to work. (I think there's probably a better way, but that's what I do.) `which eslint` shows `node_modules/.bin/eslint`.

If I run `eslint .`, I get the plugin error.

If I run `node_modules/.bin/eslint`, I get `Error while loading rule 'notice/notice': copyright is not defined`.

dstillman:
Sorry, scratch that. After opening a new terminal window, I'm getting the same behavior for both:

`Error while loading rule 'notice/notice': copyright is not defined`

retorquere:
I have a fix for that for .eslintrc here locally:

```
+                       "templateVars": {
+                               "copyright": "<undefined>",
+                               "creator": "<undefined>"
+                       },
```

retorquere:
(having `node_modules/.bin` in the path is essentially what `npm run` does)

.ci/eslint-plugin-zotero-translator/lib/rules/test-cases.js
dstillman: Yes please.

retorquere: Done.

dstillman:
Sorry, where does templateVars go? Can we add that so it's available by default?

retorquere:
I've just pushed all changes onto this PR -- it's available by default as options to the `notice/notice` rule, the values make no sense but will never be applied since the fixer can't run from a regular `eslint` call, and `teslint` will override them with meaningful values.

dstillman:
Only 129,589 problems!

So if I run `eslint --fix Amazon.js`, it doesn't seem to actually be running `--fix`. It just tells me that I can potentially fix things with `--fix`, and it doesn't change the file.

Perhaps I'm meant to run `npm run fix`, but that looks like it runs on `.`, not the specified file?

retorquere:
That's the thing -- eslint silently ignores `--fix` when a processor is active. You can run on a single file as either `npm run lint -- --fix Amazon.js` or `teslint --fix Amazon.js` (assuming you have `node_modules/.bin` in your path).

retorquere:
(♫ I've got 129,589 problems but `eslint` won't fix one ♫)

dstillman:
Either way, I get this:

```
/Users/dan/zotero-translators/.ci/eslint-plugin-zotero-translator/lib/translators.js:118
	if (!m) throw new Error('not decorated');
	        ^

Error: not decorated
    at Object.strip (/Users/dan/zotero-translators/.ci/eslint-plugin-zotero-translator/lib/translators.js:118:16)
    at Object.<anonymous> (/Users/dan/zotero-translators/.ci/eslint-plugin-zotero-translator/bin/teslint.js:116:69)
    at Module._compile (internal/modules/cjs/loader.js:738:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:749:10)
    at Module.load (internal/modules/cjs/loader.js:630:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:570:12)
    at Function.Module._load (internal/modules/cjs/loader.js:562:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:801:12)
    at internal/main/run_main_module.js:21:11
```

retorquere:
Both `eslint` and `teslint` are shallow wrappers around the eslint `CLIEngine`, but `teslint` disables the processor and decorated/strips around calls to the engine -- essentially doing what `eslint` does, but since I know that the results of decorate/strip are safe to write back, I can actually do `--fix`.

retorquere:
You get that running `teslint --fix Amazon.js`? Did you perchance use `--dump-decorated` into the root of the repo?

retorquere:
No wait that'd have `decorate` complaining, not `strip`.

dstillman:
> You get that running teslint --fix Amazon.js? 

Yes.

retorquere:
Drat -- one of the new rules added in the new .eslintrc is removing the spurious `;` I add in the decoration. I need to disable that rule during decorating, but I first need to find which it is (in addition to `no-empty`. Suggestions welcome.

retorquere:
Try the latest push. Protecting the header from eslint is tricky because `/* eslint-disable */ /* eslint-enable indent */` does not do what you'd (or at least I'd) expect; it leaves eslint entirely disabled, which means that no linting (or fixing) is applied in the header. eslint is considering an RFC to change this behavior, but in the interim I have to apply some placement trickery to make sure the enable/disable applies at the right places.

retorquere:
BTW one of the fixes applied to `Amazon.js` leaves it with

```
function attr(docOrElem, selector, attr, index) {
       var elem = index ? docOrElem.querySelectorAll(selector).item(index) : docOrElem.querySelector(selector); return elem ? elem.getAttribute(attr) : null;
} function text(docOrElem, selector, index) {
       var elem = index ? docOrElem.querySelectorAll(selector).item(index) : docOrElem.querySelector(selector); return elem ? elem.textContent : null;
}
```

which still seems pretty strange to me.

retorquere:
Oh BTW it's possible to have `teslint` report the total number of errors/warnings as `eslint` does rather than per-translator -- currently it reports per-translator as that gives feedback during run, rather than the `eslint` behavior of collecting all reports silently and reporting all at once.

retorquere:
It looks like `"no-implicit-globals": "error"` picks out all global vars & functions (such as `doWeb`) -- should I set it to `"off"`?

.ci/eslint-plugin-zotero-translator/index.js
retorquere: BTW I am probably an outlier but I've never used Scaffold.

retorquere:
I still see this marked as "changes requested", but I can't find which changes those are.

dstillman:
OK, `teslint --fix Amazon.js` is working now — thanks.

1. The `attr`/`text` line shouldn't be flagged or touched. We could tell people to add `// eslint-disable-line` to those, but I think we'll be able to remove them altogether soon, so it's probably enough to tell people to ignore those errors (and fixes) for now. But if there's a trivial way to skip that line during fixing, we could do that.
2. I'm inclined to just remove the "use strict" rule. The important things, like global assignment, are checked by ESLint anyway, and adding `"use strict";` to every file is annoying.
3. The character position (`1:37`) for `notice/notice` (and `strict`) is incorrect, which is logging an "Invalid position given by notice/notice" error in Atom. Any way to force that to be `1:1`?
4. I removed `no-implicit-globals` in master a while back.
5. teslint isn't returning any warnings, just errors. Is that intentional? Is there meant to be any difference between `eslint` and `teslint` in non-`--fix` mode?

retorquere:
1. I can't find what this refers to, sorry.
2. That works for me, and would simplify the decorator.
3. Aw crap, that's another side effect of the current way of decorating. I use a block comment, and notice/notice treats that like the license block. I'll have a look at that.
4. OK, I've merged that change in.
5. There isn't meant to be, the eslint function to get the error results modifies the results to strip out non-errors. That's fixed and pushed.
6. Fixed and pushed.

retorquere:
point 3. is fixed and pushed.

retorquere:
So that leaves point 2. If the directive in `.eslintrc` can be changed to `"strict": [ "error", "never" ],`, I can remove the `use strict` hoisting I do now.

retorquere:
(I don't know if there's a way at all to disable fixing of 1. if `--fix` is specified -- that's handled by a standard eslint rule. AFAIK, the only way to prevent that fix is to either tweak what the rule considers valid layout, or to add eslint disable directives there)

dstillman:
I've removed the strict rule on master.

retorquere:
I've merged the change on master and removed the hoisting.

.ci/eslint-plugin-zotero-translator/lib/translators.js
```diff
@@ -84,6 +84,30 @@ function decorate(source) {
...
+		'no-unused-expressions',
+		'no-void',
+		'object-curly-spacing',
+		'semi',
+		'semi-spacing',
+		'space-before-blocks',
+		'space-before-function-paren',
+		'space-infix-ops',
+	];
+	code = code.replace(/(\n\/\* FW LINE 59:b820c6d \*\/[^\n]+)(\n)/, `$1 // eslint-disable-line ${tfw.join(', ')}$2`);
```
dstillman: Can't it just be `// eslint-disable-line`? We don't need to check for anything in these blocks.

.ci/eslint-plugin-zotero-translator/lib/rules/translator-framework.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+
+const translators = require('../translators');
+
+module.exports = {
+	meta: {
+		type: 'problem',
+		fixable: 'code',
+		docs: {
+			description: 'checks for AGPL license',
```
dstillman: Is this the wrong description?

dstillman:
`zotero-translator/last-updated` (if that's the one that checks whether lastUpdated has changed) should be a warning, not an error.

.ci/eslint-plugin-zotero-translator/lib/translators.js
retorquere: That's possible, but then we also lose the opportunity to report it's being used.

.ci/eslint-plugin-zotero-translator/lib/rules/translator-framework.js
retorquere: Most definitely.

retorquere:
last-updated is now a warning

dstillman:
Thanks so much for all your work on this!

retorquere:
Glad to be of help.

PR #1897, opened 2019-03-04
[DO NOT MERGE] Improved generic translation

mrtcode:


Generic.js
```diff
@@ -0,0 +1,499 @@
...
+	}
+}
+
+var exports = {
+	"doWeb": doWeb,
+	"detectWeb": detectWeb
+}
+
+/** BEGIN TEST CASES **/
+var testCases = [];
```
mvolz: It's be helpful to have test cases here.

zuphilip:
What is the status/plan for this PR which relies on the new functions for JSON-LD in `ZU` https://github.com/zotero/zotero/pull/1656 ? CC @mrtcode @dstillman

mvolz:
> OK, this is working well enough to test with [zotero/zotero#2894](https://github.com/zotero/zotero/pull/2894) and [zotero/utilities#20](https://github.com/zotero/utilities/pull/20) checked out.

Great! Adding json-ld seems it would help a lot, see this example where the json-ld was much better than using the byline: https://phabricator.wikimedia.org/T432076

PR #1908, opened 2019-03-12
Update GaleGDC.js

jimmiazek:
New Authorized Gale Cengage Zotero Support File.

Please review and update this file.  This update authorizes support for gale.com, galegroup.com, and supported gale development servers (ggtest.com).

Jim Miazek
Software Engineer IV
Gale Cengage Learning
jim.miazek@cengage.com

### DIFF
--- a/GaleGDC.js
+++ b/GaleGDC.js
@@ -1,19 +1,19 @@
 {
 	"translatorID": "04e63564-b92b-41cd-a9d5-366a02056d10",
 	"label": "GaleGDC",
-	"creator": "GaleGDC",
-	"target": "/gdc/ncco|/gdc/xsearch|/gdc/artemis",
+	"creator": "Jim Miazek (Gale Cengage), Rui Wang (Gale Cengage), Gale Online Product Development",
+	"target": "gale.com|galegroup.com|ggtest.com",
 	"minVersion": "3.0",
 	"maxVersion": "",
-	"priority": 270,
+	"priority": 300,
 	"inRepository": true,
 	"translatorType": 4,
-	"browserSupport": "gcsb",
-	"lastUpdated": "2014-08-26 03:45:45"
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2019-03-12 12:00:00"
 }
 
 /*
- * Gale GDC Copyright (C) 2011 Gale GDC
+ * Gale GDC Copyright (C) 2019 Gale GDC
  *
  * This program is free software: you can redistribute it and/or modify it under the terms of the
  * GNU General Public License as published by the Free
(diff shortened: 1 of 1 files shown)


adam3smith:
Could you specify for which Gale databases this will work and/or has been tested? We're seeing very different page structures for different Gale databases and have two well-working translators covering most (all?) of those except GDC (which the previous version of this one covered.)

jimmiazek:
Hi Sebastian,

My name is Jim Miazek and I am a Software Engineer at Cengage Learning. My team supports the development for all Gale Databases.

There are 3 connectors for Gale Databases.  Our team supports GaleGDC.js and it supports 30+ archive products.  We plan to extend that support for over 100 more. The problem that we are running into is that there are 2 other web translators (Galegroup.js and Gale Databases.js) that are not supported by us.  This is important because the documents from these translators are not correct and are attributed to Google Scholar links.

Instead we will support with *auth-free* links for our customers using Zotero.  This will be a better experience for our customers ensuring that the provided links will always work for them.

Our code is test protected and is under 100% code coverage – note, we don’t publish our Spec.  Also, we follow our eslint rules before publishing our code.

We ask that the “Galegroup.js” and “Gale Databases.js” be removed.  The links published by Zotero from these two translators are unauthorized for use.

Sincerely,

Jim Miazek
Software Engineer IV
Jim.miazek@cengage.com<mailto:Jim.miazek@cengage.com>
Gale, A Cengage Company
27500 Drake Road, Farmington Hills, Michigan 48331

jimmiazek:
So what next?  Your eslintrc rules are different.  No problem, we can change.  But most of these errors have to do with the JSON object that is declared before the code.  

What do we do now?

adam3smith:
don't worry about the automated errors for now. I'll need to review the code & performance first

jimmiazek:
As mentioned earlier, we would like the other translators taken down.  The links used are not correct and need to be customized to our user's login.   

Our products will be under going a migration to new platforms with different domain including in some cases, new path  and URI.  We would like to ensure that the experience will work for all of our products, not just a certain few.  That is why, as a company, we intend to support Zotero web translation, ourselves.  

If you wish to speak in conference then please contact me at jim.miazek@cengage.com.

adam3smith:
I'm absolutely happy to take the existing translators down provided yours produce equivalent or higher-quality import, yes. (Not sure what links you're referring to, though? Just the test cases?)

jimmiazek:
The problem is that the targets (in these translators) from these other connecters are matching our production servers which we host 170 database applications.  The URLs used and are saved in Zotero are incorrect -- They contain "&sid=googlescholar" in the URL.  These links need to be removed ASAP!  Therefore these translators need to be removed ASAP!

adam3smith:
I assume you don't blindly accept other people's code into your production platform and neither does Zotero. We'll review this as speedily as possible, but we also won't rush code review cc @dstillman . 

As I say above, I'll look at your code and see how it performs and am happy to accept it provided it provides the same quality and coverage of import.

If there are severe problems with Zotero's Gale import, we will obviously act with the appropriate speed on those, but I'm not seeing that. The only links that Zotero saves to the URL field are taken from the RIS files provided by Gale (so what users would download when using citation tools --> Export to --> Reference Manager) and they do not contain `&sid=googlescholar -- I've just tested this to verify and this makes sense as the translator is written to mimic the behavior of a user manually exporting to RIS. Could you specify where you're seeing the behavior you describe? 
They only way I can see that &sid=googlescholar would make it into Zotero is if people download Gale citations directly from google scholar, not accessing your database at all, and in that case, of course, your translators wouldn't change that.

adam3smith:
Hi @jimmiazek I've started testing this, but so far with very little success. I started by testing simple individual pages from two popular Gale databases, Academic One File and Central Reference Center GOLD, that I have access to (through Syracuse University in case that matters). For both of these (they're "Omni" databases in your terminology I believe), `detectWeb` returns false for single items and when I run `doWeb` I get
`TypeError: doc.getElementById(...) is null`
referring to this line:
`var citationData = doc.getElementById('dviTools').querySelector('input[class="citationToolsData"]');`
and I do in fact not see dviTools in an id (or anywhere else) on those pages. 

Detect and import works fine with the current Gale Database translator, FWIW. You mention above that your translator works for a subset of databases, could you point me to those? Or how would you like to proceed? 
Thanks!

jimmiazek:
Please disregard this pull request.   We decided to keep this translator as is to support Gale DVI legacy products.  A future translator update will be submitted with our next production release of our products.
Thank you.

PR #1911, opened 2019-03-26
Re-write ACL translator for new site

GuyAglionby:
As in title. Includes conference name change from #1797. Makes use of hidden preferences to decide whether to save supplementary content. Wasn't sure the best way of determining mimetype, but thought this seemed reasonable given how nicely structured the Anthology data is.

Thanks for any comments!


Evpok:
So, given this, what should I configure to get the supplementary materials ?

GuyAglionby:
You'll need to enable the `translators.attachSupplementary` hidden property in the connector in your browser. There are instructions for that [here](https://www.zotero.org/support/preferences/hidden_preferences#zotero_connector)

GuyAglionby:
@adam3smith when you have time, would it be possible to have a look to see if any changes are needed here please? #1702 has some suggestions for sentence-casing the title but unfortunately I don't have a lot of time at the moment to implement this -- could it be integrated at a later date? Thanks!

ACLWeb.js
```diff
@@ -34,175 +34,456 @@
...
 		return 'multiple';
 	}
+	return false;
 }
 
 function doWeb(doc, url) {
 	if (detectWeb(doc, url) === 'multiple') {
-		Zotero.selectItems(extractFullProceedings(doc), function (selected) {
+		Zotero.selectItems(getSearchResults(doc, url), function (selected) {
 			if (!selected) {
```
adam3smith: @zuphilip -- I'm understanding correctly that this simply isn't needed (anymore) and can be removed together with the `return true; inserted further down to fix linting?

adam3smith:
I have one [more general question](https://github.com/zotero/translators/pull/1911/files#r294110559) on this, but otherwise this looks great. Thanks for adding the supplementary info.

ACLWeb.js
dstillman: As far as I know, the `selectItems` callback doesn't expect a return value, so `return true` shouldn't be needed.

And I would hope that the argument to the callback is always an object, such that `if (!selected)` isn't necessary. If that's not the case we should fix that.

GuyAglionby: If nothing is checked in the selection dialog, the callback in `selectItems` is called with `null`, so `Object.keys` gives a `TypeError: can't convert null to object`. 

I have cleaned up that code wrt. returns. The build on this new commit failed because of linting problems with other translators, not sure why it ran those.

dstillman: Oh, that's a shame. We should fix that and make a note of what version has the fix so we can do a cut-off down the line and stop including the extra lines. I've created an [issue](https://github.com/zotero/zotero/issues/1705). Thanks.

nschneid:
Is this ready to merge? Travis shows just a few style warnings:

```
/home/travis/build/zotero/translators/ACLWeb.js
   67:2  warning  Closing curly brace appears on the same line as the subsequent block  brace-style
   82:2  warning  Closing curly brace appears on the same line as the subsequent block  brace-style
   86:2  warning  Closing curly brace appears on the same line as the subsequent block  brace-style
   91:2  warning  Closing curly brace appears on the same line as the subsequent block  brace-style
  153:4  warning  Closing curly brace appears on the same line as the subsequent block  brace-style
  155:4  warning  Closing curly brace appears on the same line as the subsequent block  brace-style
  157:4  warning  Closing curly brace appears on the same line as the subsequent block  brace-style
  159:4  warning  Closing curly brace appears on the same line as the subsequent block  brace-style
  161:4  warning  Closing curly brace appears on the same line as the subsequent block  brace-style
  164:3  warning  Closing curly brace appears on the same line as the subsequent block  brace-style
  167:2  warning  Closing curly brace appears on the same line as the subsequent block  brace-style
  183:2  warning  Closing curly brace appears on the same line as the subsequent block  brace-style
```

zuphilip:
I can fix the styling issues in a final step (no problem for those). - However, I have a question about extracting the venues. It seems that student papers and workshop papers etc. are returning empty string in the `getVenue` function by purpose. That doesn't seem correct for me. Maybe a workshop is a satellite event of a larger conference, but it is still part of that conference. Could we maybe instead add the conference name together with some additional information in such cases?

zuphilip:
BTW what happened to https://aclanthology.coli.uni-saarland.de/ ? We have another translator `ACLAnthology.js` for that, but maybe that is not anymore needed, because everything seems to be redirected to ACL web at the moment...

nschneid:
> Maybe a workshop is a satellite event of a larger conference, but it is still part of that conference. Could we maybe instead add the conference name together with some additional information in such cases?

We are using the conference name field for the abbreviated name for displaying a short form of the proceedings title. With workshops we want to use the full proceedings title since abbreviated names aren't as well known, so the conference name field is blank.

> BTW what happened to https://aclanthology.coli.uni-saarland.de/ ? We have another translator `ACLAnthology.js` for that, but maybe that is not anymore needed, because everything seems to be redirected to ACL web at the moment...

That was the transitional home of the Anthology during its software rewrite. It is no longer in use so the other translator can be removed.

zuphilip:
I also see that the conference name contains the year, e.g. "NAACL 2018", but not for the special cases with SEMEVAL etc. We should either do this always or never.

nschneid:
Agreed, let's always include the year.

zuphilip:
Here we go. The first 10 commits need to be squashed together. I can do that in the next days if nobody objects to the last changes or @GuyAglionby wants to rebase himself.

GuyAglionby:
Great, thanks for the feedback and your changes! Have squashed.

zuphilip:
Thank you very much!

PR #1914, opened 2019-03-28
report non-AGPL rather than saying license is missing

retorquere:


.ci/eslint-plugin-zotero-translator/lib/rules/license.js
```diff
@@ -26,20 +26,31 @@ module.exports = {
...
+						if (firstComment || comment.loc.start.line > header.followingStatement.start.line) {
 							context.report({
 								loc: comment.loc,
 								message: 'Preferred to have license block at the top'
 							});
 						}
 						return;
 					}
+
+					if (comment.value.match(/copyright/i)) copyright = comment;
```
dstillman: Braces

retorquere: Braces added, but it passed the existing eslint rules.

PR #1915, opened 2019-03-28
Runs eslint only on changed translators (closes #1900)

### LINKED ISSUE #1900: Check only modified files in CI
Unless and until we go through all translators to fix linter errors ([which I'm in no rush to do](https://github.com/zotero/translators/pull/1838#issuecomment-469978240)), we're never going to have passing tests. It would be more useful (and it would certainly be quicker) to run ESLint only against the files modified in the commit range, so that we can rely on the checks on GitHub without having to check the output.

https://github.com/zotero/translators/pull/1895 probably needs to land first.

/cc @retorquere

retorquere:


dstillman:
Can you add a descriptive commit line?

dstillman:
(It should be possible to see what a commit does without clicking through to another issue.)

PR #1916, opened 2019-03-28
Allow unused "detectImport"

retorquere:


retorquere:
Hold on, don't merge, I've found a naming issue. Will be fixed in a few.

retorquere:
OK, fixed.

.eslintrc
```diff
@@ -249,7 +249,7 @@
 			"error",
 			{
 				"argsIgnorePattern": "^_",
-				"varsIgnorePattern": "^doWeb$|^detectWeb$|^doImport$|^doExport$"
+				"varsIgnorePattern": "^detectWeb$|^doWeb$|^detectImport$|^doImport$|^doExport$"
```
dstillman: These shouldn't be here at all. The rules copied from the client will likely change further, so the whole block should stay the same.

dstillman:
But I'm not clear what this is actually doing. Can you explain?

.eslintrc
retorquere: But these exceptions are specifically for translators. Translators will always have unused functions by one of these names. All import translations have doImport, but from the POV of eslint, it's not ever called.

retorquere:
It checks consistency of what a translator declared it will do by its translatortype, and what it actually does in terms of handlers.

.eslintrc
retorquere: Rather than copying, it may be worth considering inheriting and overriding the rules.

dstillman: Ugh, I didn't realize that no-unused-vars also applies to functions (and there's no way to change that).

Is there a way to inherit rules other than [publishing it to NPM](https://eslint.org/docs/developer-guide/shareable-configs)?

retorquere: The only other way I know is the file: trick I use for the linter, but that's going to be awkward because then you will always have to checkout the translation server. Another option is to make a script that derives the one from the other and just check the results in, that way you at least don't duplicate work.

retorquere: The other way to work around this is to add a spurious call to the handlers by the decorator, but that wouldn't have my preference.

retorquere: The choice of the `no-unused-vars` rule seems to case a wider net than technically required -- `const X = function() {}` and `function X() {}` are easily distinguished in the AST. That said, for many purposes you'd want them to be treated as equivalent, such as when testing for shadowing/redefinition.

dstillman: I ended up just publishing the base config as a scoped module and inheriting it, so we can keep this as an override.

PR #1917, opened 2019-03-28
Fixes most MODS import tests

retorquere:
One still remains but I don't know MODS well enough to decide what to do; I've sent a message to Simon to ask.


dstillman:
Simon isn't involved with Zotero development anymore, so if you have a question you should ask it here and hopefully someone will know.

dstillman:
As we've discussed, lint changes should be in separate commits so they can be reviewed quickly separately from actual changes. Can you fix that and force-push?

retorquere:
So if I understand correctly - you prefer changes to translators in one PR, but one PR with two commits - one that does the lint changes, one that does functional changes?

dstillman:
That's right.

retorquere:
Alright, I didn't get that from the earlier discussion, but I'll do that Sunday.

retorquere:
OK, can you look at the lint reports first before I stack the other changes on to them?

dstillman:
It won't necessarily be me reviewing this, but there's no need to do this in stages. Git makes it easy to add changes to previous commits:

https://fle.github.io/git-tip-keep-your-branch-clean-with-fixup-and-autosquash.html

retorquere:
As much as I love git, rebase still terrifies me. If it's OK with you (or whomever might end up reviewing this), I'd prefer to wait and stack the commits using regular add.

adam3smith:
@retorquere -- I had a look and all the changes look fine. I think the remaining linting error should disappear on your next push as we added `detectImport` to the exceptions a while ago.

adam3smith:
If you have substantive changes you want to make -- I have this one that I'd want to also include: https://github.com/adam3smith/translators/commit/783ac38a77a307fdfe65c4a1b5d9e6bfda4af144

retorquere:
I was going to address the import tests in this PR, but my understanding of how they're going to be used has changed -- I thought the translation-server would be the test runner for the `testCases`, but the import tests in the translation server are directed towards validating translation-server itself (if I understand @dstillman correctly), not towards validating what's produced by the translator with the `testCases` (which is subtly but importantly different). I currently don't know how to do the latter, but any "substantive" change I was going to make would be directed at making the import tests pass.

At this stage, it may be easier to just close this PR and work from https://github.com/adam3smith/translators/commit/783ac38a77a307fdfe65c4a1b5d9e6bfda4af144

dstillman:
One way or another, please be sure to rebase to remove the merge commit before merging. Again, there should almost never be merge commits in PRs, and certainly never merges back from the master branch.

adam3smith:
Closed via #1928 

@retorquere -- I'm not sure what you were referring to with the import tests. The way we create and run them is in Scaffold, and all tests pass in the current MODS translator, so nothing to fix there, but I may be misunderstanding.

retorquere:
The misunderstanding was from me, and you clarified it earlier. I knew scaffold was *a* way to run the tests but I didn't realize it was *the* way to run the tests. I thought I was working towards a way to have the import tests be runnable automatically so that they could, in principle, be ran in CI (so as to prevent something like my gaffe with the BibTeX translator happening again), but as you pointed out earlier, that's not strictly what the tests on translation-server are for.

dstillman:
You're still confusing the translation-server tests and the translator tests. As I've noted, translator tests are [already run automatically](https://translator-tests.zotero.org/), and they [could be run](https://github.com/zotero/translators/issues/1310) via CI in this repo.

retorquere:
Which is exactly my point. I *was* confusing them, and you corrected that.

PR #1921, opened 2019-03-31
add eslint-fixer for common import-test errors

retorquere:
add eslint-fixer for common import-test errors. Many (perhaps most) existing import translators have tests that don't conform to what the translation-server import test expect.

### DIFF
--- a/.ci/eslint-plugin-zotero-translator/lib/rules/test-cases.js
+++ b/.ci/eslint-plugin-zotero-translator/lib/rules/test-cases.js
@@ -1,6 +1,95 @@
 'use strict';
 
 const translators = require('../translators').cache;
+const astring = require('astring');
+
+function verifyWeb(context, astNode, tc, prefix) {
+	if (tc.items !== 'multiple' && !Array.isArray(tc.items)) {
+		context.report(astNode, `${prefix} needs items`);
+	}
+
+	if (typeof tc.url !== 'string') {
+		context.report(astNode, `${prefix} test needs url`);
+	}
+}
+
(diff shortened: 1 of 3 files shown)


dstillman:
Can you be sure to include descriptions of what these PRs do?

From a quick look, I don't think we want any of this. Test cases aren't meant to be edited manually — they're generated manually by Scaffold, and any problems with the output needs to be addressed there.

retorquere:
That's OK by me, but so far, none of the import tests I looked at conformed to the output of `itemToAPIJSON`. Using `update` helped, but didn't fix everything.

This is 100% your call, but personally I find the trip through scaffold cumbersome. I have to copy the translator out of the git clone into my Zotero dir, run scaffold, and then copy it back so I can run the linter. Is there a way to integrate these? Can I point Zotero to a different translators dir?

dstillman:
Zotero translators don't generate API JSON — they generate a similar, older, internal JSON format. `itemToAPIJSON` converts that format into API JSON. So the test case data needs to be run through that as well.

retorquere:
I've tried that, but that means that when the test case says it generates some output (like a bookTitle in the case of MODS), the normalization through itemToAPIJSON means that the fact that that content will be discarded goes undetected, because it's scrubbed from the generated output before comparison.

retorquere:
From my POV it'd be better to have an eslint fixer drop these fields than having the test runner drop them -- at least with the fixer, a `git diff` will show that it's happening.

But, as I've said before, this is just my opinion, and it's your call.

PR #1923, opened 2019-04-01
Allow unused search functions

adam3smith:
Same logic as https://github.com/zotero/translators/pull/1916

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -23,7 +23,7 @@
 			"error",
 			{
 				"argsIgnorePattern": "^_",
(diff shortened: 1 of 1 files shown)


.eslintrc
```diff
@@ -23,7 +23,7 @@
 			"error",
 			{
 				"argsIgnorePattern": "^_",
-				"varsIgnorePattern": "^detectWeb$|^doWeb$|^detectImport$|^doImport$|^doExport$"
+				"varsIgnorePattern": "^detectWeb$|^doWeb$|^detectImport$|^doImport$|^doExport$|^doSearch$|^detectSearch$"
```
dstillman: Doesn't make a difference, but flip the order?

PR #1929, opened 2019-04-12
Update newspapers.com.js to use json string

### LINKED ISSUE #1927: newspapers.com error
Tests failing; must be a change in the source HTML. I'll look into it.

```
12:56:59 TranslatorTester: Running 1 test for newspapers.com
12:56:59 TranslatorTester: Running newspapers.com Test 1
12:56:59 TranslatorTester: Translating https://www.newspapers.com/clip/7960447/my_day_eleanor_roosevelt/
12:56:59 Translation using newspapers.com failed: 
         TypeError: text(...) is null
         
         doWeb@newspapers.com:96:17
         
         url => https://www.newspapers.com/clip/7960447/my_day_eleanor_roosevelt/
         downloadAssociatedFiles => true
         automaticSnapshots => true
12:56:59 TranslatorTester: newspapers.com Test 1: failed (Translation failed: TypeError: text(...) is null)```

### PR DESCRIPTION
**Pull request by pbinkley**
Closes #1927

### DIFF
--- a/newspapers.com.js
+++ b/newspapers.com.js
@@ -9,13 +9,13 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2019-04-12 03:06:00"
+	"lastUpdated": "2019-04-17 01:46:54"
 }
 
 /*
 	***** BEGIN LICENSE BLOCK *****
 
-	Copyright © 2017 Peter Binkley
+	Copyright © 2017-2019 Peter Binkley
 
 	This file is part of Zotero.
 
@@ -35,37 +35,31 @@
 	***** END LICENSE BLOCK *****
 */
 
-
-// attr()/text() v2
-function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null;}function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null;}
-
-
-function detectWeb(doc, url) {
+function detectWeb(_doc, _url) {
 	return "newspaperArticle";
 }
 
-function doWeb(doc, url) {
+function doWeb(doc, _url) {
 	var newItem = new Zotero.Item("newspaperArticle");
 	var scripts = doc.getElementsByTagName("script");
 	var json = '';
-	var testre = new RegExp('^(var staPageDetail).*', 'm');
-	var jsonre = /var staPageDetail = JSON.parse\(\"(.+?)\"\)\;/;
-	for (var i = 0 ; i < scripts.length ; i++) {
+	var jsonre = /var staPageDetail = JSON.parse\((.+?)\);/;
+	for (var i = 0; i < scripts.length; i++) {
 		var arr = scripts[i].textContent.match(jsonre);
 		if (arr) {
-			// we have to unescape the json string this way, since javascript 
-			// unescape doesn't work here
-			json = arr[1].replace(/\\\//g,'/').replace(/\\\"/g,'"');
+			json = arr[1];
 			break;
 		}
 	}
-	details = JSON.parse(json);
 	
+	// one JSON.parse to unstringify the json string, and one to parse it into an object
+	var details = JSON.parse(JSON.parse(json));
+
 	var metaArr = {};
 	var metaTags = doc.getElementsByTagName("meta");
-	for (var j = 0 ; j < metaTags.length ; j++) {
-		if (metaTags[j].getAttribute("property")) {
-			metaArr[metaTags[j].getAttribute("property")] = metaTags[j].getAttribute("content");
+	for (let metaTag of metaTags) {
+		if (metaTag.getAttribute("property")) {
+			metaArr[metaTag.getAttribute("property")] = metaTag.getAttribute("content");
 		}
 	}
 	newItem.title = details.citation.title;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by pbinkley** [comment]
I'm not sure whether the ```attr()``` function is still needed, or what to do about the unused ```doc``` and  ```url``` params that Travis doesn't like. I'd be happy to clean up further if someone can advise me.

File: newspapers.com.js [hunk 1]
```diff
@@ -66,15 +81,12 @@ function doWeb(doc, url) {
 		newItem.title = tokens[0].trim();
 		// multiple authors are separated with semicolons
 		var authors = author.split("; ");
-		for (i=0; i<authors.length; i++) {
+		for (i = 0; i < authors.length; i++) {
```
dstillman: `let i` or, better, `let author of authors`

newspapers.com.js
```diff
@@ -46,15 +46,30 @@ function detectWeb(doc, url) {
...
 	var newItem = new Zotero.Item("newspaperArticle");
+	var scripts = doc.getElementsByTagName("script");
+	var json = '';
+	var jsonre = /var staPageDetail = JSON.parse\("(.+?)"\);/;
+	for (var i = 0; i < scripts.length; i++) {
+		var arr = scripts[i].textContent.match(jsonre);
+		if (arr) {
+			// we have to unescape the json string this way, since javascript
+			// unescape doesn't work here
+			json = arr[1].replace(/\\\//g, "/").replace(/\\"/g, '"').replace(/\\'/g, "'");
```
dstillman: No need for this. You can just use `JSON.parse()` the same way they are. You just need to capture the double quotes in the regexp above.

newspapers.com.js
```diff
@@ -46,15 +46,30 @@ function detectWeb(doc, url) {
...
+		}
+	}
+	var details = JSON.parse(json);
+	
 	var metaArr = {};
 	var metaTags = doc.getElementsByTagName("meta");
-	for (var i = 0 ; i < metaTags.length ; i++) {
-		if (metaTags[i].getAttribute("property")) {
-			metaArr[metaTags[i].getAttribute("property")] = metaTags[i].getAttribute("content");
+	for (var j = 0; j < metaTags.length; j++) {
```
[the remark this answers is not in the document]
pbinkley: done

pbinkley:
Thanks for the suggestions. I've cleaned up the code in accordance with them. The ```JSON.parse()``` problem is weird: because the source string is stringified (with escaped slashes etc.), I have to apply ```JSON.parse``` twice to get it into an object. If there's a cleaner way to achieve this I'd love to learn it.

adam3smith:
attr and text are no longer needed. The functions will work without the `url` parameter, but I'm pretty sure `detectWeb` does need `doc` so we'll need to escape the eslint rule there. 
There was also something wonky about the tests -- somehow they had duplicated which caused the linter to fail.

newspapers.com.js
```diff
@@ -35,46 +35,58 @@
 	***** END LICENSE BLOCK *****
 */
 
-
-// attr()/text() v2
-function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null;}function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null;}
-
-
-function detectWeb(doc, url) {
+function detectWeb(doc) {
```
adam3smith: @dstillman what's your view on this? it's easy enough to eslint-escape locally, but this will be a pretty common issue, so might be worth changing globally? On the other hand, the may well be actual cases where `doc` is passed unnecessarily that'd be nice to catch.

dstillman:
> The JSON.parse() problem is weird: because the source string is stringified (with escaped slashes etc.), I have to apply JSON.parse twice to get it into an object.

Sorry, that's right. Since we're treating the contents of this script as a string, the JSON string is itself stringified from the perspective of the translator (as opposed to the raw JS we see via View Source). There's no need for the conditional — it just needs to be run through `JSON.parse()` twice.

newspapers.com.js
dstillman: It'd be fine on a technical level to just remove the unneeded arguments, but our ESLint rules allow unused arguments to be prefixed with an underscore (`function detectWeb(_doc, _url)`), the idea being that it can be helpful to leave in the available parameters in case someone updating the code later wants to make use of them without having to remember or look up the function signature. I'm ambivalent about this, but that's the thinking behind it, at least.

newspapers.com.js
```diff
@@ -35,46 +35,58 @@
...
+	// one JSON.parse to unstringify the json string, and one to parse it into an object
+	var details = JSON.parse(json);
+	if (typeof details === "string") {
+		details = JSON.parse(details);
+	}
+
 	var metaArr = {};
 	var metaTags = doc.getElementsByTagName("meta");
-	for (var i = 0 ; i < metaTags.length ; i++) {
+	for (i = 0; i < metaTags.length; i++) {
```
dstillman: This is still wrong — it would mean that if the `var i` above was changed or removed this would become a global assignment. These loops should all use either `for (let i = 0; …)` or `for (let foo of foos)`.

newspapers.com.js
```diff
@@ -35,46 +35,58 @@
...
+function doWeb(doc) {
 	var newItem = new Zotero.Item("newspaperArticle");
+	var scripts = doc.getElementsByTagName("script");
+	var json = '';
+	var jsonre = /var staPageDetail = JSON.parse\((.+?)\);/;
+	for (var i = 0; i < scripts.length; i++) {
+		var arr = scripts[i].textContent.match(jsonre);
+		if (arr) {
+			// we have to unescape the json string this way, since javascript
+			// unescape doesn't work here
```
dstillman: Remove comment

newspapers.com.js
pbinkley: I've kept it with @adam3smith 's version for simplicity

newspapers.com.js
pbinkley: I've gone with ```for (let metaTag in metaTags)```

dstillman: If you really mean `for … in`, that's incorrect — that does something different. See the two options I give above.

newspapers.com.js
dstillman: I'm not sure what you mean, but `detectWeb(doc)` will still trigger a lint error, since `doc` is unused. If we're going to keep the lint rule, this should be `function detectWeb(_doc, _url)`.

pbinkley:
I think I've made all the requested changes, but let me know if there's more that needs doing.

pbinkley:
Is any more work needed on this?

pbinkley:
I've just committed a fix for the case where the json string includes an apostrophe: the source escapes it with a backslash, which JSON.parse treats as an invalid character. I've added a test for this situation.

adam3smith:
(I don't think this needs anything else; I'm just a bit underwater with reviews.

adam3smith:
Thanks & sorry for the long wait!

PR #1933, opened 2019-04-20
Update arxiv xpath

GuyAglionby:


adam3smith:
Thanks! Let's keep an eye on this -- I'm not 100% happy with the xpath (i.e. worried it might be a bit unstable relying on the "Cite as:" text), but not sure there's something better.

PR #1940, opened 2019-05-03
save full-res versions of any images in tweet

wildthyme:
a snapshot of the page only saves the low-res version of tweet images. this downloads and adds them as additional attachments

### DIFF
--- a/Twitter.js
+++ b/Twitter.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2018-05-11 23:28:19"
+	"lastUpdated": "2019-05-03 17:29:55"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


wildthyme:
RIP i messed up the style ig

zuphilip:
I am hesitant that saving the images in high resolution from Twitter is that common for the users of reference management software. @adam3smith Was anything like this asked before in the forum?

zuphilip:
Maybe https://www.zotero.org/support/preferences/hidden_preferences#zotero_connector

adam3smith:
@wildthyme Do you have an example where this matters? As far as I can tell, the Snapshot does save the full size pic, though there's not easy way to get it to show.

zuphilip:
Closing this pull request because of inactivity. @wildthyme, if you want to continue here, then feel free to reopen it.

PR #1941, opened 2019-05-04
Add The Art Newspaper.js

owcz:
Handles article pages, home/section page CMS, search. Pulls magazine issue info when available.


retorquere:
(the linting fails because of how the linter is called, this has since been fixed on master)

PR #1942, opened 2019-05-07
Suppress error when Bibliontology RDF detect fails

retorquere:
Bibliontology RDF doesn't catch the error in detectImport so it puts an error in the log for every detect. Not a major issue in the grand scheme of things but it pollutes my test logs (which do a lot of imports) pretty severely.

### DIFF
--- a/Bibliontology RDF.js
+++ b/Bibliontology RDF.js
@@ -785,7 +785,15 @@ function getStatementsByDefinition(definition, node) {
 
 function detectImport() {
 	// look for a bibo item type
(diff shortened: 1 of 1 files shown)


retorquere:
That was unexpected -- I ran lint locally before submitting the PR.

retorquere:
Wait -- I fetched upstream and branched from upstream/master, but it thinks I changed RDF.js?

retorquere:

AbeJellinek:
I apparently do not know what month it is...

Going to merge this but undo the lint stuff and do that separately. It's too complicated to reconcile with changes that have happened in the mean time.

PR #1943, opened 2019-05-07
silently fail detect for MODS

retorquere:
MODS doesn't catch the error in detectImport so it puts an error in the log for every detect. Not a major issue in the grand scheme of things but it pollutes my test logs (which do a lot of imports) pretty severely.


adam3smith:
Thanks!

PR #1953, opened 2019-06-09
lint names with spaces

retorquere:
The previous invocation of teslint would run

```
teslint "Primo" "Normalized" "XML.js"
```

rather than

```
teslint "Primo Normalized XML.js"
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,6 @@ language: node_js
 node_js:
   - "8"
 script:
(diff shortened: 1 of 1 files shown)


adam3smith:
Looks good to me, but my shell scripting is lousy. @dstillman could you please do a quick sanity check and merge if this looks OK?

adam3smith:
Thanks!

PR #1957, opened 2019-06-11
Fix semantic scholar arXiv links and update tests

GuyAglionby:
As description. This change should also preclude any more breakages from similar changes on their end.

Thanks!

### DIFF
--- a/Semantic Scholar.js
+++ b/Semantic Scholar.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2019-03-30 16:44:44"
+	"lastUpdated": "2019-07-07 21:59:05"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


Semantic Scholar.js
```diff
@@ -118,12 +118,10 @@ function parseDocument(doc, url) {
 			item.issue = volumeAndIssue[1];
 		}
 		
-		if (rawData.hasPdf) {
-			let paperLink = rawData.links.filter(function (link) {
-				return link.linkType === 's2';
-			})[0].url;
+		if (rawData.hasPdf && (rawData.primaryPaperLink.linkType === 's2'
+			|| rawData.primaryPaperLink.linkType == 'arxiv')) {
```
zuphilip: Do we want the PDFs from arXiv also automatically attached here when saving from SemanticScholar? Alternatively we could also just link to them.

Anyway, we can for save the arXiv ID in the extra field as well, cf. https://github.com/zotero/translators/blob/master/arXiv.org.js#L188

GuyAglionby: Thanks for the review! I'd say yes, as I'm not sure there's much functionally different between the hosting locations. I think it fits with semantic scholar's role as an aggregator to grab the papers too.

Semantic Scholar.js
```diff
@@ -118,15 +118,19 @@ function parseDocument(doc, url) {
...
 			item.attachments.push({
-				url: paperLink,
+				url: rawData.primaryPaperLink.url,
 				title: "Full Text PDF",
 				mimeType: 'application/pdf'
 			});
+			if (rawData.primaryPaperLink.linkType == 'arxiv') {
+				let arxivId = rawData.primaryPaperLink.url.match(/\d{4}\.\d{5}/);
+				if (arxivId.length >= 1) {
+					item.extra = 'arXiv: ' + arxivId[0];
```
zuphilip: There could already be something in the extra field from bibtex translator, which we don't want to override  then. Can you change this into something like
```js
if (item.extra) {
   item.extra += "\narXiv: " + arxivId[0];
}
else {
   item.extra = 'arXiv: ' + arxivId[0];
}
```
(untested code)

GuyAglionby: Thanks, added

zuphilip:
Thank you!

PR #1958, opened 2019-06-11
eslint rule: report trailing semi-colon on testcases

retorquere:
Test cases should not have a trailing semicolon, this rule reports these as errors.

### DIFF
--- a/.ci/eslint-plugin-zotero-translator/lib/rules/test-cases.js
+++ b/.ci/eslint-plugin-zotero-translator/lib/rules/test-cases.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const translators = require('../translators').cache;
+const astUtils = require("eslint/lib/util/ast-utils");
 
 module.exports = {
 	meta: {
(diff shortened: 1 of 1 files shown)


zuphilip:
Can/should we do a more general test for these things to being valid JSON? I thought we had something like that previously...

retorquere:
It should [already do this](https://github.com/zotero/translators/blob/master/.ci/eslint-plugin-zotero-translator/lib/translators.js#L125).

zuphilip:
Okay, such a trailing semi-colon in the test cases will cause Scaffold to refuse reading the test cases. Thus, it would be good to have such a rule.

retorquere:
to be clear -- the most likely culprit to be adding these is me, but I think it was at some point added by `teslint --fix`. I haven't seen that again, but the spate of occurrences we've just went through was almost certainly added by `--fix`.

I just figured we had the rule already, now it checks one additional thing, and it was an easy check to add.

zuphilip:
Thank you!

PR #1960, opened 2019-06-13
Archives Canada.js -- fix eslint warnings

retorquere:


adam3smith:
I'll take this -- but the site referenced doesn't exist anymore. There's a good chance we'd need to completely rewrite anyway.

PR #1961, opened 2019-06-14
eslint rule: don't allow overriding globals

retorquere:
Don't allow overriding globals specified in eslintrc

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -30,7 +30,6 @@
 		],
 
 		"no-redeclare": ["error", {"builtinGlobals": true}],
-		"no-global-assign": "error",
 
 		"zotero-translator/not-executable": "error",
 		"zotero-translator/header-valid-json": "error",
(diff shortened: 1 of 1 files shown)


adam3smith:
(this makes a lot of sense to me)

.eslintrc
```diff
@@ -28,7 +28,10 @@
 				"varsIgnorePattern": "^detectWeb$|^doWeb$|^detectImport$|^doImport$|^doExport$|^detectSearch$|^doSearch$"
 			}
 		],
-		
+
+		"no-redeclare": ["error", {"builtinGlobals": true}],
+		"no-global-assign": "error",
```
dstillman: `no-global-assign` is already included by way of `eslint:recommended` in `@zotero/eslint-config` (as is `no-redeclare`, but without `"builtinGlobals": true`), so is can be removed here.

PR #1974, opened 2019-07-14
ACLWeb: Adding -HLT to conference name based on proceedings title

nschneid:
https://github.com/acl-org/acl-anthology/issues/453


zuphilip:
Thank you @nschneid ! :rocket:

PR #1977, opened 2019-07-17
 Oxford English Dictionary: remove translator framework

retorquere:
How do I test translators that require authentication in scaffold? I have access through an institutional proxy.

### DIFF
--- a/Oxford English Dictionary.js
+++ b/Oxford English Dictionary.js
@@ -1,7 +1,7 @@
 {
 	"translatorID": "d3ee2368-04d7-4b4d-a8f3-c20c3f5234a9",
 	"label": "Oxford English Dictionary",
-	"creator": "Sebastian Karcher",
+	"creator": "Sebastian Karcher and Emiliano Heyns",
 	"target": "^https?://(www\\.)?oed\\.com/",
 	"minVersion": "3.0",
 	"maxVersion": "",
@@ -9,13 +9,13 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2019-07-17 00:07:32"
+	"lastUpdated": "2021-06-14 21:18:33"
 }
 
 /*
 	***** BEGIN LICENSE BLOCK *****
 	
-	Copyright © 2013 Sebastian Karcher
+	Copyright © 2013-2021 Sebastian Karcher and Emiliano Heyns
 	This file is part of Zotero.
 	
 	Zotero is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


dstillman:
If you have direct, real proxy, or VPN access, it'll just work. If you only have a web-based proxy, you'd need to save the translator, reset translators in the connector preferences, and test pages manually in the browser.

retorquere:
I have only web-based. I'll give that procedure a go.

retorquere:
Well that's not great. The web-based proxy changes the url from http://www.oed.com/ into https://www-oed-com.vu-nl.idm.oclc.org/, so the url-based regex would fail. Suggestions? How have other translators dealt with this?

dstillman:
That’s handled by the connector’s proxy support — the translator doesn’t have to deal with it. EZProxy proxies should be detected automatically. If yours is something different you may need to configure it manually in the connector prefs.

retorquere:
I'm using EZproxy, but the Zotero icon doesn't change and OED is not one of the options under the Zotero browser button after updating the translators.

retorquere:
I've submitted the connector log with debug Report ID: 5040489

dstillman:
You should just configure your proxy in the Proxies pane of the connector preferences.

retorquere:
I have this in my proxies: `%h.vu-nl.idm.oclc.org/%p`

dstillman:
Are "Automatically associate new hosts" and "Automatically convert between dots and hyphens in proxied hostnames" checked?

If so, we'd want to see a <a href="https://www.zotero.org/support/debug_output#debug_output_logging">Debug ID</a> for loading the page when it's not detected. Make sure you've saved the translator in Scaffold and run Reset Translators first.

retorquere:
Yes to both -- but I also see the yellow top-bar saying "Zotero redirected" etc while I'm not being redirected.

The "Entry" test now passes, but the multi-test does not. Still looking into that.

retorquere:
I've logged the url from `detectWeb`, and I'm being handed https://www-oed-com.vu-nl.idm.oclc.org/search?searchType=dictionary&q=labor&_searchBtn=Search, so the regex fails.

retorquere:
Changing the check to `/\/search\?/` seems to fix it, but previously the translator did check for the full URL.

dstillman:
No, the target needs to match on the domain.

Can you provide a <a href="https://www.zotero.org/support/debug_output#debug_output_logging">Debug ID</a> from the connector for loading the page?

retorquere:
D982078648

dstillman:
Oh, you're talking about the `url` in `detectWeb`. That does give you the unproxied URL, but note that the previous version of the translator did, in fact, match only on the path, not the full URL. It's the job of the `target` to match the domain, and that comparison is to the proxied URL, which is working properly (or you wouldn't be getting to `detectWeb` at all). I'm not sure why the `url` in `detectWeb` isn't the proxied URL, but it doesn't really matter — there shouldn't really be a need to check the domain there.

In this case, we do need to make the `www.` in the target optional, however, since OED doesn't redirect.

dstillman:
Oh, sorry, you're right — the MultiScraper did check the full URL. Yeah, don't worry about that — that was wrong. Checking just the path is fine, since the domain was already checked by the target.

retorquere:
In that case it should be done as it is now.

dstillman:
> In this case, we do need to make the `www.` in the target optional, however, since OED doesn't redirect.

dstillman:
In other words: `"^https?://(www\\.)?oed\\.com"`

dstillman:
Oh, while we're here, the target should end with `/`.

zuphilip:
Is this ready for a review?

retorquere:
It is for my part.

AbeJellinek:
This looks great, thanks. Updated for a slight search page layout change and rebased.

PR #1981, opened 2019-07-25
Fix Eastview multiples

adam3smith:
and some validation changes


zuphilip:
The fix looks fine, but there are some more linting issues. Can you fix them as well? The easiest is to run  `npm run lint -- Eastview.js --fix` and then fix the remaining ones manually.

zuphilip:
I fixed the remaining linting issues and add another check for the existence of `header`. @adam3smith Anything else left to do here?  Is there ready to merge now?

PR #1988, opened 2019-08-11
Fix misspelled foaf property name

glaserL:
the `Zotero RDF.js` translator outputs a misspelled property name from the foaf namespace. 
`foaf:givenname` should be `foaf:givenName`.
In the future, `foaf:surname` which is considered archaic could be replaced by `foaf:familyName` which is in testing right now. See [here](http://xmlns.com/foaf/spec/#term_familyName).


zuphilip:
Okay, that is a good catch!

@glaserL You need to bump the timestamp at the beginning. (Otherwise the changes will not be shiped to the individual Zotero clients.)

Have you tested to export from Zotero and then import the RDF again? I think we handle this case already in the import general RDF translator, but a test should give us a quick verification.

Moreover, we can change this also in the Bibliontology RDF translator.

glaserL:
@zuphilip Thank you very much for the fast reply! I will bump the timestamp and try again. I should not change the translatorID field as well, correct?

I did try exporting from and reimporting to zotero and it works correctly.

A general question: There are other issues with the Zotero RDF export, e.g. the author and editor lists contain blank nodes which make it, as far as I know, impossible to link to them from other resources. 
Unfortunately my js skills reach their limit with string replacement; what would be a good way to report those issues and potentially help fixing them?

Thank you very much for your help

zuphilip:
Great! Yes, `translatorID` has to stay always the same.

Maybe the best place for general remarks about Zotero RDF translator is then the [Zotero forum](https://forums.zotero.org/).

zuphilip:
> In the future, foaf:surname which is considered archaic could be replaced by foaf:familyName which is in testing right now. See http://xmlns.com/foaf/spec/#term_surname

Any reason not to do that immediately? @glaserL @dstillman

zuphilip:
Thank you @glaserL !

Issue #1998, opened 2019-08-31
ACLWeb: Anthology now redirects to canonical+slash links


nschneid:
The translator expects ACL Anthology paper metadata pages to have a nested URL under "papers" like https://www.aclweb.org/anthology/papers/P/P13/P13-2103, but these now redirect to the canonical URL ending in a slash, i.e. https://www.aclweb.org/anthology/P13-2103/. (Eventually the final slash will be stripped, but right now the canonical URL without the slash points to the PDF: https://github.com/acl-org/acl-anthology/issues/480)

I will see if I can make a quick fix that relies on the current format for canonical paper IDs (though that too will change eventually for future papers: https://github.com/acl-org/acl-anthology/issues/291).

PR #1999, opened 2019-08-31
ACLWeb: new URL scheme for papers (#1998)

nschneid:
#1998

### DIFF
--- a/ACLWeb.js
+++ b/ACLWeb.js
@@ -12,6 +12,7 @@
 	"lastUpdated": "2019-07-14 19:22:00"
 }
 
+
 /*
 	***** BEGIN LICENSE BLOCK *****
 
--- a/ACLWeb.js
+++ b/ACLWeb.js
@@ -9,9 +9,8 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2019-07-14 19:22:00"
+	"lastUpdated": "2019-08-31 13:36:00"
 }
-
 
 /*
 	***** BEGIN LICENSE BLOCK *****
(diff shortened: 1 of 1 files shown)


nschneid:
@GuyAglionby Any idea why Travis build isn't failing when I put in an invalid URL for a test case?

zuphilip:
Travis does currently no look into the test cases, but you should check them yourself before pushing here. There are also some ignorable differences when running the checks, e.g. some language-dependent information.

Let me know when you are ready with this PR.

nschneid:
@zuphilip This should get the functionality back to where it was with up-to-date URLs. (Importing the PDF directly may still be broken on Firefox.)

ACLWeb.js
```diff
@@ -58,7 +58,9 @@ var ext2mime = {
 };
 
 function detectWeb(doc, url) {
-	if (doc.contentType === 'application/pdf' || url.endsWith('.bib') || url.includes('/papers/')) {
+	let paperIdUrl = /[/][A-Z][0-9]{2}-[0-9]{4}[/]?$/;
```
zuphilip: That is fine, but you could also use the abbreviation `\d` for `[0-9]`.

ACLWeb.js
```diff
@@ -86,7 +88,7 @@ function doWeb(doc, url) {
 		let paperURL = url.replace('.bib', '');
 		ZU.processDocuments(paperURL, scrape);
 	}
-	else if (doc.contentType === 'application/pdf') {
+	else if (doc.contentType === 'application/pdf' || url.endsWith('.pdf')) {
```
zuphilip: Are both checks here needed?

nschneid: I'm not sure... @GuyAglionby mentioned something about a `contentType` bug in Firefox that I figure this might fix once further URL changes are made on the site.

zuphilip:
Thank you!

PR #2010, opened 2019-09-17
Update item for Bioconductor.js

hubentu:
Based on feedback, the item 'system' is changed to 'company', so the Bioconductor with version can be shown in the generated Bibliography.

### DIFF
--- a/Bioconductor.js
+++ b/Bioconductor.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2019-09-16 15:14:13"
+	"lastUpdated": "2019-09-17 16:47:07"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


zuphilip:
@hubentu Where do you get this feedback? Which citation style are you looking to?

The previous commits were merged together which now creates this conflict. You should have started with the new clean version from here. However, I may simply cherry-pick the last commit(s) then...

hubentu:
@zuphilip Sorry for the inconvenience... I got feedback from one of the Bioconductor core team. We tried different styles, such as, IEEE, Nature, BMC bioinformatics... The 'company' item will be shown in the citation, but the 'system' won't.

For example, the BMC Bioinformatics style:

With system item:
1. Morgan M, Obenchain V, Hester J, Pagès H. SummarizedExperiment: SummarizedExperiment container. 2019. doi:10.18129/B9.bioc.SummarizedExperiment.

Use company item instead:
1. Morgan M, Obenchain V, Hester J, Pagès H. SummarizedExperiment: SummarizedExperiment container. Bioconductor version: Release (3.9); 2019. doi:10.18129/B9.bioc.SummarizedExperiment.

hubentu:
@zuphilip I have updated my master to fix the conflicts, so the only differences are the Bioconductor.js. Thanks!

zuphilip:
Okay, thank you for the information. Yes, it seems you are right. The name of this field  `company` is just a little strange here and maybe also in general. For CSL it is mapped to  `publisher` and therefore shown in a lot of styles. Better names of this field in Zotero are IMO something like "repository" or "hosting platform".

zuphilip:
Thank you! This is now merged again.

PR #2011, opened 2019-09-26
support download multiple items from a conference page of IEEE Xplore

fzyz999:
For example: https://ieeexplore.ieee.org/xpl/conhome/7048058/proceeding

The page contains all the papers from specific conference. However, we have to download the papers one by one because `IEEE Xplore.js` doesn't reconize it as `multiple` type. 

`target` and `detectWeb` have been modified for supporting this feature

### DIFF
--- a/IEEE Xplore.js
+++ b/IEEE Xplore.js
@@ -2,18 +2,18 @@
 	"translatorID": "92d4ed84-8d0-4d3c-941f-d4b9124cfbb",
 	"label": "IEEE Xplore",
 	"creator": "Simon Kornblith, Michael Berkowitz, Bastian Koenings, and Avram Lyon",
-	"target": "^https?://([^/]+\\.)?ieeexplore\\.ieee\\.org/([^#]+[&?]arnumber=\\d+|(abstract/)?document/|search/(searchresult|selected)\\.jsp|xpl/(mostRecentIssue|tocresult)\\.jsp\\?)",
+	"target": "^https?://([^/]+\\.)?ieeexplore\\.ieee\\.org/([^#]+[&?]arnumber=\\d+|(abstract/)?document/|search/(searchresult|selected)\\.jsp|xpl/(mostRecentIssue|tocresult)\\.jsp\\?|xpl/conhome/\\d+/proceeding)",
 	"minVersion": "4.0",
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2018-11-18 13:57:41"
+	"lastUpdated": "2019-09-30 01:44:41"
 }
 
 function detectWeb(doc, url) {
-	if (doc.defaultView !== doc.defaultView.top) return;
+	if (doc.defaultView !== doc.defaultView.top) return false;
 	
 	if (/[?&]arnumber=(\d+)/i.test(url) || /\/document\/\d+/i.test(url)) {
 		return "journalArticle";
@@ -27,6 +27,11 @@ function detectWeb(doc, url) {
 	
 	// Search results
 	if (url.includes("/search/searchresult.jsp")) {
+		return "multiple";
+	}
+	
+	// conference list results
+	if (url.includes("xpl/conhome") && url.includes("proceeding")) {
 		return "multiple";
 	}
 
@@ -44,6 +49,7 @@ function detectWeb(doc, url) {
 		return "multiple";
 	}
 	*/
+	return false;
 }
 
 function getSearchResults(doc, checkOnly) {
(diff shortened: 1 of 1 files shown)


IEEE Xplore.js
```diff
@@ -516,10 +528,9 @@ var testCases = [
 						"creatorType": "author"
 					}
 				],
-				"date": "January 1991",
+				"date": "一月 1991",
```
dstillman: I assume this is a localization based on your IP address. If so, we'd want to keep it set to the English version.

fzyz999: OK, I change it from `一月` to `January` manually because `一月` has exactly the same meaning as `January`. You are right. Geographic location of my IP address is China. So when I click `update` in Scaffold, it change `January` to `一月`. Sorry, I wasn't noticing this.

IEEE Xplore.js
```diff
@@ -2,18 +2,18 @@
 	"translatorID": "92d4ed84-8d0-4d3c-941f-d4b9124cfbb",
 	"label": "IEEE Xplore",
 	"creator": "Simon Kornblith, Michael Berkowitz, Bastian Koenings, and Avram Lyon",
-	"target": "^https?://([^/]+\\.)?ieeexplore\\.ieee\\.org/([^#]+[&?]arnumber=\\d+|(abstract/)?document/|search/(searchresult|selected)\\.jsp|xpl/(mostRecentIssue|tocresult)\\.jsp\\?)",
+	"target": "^https?://([^/]+\\.)?ieeexplore\\.ieee\\.org/([^#]+[&?]arnumber=\\d+|(abstract/)?document/|search/(searchresult|selected)\\.jsp|xpl/((mostRecentIssue|tocresult)\\.jsp\\?|conhome/\\d+/proceeding))",
```
zuphilip: With the new parenthesis this is quite hard to understand at first. Can we change it to easier expression by repeating the `xpl/`, i.e. just adding `|xpl/conhome/\\d+/proceeding` at the end?

IEEE Xplore.js
```diff
@@ -518,8 +530,7 @@ var testCases = [
 				],
 				"date": "January 1991",
 				"DOI": "10.1109/78.80767",
-				"ISSN": "1053-587X",
```
zuphilip: Why is the ISSN now missing? Did you also test the other examples or only this one?

IEEE Xplore.js
fzyz999: Thanks for your suggestion. I have changed it as you suggested.

IEEE Xplore.js
fzyz999: To be honest, I only test my code by visiting `https://ieeexplore.ieee.org/xpl/conhome/7048058/proceeding` and try to downloads all papers. The page was detected by the translator and I downloads all papers from the conference successfully. So I think my code work correctly. The only thing my code will do is recognizing pages whose url matches `xpl/conhome/\\d+/proceeding` as `multiple`. It won't disturb other test cases. 

I tried to pass all test cases. But when I click `run` button of Scaffold, it always report "TranslatorTester: IEEE Xplore Test 1: failed (Detection failed)" (for all of tests). I have no idea why this happens. When I paste the url of a test(`https://ieeexplore.ieee.org/document/80767/`) to `Browser` tab, everything works fine. I click `Test Regex`, it  outputs `===>true<=== (boolean)` . Then I click `Run detectWeb`, it outputs `detectWeb returned type "journalArticle"`. It seems that the detection process is correct. Why it output `Detection failed` when I click `run` ? I have no idea, so I give up.

I think the reason of why the ISSN is missing is that the procedure of parsing IEEE web pages is outdated. Maybe the position of some elements of the web page is changed. I'll try my best to handle this if needed.

fzyz999: > Why is the ISSN now missing? Did you also test the other examples or only this one?

After some debug, I found that ISSN is missing because IEEE Xplorer no longer provide it. When I tried to click `Download citation` and select `BibTex` output format manually, I got:
```
@ARTICLE{80767,
author={A. M. {Haimovich} and Y. {Bar-Ness}},
journal={IEEE Transactions on Signal Processing},
title={An eigenanalysis interference canceler},
year={1991},
volume={39},
number={1},
pages={76-84},
keywords={adaptive filters;eigenvalues and eigenfunctions;filtering and prediction theory;interference suppression;signal processing;eigenanalysis methods;adaptive implementation;interference cancellation;array processing;fast converging algorithm;Interference cancellation;Background noise;Sensor arrays;Signal to noise ratio;Jamming;Array signal processing;Steady-state;Noise cancellation;Direction of arrival estimation;Optimization methods},
doi={10.1109/78.80767},
ISSN={},
month={Jan},}
```
As can be seen above, `ISSN` is missing. `IEEE Xplorer.js` extract `ISSN` information from the downloaded BibTex citation. Thus, the ISSN is now missing because it is missing in the downloaded BibTex citation.

zuphilip: Okay, I looked into this and tried to extract the ISSN from other data in the website. The tests for multiple within Scaffold might fail due to some dynamic content loading.

zuphilip:
I pushed directly into your branch an update which also fixes all the remaining linting issues. This PR should be ready to merge now.

IEEE Xplore.js
fzyz999: Thanks ^_^

zuphilip:
:rocket: Thank you very much @fzyz999 !

PR #2013, opened 2019-10-01
Translator for globes

eranroz:
A custom translator for globes, for properly extracting the author

### DIFF
--- a/Globes.js
+++ b/Globes.js
@@ -30,8 +30,17 @@
 	<http://www.gnu.org/licenses/>.
 */
 
-function detectWeb(_doc, _url) {
-	return 'newspaperArticle';
+
+// attr()/text() v2
+// eslint-disable-next-line
+function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null;}function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null;}
+
+
+function detectWeb(doc, _url) {
+	if (attr(doc, 'meta[property="og:type"]', 'content') == "article") {
+		return 'newspaperArticle';
+	}
+	return false;
 }
 function doWeb(doc, url) {
 	var i;
--- a/Globes.js
+++ b/Globes.js
@@ -1,0 +1,234 @@
+{
+	"translatorID": "4e9c34ef-89bb-4c7e-97dd-abb61d8b9254",
+	"label": "Globes",
+	"creator": "Eran Rosenthal",
+	"target": "^https?://(www\\.)?(en\\.)?globes\\.co\\.il/",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2019-12-21 07:49:34"
+}
+
+/**
+	Copyright (c) 2019 Eran Rosenthal
+	
+	This program is free software: you can redistribute it and/or
+	modify it under the terms of the GNU Affero General Public License
+	as published by the Free Software Foundation, either version 3 of
+	the License, or (at your option) any later version.
+	
+	This program is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+	Affero General Public License for more details.
+	
+	You should have received a copy of the GNU Affero General Public
+	License along with this program. If not, see
+	<http://www.gnu.org/licenses/>.
+*/
+
+function detectWeb(_doc, _url) {
+	return 'newspaperArticle';
+}
+function doWeb(doc, url) {
+	var i;
+	var item = new Zotero.Item('newspaperArticle');
+	
+	item.title = ZU.xpathText(doc, '//meta[@property="og:title"]/@content');
+	item.publicationTitle = 'Globes';
+	item.url = url;
+	item.abstractNote = ZU.xpathText(doc, '//meta[@name="description"]/@content');
+	item.section = ZU.xpathText(doc, '//meta[@property="article:section"]/@content');
+	var tags = ZU.xpath(doc, '//meta[@property="article:tag"]');
+	for (i = 0; i < tags.length; i++) {
+		item.tags.push(tags[i].content);
+	}
+
+	var json = ZU.xpathText(doc, '//script[@type="application/ld+json"]');
+	var data = JSON.parse(json);
+
+	if (data) {
+		if (data.author && Array.isArray(data.author)) {
+			for (i = 0; i < data.author.length; i++) {
+				// note the author sometimes include date in form of Month Year
+				// and possibly Day Month Year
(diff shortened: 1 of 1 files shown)


Globes.js
```diff
@@ -0,0 +1,94 @@
...
+function doWeb(doc, url) { 
+	var item = new Zotero.Item('newspaperArticle');
+	
+	item.title = ZU.xpathText(doc, '//meta[@property="og:title"]/@content');
+	item.publicationTitle = 'Globes';
+	item.url = url;
+	item.language = 'he';
+	var abstract = ZU.xpathText(doc, '//meta[@name="description"]/@content');
+	if (abstract) {
+		item.abstractNote = abstract;
```
zuphilip: You can assign this also directly, if no postprocessing is needed.

Globes.js
```diff
@@ -0,0 +1,94 @@
...
+	}
+	
+	var json = ZU.xpathText(doc, '//script[@type="application/ld+json"]');
+	var data = JSON.parse(json);
+	if (data) {
+		
+		if (data.author && Array.isArray(data.author)) {
+			// note only the 0 item is a real author
+			item.creators.push(Zotero.Utilities.cleanAuthor(data.author[0].name, 'author'));
```
zuphilip: How about articles with multiple authors? I guess we should restrict this in another way...

Globes.js
```diff
@@ -0,0 +1,94 @@
...
+				"date": "2019-01-10",
+				"abstractNote": "גורמים עלומים בצפון-מזרח תאילנד מנהלים כבר שנים מערך מתוחכם ומורכב להפליא של ציד בלתי חוקיים, ומרוויחים הון עתק מהכליה שהם ממיטים על כמה מהמינים האקזוטיים בעולם ● האם האדם היחיד שנאבק בתופעה יצליח להפיל את סינדיקט ההברחות שאיש לא יכול לו?&nbsp;●&nbsp;גיליון שכולו קנאה&nbsp;&nbsp;",
+				"language": "he",
+				"libraryCatalog": "Globes",
+				"publicationTitle": "Globes",
+				"url": "https://www.globes.co.il/news/article.aspx?did=1001302127",
+				"attachments": [],
+				"tags": [],
+				"notes": [],
+				"seeAlso": []
```
zuphilip: Can we save section within the newspaper where this article appeared as well? I.e.
```html
<meta property="article:section" content="מגזין G">
```

Globes.js
```diff
@@ -0,0 +1,94 @@
...
+					}
+				],
+				"date": "2019-01-10",
+				"abstractNote": "גורמים עלומים בצפון-מזרח תאילנד מנהלים כבר שנים מערך מתוחכם ומורכב להפליא של ציד בלתי חוקיים, ומרוויחים הון עתק מהכליה שהם ממיטים על כמה מהמינים האקזוטיים בעולם ● האם האדם היחיד שנאבק בתופעה יצליח להפיל את סינדיקט ההברחות שאיש לא יכול לו?&nbsp;●&nbsp;גיליון שכולו קנאה&nbsp;&nbsp;",
+				"language": "he",
+				"libraryCatalog": "Globes",
+				"publicationTitle": "Globes",
+				"url": "https://www.globes.co.il/news/article.aspx?did=1001302127",
+				"attachments": [],
+				"tags": [],
```
zuphilip: Can we add some tags as well? Maybe from here
```html
<meta property="article:tag" content="גיליון שכולו קנאה">
```

Globes.js
```diff
@@ -0,0 +1,94 @@
...
+						"creatorType": "author"
+					}
+				],
+				"date": "2019-01-10",
+				"abstractNote": "גורמים עלומים בצפון-מזרח תאילנד מנהלים כבר שנים מערך מתוחכם ומורכב להפליא של ציד בלתי חוקיים, ומרוויחים הון עתק מהכליה שהם ממיטים על כמה מהמינים האקזוטיים בעולם ● האם האדם היחיד שנאבק בתופעה יצליח להפיל את סינדיקט ההברחות שאיש לא יכול לו?&nbsp;●&nbsp;גיליון שכולו קנאה&nbsp;&nbsp;",
+				"language": "he",
+				"libraryCatalog": "Globes",
+				"publicationTitle": "Globes",
+				"url": "https://www.globes.co.il/news/article.aspx?did=1001302127",
+				"attachments": [],
```
zuphilip: Let us also save a snapshot, i.e. something like
```js
			item.attachments.push({
				title: "Snapshot",
				document: doc,
				mimeType:"text/html"
			});
```

Globes.js
```diff
@@ -0,0 +1,94 @@
...
+function detectWeb(doc, url) {
+	return 'newspaperArticle';
+}
+function doWeb(doc, url) { 
+	var item = new Zotero.Item('newspaperArticle');
+	
+	item.title = ZU.xpathText(doc, '//meta[@property="og:title"]/@content');
+	item.publicationTitle = 'Globes';
+	item.url = url;
+	item.language = 'he';
```
zuphilip: Can you also handle the English version of the Globes, e.g. https://en.globes.co.il/en/article-will-competition-authority-approve-taboola-outbrain-merger-1001302929 ?

Globes.js
zuphilip: This then should go into the field `section`.

Globes.js
eranroz: added to test cases few different cases - one where multiple "authors" are real author, and date and one where there are really 2 authors.

Globes.js
eranroz: Done

Globes.js
eranroz: Done

Globes.js
eranroz: Done

Globes.js
eranroz: Done

Globes.js
```diff
@@ -0,0 +1,248 @@
...
+				"creators": [
+					{
+						"firstName": "ג'ושוע",
+						"lastName": "האמר",
+						"creatorType": "author"
+					},
+					{
+						"firstName": "",
+						"lastName": "GQ",
+						"creatorType": "author"
```
zuphilip: Is this a real author or should we rather filter this one out?

eranroz: this a translated article (original aticle:  https://www.gq.com/story/hunting-rhino-horn-cartel-of-thailand ) and the credits are to GQ. 
I'm not sure how to determine it programmatically (other than heuristics of having both firstName and lastName?)

Globes.js
```diff
@@ -0,0 +1,248 @@
+{
+	"translatorID": "4e9c34ef-89bb-4c7e-97dd-abb61d8b9254",
+	"label": "Globes",
+	"creator": "Eran Rosenthal",
+	"target": "^https?://www\\.globes\\.co\\.il/news/article.aspx",
```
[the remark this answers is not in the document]
zuphilip: @eranroz That seems still open.

Globes.js
zuphilip: I would suggest only to not take it as author if it is exactly `GQ`. (If we have more cases like this this, then we may see better about which pattern might be suitable.)

Globes.js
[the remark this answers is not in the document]
eranroz: Fixed

Globes.js
eranroz: Seems the site changes the metadata in that item - updating the test itself

zuphilip:
Great! I did a small improvement in the detection and merge this now! :rocket:

Thank you very much @eranroz !

PR #2014, opened 2019-10-01
Wiley: PDFs not saved

### LINKED ISSUE #2004: Wiley PDF not always saved
There is no PDF saved for the Open Access article here from Wiley: https://besjournals.onlinelibrary.wiley.com/doi/10.1002/pan3.10038 cf. https://forums.zotero.org/discussion/78966/no-pdf-downloading-on-wiley . The same is true for other examples I tested.

I can replicate this issue on my Windows machine. It seems that the translator correctly tries to use the URL https://besjournals.onlinelibrary.wiley.com/doi/pdf/10.1002/pan3.10038  as the `pdfUrl`. However, in the end the connector fails with a red x.

This URL works in the browser for me without any problem, but when I trying to curl for it I see:
```
$ curl -I https://besjournals.onlinelibrary.wiley.com/doi/pdf/10.1002/pan3.10038
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--     0HTTP/2 302
server: AtyponWS/7.1
cache-control: private
x-xss-protection: 1; mode=block
x-content-type-options: nosniff
strict-transport-security: max-age=16070400
x-frame-options: SAMEORIGIN
p3p: CP="NOI DSP ADM OUR IND OTC"
location: https://besjournals.onlinelibrary.wiley.com/doi/pdf/10.1002/pan3.10038?cookieSet=1
set-cookie: I2KBRCK=1; domain=.onlinelibrary.wiley.com; path=/; secure; expires=Mon, 07-Sep-2020 18:13:16 GMT
content-type: text/html; charset=utf-8
date: Sun, 08 Sep 2019 18:13:16 GMT
```

@dstillman Is this maybe the problem? How could we overcome this?

dstillman:
This fixes #2004 (see https://github.com/zotero/translators/issues/2004#issuecomment-537209815), but I don't know if other pages are still using iframes. If so, we can restore some of the older logic.

We can clean up this logic soon when we have `DOMParser` and `async`/`await`.

### DIFF
--- a/Wiley Online Library.js
+++ b/Wiley Online Library.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2019-10-01 20:37:08"
+	"lastUpdated": "2019-10-01 20:54:57"
 }
 
 /*
@@ -29,6 +29,9 @@
    You should have received a copy of the GNU Affero General Public License
    along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
+
+// attr()/text() v2
+function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null;}function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null;}
 
 function fixCase(authorName) {
 	if (typeof authorName != 'string') return authorName;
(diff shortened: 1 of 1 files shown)


dstillman:
Actually, I guess it's just in child translators where DOMParser fails now. I can clean this up to use that.

dstillman:
Don't merge — I found a page where this doesn't work.

dstillman:
Wait, the site seems to have changed since a few hours ago, at least for me. The PDF URL was previously set on the `data` attribute of the `<object>` element, but it's now being set as `src` on the `<embed>` via JS, so parsing it from a `doGet()` response would now be more error-prone:

```javascript
<object id="pdfFrameObject" data width="100%" height="100%" type="application/pdf">
	<embed id="pdfFrame" src type="application/pdf"/>
</object>

<script>
	function isIOSDevice() {

		var iDevices = [
			'iPad Simulator',
			'iPhone Simulator',
			'iPod Simulator',
			'iPad',
			'iPhone',
			'iPod'
		];

		if (!!navigator.platform) {
			while (iDevices.length) {
				if (navigator.platform === iDevices.pop()) {
					return true;
				}
			}
		}
		return false;
	}

	var src = "/doi/pdfdirect/10.1002/pan3.10038";

	if (isIOSDevice()) {
		src = "/templates/jsp/_ux3/_pericles/pdf-viewer/web/viewer.html?file=/doi/pdfdirect/10.1002/pan3.10038";
	}


	document.getElementById("pdfFrame").setAttribute("src", src);
	document.getElementById("pdfFrameObject").setAttribute("data", src);
</script>
```

Maybe we should just look for a `pdfdirect` URL bounded by quotes and change it if they break it again later?

dstillman:
Or, better yet, we could just replace `pdf` with `pdfdirect` in the `citation_pdf_url`.

dstillman:
This seems to work for me.

dstillman:
OK, now fixed while viewing PDFs as well. I'll squash and merge unless anyone has objections.

zuphilip:
Thank you! No objection from my side. (But I have not time currently to test it myself.)

PR #2021, opened 2019-10-13
allow crlf (mainly so it can be detected and warned against)

retorquere:


.ci/eslint-plugin-zotero-translator/lib/translators.js
```diff
@@ -143,15 +145,15 @@ function tryFormatJSON(raw) {
 		return raw;
 	}
 }
-function strip(source) {
+function strip(source, newline) {
```
adam3smith: I'm getting some `undefined` inserted into the translators when fixing. I'm pretty sure that's because the `stip` function is also called from within `teslint`, but only with one argument and then `newline` is undefined in the replace below. Sorry for not trying to fix this, but I don't know this code at all so it'd take me a long time and I'm guessing it's a quick fix for you

zuphilip:
If I understand this correctly, then you want to normalize the line breaks. Is this really needed? - I think it is not uncommon to checkout repositories under Windows with CR+LF line endings but then configure git such that it will save them in commits as LF only.

Will you then warn on every line with line endings CR+LF? Or will this commit rather silently replace all those line endings?

retorquere:
It's not me doing the warning (or erroring in this case); the linter config inherits `@zotero/eslint-config`, which includes `"linebreak-style": [ "error", "unix" ],`. I agree setting up git as you describe it would make sure the linter would succeed when ran on Travis, but I can't really do "ignore the linebreak rule when you're on windows and have core.autocrlf set properly" so it can be ran locally on windows with that setting.

dstillman:
Can you explain the point of this patch?

retorquere:
The point of this patch is that the regex failed on `\r\n`, which means it wouldn't check the file at all: https://groups.google.com/forum/#!topic/zotero-dev/7aX8buzIPIM

adam3smith:
To rephrase the use case here (since this was prompted by me): without this, linting is not possible at all in Windows with a standard git config.

PR #2023, opened 2019-10-15
Update CNKI translator

l0o0:
1. Update target regex to match CNKI url prefix via proxy
2. Add snapshot attachment by default. PDF and CAJ file links will be added when the url is accessible
3. Use `doGet` to get refoworks from CNKI. `doPost` may have CSRF
    problem via proxy.
4. Add function 'getIDFromRef' to get `dbname` and `filename` from
    reference links. `getIDFromUrl` can not get `dbname` and `filename` when
    the article is pre-released.
5. Change `journal article` field value `callNumber`. `CN` in CNKI
    refwork format is not the `callNumber`, it's a Chinese version of ISSN.
    Move `CN` to extra field.

### DIFF
--- a/CNKI.js
+++ b/CNKI.js
@@ -1,22 +1,22 @@
 {
 	"translatorID": "5c95b67b-41c5-4f55-b71a-48d5d7183063",
 	"label": "CNKI",
-	"creator": "Aurimas Vinckevicius",
-	"target": "^https?://([^/]+\\.)?cnki\\.net",
+	"creator": "l0o0 linxzh1989@gmail.com",
+	"target": "^https?://kns",
 	"minVersion": "3.0",
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcs",
-	"lastUpdated": "2018-09-08 22:09:39"
+	"lastUpdated": "2019-10-15 03:24:45"
 }
 
 /*
 	***** BEGIN LICENSE BLOCK *****
 
 	CNKI(China National Knowledge Infrastructure) Translator
-	Copyright © 2013 Aurimas Vinckevicius
+	Copyright © 2013 Aurimas Vinckevicius, updated by Lin Xingzhong linxzh1989@gmail.com
 
 	This file is part of Zotero.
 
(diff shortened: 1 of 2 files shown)


dstillman:
> Update target regex to match CNKI url prefix via proxy

[As I explained](https://forums.zotero.org/discussion/comment/341200/#Comment_341200) in the forums, this is incorrect and needs to be undone. The target is automatically compared against the unproxied URL.

CNKI.js
```diff
@@ -41,33 +41,33 @@
...
-					//Zotero doesn't do well with mixed line endings. Make everything \n
-					.replace(/\r\n?/g, '\n')
-					//split authors
-					.replace(/^(A[1-4]|U2)\s*([^\r\n]+)/gm, function(m, tag, authors) {
+	ZU.doGet('https://kns.cnki.net/kns/ViewPage/viewsave.aspx?displayMode=Refworks' + '&' +  postData, 
+		function(text) {
+			var parser = new DOMParser();
+			var html = parser.parseFromString(text, "text/html")
+			var text = ZU.xpath(html, "//table[@class='mainTable']//td")[0].innerHTML;
+			var text = text.replace(/<br>/g, '\n');
```
dstillman: Can you just use `.textContent` instead of `.innerHTML` and manual tag parsing? Replacing `<br>` manually is going to be much more fragile.

l0o0:
https://kns.cnki.net/KCMS/detail/11.1809.S.20191010.1139.002.html, this is a pre-released article for the test.

CNKI.js
l0o0: I've tried to use `.textContent`, but the returned result is 'RT DissertationSR 1A1 戈亚妮A3 黄毅勤PB 首都经济贸易大学T1 中小企业融资方式的比较与选择CL 硕士YR 2018K1 中小企业;融资方式;比较分析;选择建议...`. There are no line breaks and hard to parse.  

As you mentioned in the forum, there are no proxy entries in the Proxies tab after I use the proxy.

dstillman: OK, then you may need to configure the proxy manually. But the regex should just reflect the target site, without taking any proxy into account.

l0o0: Could you show me some proxy config examples? I really don't know how to set proxy manually in Zotero connector.

dstillman: Ask that in the forums.

CNKI.js
```diff
@@ -173,6 +181,72 @@ function doWeb(doc, url) {
 	}
 }
 
+function scrape(ids, doc, url, itemInfo) {
```
zuphilip: This function is defined twice! You should delete one of them.

CNKI.js
```diff
@@ -232,14 +306,134 @@ function scrape(ids, doc, url, itemInfo) {
 			}
 			
 			i++;
+			
+			// CN 中国刊物编号，非refworks中的callNumber
+			if (newItem.callNumber){
+				newItem.extra = 'CN ' + newItem.callNumber;
+				newItem.callNumber = "";
```
zuphilip: Add an (English) explanation as a comment why you moving this information. Moreover, I am not sure it is worth to save this journal identification number at all. Can you say a little bit more, why this could be useful for someone?

CNKI.js
```diff
@@ -1,22 +1,22 @@
 {
 	"translatorID": "5c95b67b-41c5-4f55-b71a-48d5d7183063",
 	"label": "CNKI",
-	"creator": "Aurimas Vinckevicius",
-	"target": "^https?://([^/]+\\.)?cnki\\.net",
+	"creator": "l0o0 linxzh1989@gmail.com",
+	"target": "^https?://kns",
```
zuphilip: No!

zuphilip:
This is superseeded by #2025. However, I want to second that not continuing here but opening another PR is not very helpful for us. :angry:

PR #2025, opened 2019-10-17
Update CNKI translator

l0o0:
1. Change post url from http to https
2. Fix `callNumber` field error. The `CN` section from CNKI in refworks format is the Chinese version of ISSN. Move this to `extra` field
3. Add PDF or CAJ attachment when the file is accessible.
4. Add function to get metadata when the paper is pre-released


dstillman:
Please push these changes to the original pull request rather than opening a new one. (The target is also still wrong.)

dstillman:
(If you need help doing that, let us know.)

l0o0:
> 
> 
> (If you need help doing that, let us know.)

Yes, I need your help. I have deleted the repo forked from Zotero and create a new repo.

dstillman:
OK, I'm not sure if you can restore that, so I'll reopen this one, but note that you always want to push changes to the original pull request. There's never a need to start a new one, even if you make completely different changes.

l0o0:
> 
> 
> OK, I'm not sure if you can restore that, so I'll reopen this one, but note that you always want to push changes to the original pull request. There's never a need to start a new one, even if you make completely different changes.

OK, I will change the regex and submit the change.

CNKI.js
```diff
@@ -2,14 +2,14 @@
 	"translatorID": "5c95b67b-41c5-4f55-b71a-48d5d7183063",
 	"label": "CNKI",
 	"creator": "Aurimas Vinckevicius",
-	"target": "^https?://([^/]+\\.)?cnki\\.net",
+	"target": "^?://([^/]+\\.)?cnki\\.net",
```
dstillman: This needs to be reverted to its original value.

l0o0: OK, I found many errors from the CI check result. Do I need to fix all the errors?

zuphilip: You don't have to fix linting errors. We can do that as well. However, it is difficult now to see, which functional changed you did now. Are your first two commits the functional changes and all the remaining ones for fixing linting issues according to Travis?

l0o0: Hi @zuphilip  Sorry about that.  The changes:
1. added return value in `getIDFromURL`
2. added new function `getIDFromRef`, `getPDF`, `getCAJ` and `getAttachments` 
3. Change code in  `getIDFromPage` to get `dbname` and `filename` pre-released article.
4. Change code in `scrape`, fix `CN` field error  and add attachments.

zuphilip:
Are the test cases working for you in Scaffold? Or browsing within Scaffold and then trying out the extraction?

l0o0:
> 
> 
> Are the test cases working for you in Scaffold? Or browsing within Scaffold and then trying out the extraction?

Hi, I tried these in scaffold test and failed.  But when I first wrote this code and passed the test in scaffold. I don't know why these happen. I will try to fix this error and make more tests in scaffold.   

The error is caused by the code to detect user login status HTML elements.  But these elements are dynamically created by js code.  It looks like Zotero can't find the whole HTML document after js modified.  I want to download pdf or caj file when these files are accessible after login.  Is it OK to get 
user cookies in Zotero according to Zotero's private rules?

dstillman:
You don’t need to do anything manually to support cookie-based sites. The Zotero Connector forwards cookies to the Zotero app, where they’re used automatically for attachment requests.

l0o0:
> 
> 
> You don’t need to do anything manually to support cookie-based sites. The Zotero Connector forwards cookies to the Zotero app, where they’re used automatically for attachment requests.

Yes, I don't need to detect user login status. Zotero will download the url and detect the file type by mimeType. Zotero will not save the attachments when mimeType doesn't match.

l0o0:
@dstillman It can detect PDF mimeType, but can not work for caj mimeType "application/caj". Is caj mimeType not supported by Zotero?   

Parts of the response header returned by CNKI server when downloading caj file is 
```
Cache-Control: private  
Content-Length: 1367818  
Content-Type: application/caj   
Accept-Ranges: bytes  
```  
When you not login the CNKI, the caj download url will redirect to a login webpage. The Zotero will save a caj attachment, but this attachment is a login HTML page that redirected after caj url.   
Parts of header returned by redict to login HTML:
```
Cache-Control: private
Content-Type: text/html; charset=utf-8
Server: Microsoft-IIS/7.5
X-AspNet-Version: 4.0.30319
X-Powered-By: ASP.NET
Date: Mon, 28 Oct 2019 02:25:22 GMT
Connection: close
Content-Length: 12613
```

dstillman:
Are you referring to the logic that avoids saving a PDF when the response isn't application/pdf? So that it will just fail instead of saving a login page? Is that what you're hoping to do for CAJ files, or am I misunderstanding?

l0o0:
@dstillman Yes, I try to detect user login status then decide whether to save PDF or CAJ. If PDF and CAJ are available, PDF is preferred.  I don't have a good solution to detect user status by now.  

But I found Zotero fails to download PDF attachment when I logout. I think Zotero will compare mimeType in attachment and the downloaded file type. If they are not matched after compared, Zotero will raise some kind of error message.   
图片

When I logout and want to download a CAJ file. A login page after redirected will save in Zotero. Zotero doesn't seem like to compare mimeType and file type when downloading the CAJ file.

dstillman:
Right, there's [code in Zotero](https://github.com/zotero/zotero/blob/d96fe03fa28037adf73574a735bc0b1952cf7293/chrome/content/zotero/xpcom/attachments.js#L884-L894) specifically for PDFs to verify that the content type is `application/pdf`, since it's common for a PDF URL provided by a translator to be inaccessible. When that happens, the red X shows up.

We could add a similar thing for CAJ, though given that I've never heard of CAJ before this thread, I'd sort of prefer if the translator could just detect whether the user is logged in based on the page content. Are you saying that's not possible? The PDF verification is meant as a generic solution to a common problem, but it's still preferable for the translator to determine itself whether the file is available.

l0o0:
The user info content in HTML page is generated by JS according to cookies. It seems quite a tough task for me to detect user login status.  Is it possible to use `XMLHttpRequest` to get the content return by CAJ download link then detect user status by the content?

dstillman:
It shouldn't matter that it's generated by JS as long as users are only saving once the page is done loading (and you can use `monitorDOMChanges()` to wait to show the save icon until the page is ready — see some other translators for examples). All you need to do is look at the page structure at the time of saving for whatever indicates that the page is loaded.

dstillman:
E.g., presumably "登录" doesn't appear in the page structure when the user is logged in.

l0o0:
> 
> 
> E.g., presumably "登录" doesn't appear in the page structure when the user is logged in.

I have tried this element before. Maybe I can try to wait for seconds (delayed by `monitorDOMChanges()`) . Thanks for your help @dstillman

l0o0:
```javascript
	// Z.debug(doc);
	var id = getIDFromPage(doc, url);
	var items = getItemsFromSearchResults(doc, url);
	Z.debug(id);
	Z.monitorDOMChanges(
		ZU.xpath(doc, "//div[@class='ecp_top-nav']")[0], 
		{subtree: true, attributes: true, attributeFilter: ['style']}
	);
	Z.debug(ZU.xpath(doc, "//div[@id='Ecp_top_login']")[0].innerText);
	if (id) {
		return getTypeFromDBName(id.dbname);
	} else if (items) {
		return "multiple";
	} else {
		return false;
	}
}
```
`monitorDOMChanges()` is added in `detectWeb()`.  In Scaffold, Run detectWeb will display element text '登录', but an error occurs when I run tests in Testing tab .

图片

I think the element monitored in `monitorDOMChanges()` is correct. Some elements will be hidden, so I add a style filter.
图片  
It looks like this element `ZU.xpath(doc, "//div[@id='Ecp_top_login']")[0]` created by JS doesn't exist when run tests in testing tab.

zuphilip:
@dstillman Can you do a code review here? (I have problems to follow the discussion here and would prefer to spend my time somewhere else.)

dstillman:
Unless it's somehow related to CNKI, can you open a separate PR for Wanfang Data?

CNKI.js
```diff
@@ -215,29 +242,88 @@ function scrape(ids, doc, url, itemInfo) {
...
+}
+
+// add pdf or caj to attachments, default is pdf
+function getAttachments(doc, item) {
+	var attachments = [{
+		url: item.url,
+		title: item.title,
+		mimeType: "text/html",
+		snapshot: true
+	}];
```
dstillman: Is it worth saving a snapshot? Generally, a snapshot is useful if 1) the page contains full-text content, particularly if it might later become unavailable (e.g., a newspaper/magazine article that might go behind a paywall) or 2) the page contains text that might be useful for searching. (Here, it looks like the page content is also dynamically loaded, and the way snapshots are currently saved doesn't make saving such pages very useful.)

l0o0:
> 
> 
> Unless it's somehow related to CNKI, can you open a separate PR for Wanfang Data?

I am so sorry to add a WanFang data translator. I will create a new pull request after I finish it.

CNKI.js
```diff
@@ -39,32 +39,32 @@
...
+				text = text.replace(/^(A[1-4]|U2)\s*([^\r\n]+)/gm, 
+						function(m, tag, authors) {
+							var authors = authors.split(/\s*[;，,]\s*/); //that's a special comma
+							if (!authors[authors.length-1].trim()) authors.pop();
+							return tag + ' ' + authors.join('\n' + tag + ' ');
+						}
+					);
+				next(text);
+			}
+		);
```
dstillman: What was the reason for changing this? Calling 'doPost" within a loop isn't ideal, because it results in all requests being made at the same time (or possibly limited by a per-server connection limit in the network layer) rather than run serially. While this can be faster, it makes sites more likely to block requests.

dstillman:
> I am so sorry to add a WanFang data translator. I will create a new pull request after I finish it.

No problem. For now, you should just remove those changes from this PR so that we can move forward with this one.

CNKI.js
l0o0: CNKI post API has changed in the search results. A new parameter is added, it looks like this parameter is from cookies. There is less information for me to figure out how this parameter is generated.

l0o0:
@dstillman I have `git rm Wanfang.js` and restored the old version.

CNKI.js
dstillman: I just mean the change in the last commit from a single POST request to multiple requests. Is it no longer possible to make a single POST request? If not, it'd still be better to run this serially by `shift()`ing off a list of ids and processing each one in turn until there are no more, at which point `next` can be called. I can provide an example if necessary. (We're working on `async`/`await` support in translators to make this easier, but for now it has to be done the awkward callback way.)

dstillman: Let me know if you want help with this. I'd love to get this fix in, but I don't want to release a version that will make up to 50 simultaneous POST requests and potentially get people blocked by the site.

l0o0: @dstillman Sorry for the delay, I was in the hospital a few days ago. An encrypted parameter is needed when getting multiple items by a single POST. I don't know how the string is encrypted, maybe this encrypted string is generated based on cookies on search page.  I hope you can give me some examples of `async`/`await` in translators.

dstillman: Oh, sorry to hear that.

I'm not suggesting making a single POST. I'm just saying this needs to be structured so that the individual POST requests are serialized instead of run concurrently, in the way I describe above. Let me know if you need an example of that.

`async`/`await` isn't yet available.

dstillman: Actually, I'll take care of this. I'll have an update soon.

dstillman:
Merged. I fixed the POST serialization issue and also fixed single-page saves in Chrome, which weren't working because `onclick` was only available via an attribute, not a property.

Thank you for working on this!

CNKI.js
l0o0: @dstillman Sorry, I don't quite understand what you mean.  I am a beginner in javascript 😂.  I don't know how to send posts in serialized. I hope you can give me a hand 😊

dstillman: I already fixed this and committed the translator. See my message below.

PR #2028, opened 2019-10-17
Added the possibility to select which fields should be exported to TeX

bonassifabio:
It would be awesome to let the users choose what kind of fields should be exported when BibTeX/BibLaTeX format is chosen. While I perfectly get that it could be done via proper citation styles, or even with _Better BibTeX for Zotero_, I think that giving a much easier way to do that would be willingly accepted by the users.

In particular, in the commit I added the possibility to enable or disable the export of the following fields:

- URL and URL-date
- DOI
- Abstract (abstract field always screwed up all my exports)
- Keywords
- Other informations, such as ISBN, language, number of volumes, number of pages, acc

Also the Zotero repository should be updated [introducing the corresponding setting strings](https://github.com/zotero/zotero/commit/7909f1dde599390b26569f3b175d8cf7479fa7c1)

Thank you for considering my request.

### DIFF
--- a/BibLaTeX.js
+++ b/BibLaTeX.js
@@ -13,6 +13,11 @@
 		"exportCharset": "UTF-8",
 		"exportNotes": false,
 		"exportFileData": false,
+		"exportAbstract": false,
+		"exportUrl": false,
+		"exportDOI": false,
+		"exportExtraInfo": false,
+		"exportKeywords": false,
 		"useJournalAbbreviation": false
 	},
 	"lastUpdated": "2019-01-31 13:16:00"
@@ -46,7 +51,6 @@
 //%t = first word of title
 var citeKeyFormat = "%a_%t_%y";
 
-
 var fieldMap = {
 	location: "place",
 	chapter: "chapter",
@@ -67,6 +71,31 @@ var fieldMap = {
 	eventtitle: "conferenceName",
 	pages: "pages",
 	pagetotal: "numPages"
+};
+
+var blockFieldsExtra = {
+	location: "place",
+	rights: "rights", //it's rights in zotero nowadays
+	isbn: "ISBN",
+	issn: "ISSN",
+	series: "series",
+	shorttitle: "shortTitle",
+	holder: "assignee",
+	volumes: "numberOfVolumes",
+	version: "version",
+	pagetotal: "numPages"
+};
+
+var blockFieldsUrl = {
+	url: "url"
+};
+
+var blockFieldsDOI = {
+	doi: "DOI"
+};
+
+var blockFieldsAbstract = {
+	abstract: "abstractNote"
 };
 //more conversions done below with special rules
 
@@ -516,7 +545,9 @@ function doExport() {
 		first = false;
 
 		for (var field in fieldMap) {
-			if (item[fieldMap[field]]) {
+			if (blockFieldsUrl[field] && Zotero.getOption("exportUrl") || blockFieldsDOI[field] && Zotero.getOption("exportDOI") || blockFieldsExtra[field] && Zotero.getOption("exportExtraInfo") || blockFieldsAbstract[field] && Zotero.getOption("exportAbstract")) {
+				continue;
+			} else if (item[fieldMap[field]]) {
 				writeField(field, item[fieldMap[field]]);
 			}
 		}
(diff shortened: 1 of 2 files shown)


dstillman:
Thanks, but not going to take this. The people using BibTeX in Zotero without BBT generally just use it as a standard exchange format, and I don't think this sort of configurability is necessary or desirable for that kind of usage. Data doesn't have to be used in all contexts, but throwing it out seems like bad practice to me in most situations. @retorquere may be willing to consider this — or something like it — in BBT, if it's not already possible.

retorquere:
https://retorque.re/zotero-better-bibtex/installation/preferences/export/#fields-to-omit-from-export-comma-separated

retorquere:
Throwing it out is indeed bad practice when the aim is data exchange, but I'd say for data exchange, the only redeeming feature of BibTeX is that is is widely accepted; it is ill-standardised in both what is generated by the myriad implementations, and how it is interpreted back by that same myriad. If you want data exchange I'd generally recommend RIS or CSL. Exporting to BibTeX, whether with stock Zotero or even with all fields enabled in BBT, already incurs data loss, as not all fields find a home in bibtex.

I reckon BBT users are generally not using it as a data exchange format though, but stock Zotero export and BBT live happily side by side in this regard I think.

dstillman:
> it is ill-standardised in both what is generated by the myriad implementations, and how it is interpreted back by that same myriad

Unfortunately that applies equally, if not more so, to RIS. Even EndNote does different things in each version. When there isn't a better option, we recommend them both pretty equally.

retorquere:
Did not know that. MODS then? Even CSL doesn't carry everything Zotero does.

dstillman:
I think MODS is better, but most programs don't support it. And no, I've personally never thought of CSL-JSON as an exchange format unless your concern is entirely what's citable.

The only (nearly?) lossless format is Zotero RDF, but that's a nightmare to support and not realistically going to be used by anything else. In theory something with no knowledge of Zotero that supported various RDF ontologies would be able import some data from it, but in the real world it's pretty much just for moving stuff between Zotero installs.

Our API JSON is entirely lossless and much easier to support, but that's tightly coupled to our data model and not something I've ever particularly wanted to expose as an exchange format.

zuphilip:
@dstillman Could the paths in BibTeX to the attachments been made optional? I see the complete path on my computer or network drive to this attachment, which I may not want to share with someone else. E.g.
```bibtex
file = {:C$\backslash$:/Users/amitrani/Documents/Mendeley/Yang, Leong - 2017 - Using Reward Programs as Public Policy Insights from a Field Experiment on Subway Trains.pdf:pdf},
```

dstillman:
If BibTeX is a central part of your workflow, you should be using BBT. Zotero isn't a BibTeX manager. As @retorquere said, this is already possible in BBT.

mirh:
> The people using BibTeX in Zotero without BBT generally just use it as a standard exchange format, and I don't think this sort of configurability is necessary or desirable for that kind of usage. Data doesn't have to be used in all contexts, but throwing it out seems like bad practice to me in most situations.

This is all legit and dandy
But for the love of me I cannot understand what "file" gives. It specifically stores *my own* data path to the file, and no way in hell that other people out there are gonna have the same document in the same position.

retorquere:
You can still use JabRef on the same system, and it would understand those paths.

dstillman:
> Could the paths in BibTeX to the attachments been made optional? I see the complete path on my computer or network drive to this attachment, which I may not want to share with someone else.

Sorry, missed this. That's a fair point, and I'd be fine with an option to exclude file paths. The problem is that there's already an "Export FIles" option, and the file paths option would really only make sense when "Export FIles" was disabled. But I don't think we have that sort of conditional configurability in the export dialog now, so we would have to either add generic handling to allow translators to define conditional options or hard-code some logic that applies specifically to "Export Files".

UI-wise, I guess when "Export Files" was checked, "Include File Paths" below it would be disabled and checked, and if "Export Files" was unchecked, "Include Files Paths" would become enabled and could be unchecked by the user?

I'll reopen this for this specific issue.

zuphilip:
The export dialog for BibTeX looks currently like this:

export-dialog

The new option could be named to `Export Paths to the Files`. Then there are four options to toggle these two things:
1. Check `Export Files` AND check `Export Paths to Files` (=current behavior where `Export Files` is checked)
2. Check `Export Files` but NOT check `Export Paths to Files`
3. NOT check `Export Files` but check `Export Paths to Files` (=current behavior where `Export Files` is NOT checked)
4. NOT check `Export Files` AND NOT check `Export Paths to Files`

The path is then either a local path to the newly exported files or the global path where the files are at the moment. The current available options are number 1 and 3, and you still only need to click once to switch between those. Number 2 may be a strange choice, because you want to export the files but not export paths to the files, but why not provide that option as well? 

The difference to what you suggested is, that you assume implicitly that if someone wants to export files then they also want to export the paths to these files. This has now just clicked explicitly on. But it avoids any conditional behavior this way. What do you think?

retorquere:
An alternative would be to offer a 3-option dropdown menu. Option 2 doesn't make much sense to me tbh.

zuphilip:
Another request from the forum to make it possible to toggle of the paths in BibTeX: https://forums.zotero.org/discussion/80399/dont-export-file-field-to-bibtex

PR #2035, opened 2019-10-22
Added translator for Public broadcasting of Latvia (lsm.lv)

papuass:


Public broadcasting of Latvia.js
```diff
@@ -0,0 +1,335 @@
+{
+	"translatorID": "dab3903a-0289-46f8-a74a-eccf7bdb86b6",
+	"label": "Public broadcasting of Latvia",
```
zuphilip: You should be more specific here. I would suggest to use here `Latvijas Sabiedriskie Mediji` which also resolves to the acronym of the url `lsm`.

Public broadcasting of Latvia.js
```diff
@@ -0,0 +1,335 @@
...
+	return date.toISOString().split('T')[0];
+}
+
+function scrape(doc, url) {
+	var translator = Zotero.loadTranslator('web');
+	// Embedded Metadata
+	translator.setTranslator('951c027d-74ac-47d4-a107-9c3069ab7b48');
+
+	translator.setHandler('itemDone', function (obj, item) {
+		item.publicationTitle = ZU.xpathText(doc, '//section[@class="globalnav__item globalnav__item-sitename"]');
```
zuphilip: Can this be anything else then `Latvijas Sabiedriskie Mediji`?

All test cases have the same value and maybe we should just set it automatic to this.

Public broadcasting of Latvia.js
```diff
@@ -0,0 +1,335 @@
...
+						author = author.substring(0, author.indexOf('('));
+					}
+					item.creators.push(ZU.cleanAuthor(author, "author"));
+				}
+			}
+		}
+		
+		// Date string is of the format:
+		// Publicerat onsdag 15 november 2017 kl 00.11
+		// /html/body/main/section[8]/div/div[1]/section[1]/article/div[3]/section[1]
```
zuphilip: Delete old code

Public broadcasting of Latvia.js
```diff
@@ -0,0 +1,335 @@
...
+		var dateString = ZU.xpathText(doc, '//div[@class="article__info"]/section[1]');
+		if (dateString.indexOf('Šodien') > 0) {
+			item.date = formatDate(new Date());
+		} else if (dateString.indexOf('Vakar') > 0) {
+			item.date = formatDate(new Date(new Date().setDate(new Date().getDate() - 1)));
```
zuphilip: I guess this strings means today and yesterday. Please add an English comment for this. Moreover, the second one looks too complicated. I guess that similar things like in https://github.com/zotero/zotero/blob/eca2822651ac191752e0f055eb5c06d386dd5197/chrome/content/zotero/xpcom/date.js#L276 should work here.

Public broadcasting of Latvia.js
```diff
@@ -0,0 +1,335 @@
...
+					jūlijs: "07",
+					augusts: "08",
+					septembris: "09",
+					oktobris: "10",
+					novembris: "11",
+					decembris: "12"
+				};
+				var month = months[dateParts[2]];
+				var day = dateParts[1];
+				item.date = year + "-" + month + "-" + day;
```
zuphilip: It looks that you are re-implementing things which should already be covered with Zotero's date functionality. However, Latvian terms might be missing. @dstillman Should this go into the file https://github.com/zotero/zotero/blob/master/resource/schema/dateFormats.json instead? Then use `ZU.strToISO()`.

Public broadcasting of Latvia.js
```diff
@@ -0,0 +1,335 @@
...
+			}
+		}
+		item.tags = [];
```
zuphilip: Why are you deleting any possible tags?

Public broadcasting of Latvia.js
```diff
@@ -0,0 +1,335 @@
...
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+function detectWeb(doc, url) {
+	if (url.includes('/raksts/')) {
+		return "newspaperArticle";
+	}
+	return false;
```
zuphilip: You could also look at some multiples pages, e.g. search, but I guess this is not in your focus.

papuass:
Thanks, will get on it. Most of you suggestions seem to be valid concerns.

zuphilip:
It would be great if you can also make sure that the other language versions work within the same translator. I hope that they the same XPath can be used and possibly only small adjustments are needed.

@dstillman Let us know about parsing the dates in Latvian (see my question above).

AbeJellinek:
Hi @papuass, this looks great and I'd love to get it merged! Any progress?

AbeJellinek:
@papuass: Had a chance to take a look at this?

Issue #1913, opened 2019-03-28
Remove attr()/text() lines


dstillman:
These lines cause spurious errors in ESLint and interfere with `--fix`.

The shims aren't necessary in Zotero 5, but removing them will break translators in any remaining Zotero 4 installations, so before removing these lines we should stop pushing translator updates to Zotero 4. We need to do this anyway for [async translators](https://github.com/zotero/zotero/pull/1609).

dstillman:
Translators are no longer being sent to <5.0.34, so we can now remove these lines from translators.

retorquere:
https://github.com/zotero/translators/pull/1961

retorquere:
So does this mean we're getting async translators soon?

dstillman:
OK, these actually can't be removed yet when they need to work on pages other than the initial one.

There's both a conceptual problem and a technical one:

1) The `doc` can't be implicit if the function is supposed to be operating on something other than the original page. The function will automatically treat the first element as `docOrElem`, but the plan was for that to always be implicit, which doesn't make sense, and since `doc` will sometimes be needed, it seems error-prone to allow it to ever be omitted.

2) Passing of a `doc` that's not the original page doesn't seem to currently work in Zotero — I get an `Opaque` object in the debugger. It probably does work in the Connector and translation-server, and it will work once we remove the translator sandbox, but until that's done and we cut off translator updates for older versions, we'll need to rely on the shims.

dstillman:
While we're rethinking these, we could also consider always returning a string for these, as @retorquere was doing in https://github.com/zotero/translators/pull/1964, and we could consider trimming automatically. I'm not sure if there's any value in returning `null` over an empty string. Does a translator ever need to know that an element didn't exist instead of its being empty?

dstillman:
Coming back to this, I'm not really seeing the optional `docOrElem` as a problem. Maybe I'm overlooking something, but it seems like a reasonable optimization for the probable majority of situations where the main document is the one being used, and being able to just use `text('.foo')` will be really nice.

We've long since switched to the different sandbox approach, so the `Opaque` problem no longer occurs.

I do think trimming automatically and always returning a string instead of `null` makes sense — a translator can always use `querySelector()` if it really needs to know if a node exists. But even after we add that, translators won't be able to rely on that behavior for a while, so we'll need to be careful to watch for problems there, or translators will need to include a v3 polyfill.

[...]

PR #2037, opened 2019-10-24
New Emerald Insight Translator

### LINKED ISSUE #2036: Emerald Journals
They changed their site -- it think we used to cover this with one of the default translators, but doesn't look like that's currently working. EM imports nicely but without PDF

https://www.emerald.com/insight/content/doi/10.1108/IJRDM-08-2017-0180/full/html

adam3smith:
Closes #2036

@dstillman do we still need to include `attr` and `text` code in translators? If not, can we add them to linting as defined?

### DIFF
--- a/Emerald Insight.js
+++ b/Emerald Insight.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2019-10-28 02:10:06"
+	"lastUpdated": "2019-10-29 01:43:27"
 }
 
 /*
@@ -40,7 +40,7 @@ function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelec
 
 
 function detectWeb(doc, url) {
-	// ensure that we only detect on items where Scrape will (most likely) work
+	// ensure that we only detect where scrape will (most likely) work
 	if (url.includes('/content/doi/') && (url.search(/\/(10\.[^#?/]+\/[^#?/]+)\//) != -1 || url.includes("/full"))) {
 		if (attr(doc, 'meta[name="dc.Type"]', 'content') == "book-part") {
 			return "bookSection";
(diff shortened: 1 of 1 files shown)


dstillman:
> do we still need to include attr and text code in translators? If not, can we add them to linting as defined?

[We do](https://github.com/zotero/translators/issues/1913#issuecomment-505618869), unfortunately.

Emerald Insight.js
```diff
@@ -0,0 +1,240 @@
...
+	var DOI = url.match(/\/(10\.[^#?/]+\/[^#?/]+)\//)[1];
+	var risURL = "/insight/content/doi/" + DOI + "/full/ris";
+	// Z.debug(risURL);
+	var pdfURL = attr(doc, 'a.intent_pdf_link', 'href');
+	// Z.debug("pdfURL: " + pdfURL);
+	ZU.doGet(risURL, function (text) {
+		// they number authors in their RIS...
+		text = text.replace(/A\d+\s+-/g, "AU  -");
+		// add a comma after the last name
+		text = text.replace(/AU\s\s-\s(\w+)/g, "AU  - $1,");
```
zuphilip: I think we should test here first that no comma is present at the moment (because they might fix that at some point). Moreover, can you extend the comment for this a little more?

It seems to me that they are always treating the last word of the author string as the last name and everything else as the first name(s), which they then put into RIS data as "lastname firstname(s)" without a comma. The splitting does not need to be correct, e.g. https://www.emerald.com/insight/content/doi/10.1108/S1572-832320170000026008/full/html , but it is the only thing we can get from their metadata.

Emerald Insight.js
```diff
@@ -0,0 +1,240 @@
...
+				"publicationTitle": "International Journal of Prisoner Health",
+				"url": "https://doi.org/10.1108/IJPH-07-2016-0028",
+				"volume": "13",
+				"attachments": [
+					{
+						"title": "Full Text PDF",
+						"mimeType": "application/pdf"
+					}
+				],
+				"tags": [],
```
zuphilip: There are keywords which could be saved as tags, i.e. look for `doc.querySelectorAll('li .intent_text');`.

Emerald Insight.js
```diff
@@ -0,0 +1,240 @@
...
+						"firstName": "Brandon",
+						"creatorType": "author"
+					},
+					{
+						"lastName": "Allen",
+						"firstName": "Scott A.",
+						"creatorType": "author"
+					}
+				],
+				"date": "January 1, 2017",
```
zuphilip: Should we make this language independent with `ZU.strToISO`? (Maybe even in the RIS translator itself?)

Emerald Insight.js
```diff
@@ -0,0 +1,240 @@
...
+					},
+					{
+						"lastName": "Malone",
+						"firstName": "Stephanie",
+						"creatorType": "author"
+					}
+				],
+				"date": "January 1, 2015",
+				"ISBN": "9781784415877 9781784415884",
+				"abstractNote": "AbstractOriginality/value\nThis technique creates opportunities for students to have unique assignments encouraging student to student teaching and can be applied to assignments in any accounting course (undergraduate and graduate). This testing method has been used in Intermediate I and II, Individual Taxation, and Corporate Taxation.",
```
zuphilip: Maybe it is cleaner to take here
```js
ZU.cleanInternal(doc.getElementById('abstract').textContent);
```
for the abstract. See also https://www.emerald.com/insight/content/doi/10.1108/07419051111154758/full/html for more subcontent under abstract.

Emerald Insight.js
```diff
@@ -0,0 +1,240 @@
...
+					{
+						"title": "Snapshot"
+					}
+				],
+				"tags": [],
+				"notes": [],
+				"seeAlso": []
+			}
+		]
+	},
```
zuphilip: Can we also add `en` as language for all publications from Emerald?

Emerald Insight.js
adam3smith: Done and done.

Emerald Insight.js
```diff
@@ -0,0 +1,240 @@
...
+		});
+	}
+	else {
+		scrape(doc, url);
+	}
+}
+
+
+function scrape(doc, url) {
+	var DOI = url.match(/\/(10\.[^#?/]+\/[^#?/]+)\//)[1];
```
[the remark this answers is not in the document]
adam3smith: improved check in `detectWeb` and added one more fallback here. Also added the scrape function to work on PDF pages.

Emerald Insight.js
adam3smith: I'll impement this here -- let's think through some RIS scenarios together. Generally sounds like we should.

Emerald Insight.js
```diff
@@ -0,0 +1,240 @@
...
+		// search results
+		rows = doc.querySelectorAll('h2>a.intent_link');
+	}
+	else {
+		// journal issue
+		rows = doc.querySelectorAll('.intent_issue_item h4>a');
+		if (!rows.length) {
+			// book
+			rows = doc.querySelectorAll('li.intent_book_chapter>a');
+		}
```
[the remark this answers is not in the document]
adam3smith: done

Emerald Insight.js
adam3smith: Done (won't use this when scraping from PDF, which I deemed acceptable)

Emerald Insight.js
adam3smith: Done. Won't use when scraping from PDF view which I think is acceptable.

Emerald Insight.js
adam3smith: Unfortunately not. Emerald also publishes in other languages (I found German and Spanish at a minimum) and there are no usable language tag on those.

adam3smith:
Thanks -- take another look please if you have a moment.

Emerald Insight.js
zuphilip: Okay, that is pity but it is the publisher's fault we cannot do better.

zuphilip:
That all looks good, but can you update the test cases as well?

zuphilip:
Thank you very much @adam3smith !

PR #2039, opened 2019-10-26
Douban: deal with url item and update unit tests

kun-g:
Safari can't open this [link](https://www.douban.com/url/9552666/) where Chrome can.
This is why some [Dou List](https://www.douban.com/doulist/120664512/) which contains URL item can't be scraped (I guess)
I fixed this issue by add the URL item as a webpage without snapshot.

### DIFF
--- a/Douban.js
+++ b/Douban.js
@@ -9,23 +9,23 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2017-01-01 16:52:39"
+	"lastUpdated": "2019-10-28 20:20:19"
 }
 
 /*
    Douban Translator
    Copyright (C) 2009-2010 TAO Cheng, acestrong@gmail.com
-   
+
    This program is free software: you can redistribute it and/or modify
    it under the terms of the GNU General Public License as published by
    the Free Software Foundation, either version 3 of the License, or
    (at your option) any later version.
-   
+
    This program is distributed in the hope that it will be useful,
    but WITHOUT ANY WARRANTY; without even the implied warranty of
    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
    GNU General Public License for more details.
-   
+
    You should have received a copy of the GNU General Public License
    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
@@ -62,9 +62,10 @@ function trimMultispace(text) {
 
 // #############################
 // ##### Scraper functions #####
-// ############################# 
+// #############################
 
 function scrapeAndParse(doc, url) {
+	// Z.debug({ url })
 Zotero.Utilities.HTTP.doGet(url, function(page){
 	//Z.debug(page)
 	var pattern;
@@ -82,7 +83,7 @@ Zotero.Utilities.HTTP.doGet(url, function(page){
 		newItem.title = Zotero.Utilities.trim(trimTags(title));
 //		Zotero.debug("title: "+title);
 	}
-	
+
 	// 又名
 	pattern = /<span [^>]*?>又名:(.*?)<\/span>/;
 	if (pattern.test(page)) {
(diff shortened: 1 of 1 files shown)


zuphilip:
For https://www.douban.com/url/9552666/ I see an error page in Firefox AND Chrome and then after some seconds get redirected to the main website. I guess that you are logged in one browser and only then you have access to your bookmarked url from that list.

I don't see much advantages to save just the url and title for such a website in Zotero. Moreover, most of these links will not work, when you are not the one which created the list and you are logged in.

We could simply exclude these cases and not even make make them appear in the multiple list, i.e. something like
```js
function doWeb(doc, url) {
	var articles = [];
	var items = {};
		var titles = ZU.xpath(doc, '//div[@class="title"]/a');
		for (let title of titles) {
			if (!title.href || title.href.includes("/url/")) continue;
			items[title.href] = title.textContent;
		}
		Zotero.selectItems(items, function (items) {
```

Then the Dou list will still work.

kun-g:
> For https://www.douban.com/url/9552666/ I see an error page in Firefox AND Chrome and then after some seconds get redirected to the main website. I guess that you are logged in one browser and only then you have access to your bookmarked url from that list.
> 
> I don't see much advantages to save just the url and title for such a website in Zotero. Moreover, most of these links will not work, when you are not the one which created the list and you are logged in.
> 
> We could simply exclude these cases and not even make make them appear in the multiple list, i.e. something like
> 
> ```js
> function doWeb(doc, url) {
> 	var articles = [];
> 	var items = {};
> 	if (detectWeb(doc, url) == "multiple") {
> 		var titles = ZU.xpath(doc, '//div[@class="title"]/a');
> 		for (let title of titles) {
> 			if (!title.href || title.href.includes("/url/")) continue;
> 			items[title.href] = title.textContent;
> 		}
> 		Zotero.selectItems(items, function (items) {
> ```
> 
> Then the Dou list will still work.
You are right, I logged out and fail to open the page in Chrome.

I've considered this approach at the beginning. Ignore the link is easy and fast, but Dou list is a collection of a topic, the links are part of the topics.
In the other hand, saving the link offers choices to the users, select it or not, the translator works fine.

So, I still think saving the link is a better solution.

zuphilip:
Providing such an option will give a promise to the users of Zotero that this will work fine. However, that is not the case here for at least two main reasons: 1) it is likely that the link will not open at all, because you are not logged in or you are not the user saved that link; 2) the link is not one that would be need to citing, e.g. you would provide that link https://www.zotero.org/support/dev/translators in any citation of that page and never https://www.douban.com/url/9552666/.

Thus, it is better to exclude this completely in multiples. If any user wants to save the bookmarked webpage then it is much better if they click on the link and try to save it from that page rather than from the Dou list which does not contain enough useful information.

kun-g:
OK, got your point 👍

Douban.js
```diff
@@ -2,30 +2,30 @@
 	"translatorID": "fc353b26-8911-4c34-9196-f6f567c93901",
 	"label": "Douban",
 	"creator": "Ace Strong<acestrong@gmail.com>",
-	"target": "^https?://(www|book)\\.douban\\.com/(subject|doulist|people/[a-zA-Z._]*/(do|wish|collect)|.*?status=(do|wish|collect)|group/[0-9]*?/collection|tag)",
+	"target": "^https?://(www|book)\\.douban\\.com/(subject|doulist|people|url/[a-zA-Z._]*/(do|wish|collect)|.*?status=(do|wish|collect)|group/[0-9]*?/collection|tag)",
```
zuphilip: That is not anymore needed, right?

Douban.js
```diff
@@ -229,13 +233,9 @@ function doWeb(doc, url) {
...
 	}
 }
+
 /** BEGIN TEST CASES **/
 var testCases = [
-	{
-		"type": "web",
-		"url": "http://book.douban.com/subject_search?search_text=Murakami&cat=1001",
-		"items": "multiple"
-	},
```
zuphilip: Is that not anymore working or why did you delete it?

Douban.js
kun-g: Yes，I'll remove it.

Douban.js
kun-g: It's not working. It's using some delayed loading technique, I can't solve it yet.

zuphilip: Okay, I will add a note about this.

zuphilip:
Thank you @kun-g !

PR #2043, opened 2019-10-27
Add MIT Press translator

GuyAglionby:
As title. Only thing I'm not sure about is how to handle the book subtitle. For example, for title "Elements of Causal Inference" with subtitle "Foundations and Learning Algorithms", would the appropriate `title` be "Elements of Causal Inference: Foundations and Learning Algorithms" and `shortTitle` "Elements of Causal Inference"? Or should the `title` just be "Elements of Causal Inference"

Thanks!

### DIFF
--- a/MIT Press Books.js
+++ b/MIT Press Books.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2019-12-28 18:25:46"
+	"lastUpdated": "2021-06-28 22:21:37"
 }
 
 /*
@@ -76,13 +76,7 @@ function scrape(doc, url) {
 	let editionRegex = /, (([\w ]+) edition( [\w ]+)?)/i;
 	let matchedEdition = title.match(editionRegex);
 	if (matchedEdition) {
-		// Only include "Edition" in the field if there's text after it
-		if (matchedEdition[3]) {
-			item.edition = matchedEdition[1];
-		}
-		else {
-			item.edition = matchedEdition[2];
-		}
+		item.edition = cleanEdition(matchedEdition[1]);
 		title = title.replace(matchedEdition[0], '');
 	}
 	
@@ -147,7 +141,68 @@ function scrape(doc, url) {
 		}
 	}
 	
-	item.complete();
+	let seriesURL = attr(doc, '.book__series a', 'href');
+	if (seriesURL) {
+		ZU.processDocuments(seriesURL, function (seriesDoc) {
+			let seriesEditors = text(seriesDoc, '.series__editors')
+				.split(/,| and /);
+			for (let seriesEditor of seriesEditors) {
+				let creator = ZU.cleanAuthor(seriesEditor, 'seriesEditor');
+				
+				// sometimes series editors are also editors of individual
+				// volumes, and it doesn't make sense to include the same name
+				// twice. not an efficient approach but we're dealing with 4-5
+				// contributors max.
+				let duplicate = false;
+				for (let other of item.creators) {
+					if (other.firstName == creator.firstName && other.lastName == creator.lastName) {
+						duplicate = true;
+					}
+				}
+				
+				if (!duplicate) {
+					item.creators.push(creator);
+				}
+			}
+			item.complete();
+		});
+	}
+	else {
+		item.complete();
+	}
+}
+
+function cleanEdition(text) {
+	if (!text) return text;
+	
+	// from Taylor & Francis eBooks translator, slightly adapted
+	
+	const ordinals = {
+		first: "1",
+		second: "2",
+		third: "3",
+		fourth: "4",
+		fifth: "5",
+		sixth: "6",
+		seventh: "7",
+		eighth: "8",
+		ninth: "9",
+		tenth: "10"
+	};
+	
+	text = ZU.trimInternal(text).replace(/[[\]]/g, '');
+	// this somewhat complicated regex tries to isolate the number (spelled out
+	// or not) and make sure that it isn't followed by any extra info
+	let matches = text
+		.match(/^(?:(?:([0-9]+)(?:st|nd|rd|th)?)|(first|second|third|fourth|fifth|sixth|seventh|eighth|ninth|tenth))(?:\s?ed?\.?|\sedition)?$/i);
+	if (matches) {
+		let edition = matches[1] || matches[2];
+		edition = ordinals[edition.toLowerCase()] || edition;
+		return edition == "1" ? null : edition;
+	}
+	else {
+		return text;
+	}
 }
 
 function getSearchResults(doc, checkOnly) {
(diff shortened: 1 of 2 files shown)


MIT Press.js
```diff
@@ -0,0 +1,206 @@
+{
+	"translatorID": "ac277fbe-000c-46da-b145-fbe799d17eda",
+	"label": "MIT Press",
```
adam3smith: Should be called "MIT Press Books" to distinguish it from existing "MIT Press Journals" translator

MIT Press.js
```diff
@@ -0,0 +1,206 @@
...
+	"creator": "Guy Aglionby",
+	"target": "https://(www\\.)?mitpress\\.mit\\.edu/",
```
adam3smith: again, to distinguish it from journals, make the regex more restrictive (just re-using the bits from detect Web, which you may then be able to simplify)

MIT Press.js
```diff
@@ -0,0 +1,206 @@
...
+		});
+	}
+	else {
+		scrape(doc);
+	}
+}
+
+function scrape(doc, url) {
+	let authorsElement = ZU.xpath(doc, '//span[@class = "book__authors"]/p[1]/a');
+	let authors = authorsElement.map(function (author) {
```
adam3smith: Could we try to distinguish editors by using the "Edited by" wording, e.g. https://mitpress.mit.edu/books/construction-site-possible-worlds

adam3smith: Example of book with editor and author: https://mitpress.mit.edu/books/kings-and-things
Example of book with author and translator: https://mitpress.mit.edu/books/ribbon-olympias-throat

MIT Press.js
```diff
@@ -0,0 +1,206 @@
...
+	let openAccessUrl = ZU.xpathText(doc, '//div[contains(@class, "open-access")]/a/@href');
+	let series = ZU.xpathText(doc, '//p[@class = "book__series"]/a');
+	
+	let item = new Zotero.Item('book');
+	item.creators = authors;
+	item.url = url;
+	item.place = 'Cambridge, MA, USA';
+	item.publisher = 'MIT Press';
+	item.language = 'en';
+	item.title = ZU.xpathText(doc, '//h1[@class = "book__title"]').trim();
```
adam3smith: I think we do want the subtitle after : as per your question. E.g. for the causal inference book, that's how it is cataloged by LoC.

GuyAglionby:
Thanks for the feedback! Made changes as requested. Added some new tests to include the multiple contributor types + cases where two elements for the book series were present.

MIT Press Books.js
```diff
@@ -0,0 +1,298 @@
...
+	item.publisher = 'MIT Press';
+	item.language = 'en';
+	item.date = ZU.xpathText(doc, '(//time[@property = "publishDate"]/@content)[1]');
+	item.ISBN = ZU.xpathText(doc, '(//span[@property = "isbn"])[1]');
+	item.numPages = ZU.xpathText(doc, '(//span[@property = "numPages"])[1]');
+	
+	let title = ZU.xpathText(doc, '//h1[@class = "book__title"]').trim();
+	let subtitle = ZU.xpathText(doc, '//h2[@class = "book__subtitle"]').trim();
+	if (subtitle) {
+		item.title = [title, subtitle].join(': ');
```
zuphilip: Move the `trim()` from the `subtitle` inside the `if` block, because otherwise there is a potential that you try to trim from `null` which will fail.

MIT Press Books.js
```diff
@@ -0,0 +1,298 @@
...
+				"attachments": [
+					{
+						"title": "Full Text PDF",
+						"mimeType": "application/pdf"
+					}
+				],
+				"tags": [],
+				"notes": [],
+				"seeAlso": []
+			}
```
zuphilip: You could also add the short abstract from the `div` of class `book__blurb`.

MIT Press Books.js
```diff
@@ -0,0 +1,298 @@
...
+					}
+				],
+				"date": "2020-05-12",
+				"ISBN": "9781913029579",
+				"language": "en",
+				"libraryCatalog": "MIT Press Books",
+				"numPages": "136",
+				"place": "Cambridge, MA, USA",
+				"publisher": "MIT Press",
+				"series": "Urbanomic",
```
zuphilip: Actually, this looks like we should better save this as the publisher info, cf. https://www.urbanomic.com/book/construction-site-possible-worlds/ .

MIT Press Books.js
```diff
@@ -0,0 +1,298 @@
...
+		"url": "https://mitpress.mit.edu/contributors/ian-goodfellow",
+		"items": "multiple"
+	},
+	{
+		"type": "web",
+		"url": "https://mitpress.mit.edu/books/sciences-artificial-reissue-third-edition-new-introduction-john-laird",
+		"items": [
+			{
+				"itemType": "book",
+				"title": "The Sciences of the Artificial, Reissue Of The Third Edition With A New Introduction By John Laird",
```
zuphilip: It looks like the edition (as well as volume) info is saved as part of the title. This would be good to try to split it of, if we can. This example here looks extremely challenging and therefore I am not sure that we should tackle it at all. But easier example are maybe
- Foundations of Machine Learning, Second Edition: https://mitpress.mit.edu/books/foundations-machine-learning-second-edition
- Collapse, Volume 8: https://mitpress.mit.edu/books/collapse-volume-8
- Introduction to Algorithms, Third Edition: https://mitpress.mit.edu/books/introduction-algorithms-third-edition

Can we fix at least these ans similar examples?

GuyAglionby:
Thanks for the comments and sorry for the delay here. I've updated this so that the publisher and series are appropriately set, and included the blurb as an abstract. 

I had a look on the site and they're very regular on how volumes and editions are noted -- they're always of the form `Book Title, Volume \d+` or `Book Title, [\w ]+Edition([\w ]+)?`. Volume I just extracted the number; for edition I had a look at what the other translators do and only include "Edition" if there's text after it (e.g. [Reissue Of The Third Edition With A New Introduction By John Laird](https://github.com/zotero/translators/pull/2043/commits/8a7195b94df2db6fc339e3589558ed1fc6a13fcb#diff-58d461527180df476ba1d3d619f51ea5R257)), otherwise exclude it (e.g. [Revised And Expanded](https://github.com/zotero/translators/pull/2043/commits/50d0ea1936778cbe96048a8e68d971d458b4532d#diff-58d461527180df476ba1d3d619f51ea5R516)). This is easily changed if another way would be preferable. In all cases all the text is removed from the title.

GuyAglionby:
@zuphilip I hope the changes address your concerns -- please let me know if there's anything else that needs doing!

GuyAglionby:
@adam3smith @zuphilip please let me know if there's anything else!

boan-anbo:
Bumping this as I was checking if anyone already made the MIT press translator. It's used pretty frequently. If there is anything I can help on top of @GuyAglionby's PR let me know.

AbeJellinek:
Sorry for the delay on this. Great translator, and very much needed. I tweaked the edition handling so that we get numbers, not spelled-out ordinals, and added series editors. I also removed a 404'd test.

AbeJellinek:
I think all is good @boan-anbo, thanks for offering help! I've just been slow to get to this one.

PR #2050, opened 2019-11-05
new translator for Duxiu

boan-anbo:
New translator for Duxiu (读秀; duxiu.com), a database for Chinese books. Currently for books only.

### DIFF
--- a/Duxiu.js
+++ b/Duxiu.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2019-11-05 14:51:51"
+	"lastUpdated": "2019-11-08 18:37:44"
 }
 
 /*
@@ -120,7 +120,7 @@ function scrapeAndParse(doc, url) {
 			// Zotero.debug(authorNames);
 			
 			// list of role titles. used to remove the role title from the name of the creator.
-			var titleMask = /本书主编$|本书副主编$|总主编$|总编辑$|总编$|编译$|编著$|主编$|副主编$|改编$|编$|著$|译$|选编$|摄影$|整理$|执笔$|合著$|撰$|编纂$|纂$|辑$|集注$|编辑$|原著$|主译$/;
+			var titleMask = /本书主编$|本书副主编$|总主编$|总编辑$|总编$|编译$|编著$|主编$|副主编$|改编$|编$|著$|译$|选编$|摄影$|整理$|执笔$|合著$|撰$|编纂$|纂$|辑$|集注$|编辑$|原著$|主译$|绘$/;
 
 			for (let i = 0; i < authorNames.length; i++) {
 				var assignedRole = "";
@@ -160,7 +160,9 @@ function scrapeAndParse(doc, url) {
 					case '著':
 					case '执笔':
 					case '撰':
+					case '绘':
 					case '纂':
+					case '摄影':
 					case '集解':
 					case '集注':
 						newItem.creators.push({ lastName: assignedName,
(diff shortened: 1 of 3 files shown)


boan-anbo:
Is there anything wrong with the request (my first pull), or is it unwanted? plz let me know. thx.

zuphilip:
I am sorry, but I haven't yet had the time to look at it closer. New translators need more time to review. Don't worry about the linting issues from Travis; we can deal with them later. I will try to look your PR soon...

Duxiu.js
```diff
@@ -0,0 +1,423 @@
...
+
+		newItem.complete();
+	});
+}
+
+// the list from which to pick the best role for a given creator. Do not add variants of strings that end with 著,译，编
+var rolelist = ["总主编", "总编辑", "总编", "编著", "编译", "编", "整理", "执笔", "译", "著", "撰", "纂", "集解", "辑", "编辑", "集注","绘"];
+
+function trimTags(text) {
+	return text.replace(/(<.*?>)|\t|\r|(隐藏更多)|&nbsp;|/g, "");
```
zuphilip: Look at `ZU.cleanTags()` and maybe also `ZU.trimInternal()` which should to a similar thing.

boan-anbo: Thanks so much @zuphilip ! I can make the corrections later today or tomorrow (such as double calling the website and unnecessary lines). But I do have a follow-up question about using RegEx:

I realized using all regex is not as stable, but the problem I had is that the Duxiu entry html structure is messy. The info section is like (if the Chinese characters are translated into English):
```
<dl>
<dt>TITLE</dt>
<dd>
PUBLICATION DATE : 2011.10&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
P&nbsp;A&nbsp;G&nbsp;E : 549</dd>
<dd>SERIES</dd>
<dd>ISBN: 12345678910</dd>
...
</dl>
```
I know I can use xPath to pick out `<dt>` (the title), but can I do it with the `<dd>` (all the other entries)? 

Or should I mix xPath and regex? Like pick out `<dl>` first and then parse all the `<dd>` using shorter regex?

This point was actually raised earlier on the forum by someone else: https://forums.zotero.org/discussion/76178/can-the-zotero-connector-catch-the-information-of-duxiu
And I quote from adamsmith:
> but as a data structure, it's a complete mess: all of this is in unnamed HTML nodes, so the only way to get to it is by parsing plain text using regular expressions, which is super fragile. It's doable, but it's not something I prioritize given how easily it breaks.

Which is why I started with all regex. But I would be happy to switch to xPath if possible.

And to answer your earlier question:
1. The database can be accessed fully with subscription and partially (few functionalities) with "trial option". To use trial option, it asks you to click on a button (the button below on the right) to store the session. This can be done in a sys browser, but not always in Scaffold browser. 
image

 2. In practice, it has been working alright for the past week in daily uses, and I have added support for journal articles. I'll wait for your comment on the question above before making another commit.

Thanks again!

zuphilip: Thank you very much! I click on one of the links below the blue box and can enter the database as well (for me the link was on the left). That is useful, as maintenance of this translator is then also easier.

Yes, a mix with xpath and regular expression might be the way to go here. On a general level I think it should be easy to get all the main information parts with xpaths and decide what they are about. The more fine-grained information could then be parsed with some regex, i.e. something like
```js
var informations= doc.querySelectorAll('dl>dd');
for (let info of informations) {
  let content = info.textContent;
  if (!content.includes(':')) continue;
  let field = content.split(':')[0].trim();
  let value = content.split(':').shift().join(':').trim();
  // information about publisher, place, year
  // e.g. 出版发行 : 上海：中华书局 , 1946.09
  if (field === "出版发行") {
    let m = value.match(...);
...
```
(Note, this code is not tested)

Let me know if you have more questions or when this is ready for another review.

ysmintor:
@boan-anbo You really done a great job for user who use Duxiu.

fanzhen1:
讀秀的translator好像失效了，不知道是zotero更新還是讀秀的原因。麻煩大佬有空看一下
image

PR #2051, opened 2019-11-07
Update xpaths for ACLWeb for change in site

GuyAglionby:
As title -- the site layout was updated.

### DIFF
--- a/ACLWeb.js
+++ b/ACLWeb.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2019-08-31 18:53:37"
+	"lastUpdated": "2019-11-07 17:17:11"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


ACLWeb.js
```diff
@@ -193,14 +193,14 @@ function getSearchResults(doc, url) {
 	let items = {};
 	if (url.includes('/search/')) {
 		// e.g. https://www.aclweb.org/anthology/search/?q=foo+bar
-		papers = ZU.xpath(doc, '//div[@class = "gsc-wrapper"]//td//a[contains(@class, "gs-title")]');
+		papers = ZU.xpath(doc, '//div[contains(@class, "gsc-webResult")]//div[@class = "gs-title"]/a');
```
zuphilip: Did you change to `@class=` instead of only `contains` by a special reasons? I would prefer to use conains for class attributes, as this is more stable also when another class is added in the future.

GuyAglionby: No special reason, but good point. Have changed back

zuphilip:
Thank you!

PR #2054, opened 2019-11-10
Twitter: Support localization where am/pm are lowercase

tfiers:
Tests run succesfully now.

### DIFF
--- a/Twitter.js
+++ b/Twitter.js
@@ -13,6 +13,8 @@
 }
 
 /*
+	***** BEGIN LICENSE BLOCK *****
+
    Twitter Translator
    Copyright (C) 2011 Avram Lyon, ajlyon@gmail.com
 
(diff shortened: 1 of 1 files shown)


dstillman:
I think this is a misunderstanding. I'm not sure what this fixes, but it certainly wouldn't add support for the new Twitter layout, which is a complete redesign of the site, with entirely different — and essentially unparseable — DOM. It seems like you're still viewing the old version.

dstillman:
And I would guess that the am/AM thing is non-US localization, so this wouldn't be an appropriate fix for that either. I still get "AM" in the old design (and the new one).

dstillman:
So while we'd take a patch to make the date detection case-insensitive (though it'd be good to figure out what's triggering the difference — `Accept-Language`, IP geolocation, a Twitter localization setting?), saving on Twitter will be broken in general for most people (possibly anyone logged in?) at this point.

zuphilip:
In the new Twitter UI I see either `<time datetime=...>` or `data-time-ms` which would be easier to parse. Are they not present in the old UI?

dstillman:
`data-time` and `data-time-ms` attributes do appear to exist in the old UI.

dstillman:
(And in the new UI, it's unfortunately not that simple. If you're focused on a reply, only the original tweet has a `<time>`. The reply you're focused on just has a `<span>` with gibberish CSS class names.)

zuphilip:
@dstillman I switched Zotero to English (UK), restart it, and then opened the Twitter website in Scaffold and see there the small `am`. It is large with the English ((not UK))  settings.

tfiers:
Hi @dstillman and @zuphilip. Thanks for the quick responses.
Last commit makes the date parsing case-insensitive.

I indeed got the old UI when running tests.

For the new UI with the gibberish CSS classes, seems like a combination of searching for <article> and navigating lots of div>div>div> hierarchies might do it?

zuphilip:
Thank you @tfiers for the support of different localization of am/pm! I squashed the commits together and merged it. :rocket:

The new website layout of Twitter is a different story, but we can continue this in the other issue.

PR #2055, opened 2019-11-11
Add translator for new NASA ADS 

thostetler:
New translator for the new NASA ADS site, using the same RIS export endpoint as the old translator.


NASA ADS 2.js
```diff
@@ -0,0 +1,185 @@
...
+		}
+	}
+	return entries;
+}
+
+function detectWeb(doc, url) {
+	if (url.includes("/search/")) {
+		return "multiple";
+	}
+	else if (url.includes("/abs/")) {
```
adam3smith: How hard would it be to detect some of the other content types here? It's not super critical to detect the right type, but if e.g. matching by the id would get us mostly there, it's worth the effort (e.g. it looks like all PhD Thesis have PhDT in the ID etc.

zuphilip:
Is the old translator for NASA ADS then still working? Or should we rather replace the old one with the code here? I don't like the label here "NASA ADS 2"

NASA ADS 2.js
thostetler: Yep, I can make that change.  I know for sure we can extract `PhDT` and `MsT`

thostetler:
@zuphilip The old system has been shutdown now, so the other translator won't work.  
We could replace the old translator I believe

dstillman:
OK, if this is a replacement let's do this as a modification of the existing file with the same translatorID and title. (You can make that change in this PR without opening a new one.)

thostetler:
@dstillman sure I can make that change

NASA ADS.js
```diff
@@ -1,22 +1,22 @@
 {
 	"translatorID": "7987b420-e8cb-4bea-8ef7-61c2377cd686",
 	"label": "NASA ADS",
-	"creator": "Philipp Zumstein, Asa Kusuma and Ramesh Srigiriraju",
-	"target": "^https?://(ukads|cdsads|ads|adsabs|esoads|adswww|www\\.ads)\\.(inasan|iucaa\\.ernet|nottingham\\.ac|harvard|eso|u-strasbg|nao\\.ac|astro\\.puc|bao\\.ac|on|kasi\\.re|grangenet|lipi\\.go|mao\\.kiev)\\.(edu|org|net|fr|jp|cl|id|uk|cn|ua|in|ru|br|kr)/(cgi-bin|abs)/",
+	"creator": "Philipp Zumstein, Asa Kusuma, Ramesh Srigiriraju and Tim Hostetler",
```
adam3smith: Since this is a rewrite, you should be the only creator I think

NASA ADS.js
```diff
@@ -1,22 +1,22 @@
...
-	"lastUpdated": "2018-02-17 17:39:33"
+	"lastUpdated": "2019-11-20 22:29:34"
 }
 
 /*
 	***** BEGIN LICENSE BLOCK *****
 
-	Copyright © 2018 Philipp Zumstein
-	
+	Copyright © 2019 Philipp Zumstein
```
adam3smith: this should be you accordingly

NASA ADS.js
```diff
@@ -35,131 +35,228 @@
...
-						mimeType: "application/pdf"
-					}];
-				}
-			}
-			item.complete();
-		});	
-		translator.translate();
-	});
+		if (bibstem.startsWith("MsT")) {
+			return "Masters thesis";
```
adam3smith: Only `thesis` is a valid Zotero item type. You could see if you can get the Master vs. Ph.D. thesis into the `type` field in Zotero

NASA ADS.js
```diff
@@ -35,131 +35,228 @@
...
+				"seeAlso": []
+			}
+		]
+	},
+	{
+		"type": "web",
+		"url": "https://ui.adsabs.harvard.edu/abs/2019PhDT........69B/abstract",
+		"items": [
+			{
+				"itemType": "Ph.D. thesis",
```
adam3smith: And this test would need to update accordingly

NASA ADS.js
thostetler: So I have a question on this, I updated the itemType returned here to be `thesis`.

```
const bibstem = extractId(url).slice(4);
if (bibstem.startsWith("MsT") || bibstem.startsWith("PhDT")) {
  return "thesis";
}
return "journalArticle";
```

I can get the correct output: `detectWeb returned type "thesis"` when running `detectWeb` from the editor.  But when I run the `doWeb` function I always get:

```
Returned item:
{
  "itemType": "journalArticle"
  ...
}
```
Is this being set by the RIS importer? or is there something that I'm forgetting to set prior to passing that text?

adam3smith: Yes, I'm pretty sure that's the RIS importer. You have 
`TY  - Thesis/Dissertation` in the RIS, which isn't valid and Zotero uses journalArticle as a generic fallback for those cases. It should instead be
`TY  - THES` It'd be best to change it in the RIS so that everyone benefits, but if that's impossible, you can overried the importers item type in the translator.

zuphilip:
Thank you for replacing the file now as the old system is down. I let the remaining review to @adam3smith .

NASA ADS.js
thostetler: @adam3smith So I talked with some people on the team about this, it seems like the `THES` is just an abbreviation and that the longer version is also valid (at least this is what I've heard) so I don't think it will be changing on our end.
I've updated the PR though to do it manually, hopefully I did it the right way.
Thanks!

adam3smith:
Great, thanks so much @thostetler

PR #2056, opened 2019-11-15
Update EUR-Lex.js to translate caselaw as well

hueyy:
Case law such as https://eur-lex.europa.eu/legal-content/EN/ALL/?uri=CELEX:61962CJ0026 was not captured. It is now captured.

### DIFF
--- a/EUR-Lex.js
+++ b/EUR-Lex.js
@@ -39,6 +39,15 @@
 // attr()/text() v2
 function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null;}function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null;}
 
+function getQueryParam(url, param){
+	const queryString = url.split("?")[1]
+	const vars = queryString.split("&")
+	for (var i=0;i<vars.length;i++) {
+           const pair = vars[i].split("=");
+           if(pair[0] == param){return pair[1];}
+   }
+   return(false);
+}
 
 // the eli resource types are described at:
 // http://publications.europa.eu/mdr/resource/authority/resource-type/html/resourcetypes-eng.html
@@ -46,13 +55,27 @@ var typeMapping = {
 	'DIR': 'bill', // directive
 	'REG': 'statute', // regulation
 	'DEC': 'statute', // decision
-	'RECO': 'report', // recommodation
-	'OPI': 'report' // opinion
+	'RECO': 'report', // recommendation
+	'OPI': 'report', // opinion
+	'CASE': 'case', // case
+	'CASE_LAW': 'case', // case law
+	'OPIN_AG': 'case', // advocate general's opinion
+	'OPIN_CASE': 'case', // advocate general's opinion
+	'VIEW_AG': 'case', // advocate general's opinion
 };
 
 
 function detectWeb(doc, url) {
-	var eliTypeURI = attr(doc, 'meta[property="eli:type_document"]', 'resource');
+	const celex = getQueryParam(url, 'uri')
+	if(celex){
+		const sector = celex.slice(6, 7)
+		if(sector === '6'){ // caselaw
+			return 'case'
+		}
+	}
+	
+	
+	const eliTypeURI = attr(doc, 'meta[property="eli:type_document"]', 'resource');
 	if (eliTypeURI) {
 		var eliType = eliTypeURI.split('/').pop();
 		var eliCategory = eliType.split('_')[0];
@@ -64,7 +87,7 @@ function detectWeb(doc, url) {
 		}
 	} else if (getSearchResults(doc, true)) {
 		return "multiple";
-	}
+	}	
 }
 
 
@@ -105,6 +128,8 @@ function doWeb(doc, url) {
 			}
 			ZU.processDocuments(articles, scrape);
 		});
+	} else if (detectWeb(doc, url) === "case"){
+		scrapeCase(doc, url);
 	} else {
 		scrape(doc, url);
 	}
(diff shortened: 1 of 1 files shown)


EUR-Lex.js
```diff
@@ -37,42 +37,74 @@
 
 
 // attr()/text() v2
-function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null;}function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null;}
+function attr(docOrElem, selector, attr, index) {
+	var elem = index ? docOrElem.querySelectorAll(selector).item(index) : docOrElem.querySelector(selector); return elem ? elem.getAttribute(attr) : null;
+}
```
zuphilip: No, this can stay minimized as it was before. Just add the comment `// eslint-disable-next-line` before it, see also https://github.com/zuphilip/translators/wiki/Common-code-blocks-for-translators#replacements-for-xpath

EUR-Lex.js
```diff
@@ -37,42 +37,74 @@
...
+	DIR: 'bill', // directive
+	REG: 'statute', // regulation
+	DEC: 'statute', // decision
+	RECO: 'report', // recommendation
+	OPI: 'report', // opinion
+	CASE: 'case', // case
+	CASE_LAW: 'case', // case law
+	OPIN_AG: 'case', // advocate general's opinion
+	OPIN_CASE: 'case', // advocate general's opinion
+	VIEW_AG: 'case', // advocate general's opinion
```
zuphilip: Are all these different types in EUR-lex mapped to case?

EUR-Lex.js
```diff
@@ -37,42 +37,74 @@
...
 };
 
 
 function detectWeb(doc, url) {
-	var eliTypeURI = attr(doc, 'meta[property="eli:type_document"]', 'resource');
+	const celex = getQueryParam(url, 'uri');
+	if (celex) {
+		const sector = celex.slice(6, 7);
+		if (sector === '6') { // caselaw
```
zuphilip: Simplify
```js
if (celex && celex.charAt(6) === '6') {
```
Can you then also explain more in the comment what is the reason for this?

EUR-Lex.js
```diff
@@ -37,42 +37,74 @@
...
 		var eliType = eliTypeURI.split('/').pop();
 		var eliCategory = eliType.split('_')[0];
 		var type = typeMapping[eliCategory];
 		if (type) {
 			return type;
-		} else {
+		}
+		else {
 			Z.debug("Unknown eliType: " + eliType);
+			return null;
```
zuphilip: `return false;` instead here.

EUR-Lex.js
```diff
@@ -37,42 +37,74 @@
...
 		}
-	} else if (getSearchResults(doc, true)) {
+	}
+	else if (getSearchResults(doc, true)) {
 		return "multiple";
 	}
+
+	return null;
```
zuphilip: dito

EUR-Lex.js
```diff
@@ -103,96 +135,125 @@ function doWeb(doc, url) {
 			for (var i in items) {
 				articles.push(i);
 			}
-			ZU.processDocuments(articles, scrape);
+			return ZU.processDocuments(articles, scrape);
```
zuphilip: undo (instead replace the `return true;` from some lines above with just `return;`)

EUR-Lex.js
```diff
@@ -103,96 +135,125 @@ function doWeb(doc, url) {
...
 		});
-	} else {
-		scrape(doc, url);
 	}
+	else if (detectWeb(doc, url) === "case") {
+		return scrapeCase(doc, url);
```
zuphilip: no return here

EUR-Lex.js
```diff
@@ -103,96 +135,125 @@ function doWeb(doc, url) {
...
+	}
+	return scrape(doc, url);
```
zuphilip: no return here

EUR-Lex.js
```diff
@@ -103,96 +135,125 @@ function doWeb(doc, url) {
...
 	// determine the language we are currently looking the document at
-	var languageUrl = url.split('/')[4];
-	if (languageUrl=="AUTO") {
+	let languageUrl = url.split('/')[4];
+	if (languageUrl === "AUTO") {
 		languageUrl = autoLanguage || "EN";
 	}
-	var language = languageMapping[languageUrl] || "eng";
-	
+	const language = languageMapping[languageUrl] || "eng";
```
zuphilip: Why const?

EUR-Lex.js
```diff
@@ -103,96 +135,125 @@ function doWeb(doc, url) {
...
+
+	item.title = ZU.xpathText(doc, "//p[@id='translatedTitle']").split('.')[1].trim();
+	item.language = languageUrl.toLowerCase();
+
+
+	const parsedDate = ZU.xpathText(doc, "//div[@id='PPDates_Contents']/div[1]/dl[1]/dd[1]").split('/');
+	item.date = parsedDate[2] + '-' + parsedDate[1] + '-' + parsedDate[0];
+
+
+	item.url = url;
```
zuphilip: As far as a I see the only difference for scraping cases is the part about the date. But we can this then also integrate above in the normal scrape function with an if clause. There is IMO no need for a different function `scrapeCase`.

EUR-Lex.js
hueyy: AG Opinions are not binding but they are very closely examined together with the actual decision because they add important background, and oftentimes can help a reader to better understand the implications of a decision.

EUR-Lex.js
hueyy: It's shouldn't be mutated at any point. This makes it obvious if it is inadvertently mutated.

PR #2057, opened 2019-11-16
Adding two types of sources to Idref.js (oatao + persee)

symac:
IDref is a website describing authorities and getting bib records from multiple sources where the authority exists. This PR makes the translators compatible with more sources : 

- [OATAO](https://oatao.univ-toulouse.fr/) for example with some titles from this author : https://www.idref.fr/224733621 (the ones that appear with [the red icon](https://www.idref.fr/img/bibliographie-oatao.png))
- [Persée](https://www.persee.fr/) for example with some titles from this author : https://www.idref.fr/026905035 (the ones that appear with [the green icon](https://www.idref.fr/img/bibliographie-persee.png))

### DIFF
--- a/Idref.js
+++ b/Idref.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2019-11-16 21:34:33"
+	"lastUpdated": "2019-11-21 11:52:29"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


Idref.js
```diff
@@ -53,7 +53,7 @@ function getSearchResults(doc, checkOnly) {
 		// We need to replace the http://www.sudoc.fr/XXXXXX links are they are redirects and aren't handled correctly from subtranslator
 		href = href.replace(/http:\/\/www\.sudoc\.fr\/(.*)$/, "http://www.sudoc.abes.fr/xslt/DB=2.1//SRCH?IKT=12&TRM=$1");
 
-		if ((href.includes("www.sudoc.abes.fr")) || (href.includes("archives-ouvertes")) || (href.includes("catalogue.bnf.fr")) || (href.includes("www.theses.fr")) || (href.includes("pub.orcid.org"))) {
+		if ((href.includes("www.sudoc.abes.fr")) || (href.includes("archives-ouvertes")) || (href.includes("catalogue.bnf.fr")) || (href.includes("www.theses.fr")) || (href.includes("pub.orcid.org")) || (href.includes("www.persee.fr")) || (href.includes("oatao.univ-toulouse.fr"))) {
```
adam3smith: This is fine, but I'm wondering if it'd be possible to make sure href is just the host name of the redirected page and then test them against an array along the lines of `["catalogue.bnf.fr", "www.sudoc.abes.fr"...].includes(href)`

This would perform much better and imo be a lot more legible, but given that we're still testing against a relatively short list of strings, the current version is fine -- just let me know if you think the above is viable and you're interested in working on it or not.

adam3smith:
Ah thanks -- this is great! Will merge shortly

PR #2061, opened 2019-11-21
New translator for Cowles Foundation publications

placardo:
A translator for the documents on the Cowles Foundation website (https://cowles.yale.edu/).
The translator work on the authors' pages and on the different lists of publications (for monographs, discussion papers and published articles) and gets pdfs when available.
I tried to follow the guidelines at https://www.zotero.org/support/dev/translators.


Cowles.js
```diff
@@ -0,0 +1,316 @@
...
+		.textContent
+		.trim()
+		.toLowerCase()
+		.replace(".", "")
+		.split(" ");
+		// There are three types of documents: monographs (books), discussion papers (reports) and papers which were published
+		// This query method works on the monograph pages and the authors' pages for now
+		if (paper[0] == "cfm") {
+			href = root + "/" + paper.join("-");
+			title = row.querySelector("td.views-field-field-paper-title a").textContent;
```
adam3smith: the `text` function simplifies this a good bit.

Cowles.js
```diff
@@ -0,0 +1,316 @@
+{
+	"translatorID": "951c027d-74ac-47d4-a107-9c3069ab7b48",
+	"label": "Cowles",
```
adam3smith: "Cowles Foundation" seems like the more appropriate label & filename

Cowles.js
```diff
@@ -0,0 +1,316 @@
...
+		if (author[1] == " ed." || author[1] == " eds.") status = "editor";
+		else status = "author";
+		
+		if (author[0].search(" and ") != -1) {
+			var authors = author[0].split(" and ");
+			for (let auth of authors){
+				item.creators.push(ZU.cleanAuthor(auth,status,false));
+			}
+		}
+		else if (author[0].search(" & ") != -1) item.creators.push(ZU.cleanAuthor(author[0].split(" & ")[0],status,false));
```
adam3smith: & isn't used for multiple authors?

Cowles.js
```diff
@@ -0,0 +1,316 @@
...
+		item.seriesNumber = doc.querySelector("#page-title").textContent.match(/\d+/)[0];
+		item.libraryCatalog = "Cowles Foundation";
+		
+		var status = "";
+		var author = doc.querySelector(".field-name-field-paper-title p a")
+		.previousSibling
+		.textContent
+		.trim()
+		.split(",");
+		if (author[1] == " ed." || author[1] == " eds.") status = "editor";
```
adam3smith: there are never more than two authors? Meyer, Smith, and Jones, eds. would break this. I'd prefer something more robust. Something like `if (author.length>1){if (author.slice(-1)[0] == " ed. ")` etc. might work

Cowles.js
```diff
@@ -0,0 +1,316 @@
...
+		else status = "author";
+		
+		if (author[0].search(" and ") != -1) {
```
adam3smith: any reason to use `search` rather than `includes` here?

Cowles.js
```diff
@@ -0,0 +1,316 @@
...
+		});
+	}
+	else if (url.includes('/cfm-')) {
+		item = new Zotero.Item("book");
+		item.title = doc.querySelector('a[href*="/pub/"]').textContent;
+		item.series = "Cowles Monograph";
+		item.seriesNumber = doc.querySelector("#page-title").textContent.match(/\d+/)[0];
+		item.libraryCatalog = "Cowles Foundation";
+		
+		var status = "";
```
adam3smith: could we give this variable a better name? `creatorType` or something like that? And I'd just set it to "author" right here

Cowles.js
```diff
@@ -0,0 +1,316 @@
+{
+	"translatorID": "951c027d-74ac-47d4-a107-9c3069ab7b48",
+	"label": "Cowles",
+	"creator": "Vincent Carret",
+	"target": "^https?://(www\\.)?cowles.yale.edu/",
+	"minVersion": "3.0.4",
```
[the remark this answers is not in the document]
placardo: I was a bit quick on the metadata it was automatically filled by scaffold and I thought it was 3.0.4 for a reason... guess not

Cowles.js
placardo: There are, indeed, never more than two authors and no monographs has been published since 1981 (https://cowles.yale.edu/publications/archives/cfm)

Cowles.js
placardo: It's used in one instance in particular (see https://cowles.yale.edu/cfm-31) to indicate "_Someone_ & Associates". There are few monographs and most likely there won't be any more in the future so I adapted the code to the existing ones. Should I put "Associates" in collaborators or another field?

Cowles.js
placardo: I'm not sure I understand? (I'm really not proficient in js as you would have gathered)

Cowles.js
```diff
@@ -0,0 +1,316 @@
...
+}
+
+function scrape(doc, url){
+	var root = "https://cowles.yale.edu";
+	var pdfurl = "";
+	var item = null;
+	// Each type of document follow a different layout, which seems to be the same inside all three categories
+	if (url.includes('/cfp/')) {
+		item = new Zotero.Item("journalArticle");
+		item.title = doc.evaluate("//strong[contains(., 'CFP Paper Title')]", doc, null, XPathResult.ANY_TYPE, null).iterateNext().nextSibling.textContent;
```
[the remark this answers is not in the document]
placardo: thank you for bringing my attention to this utility

Cowles.js
placardo: I changed the filename, can you confirm that I can delete the old file? I guess it will delete all comments but I did not know how else to do it

placardo:
> Thanks!
> Some suggestions to clean up the code a bit & make it more robust

Thank you for your quick feedback! This is my first translator and I'm not used to Github so I apologize in advance for any silly mistakes

Cowles.js
adam3smith: No, OK to leave as is in that case

Cowles.js
adam3smith: Implement `text` as a function in the beginning of the translator (see https://www.mediawiki.org/wiki/Citoid/Creating_Zotero_translators#text for an explanation and https://github.com/zotero/translators/commit/2901ba9fa5b4158cf87697a979a55ee7ad867fef#diff-1dbee0bc07ad76ab1291e9571884f32bR39 for an example), then do 
`title = text(row, "td.views-field-field-paper-title a");`

adam3smith:
This all looks very good now thanks (and apologies for the delay).
Two remaining issues:
1.  The `root` that you specify -- that shouldn't be necessary. Zotero can handle relative links and they're more robust. Could you test if all of this still works entirely without root and then remove it? (If you're used to other webscraping tools e.g. in R or python, they tend to behave less like a browser and break with relative links; Zotero doesn't)

2. Please use the minimized version of `text()` and `attr()` as e.g. here: https://github.com/zotero/translators/blob/master/Emerald%20Insight.js#L37 use all three lines -- the first one identifies, the 2nd disables linting, the 3rd is the minimized code.

placardo:
I am not sure how to do without the root part because, if I understand it correctly, Zotero inserts the current url argument if I don't specify it; while it would work on the items pages or the list of publications, it would break eg on an authors' page where the current url does not match the beginning of an item's url (for example: you're on this page https://cowles.yale.edu/author/herbert-e-scarf, and select an item to import to zotero. The current url used is https://cowles.yale.edu/author and if I just append the item url for "/cfdp-2212" it throws a 404 because the actual beginning should be https://cowles.yale.edu/publications/cfdp). Same goes for the PDF, as what is scraped is an address starting at the root of the site; if I delete `var root` it sends a GET to the current URL (eg. https://cowles.yale.edu/publications/archives/cfm) + the scraped address whereas it should only send a GET to https://cowles.yale.edu/ + the scraped address...

Maybe I missed something because I did not take the time yet to see how Zotero connector works please tell me if you think I could do it using only relative links

adam3smith:
Did you test? Relative links always refer to the host, i.e what you call the root and that's what Zotero uses (or certainly _should_ use). In other words `root + "/" + variable` is essentially identical to `"/" + variable` but will behave better e.g. if URL proxies are involved or with http/https differences

placardo:
Sorry I was confused it seems I got that error by testing without the initial /, in that case apparently Zotero appends the entire current URL and not just the host. It seems to work now with the initial /

adam3smith:
Great thanks. Just FYI 

> I got that error by testing without the initial /, in that case apparently Zotero appends the entire current URL and not just the host. It seems to work now with the initial /

This is just how relative links work, e.g. that's how a browser would behave on a website.

placardo:
@adam3smith do you need any more changes on this translator?

Cowles Foundation.js
```diff
@@ -0,0 +1,322 @@
...
+
+function getSearchResults(doc, checkOnly) {
+	var items = {};
+	var found = false;
+	var rows = doc.querySelectorAll(".views-table tbody tr");
+	var href = "";
+	var title = "";
+	for (let row of rows) {
+		var paper = text(row, "td")
+		.trim()
```
AbeJellinek, line 66: Indent these lines

Cowles Foundation.js
```diff
@@ -0,0 +1,322 @@
...
+		item.series = "Cowles Monograph";
+		item.seriesNumber = text(doc, "#page-title").match(/\d+/)[0];
+		item.libraryCatalog = "Cowles Foundation";
+		
+		var creatorType = "author";
+		var author = doc.querySelector(".field-name-field-paper-title p a")
+		.previousSibling
+		.textContent
+		.trim()
+		.split(",");
```
AbeJellinek, line 163: Replace this with `.split(/, | and /)`

Cowles Foundation.js
```diff
@@ -0,0 +1,322 @@
...
+		if (author[1] == " ed." || author[1] == " eds.") creatorType = "editor";
+		else creatorType = "author";
+		
+		if (author[0].includes(" and ")) {
```
AbeJellinek, line 167: ...and you'll be able to remove this

Cowles Foundation.js
```diff
@@ -0,0 +1,322 @@
...
+			var authors = author[0].split(" and ");
+			for (let auth of authors){
+				item.creators.push(ZU.cleanAuthor(auth, creatorType, false));
+			}
+		}
+		else if (author[0].includes(" & ")) item.creators.push(ZU.cleanAuthor(author[0].split(" & ")[0], creatorType, false));
```
AbeJellinek, line 173: You could also fold the " & " splitting into the above regex.

Cowles Foundation.js
```diff
@@ -0,0 +1,322 @@
...
+				item.edition = editor[1];
+				item.publisher = editor[2];
+			}
+			else item.publisher = editor[1];
+			item.date = ZU.strToISO(editor[editor.length - 1].split(" ")[0]);
+		}
+		catch (err) {}
+		
+		pdfurl = attr(doc, 'a[href*="/pub/"]', 'href');
+		item.attachments.push({
```
AbeJellinek, line 188: Could you pull the `item.attachments.push` calls out of the various `if` branches? You're setting one `pdfurl` variable anyway.

AbeJellinek (changes requested):
Hey, thanks for contributing this. I just have a few nitpicky points about code style and duplicated logic, but otherwise this looks good and I think it's close to ready to merge. Sorry for the delay in getting to it.

vcarret:
Hey @AbeJellinek thanks for your effort to go down the list of opened pull requests!

I did the changes you requested you were right about the duplicated logic :) 

Checks are failing but the tests worked on my computer; I am a bit rusty about the linting process and maybe it changed since the last time I committed to this repo do you mind checking why it failed? I can squash the commits if you want

AbeJellinek:
No need to squash manually! I can do that when merging. The easiest way to deal with linting is like this:

```bash
$ git pull origin
$ git rebase origin/master
$ npm run lint -- "Cowles Foundation.js" --fix
```

Then commit and force push.

`npm run lint` will complain about a few things that it can't fix automatically. In this case it seems to be that you don't declare `author` before using it on line 119 and line 134.

AbeJellinek:
Oh my God, I'm sorry, I wrote `git pull origin` but meant `git fetch origin`. That's where all the irrelevant commits are coming from. A rebase on origin/master should remove them.

vcarret:
Ok sorry I did not check what I was doing but now I reset everything into the first commit, although I still don't understand what is this failing error: "fatal: ambiguous argument '577948faa947649ea40185e4730f0391f0b98535^^!': unknown revision or path not in the working tree."

PR #2062, opened 2019-11-21
New translator for the Journal of Religion & Society

### LINKED ISSUE #1581: Add support for Journal of Religion & Society
See http://moses.creighton.edu/JRS/index.html

There is only a page for each volume, e.g. http://moses.creighton.edu/JRS/toc/2017.html and no individual pages for an article. There are no metadata nor export format, thus this needs to be a plain web scraper. The abstract is also on the page, but linked with some javascript function.

vcarret:
Resolves #1581. Gets whole supplements as books on (http://moses.creighton.edu/JRS/toc/Supplement.html) and papers inside supplements as book chapters (e.g. http://moses.creighton.edu/JRS/toc/SS19.html).
For the different issues, get them as journal articles with the abstract.


JournalReligionSociety.js
```diff
@@ -0,0 +1,177 @@
...
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+
+function detectWeb(doc, url) {
+	// Three possible cases : the list of articles of an issue, the list of content of a supplement, or the list of the supplements
+	if (getSearchResults(doc, true) || url.search('/toc/SS') > -1 || url.search('/toc/Supplement') > -1) {
```
adam3smith: `getSearchResults` should be an & condition (and then the two OR conditions need to be in a parentheses -- if it doesn't return `true`, `doWeb` will fail.

JournalReligionSociety.js
```diff
@@ -0,0 +1,177 @@
+{
+	"translatorID": "e969da6b-30c9-4b26-8fcc-c2d78bce685f",
+	"label": "JournalReligionSociety",
+	"creator": "Vincent Carret",
+	"target": "^https?://(www\\.)?moses.creighton.edu/JRS",
```
adam3smith: secape all periods

JournalReligionSociety.js
```diff
@@ -0,0 +1,177 @@
+{
+	"translatorID": "e969da6b-30c9-4b26-8fcc-c2d78bce685f",
+	"label": "JournalReligionSociety",
```
adam3smith: Make the label expressive, i.e. "Journal of Religion and Society"

JournalReligionSociety.js
adam3smith: also, use `includes` for literal string matching

JournalReligionSociety.js
```diff
@@ -0,0 +1,177 @@
...
+function scrape(id, doc, url) {
+	var item = null;
+	var infoBlock = null;
+	var author = null;
+	var pdfurl = "";
+	if (url.search('/toc/SS') == -1 && url.search('/toc/Supplement') == -1) {
+		item = new Zotero.Item("journalArticle");
+		item.title = id;
+		item.publicationTitle = "Journal of Religion & Society";
+		item.date = doc.querySelector(".heading").textContent.split('(')[1].match(/\d+/)[0];
```
adam3smith: use `text` here

JournalReligionSociety.js
```diff
@@ -0,0 +1,177 @@
...
+		item.volume = doc.querySelector(".heading").textContent.split('(')[0].match(/\d+/)[0];
+		item.url = url;
+		
+		infoBlock = doc.evaluate("//p[contains(., '" + id + "')]", doc, null, XPathResult.ANY_TYPE, null)
+		.iterateNext()
+		.nextElementSibling;
+		author = infoBlock.textContent.split("\n");
+		item.creators.push(author);
```
adam3smith: I don't think you meant to put this here?

JournalReligionSociety.js
```diff
@@ -0,0 +1,177 @@
...
+	var pdfurl = "";
+	if (url.search('/toc/SS') == -1 && url.search('/toc/Supplement') == -1) {
+		item = new Zotero.Item("journalArticle");
+		item.title = id;
+		item.publicationTitle = "Journal of Religion & Society";
+		item.date = doc.querySelector(".heading").textContent.split('(')[1].match(/\d+/)[0];
+		item.volume = doc.querySelector(".heading").textContent.split('(')[0].match(/\d+/)[0];
+		item.url = url;
+		
+		infoBlock = doc.evaluate("//p[contains(., '" + id + "')]", doc, null, XPathResult.ANY_TYPE, null)
```
adam3smith: `ZU.xpathText` instead

AbeJellinek:
Thanks! Sorry this took so long to get to.

PR #2064, opened 2019-11-22
New translator for scinapse.io

### LINKED ISSUE #2027: Add Scinapse.io translator
Requested https://forums.zotero.org/discussion/78662/scinapse-io-translator-request#latest

vcarret:
Resolves #2027 . I used DOI content negotiation to retrieve the info from crossref and added the abstract when available. Works on search pages, authors pages, journals pages for batches and papers pages for individual articles. Seems to work as far as DOI can be retrieved, sometimes pdf is attached

### DIFF
--- a/scinapse.js
+++ b/scinapse.js
@@ -9,13 +9,13 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2019-12-05 22:45:05"
+	"lastUpdated": "2021-06-14 21:01:40"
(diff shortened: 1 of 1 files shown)


AbeJellinek:
Sorry for the delay on this. Great work, thanks! The DOI method wasn't always working, and some of the class names that CSS selectors targeted were dynamically generated and were causing tests to fail, so I replaced DOI with a combination of two other data source and made the selectors more general. Otherwise everything here looks great. I'm going to go ahead and merge.

PR #2067, opened 2019-11-25
Attach PDF documents when viewed directly in the wayback machine

reubot:
This pull attaches PDFs accessed in the Wayback Machine to the associated parent.

### DIFF
--- a/Internet Archive Wayback Machine.js
+++ b/Internet Archive Wayback Machine.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2014-07-05 10:50:25"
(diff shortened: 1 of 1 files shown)


zuphilip:
Can you give an example for this? If I understand it correctly, this will only change when I am on at the at a PDF saved in the InternetArchive. Thus, it is not generally attaching PDFs from the IA as the title might suggest. Will in such cases then the HTML snapshot makes sense to have additionally or should we have it only in the `else` case?

zuphilip:
Okay, then please change the saving PDF into an else case.

zuphilip:
Okay, thank you. I fixed the remaining linting issues and did actually a slight rewrite. Please let me know what you think about it. This should be ready to merge now.

zuphilip:
Thank you very much @reubot ! I changed the label and this is now merged :rocket: .

PR #2076, opened 2019-12-02
Gale Databases.js - New logic to access citation metadata

jimmiazek:


jimmiazek:
Please publish update after 12/20/2019 in order to coordinate with Gale releases.

adam3smith:
Thanks @jimmiazek . Looks good at first glance, I'll review for details next weekend the latest. 

Just to confirm the data is December 20th? I believe I had December 16th from Megan.

dstillman:
@jimmiazek: Can you revert the .gitignore change? (You can put this in a global ignore file on your system.)

jimmiazek:
@adam3smith - Original release date (12/16) has been moved to (12/20).

adam3smith:
Thanks! Looks great overall. I have no concerns about the code itself. Except two quick ones that I think are just due to the testing stage of this:
* The tests will need to get updated once the site is live. do you know already what the hostname will be? 

* Currently you have commented out the metadata JSON block. Please re-enable that

Some remaining issues on import quality: 

* I get the starting(?) page in extra occasionally, e.g. for 
https://link.qa.galetesting.com/apps/doc/A597223700/AONE?u=zotero&sid=AONE&xid=b1ebba12
coming from the `M2` tag in the RIS. I think it'd be best to just ignore the M2 tag unless it solves some other function

* Missing page numbers are (sometimes?) turned into `NA` such as on
https://link.qa.galetesting.com/apps/doc/A599824399/AONE?u=zotero&sid=AONE&xid=3ea3b5f9
that should just be removed during import

* Your `detect` function looks like it sometimes produces better item type mappings than Zotero's RIS import. E.g. for https://link.qa.galetesting.com/apps/doc/CCUUVM394664614/NCCO?u=zotero&sid=NCCO&xid=1f6d5d40 it detects (correctly) as a manuscript, but Zotero, not knowing the `UNPB` label on the RIS, imports as journal article. Unfortunately that's not always the case. E.g. in the Literature Resource center, the RIS is much better: https://link.qa.galetesting.com/apps/doc/A598538032/LitRC?u=zotero&sid=LitRC&xid=51c6b9d6 is correctly classified as a magazine article, rather than a document. 
I'm open for ideas here, but my suggestion would be to check were the two disagree, go with the zoter-label element when the RIS imports as journals and stick with the RIS otherwise.

Let me know if you have any questions about this.

jimmiazek:
@adam3smith - made the code changes that you asked to uncomment the JSON object -- I believe this is what is causing Travis to fail.  Production server is gale.com. Applied your suggestion for M2 and SP tags.

For your analysis, regarding the differences for detectWeb and doWeb.  This is something that will have to be addressed in the future.  The doWeb is using a REST client to retrieve RIS.  The detectWeb makes a different call to get different data to render the display.

adam3smith:
Thanks -- I'll take a look at why linting fails. I'd guess Windows EoL for the JSON header, but either way, I should be able to fix that quickly on my end.

Gale Databases.js
```diff
@@ -1,190 +1,211 @@
...
+	'label': 'Gale Databases',
+	'creator': 'Jim Miazek',
+	'target': '^https?://[^?&]*(?:gale|galegroup|galetesting|ggtest)\.com(?:\:\d+)?/ps/',
+	'minVersion': '3.0',
+	'maxVersion': '',
+	'priority': 100,
+	'inRepository': true,
+	'translatorType': 4,
+	'browserSupport': 'gcsibv',
+	'lastUpdated': '2019-10-11 12:00:00'
```
dstillman: You need to revert these quote changes, which will break the translator. (For historical reasons, this is JSON, not JavaScript.)

Gale Databases.js
```diff
@@ -1,190 +1,211 @@
...
+
 	You should have received a copy of the GNU Affero General Public License
 	along with Zotero.  If not, see <http://www.gnu.org/licenses/>.
-	
+
 	***** END LICENSE BLOCK *****
 */
-function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null}function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null}
 
+var GaleZotero = (function () {
```
dstillman: There's no need for this encapsulation. Just put the relevant code in `detectWeb()` and `doWeb()` and define other top-level helper functions as necessary.

Gale Databases.js
```diff
@@ -1,190 +1,211 @@
...
+var DATA_TRANSLATOR = '32d59d2d-b65a-4da4-b0a3-bdd3cfb979e7';
```
dstillman: This can just go directly in the `translator.setTranslator()` call below (as in all other translators), with a comment noting that it's RIS.

Gale Databases.js
```diff
@@ -1,190 +1,211 @@
...
-	for (var i=0; i<rows.length; i++) {
-		var href = rows[i].href;
-		var title = ZU.trimInternal(rows[i].textContent);
-		if (!href || !title) continue;
-		if (checkOnly) return true;
-		found = true;
-		items[href] = title;
+}
+
+function processMulipleEntries(entries) {
```
dstillman: Typo

Gale Databases.js
```diff
@@ -1,190 +1,211 @@
...
+			Zotero.Utilities.processDocuments(getURLs(selectedItems), processSingleEntry);
+		}
+	});
+}
+
+function createKeyValuePairs(entries) {
+	var map = {};
+	for (var item in entries) {
+		/* istanbul ignore next */
+		if (entries.hasOwnProperty(item)) {
```
dstillman: This check isn't necessary — Object.prototype will never be overridden in the contexts where this will run.

Gale Databases.js
```diff
@@ -1,190 +1,211 @@
...
+	return map;
 }
-function detectWeb(doc, url) {
-	if ((url.includes('/retrieve.do') || url.includes('/i.do?')) && text(doc, 'li#docTools-citation, li.docTools-citation')) {
-		return "journalArticle";
+
+function getURLs(selectedItems) {
+	var urls = [];
+	for (var url in selectedItems) {
+		if (selectedItems.hasOwnProperty(url)) {
```
dstillman: Not necessary

Gale Databases.js
```diff
@@ -1,190 +1,211 @@
...
+					{
+						note: "<p>Northern Illinois University</p>"
+					}
+				],
+				seeAlso: []
 			}
 		]
 	}
-]
+]; // eslint-disable-line
```
dstillman: Revert

Gale Databases.js
```diff
@@ -1,190 +1,211 @@
...
 					}
 				],
-				"notes": [],
-				"seeAlso": []
+				notes: [
+					{
+						note: "<p>Northern Illinois University</p>"
+					}
+				],
+				seeAlso: []
```
dstillman: Please revert all the quote changes in these test cases. These tests should be generated by Scaffold and shouldn't be reformatted.

Gale Databases.js
```diff
@@ -1,190 +1,211 @@
...
+function detectWeb(doc, url) { // eslint-disable-line no-unused-vars
+	return GaleZotero.detect(doc);
+}
+
+function doWeb(doc, url) { // eslint-disable-line no-unused-vars
+	return GaleZotero.process(doc);
 }
+
 /** BEGIN TEST CASES **/
+// eslint-disable-next-line no-unused-vars
```
dstillman: Remove all these eslint-disable lines. You should be using this repo's lint rules, which should handle all of this properly.

dstillman:
(Re: all of these comments, I'm guessing you were developing this code outside of the Translator Editor, with a different environment and different lint rules. We strongly recommend developing within the editor, and you should check the code with eslint within the repo, which should use the bundled lint rules.)

adam3smith:
@jimmiazek a bit confused why you closed this and deleted the branch? What´s the status of the updated site and code?

dstillman:
It looks like the account was deleted.

adam3smith:
how very odd & unfortunate...

zuphilip:
We can still take the old version of this PR and build on top of that our changes. Give me some minutes to come up with something...

adam3smith:
Thanks -- we should check that they did, in fact, update their site on the 20th. Otherwise these changes won't work.

adam3smith:
Update: yes, it looks like all of Gale is update, so this would be nice to get in.

zuphilip:
Continued in #2096.

PR #2077, opened 2019-12-06
SAGE Journals - Resolved tag issue + detection issues

### LINKED ISSUE #1789: Keywords in SAGE are missing
Currently the SAGE translator will just look for the RIS data and there are no keywords in the RIS data. However, they are in the meta tag of the website and therefore should be easy to extract. There are examples with keywords in several languages, where I would suggest to take only the keywords in one languge (e.g. English), see e.g. https://journals.sagepub.com/doi/full/10.1177/0008429818754786#articleCitationDownloadContainer

### LINKED ISSUE #2075: Sage - various issues
Including search results, e.g. https://journals.sagepub.com/action/doSearch?text1=test as reported [here](https://forums.zotero.org/discussion/80279/sage-journals#latest) but also looks like individual items don't work. 

Ideally, combine with fixing #1789

placardo:
Fix #1789 and fix #2075. This works in my browser but it doesn't work in Scaffold; when the scrape function evaluate url.match() line 85 it throws an error, because in Scaffold the URL is https://journals.sagepub.com/action/**cookieAbsent** instead of the DOI but it seems to work otherwise.

From issue #1789 : 
> There are examples with keywords in several languages, where I would suggest to take only the keywords in one languge (e.g. English), see e.g.

Does some Zotero Utility or something else exists to select only one language?

### DIFF
--- a/SAGE Journals.js
+++ b/SAGE Journals.js
@@ -59,6 +59,7 @@ function getSearchResults(doc, checkOnly) {
 		if (!href || !title) continue;
 		if (checkOnly) return true;
 		found = true;
+		href = href.replace("/doi/pdf/", "/doi/abs/");
 		items[href] = title;
 	}
 	return found ? items : false;
@@ -88,8 +89,7 @@ function scrape(doc, url) {
 	if (!doi) {
 		doi = url.match(/10\.[^?#]+/)[0];
 	}
-	var filename = ZU.xpathText(doc, '//input[@name="downloadFileName"]/@value');
-	var post = "doi=" + encodeURIComponent(doi) + "&downloadFileName=" + filename + "&include=abs&format=ris&direct=false&submit=Download+Citation";
+	var post = "doi=" + encodeURIComponent(doi) + "&include=abs&format=ris&direct=false&submit=Download+Citation";
 	var pdfurl = "//" + doc.location.host + "/doi/pdf/" + doi;
 	//Z.debug(pdfurl);
 	//Z.debug(post);
--- a/SAGE Journals.js
+++ b/SAGE Journals.js
@@ -2,14 +2,14 @@
 	"translatorID": "908c1ca2-59b6-4ad8-b026-709b7b927bda",
 	"label": "SAGE Journals",
 	"creator": "Sebastian Karcher",
-	"target": "^https?://journals\\.sagepub\\.com(/doi/((abs|full)/)?10\\.|/action/doSearch\\?|/toc/)",
+	"target": "^https?://journals\\.sagepub\\.com(/doi/((abs|full|pdf)/)?10\\.|/action/doSearch\\?|/toc/)",
 	"minVersion": "3.0",
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2018-01-14 20:53:20"
+	"lastUpdated": "2019-12-06 12:51:37"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


SAGE Journals.js
```diff
@@ -121,6 +125,11 @@ function scrape(doc, url) {
 				item.abstractNote = abstract;
 			}
 			
+			var tags = ZU.xpathText(doc, '//meta[@name="dc.Subject"]/@content');
```
adam3smith:
```suggestion
			var tags = doc.querySelectorAll('kwd-group:first-of-type a');
```

SAGE Journals.js
```diff
@@ -121,6 +125,11 @@ function scrape(doc, url) {
...
+			if (tags) {
```
adam3smith: and then something like
```
for (let tag of tags) {
  item.tags.push(tag.textContent);
}
```

(untested)

SAGE Journals.js
placardo: I guess this would work but thinking about it the example from @zuphilip may not be very representative; it's taken from a [bilingual journal published by editors from French- and English-speaking Canada](https://en.wikipedia.org/wiki/Studies_in_Religion) and I don't think that they intended for users to only use English keywords - shouldn't Zotero reflect all the meta information?

adam3smith: It's a judgement call, but I think duplicate tags are pretty useless and the nice thing about this strategy is that the journal appears to put the tags in the language of the article first, so people would get French ones here:
https://journals.sagepub.com/doi/full/10.1177/0008429819858923

placardo: You're quite right, I did not see that. Let's just hope they continue in the future...

SAGE Journals.js
```diff
@@ -2,14 +2,14 @@
 	"translatorID": "908c1ca2-59b6-4ad8-b026-709b7b927bda",
 	"label": "SAGE Journals",
 	"creator": "Sebastian Karcher",
-	"target": "^https?://journals\\.sagepub\\.com(/doi/((abs|full)/)?10\\.|/action/doSearch\\?|/toc/)",
+	"target": "^https?://journals\\.sagepub\\.com(/doi/((abs|full|pdf)/)?10\\.|/action/doSearch\\?|/toc/)",
```
adam3smith: Did you test this on a PDF page? Given that the filename for the post request (l. 91) is scraped from the page, I don't see how it would work.

placardo: I added the pdf because a lot of webpages besides doi/abs and doi/full have doi/pdf in their URL, see eg the search results included in the translator tests (https://journals.sagepub.com/action/doSearch?text1=test)
On Firefox when I'm on the actual pdf page the PDF translator is activated instead of SAGE Journals translator and on Chrome the SAGE Journal translator is sometimes activated and yet I can still import the item (eg here : https://journals.sagepub.com/doi/pdf/10.1177/003754979005400307?casa_token=8AElK4mX07EAAAAA:WxK8lh-rsHkjRwcM1Z8EldOJF-eKIOC3nCaL0gCLxHTQOohZcnxb9ykF8F7KJSboIuwQsisPgNl8cQ I think that's when I previously visited the webpage linking to the pdf), sometimes only the PDF translator shows up (eg here which goes directly to pdf without a webpage: https://journals.sagepub.com/doi/pdf/10.1177/003754979005400307)

adam3smith: Good catch on these turning up in search results, but that actually means we need to replace `/doi/pdf` in search results with `/doi/abs` to avoid import errors. Let me know if that makes sense and you can add it.

I don't have an explanation for why you'd get a working import from Chrome in the case you have -- there's really no way this should work that I can think of. The fact that it's always showing the PDF translator for Firefox is due to a restriction of the extension framework there, but that's obviously not a reason to include PDF URLs. With things working correctly, I'd expect false positives there so please rever that part both in the target and the detect function.

placardo: Actually this works without the filename argument in the POST request, if you remove it you still have a valid and complete response I guess the DOI is enough but I wasn't able to find any documentation about which arguments are necessary

adam3smith: You're right, the pdf should stay in the taget (and please do squabble if you think any of my or zuphilip's comments are wrong). I thought it redirected on paywalled pages, but it doesn't.
I've removed the unneeded post parameter (thanks for that) and redirected multiple imports to the abstract page (so we can get language, improved abstract, and where they exist, keywords). 

If you could please update tests (since they now get keywords), this is ready to merge.

adam3smith: Oh, and please fix linting errors and warnings. Should be quick.

placardo: I added the tests manually since it does not work in Scaffold (cf above) and fixed the errors and warnings

adam3smith: oh, sorry! I could have done it then. I don't have that problem, so not sure what's going on with that.

adam3smith: oh, sorry! I could have done it then. I don't have that problem, so not sure what's going on with that.

adam3smith:
Thanks!

SAGE Journals.js
placardo: np glad to help! I will look into it I'm sure there is a reasonable explanation

PR #2082, opened 2019-12-10
Translator for Library Hub Discover search results

### LINKED ISSUE #1986: UK Library Hub
Replaces the old CoPac confederated catalog. 
Site: https://discover.libraryhub.jisc.ac.uk/
Requested: https://forums.zotero.org/discussion/78478/library-hub-discover-replacing-copac-is-there-a-zotero-translator#latest

Shouldn't be too hard: site has all sorts of useful metadata export.

vcarret:
Fix #1986. This is really just a wrapper for COinS on the search results (I'm not sure how useful this is or if it is good practice to make such a limited translator - does it slow the connector's loading?). In any case I added https://discover.libraryhub.jisc.ac.uk/search?q=test as a test and it works fine, COins works seamlessly on individual items (NB: this was noted on the original issue https://forums.zotero.org/discussion/78478/library-hub-discover-replacing-copac-is-there-a-zotero-translator#latest)

### DIFF
--- a/Library Hub Discover.js
+++ b/Library Hub Discover.js
@@ -9,13 +9,13 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2019-12-16 17:25:26"
+	"lastUpdated": "2021-07-07 18:57:47"
 }
 
 /*
 	***** BEGIN LICENSE BLOCK *****
 
-	Copyright © 2019 Vincent Carret
+	Copyright © 2019-2021 Vincent Carret
 	
 	This file is part of Zotero.
 
@@ -35,9 +35,6 @@
 	***** END LICENSE BLOCK *****
 */
 
-// attr()/text() v2
-// eslint-disable-next-line
-function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null;}function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null;}
 
 var typeMapping = {
 	book: "book",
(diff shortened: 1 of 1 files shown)


adam3smith:
I think we'll actually want to check if we can do better than COinS (which is a very limited format). I suspect we can, e.g. using the provided MODS

placardo:
Do you have some kind of reference on the hierarchy between formats? This would be useful
Indeed they seem to have XML endpoints (https://discover.libraryhub.jisc.ac.uk/support/api/) I suppose I can feed this to MODS.js in the same way as the SAGE Journals translator fed it to the RIS translator?

adam3smith:
> Do you have some kind of reference on the hierarchy between formats? This would be useful

Not a reference but broadly:
1. MARC/MARCXML and MODS
2. Rich custom formats (like PNX for Primo) but those are a lot of work
3. RIS / BibTeX (both have pros and cons, depends on implementation)
4. Embedded Metadata of high quality (e.g. complete highwire tags)
5. COinS
6. Embedded Metadata of low quality (e.g. basic DC)

> Indeed they seem to have XML endpoints (https://discover.libraryhub.jisc.ac.uk/support/api/) I suppose I can feed this to MODS.js in the same way as the SAGE Journals translator fed it to the RIS translator?

Should work exactly the same, yes.

placardo:
I tried to implement it using the MODS endpoints but this is really not satisfactory, because the type is rarely well detected. Eg [this item](https://discover.libraryhub.jisc.ac.uk/id/15286374) which is obviously a feature film, is detected by MODS as a video recording, COinS thinks everything is a book and RIS / Bibtex find the right film type, but I did not find any unambiguous element on the page to detect the right type (the schema information is CreativeWork and the Library Hub info is visual, which is shared with a lot of categories). 
It does work well for some items eg sound recordings or maps where the ambiguity is reduced. I'm not sure how to proceed forward

AbeJellinek:
Thanks! Sorry for the delay in getting this merged.

PR #2083, opened 2019-12-11
BNF - Added different items in detect function + tests

vcarret:
There was no type detection on the BNF website, although the information is contained in a DC.type meta on public pages (not MARC pages oddly). I took care of matching the itemType returned by the MARC translator with the type in the DC.type tag, so that ambiguity is limited to very specific cases contained in the DC.type = "kit" category which lists several items of different types and for different physical objects (note that MARC seems to always default to "book" in those categories maybe this should be fixed to document?)
NB: on firefox the type "single" does not show an icon although it's there I don't think that's normal

### DIFF
--- a/Bibliotheque nationale de France.js
+++ b/Bibliotheque nationale de France.js
@@ -8,8 +8,8 @@
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
-	"browserSupport": "g",
-	"lastUpdated": "2017-07-07 11:47:44"
+	"browserSupport": "gc",
+	"lastUpdated": "2021-05-28 18:02:24"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


Bibliotheque nationale de France.js
```diff
@@ -43,6 +43,8 @@ var BnfClass = function () {
 	*/
 	
 	function getCreatorType(aut) {
+		// To avoid an error on certain pages where the item is not well catalogued
+		if (typeof aut['4'] === 'undefined') return undefined;
```
dstillman: Can just do `if (aut['4'] === undefined) return undefined;` these days.

placardo: What do you think of [this answer](https://stackoverflow.com/a/3550319)?

dstillman: `undefined` has been immutable for many years, so when comparing against a property it's fine to do `foo.bar === undefined`. The only time you need to use `typeof` is if a variable may not be defined (e.g., for feature detection), but that's not relevant for a property.

(This is explained well in [another answer](https://stackoverflow.com/a/22053469) on that page.

placardo: Sorry I deleted my comment before I saw your answer, after reading that answer on SO

placardo:
@dstillman what do you think about the rest of those changes? They're pretty straightforward I think and essentially cosmetic + I linted

AbeJellinek:
Alright, merged. Thanks for your contribution, @vcarret, and sorry this took so long!

PR #2089, opened 2019-12-16
Adjust Chinese Name split problem

kevin820606:
I notice that the original Airiti translator wrongly split Chinese author-name, and I adjust split place from name.length-1 to (0,1). The Chinese name appears in Airiti is in order of (Familyname - Given name)

Apart from this, vscode auto format whole script, and I think there is no impact on function.

### DIFF
--- a/Airiti.js
+++ b/Airiti.js
@@ -1,398 +1,425 @@
 {
-    "translatorID": "5f0ca39b-898a-4b1e-b98d-8cd0d6ce9801",
-    "label": "Airiti",
-    "creator": "Aurimas Vinckevicius",
-    "target": "^https?://([^/]+\\.)?airitilibrary\\.com/Publication/alDetailedMesh",
-    "minVersion": "3.0",
-    "maxVersion": "",
-    "priority": 110,
-    "inRepository": true,
-    "translatorType": 12,
-    "browserSupport": "gcsib",
-    "lastUpdated": "2018-04-17 21:16:52"
+	"translatorID": "5f0ca39b-898a-4b1e-b98d-8cd0d6ce9801",
+	"label": "Airiti",
+	"creator": "Aurimas Vinckevicius",
+	"target": "^https?://([^/]+\\.)?airitilibrary\\.com/Publication/alDetailedMesh",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 110,
+	"inRepository": true,
+	"translatorType": 12,
+	"browserSupport": "gcsib",
+	"lastUpdated": "2019-12-16 13:38:35"
 }
 
 function detectWeb(doc, url) {
-    // How can we distinguish thesis from journal article??
-    if (ZU.xpathText(doc, '/html/head/meta[@name="citation_title"]/@content')) {
-        return 'journalArticle';
-    }
+	// How can we distinguish thesis from journal article??
+	if (ZU.xpathText(doc, '/html/head/meta[@name="citation_title"]/@content')) {
+		return 'journalArticle';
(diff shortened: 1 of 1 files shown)


dstillman:
Thanks, but please revert all formatting changes.

zuphilip:
It seems that the relevant changes for splitting into last and first name for Chinese names are:
```diff
-				c.lastName = name.substr(name.length-1);
+				c.lastName = name.substr(0, 1);
				if (name.length > 1) {
-					c.firstName = name.substr(0, name.length-1);
+					c.firstName = name.substr(1, name.length);
				} else {
					delete c.firstName;
					c.fieldMode = 1;
				}
			}
```
Though this could go into one commit where you also change the timestamp and update the test cases, but not any other changes. @kevin820606 Do you want to do that?

However, I understand that there are also [compound Chinese last names](https://en.wikipedia.org/wiki/Chinese_compound_surname) like 司馬, e.g. http://www.airitilibrary.com/Publication/alDetailedMesh?DocID=1004941x-200506-23-3-54-56-a . We could try to handle the handful compount Chinese last names, which are still quite common nowadays according to the Wikipedia article, specially.

We can look at the linting issues afterwards, when the functional changes are ready.

kevin820606:
Thanks for your reviews. First of all, I apologize for the formatting and commit problem. Because this is my first time to pull and I am not familiar with coding (and everything about this, I am still learning), I don't quite aware of the auto-formatting problem. And, do you mean that I can just delete the second commit, I am not quite sure what to do about "go into one commit"

About compound Chinese last name, perhaps we can make a list to contain some common compound last name, but I think compound last names are really rare. Both in China or Taiwan, there are less than 0.5% of people with compound last name. (For example, 0.9 million people have last name "歐陽" in China, which is the most common compound last name, but that is only 0.06 % of 1.3 billion) Therefore, indeed we can build a list of the common compound last name, but perhaps there are always some names out of the list.

Once again, thanks for all the comments. I will try my best to adjust my commit through your comments.

zuphilip:
From a practical point I would suggest you to use Scaffold as the build-in editor for coding Zotero translators, see https://www.zotero.org/support/dev/translators/scaffold . Maybe simply redo the changes there and make a new commit.

> And, do you mean that I can just delete the second commit, I am not quite sure what to do about "go into one commit"

Sorry that was maybe not very clear. At the moment it doesn't matter how many commits you are doing. But we would like to have only the functional changes and no formatting changes like from VScode.

kevin820606:
Thanks for your comment! 
Perhaps I can close this request this time, and I will use Scaffold later. In addition,  I have discussed with another Chinese user about compound Chinese last name. 
Considering the frequency of compound late names and some possible problems, for example, a people named "歐陽鋒"(OU YANG FENG) could be "Feng, Ou-Yang" or "Yang Feng, Ou" which occurs sometimes, I think perhaps that remains cutting the first word is some sort of concession.

zuphilip:
Okay, thank you for the example about the Chinese last names. I am convinced now that we cannot do much about fixing these cases afterwards in Zotero translators.

> Perhaps I can close this request this time, and I will use Scaffold later. 

Sounds fine, but I suggest to leave this pull request open as long as there is not another pull request superseeding it. I will add special label here for us to remember easily. Let me know when you need any further help with this.

AbeJellinek:
Hi @kevin820606, any progress on this?

AbeJellinek:
Closing for lack of activity. It would be nice to have a generic Chinese name splitting function that handles the twenty-ish two-character surnames - maybe that can be a separate PR.

PR #2090, opened 2019-12-16
Emerald Insight - Resolved hyphen issue

### LINKED ISSUE #2087: Emerald Insight fails on book chapter
Breaks on [this book chapter](https://www.emerald.com/insight/content/doi/10.1108/S0163-786X(2012)0000033008/full/html) but appears to be fine on others

placardo:
Fix #2087 the regex did not support hyphens in last names. Note their inconsistent use of hyphens on their website (I added two test cases with different hyphens), although they appear to be the same on github one is the real deal (unicode 8208) and the other the hyphen-minus (45)

### DIFF
--- a/Emerald Insight.js
+++ b/Emerald Insight.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2019-12-25 15:26:20"
+	"lastUpdated": "2019-12-25 15:35:07"
 }
 
 /*
@@ -158,8 +158,8 @@ function scrape(doc, url) {
 						firstName: text(author, "span.given-names"),
 						lastName: text(author, "span.surname"),
 						creatorType: "author"
-				});
-			}
+					});
+				}
 				var otherContributors = item.creators.filter(creator => creator.creatorType !== "author");
 				item.creators = otherContributors.length !== 0 ? authors.concat(separateNames(otherContributors)) : authors;
 			}
@@ -193,7 +193,9 @@ function separateNames(creators) {
 			creators[i].firstName = lastName[0];
 			creators[i].lastName = lastName.slice(1).join(" ");
 		}
+		delete creators[i].fieldMode;
 	}
+	return creators;
 }
 
 /** BEGIN TEST CASES **/
--- a/Emerald Insight.js
+++ b/Emerald Insight.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2019-12-22 20:10:13"
+	"lastUpdated": "2019-12-25 15:21:24"
 }
 
 /*
@@ -146,28 +146,28 @@ function scrape(doc, url) {
 			for (let tag of tags) {
 				item.tags.push(tag.textContent);
 			}
-			// authors are in RIS as lastname firstname(s), though not necessarily correctly so
-			// trying to correct for this. Editors are firstname lastname
-			for (let i = 0; i < item.creators.length; i++) {
-				lastName = item.creators[i].lastName.split(" ");
-				if (!item.creators[i].firstName && lastName) {
-					var author = doc.querySelector("a[href*='" + lastName[0] + "'][class='contrib-search']");
-					if (author) {
-						item.creators[i].firstName = text(author, "span[class=given-names]");
-						item.creators[i].lastName = text(author, "span[class=surname]");
-					}
-					else if (item.creators[i].creatorType === "author") {
-						item.creators[i].firstName = lastName.slice(1).join(" ");
-						item.creators[i].lastName = lastName[0];
-					}
-					else {
-						Z.debug("No span tag for " + item.creators[i].lastName);
-						item.creators[i].firstName = lastName[0];
-						item.creators[i].lastName = lastName.slice(1).join(" ");
-					}
-					delete item.creators[i].fieldMode;
-				}
+			
+			var authorsNodes = doc.querySelectorAll("div > a.contrib-search");
+			if (authorsNodes.length > 0) {
+			   // prefer the authors information from the website as it contains the last and first name separately
+			   // where the RIS data does not separate them correctly (it uses a space instead of comma)
+			   // but the editors are only part of the RIS data
+			   var authors = [];
+			   for (let author of authorsNodes) {
+				  authors.push({
+					 firstName: text(author, "span.given-names"),
+					 lastName: text(author, "span.surname"),
+					 creatorType: "author"
+				  });
+			   }
+			   var otherContributors = item.creators.filter(creator => creator.creatorType !== "author");
+			   item.creators = otherContributors.length !== 0 ? authors.concat(separateNames(otherContributors)) : authors;
 			}
+			else {
+			   Z.debug("No tags available for authors");
+			   item.creators = separateNames(item.creators);
+			}
+
 			if (item.date) {
 				item.date = ZU.strToISO(item.date);
 			}
@@ -180,6 +180,21 @@ function scrape(doc, url) {
 	});
 }
 
+function separateNames(creators) {
+	for (let i = 0; i < creators.length; i++) {
+		var lastName = creators[i].lastName.split(" ");
+		// Only authors are in the format lastname firstname in RIS
+		// Other creators are firstname lastname
+		if (creators[i].creatorType === "author") {
+			creators[i].firstName = lastName.slice(1).join(" ");
+			creators[i].lastName = lastName[0];
+		}
+		else {
+			creators[i].firstName = lastName[0];
+			creators[i].lastName = lastName.slice(1).join(" ");
+		}
+	}
+}
 /** BEGIN TEST CASES **/
 var testCases = [
 	{
@@ -350,7 +365,8 @@ var testCases = [
 				"volume": "106",
 				"attachments": [
 					{
-						"title": "Snapshot"
+						"title": "Full Text PDF",
+						"mimeType": "application/pdf"
 					}
 				],
 				"tags": [
@@ -491,7 +507,8 @@ var testCases = [
 				"volume": "51",
 				"attachments": [
 					{
-						"title": "Snapshot"
+						"title": "Full Text PDF",
+						"mimeType": "application/pdf"
 					}
 				],
 				"tags": [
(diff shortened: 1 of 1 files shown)


dstillman:
Can this just be rewritten without `\w`? `\w` doesn't work on Unicode, so it shouldn't be used in the modern world.

placardo:
According to @adam3smith comment l.149 
> authors are in RIS as lastname firstname(s), though not necessarily correctly

Maybe it would be more efficient to scrape the page where we find 
```html 
<span class="given-names">Rachel V.</span>
<span class="surname">Kutz-Flamenbaum</span>
```

dstillman:
If we want to use the RIS, we can just split on the first space without using `\w`.

Maybe we should use scraping but fall back to RIS (split on the first space, and with a Z.debug() warning) if those elements aren't found, so that it continues to work if they change the page?

placardo:
Splitting on the first space would not work on most latin american names (eg https://www.emerald.com/insight/content/doi/10.1108/eb058217/full/html) but this was a problem already before, so I guess scraping with a fallback would be best.

Also from this [example](https://www.emerald.com/insight/content/doi/10.1108/S0163-786X(2012)0000033008/full/html) of a book chapter: 
1. RIS is lastname firstname for authors and firstname lastname for editors (I don' t know about other types though, nor if this is always the case)
```
AU  - Kutz-Flamenbaum Rachel V.
AU  - Staggenborg Suzanne
AU  - Duncan Brittany J.
ED  - Jennifer Earl
ED  - Deana A. Rohlinger
```
2. The spans with the name info are only for authors. So for those we can directly use RIS

About a Z.debug warning, where does that appear when the translator is used?

dstillman:
> RIS is lastname firstname for authors and firstname lastname for editors (I don' t know about other types though, nor if this is always the case)

Yeah, I noticed that. This is a mess, but we can assume this is just for `AU` until we have evidence to the contrary.

> About a Z.debug warning, where does that appear when the translator is used?

It appears in debug output – in Scaffold, in the extension's background page with Show in Console enabled in the Connector preferences, or in debug output from the Connector submitted by users.

placardo:
Is that what you had in mind? I think the only way this would break is if two authors had the same surname, then it would put the same given-name but this seems like a marginal case. The tests did not change after this update except for the added one with the editors where the order is inverted

> in the extension's background page with Show in Console enabled in the Connector preferences
This would be great I was using console.log to debug in the browser until now but I don't understand what you mean by the extension's background page; I enabled Show in Console but the Z.debug does not show up in the browser's console or in the Debug Output Logging tab of the connector's preferences

Emerald Insight.js
```diff
@@ -150,8 +150,16 @@ function scrape(doc, url) {
...
-					item.creators[i].lastName = item.creators[i].lastName.replace(/\s.+/, "");
+					var author = doc.querySelector("a[href*='" + item.creators[i].lastName.split(" ")[0] + "'][class='contrib-search']");
+					if(author) {
+						item.creators[i].firstName = text(author, "span[class=given-names]");
+						item.creators[i].lastName = text(author, "span[class=surname]");
+					}
+					else {
+						Z.debug("No span tag for " + item.creators[i].lastName);
+						item.creators[i].firstName = item.creators[i].lastName.split(" ")[0];
+						item.creators[i].lastName = item.creators[i].lastName.split(" ").slice(1).join(" ");
```
placardo: Is there a way to select all remaining elements of an array in js other than with a slice? In python I use [1:] but I didn't find any equivalent

zuphilip: `slice` is fine. But let us maybe define a variable for `item.creators[i].lastName.split(" ")` and use it instead in the condition (line 152) and subsequent places.

Emerald Insight.js
```diff
@@ -150,8 +150,16 @@ function scrape(doc, url) {
 			// trying to correct for this
 			for (let i = 0; i < item.creators.length; i++) {
 				if (!item.creators[i].firstName && item.creators[i].lastName.includes(" ")) {
-					item.creators[i].firstName = item.creators[i].lastName.match(/^\w+\s+(.+)/)[1];
-					item.creators[i].lastName = item.creators[i].lastName.replace(/\s.+/, "");
+					var author = doc.querySelector("a[href*='" + item.creators[i].lastName.split(" ")[0] + "'][class='contrib-search']");
+					if(author) {
```
zuphilip: Add a space after the `if`.

Emerald Insight.js
```diff
@@ -150,8 +150,16 @@ function scrape(doc, url) {
...
+						item.creators[i].firstName = text(author, "span[class=given-names]");
+						item.creators[i].lastName = text(author, "span[class=surname]");
+					}
+					else {
+						Z.debug("No span tag for " + item.creators[i].lastName);
+						item.creators[i].firstName = item.creators[i].lastName.split(" ")[0];
```
zuphilip: You switch here to the `firstName`, which should only been done for the editors. Can we make an additional condition and do the splitting differently for authors and editors? Moreover, I suggest to make a comment here somewhere, such that we remember also later.

placardo: > Yeah, I noticed that. This is a mess, but we can assume this is just for `AU` until we have evidence to the contrary.

I thought @dstillman previous comment implied that only authors would be lastname firstname and for others it would be firstname lastname every time, but I may have misunderstood

zuphilip: You are going through the `item.creators` list and there are the authors as well as the editors. However, for each of them you have the `creatorType` and therefore you can decide if it is an author or editor.

placardo: Ok I can add an else if in case there was no span info but the creator is still an author. I just assumed that only authors have their info in the span embedded, so that they will always be included in the first `if`

Emerald Insight.js
```diff
@@ -345,8 +350,7 @@ var testCases = [
 				"volume": "106",
 				"attachments": [
 					{
-						"title": "Full Text PDF",
-						"mimeType": "application/pdf"
+						"title": "Snapshot"
```
dstillman: I assume whoever updated the test previously had access to the PDF and therefore got a PDF here, so even if you don't this attachment entry and the one below should be reverted.

Emerald Insight.js
```diff
@@ -147,18 +147,23 @@ function scrape(doc, url) {
...
 			}
 			// authors are in RIS as lastname firstname(s), though not necessarily correctly so
-			// trying to correct for this
+			// trying to correct for this. Editors are firstname lastname
 			for (let i = 0; i < item.creators.length; i++) {
-				if (!item.creators[i].firstName && item.creators[i].lastName.includes(" ")) {
-					var author = doc.querySelector("a[href*='" + item.creators[i].lastName.split(" ")[0] + "'][class='contrib-search']");
-					if(author) {
+				lastName = item.creators[i].lastName.split(" ");
+				if (!item.creators[i].firstName && lastName) {
```
dstillman: I'm not sure I understand this part. What is this condition meant to test? Just if there's no space in the RIS name? I was suggesting that we always use the author spans when the elements are present and only fall back to splitting the RIS authors when no spans are available.

Emerald Insight.js
placardo: Actually I added those two tests as examples of hyphen breakdowns. I must have done it originally from my university connection so that I got the PDF and I thought it would be best to retain the more general case where we don't get any pdf. Could you confirm you want it reverted?

Emerald Insight.js
placardo: Originally the regex was tested on this and it needed a space to return something, I didn't threw it away in case there is nothing in the RIS data.

> I was suggesting that we always use the author spans when the elements are present and only fall back to splitting the RIS authors when no spans are available

I think that's what it does, the spans are wrapped in the `a` tag that is queried on the next line. If nothing is returned we fall back to a split of the RIS data (different for authors and editors as per @zuphilip remark)

Emerald Insight.js
dstillman: Yes, I think it'd be better to test the more important functionality, even if that means it will fail for people without that.

Emerald Insight.js
zuphilip: I think what Dan suggest here is something like this code:
```Js
var authorsNodes = doc.querySelectorAll("a.contrib-search");
if (authorsNodes.length > 0) {
   // prefer the authors information from the website as it contains the last and first name separately
   // where the RIS data does not separate them correctly (it uses a space instead of comma)
   // but the editors are only part of the RIS data
   var authors = [];
   for (let author of authorsNodes) {
      authors.push({
         firstName: text(author, "span.given-names"),
         lastName: text(author, "span.surname"),
         creatorType: "author"
      });
   }
   var otherContributors = item.creators.filter(creator => creator.creatorType !== "author");
   item.creators = authors.concat(separateNames(otherContributors));
}
else {
   Z.debug("No span tag for " + item.creators[i].lastName);
   item.creators = separateNames(item.creators);
}
```

Note: this codes is untested and you would then need to write a new function `separateNames` by copying some of the lines of the current code.

placardo: Thanks a lot @zuphilip I was a bit lost on that one

Emerald Insight.js
```diff
@@ -167,6 +180,24 @@ function scrape(doc, url) {
 	});
 }
 
+function separateNames(creators) {
+	for (let i = 0; i < creators.length; i++) {
+		var lastName = creators[i].lastName.split(" ");
+		// Only authors are in the format lastname firstname in RIS
+		// Other creators are firstname lastname
+		if (creators[i].creatorType === "author") {
```
zuphilip:
```suggestion
		if (!creators[i].firstName && lastName.length > 1) {
			if (creators[i].creatorType === "author") {
```

Emerald Insight.js
```diff
@@ -167,6 +180,24 @@ function scrape(doc, url) {
...
+			creators[i].firstName = lastName.slice(1).join(" ");
+			creators[i].lastName = lastName[0];
+		}
+		else {
+			creators[i].firstName = lastName[0];
+			creators[i].lastName = lastName.slice(1).join(" ");
+		}
+		delete creators[i].fieldMode;
```
zuphilip:
```suggestion
			delete creators[i].fieldMode;
		}
```

zuphilip: Some additional spacing should then been done as well inside this condition.

zuphilip:
LGTM! @dstillman Let us know if this is fine for you as well now.

dstillman:
Merged — thank you!

Issue #2088, opened 2019-12-16
JSTOR dropping `https://` from URLs


owcz:
Recently JSTOR has been dropping `https://` from URLs
Haven't looked into whether it's their end or the translator's but either way the translator should handle, yes?

dstillman:
Could you clarify what you mean by "dropping `https://` from URLs"?

owcz:
When using the JSTOR translator, the imported URL field no longer includes `https://` (i.e., they start with `www.`)

dstillman:
Someone will need to look at this, but in the meantime in the latest Zotero beta I've fixed the bug that caused it to open URLs without the scheme in an internal viewer instead of launching them as external webpages.

placardo:
Apparently it was dropped in the RIS data used to import items, probably in mid-november 2019. Should it be added back in the scraper? Once the bug is fixed this shouldn't pose any problem

dstillman:
> `UR  - www.jstor.org/stable/27592221`

Yes, we should add it back conditionally so that when they fix it we don't include it twice.

@adam3smith, do you know of anyone at JSTOR to report the RIS problem to?

PR #2093, opened 2019-12-20
JSTOR - https + copyright

placardo:
See #2088

### DIFF
--- a/JSTOR.js
+++ b/JSTOR.js
@@ -9,22 +9,49 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2019-02-06 21:16:48"
+	"lastUpdated": "2019-12-22 17:00:53"
 }
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2019 Simon Kornblith, Sean Takats, Michael Berkowitz, Eli Osherovich, czar
+
+	This file is part of Zotero.
+
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+// attr()/text() v2
+// eslint-disable-next-line
+function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null;}function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null;}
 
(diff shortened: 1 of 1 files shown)


JSTOR.js
```diff
@@ -193,7 +221,7 @@ function processRIS(text, jid) {
 		}
 		
 		item.url = item.url.replace('http:','https:'); // RIS still lists http addresses while JSTOR's stable URLs use https
-		
+		if(!item.url.startsWith("http")) item.url = "https://" + item.url;
```
dstillman: To be safe, let's use `if (item.url && !item.url.startsWith("http"))`, in case the URL is ever empty.

(Also missing the space after `if` for the linter.)

placardo: Ok also I just commented the last function because it threw an error (not used) should it just be deleted? For the copyright I used the names in the metadata and put 2019 as copyright year

dstillman: Yeah, it looks like that function can be removed after fb9e9e3d0ca4a6155b16d0f71940915d9e096004.

placardo: Do you remember every commit or did you have to search? (if so how do you go about to find a commit on this problem as quickly?)

adam3smith: `git blame` or the equivalent "Blame" button in the github gui is very helpful for this sort of thing.

adam3smith:
This lgtm -- could you squash the last three commits into one?

placardo:
Well apparently that did not work - in the interest of time saving could you tell me how to squash my commits? I thought you did this on your end when you merged

JSTOR.js
placardo: thanks this looks like a time saver

dstillman: In a case like this, where you want to find something that was deleted, `git log -S finalizeItem JSTOR.js` is a good option.

dstillman:
I'll squash this appropriately.

dstillman:
Actually, @placardo, could I ask you to separate these into appropriate commits? It'll be something that we'll want for any translator changes, so it'd be good to understand the git workflow (though it'll get much easier once most translators have been linted). And while I could do this, I can't do so easily while crediting you for the commits.

What you can do here is do `git reset <hash of your first commit>^` (note the `^` at the end), which will reset your index to the commit before your first commit (i.e., wipe out all your commit messages) while leaving all the file changes intact. You can then use `git add -p` to add back parts of the files that make up logical commits. Sometimes, if there are adjacent changed lines, it's necessary to temporarily revert/delete a line in your editor, save the file, do a `git add -p` to get just the line(s) you want, and then undo the changes in the editor. After collecting all the lines you want to add to a given commit and reviewing with `git diff --cached`, do a commit, and then use `git add -p` again for the next commit.

It'd be OK to have just two commits: one with all the functional changes (described in the commit message) and another with just the linting changes. What's important is that we can easily see just the functional changes separate from linting changes.

Once you've reviewed your git log locally to see that you have two commits, you can force-push back to this branch (and post here, since we may not get a notification).

Let me know if you have any trouble with this. Once you get the hang of it, it should only take a few seconds.

zuphilip:
(An alternative way here would be to use interactive rebase `git rebase -i` and let the first commit as it is and squash all the other commits into the second one.)

placardo:
Thanks a lot @dstillman for your explanation! (I tried the interactive rebase but I couldn't get it right) `git add -p` allows to edit directly the hunks so this is very practical. I put the new copyright and the updated tests with the linting commit, so that only the added http is in the first one tell me if this was not right

dstillman:
Yeah, I didn't suggest interactive rebase (or do it myself) because the changes were a bit more mixed up — there are a few more things from the second commit that I might've separated from the linting (e.g., the `attr`/`text` change, the test change) — but we don't need to obsess about this. Merged. Thank you!

dstillman:
Oh, I forgot to mention it here, but as a convention we tend to prefix each commit with the translator name (e.g., "JSTOR: ").

Issue #1811, opened 2019-01-23
ACM Digital Library Next version


javidcf:
ACM is testing a beta of the new website for their digital library, hosted under https://dlnext.acm.org. See for example: [The status of the P versus NP problem](https://dlnext.acm.org/doi/10.1145/1562164.1562186).

It seems still incomplete, PDF links redirect you to the old version and I'm not sure if the new pages have all the information yet. However, I'm filing the issue to have it present in the backlog and, with any luck, maybe have the translator implemented before it gets out of beta.

GuyAglionby:
The new version comes out of beta on 1 January, I'll see if I can get something together before then (unless anyone else has already made a start!).

zuphilip:
I don't think anybody has started on this and therefore it would be perfect if you @GuyAglionby can work on this!

anjiro:
Helpful for whomever works on this: you can get a dump of the metadata for a paper with just the DOI, which is in the URL for the paper itself:

https://dl.acm.org/action/exportCiteProcCitation?dois=10.1145/1562164.1562186&targetFile=custom-bibtex&format=bibTex

GuyAglionby:
Thanks @anjiro!

zuphilip:
For the reference: https://forums.zotero.org/discussion/80692/acm-dl-has-changed-zotero-connector-download-pdf-failed#latest

zuphilip:
Fixed now thanks to @GuyAglionby !

PR #2097, opened 2019-12-27
support download mutiple items from a conference page on ACM Digital …

fzyz999:
This patch is designed for enabling users to downloading multiple items from ACM Digital Library.  Thus, users  can easily download all papers which are released on a conference. For example, if someone want to download all papers of ISCA19, he can visit [https://dl.acm.org/citation.cfm?id=3307650](https://dl.acm.org/citation.cfm?id=3307650) , clicking `Table of Contents` and downloading all the papers listed in the tab (This patch can detect the conference page and enable downloading multiple items).

### DIFF
--- a/ACM Digital Library.js
+++ b/ACM Digital Library.js
@@ -32,20 +32,28 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 
 // attr()/text() v2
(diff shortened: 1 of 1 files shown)


zuphilip:
Currently, we are extracting the proceeding as a whole on this website. That is not wrong, but maybe most persons are more interested in the proceeding articles itself than the collection of them. However, I am not sure how many people want to add several (or all) papers from the table of contents without clicking on these papers first. But I am fine to switch to multiples on proceedings page instead of extracting only the proceeding itself. Should we make the complete proceeding as the first option? What do you think about such changes @adam3smith?

However, we should first support ACM's new design (see #1811) which will come to production in some days. Therefore I suggest to put this issue on the hold for the moment.

fzyz999:
I find that the new ACM translator already support downloading multiple articles from a proceeding. So I think this pull-request can be closed. Thanks. I really love this feature.

PR #2099, opened 2019-12-27
Fix PubMed failing tests

GuyAglionby:
Tests were failing because:

* E-utils URI was filled in with an object rather than a string
* Affiliations XPath had been updated
    * I kept the old path as an option in case it's still used somewhere; I couldn't find any instance of this but understand that PubMed is quite large so may have missed something.
* Some of the underlying info has changed

Thanks!

### DIFF
--- a/PubMed.js
+++ b/PubMed.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 12,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2019-12-27 18:42:40"
+	"lastUpdated": "2019-12-28 15:34:11"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


PubMed.js
```diff
@@ -198,10 +198,16 @@ function detectWeb(doc, url) {
...
+	if (bookCitation.length > 0 && ZU.xpath(doc, './/div[@class="affiliations"]')) {
 		// For a bookSection there are the affiliations of the authors of this
 		// section as well as the affiliations of the book authors.
-		var bookAffiliations = ZU.xpath(doc.getElementById('full-authors'), './/div[@class="affiliations"]/h3[@class="title"]').length > 1;
+		var bookAffiliations;
+		if (doc.getElementById('full-authors')) {
+			bookAffiliations = ZU.xpath(doc.getElementById('full-authors'), './/div[@class="affiliations"]/h3[@class="title"]').length > 1;
+		}
+		else {
+			bookAffiliations = ZU.xpath(doc, '//div[@id="full-view-heading"]//div[@class="affiliations"]//li').length > 1;
```
zuphilip: No, this looks wrong. We want to distinguish whether there are two affiliation sections (one for the chapter authors and one for the book authors) but not whether there are multiple affiliations. Thus, the final `li` should be deleted.

PubMed.js
```diff
@@ -198,10 +198,16 @@ function detectWeb(doc, url) {
...
+		}
 		return bookAffiliations ? "bookSection" : "book";
```
zuphilip: How about simplifying this into something like:
```js
// For a bookSection there is the affiliations section of the authors of this
// section as well as another affiliation sections for the book authors.
var isChapter = doc.querySelectorAll('full-view-heading div.affiliations, #full-view-heading div.affiliations).length > 1;
return isChapter ? "bookSection" : "book";
```
(untested code)? Shouldn't this do what we want?

GuyAglionby:
Thanks for the comments -- I misunderstood the idea of checking multiple affiliations, which the amended comment clears up. Your suggested query works well, though I think `doc.querySelectorAll('#full-view-heading div.affiliations')` would suffice?

Have tested this on the old and new sites for a book ([old](https://www.ncbi.nlm.nih.gov/pubmed/20821847), [new](https://pubmed.ncbi.nlm.nih.gov/20821847-endocrinology-an-integrated-approach/)) and book section ([old](https://www.ncbi.nlm.nih.gov/pubmed/21249758), [new](https://pubmed.ncbi.nlm.nih.gov/21249758-implications-of-peutz-jeghers-syndrome-in-children-and-adolescents)), and detectWeb acts as expected in all cases.

For reference, the failing tests that the getUID change fixes were https://www.ncbi.nlm.nih.gov/books/NBK26374/ and https://www.ncbi.nlm.nih.gov/books/NBK1825/ (i.e. `HTTP GET https://eutils.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi?db=PubMed&tool=Zotero&retmode=xml&rettype=citation&id=[object Object] failed with status code 400`)

zuphilip:
Perfect! :rocket: Thank you very much @GuyAglionby !

PR #2101, opened 2019-12-31
Rewrite ACM translator for new site

GuyAglionby:
Supports the new ACM site (#1811), which has changed over already. The tests don't run (see below), but I've manually tested that it all works. 

A couple of notes: I return that magazines and newsletters are `journalArticle` as this is the type exported by their CSL. I also wasn't sure what the best way of dealing with ISBN was in cases where it is provided but the appropriate item type doesn't have a field for it, so put it in the extras. 

With regard to the conversation in #2097, a conference proceedings page will bring up a dialog for users to select papers in that proceedings rather than download the proceedings information. This is easily changed depending on what we decide to be most reasonable. 

There is a slight problem with the new site in that it does not work without the cookie `cookiePolicy=accept` set: it just redirects to https://dl.acm.org/action/cookieAbsent. As a result I've not been able to run the translator in Scaffold, either manually or with the tests. Manually setting the headers in `doGet` to include the cookie doesn't work either, as it looks like XMLHttpRequest doesn't allow that. In any case I don't think that would address the problem with the tests.

### DIFF
--- a/ACM Digital Library.js
+++ b/ACM Digital Library.js
@@ -1,203 +1,200 @@
 {
 	"translatorID": "f3f092bf-ae09-4be6-8855-a22ddd817925",
 	"label": "ACM Digital Library",
-	"creator": "Simon Kornblith, Michael Berkowitz, John McCaffery, and Sebastian Karcher",
-	"target": "^https?://([^/]+\\.)?dl\\.acm\\.org/(results|citation|author_page|ccs/ccs)\\.cfm",
+	"creator": "Guy Aglionby",
+	"target": "^https://dl\\.acm\\.org/(doi|do|profile|toc|topic|keyword|action/doSearch|acmbooks)",
 	"minVersion": "3.0",
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2019-10-03 09:45:46"
+	"lastUpdated": "2019-12-31 14:12:38"
 }
 
 /*
-ACM Digital Library Translator
-Copyright (C) 2011 Sebastian Karcher and CHNM
+	***** BEGIN LICENSE BLOCK *****
 
-This program is free software: you can redistribute it and/or modify
-it under the terms of the GNU General Public License as published by
-the Free Software Foundation, either version 3 of the License, or
-(at your option) any later version.
+	Copyright © 2019 Guy Aglionby
+	This file is part of Zotero.
 
-This program is distributed in the hope that it will be useful,
-but WITHOUT ANY WARRANTY; without even the implied warranty of
-MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
-GNU General Public License for more details.
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
 
-You should have received a copy of the GNU General Public License
-along with this program. If not, see <http://www.gnu.org/licenses/>.
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero.  If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
 */
 
-
-// attr()/text() v2
-function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null;}function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null;}
-
-
 function detectWeb(doc, url) {
-	if (url.includes("/results.cfm") || url.includes("/author_page.cfm") || url.includes("/ccs/ccs.cfm")) {
-		return getSearchResults(doc, true) ? 'multiple' : false;
-	} else if (url.includes("/citation.cfm")) {
-		return getArticleType(doc);
+	if ((url.includes('/doi/') || url.includes('/do/')) && !url.includes('/doi/proceedings')) {
+		let extractedContext = doc.querySelector('meta[name=pbContext]').content;
+		let subtypeRegex = /csubtype:string:(\w+)/;
+		let subtype = extractedContext.match(subtypeRegex)[1].toLowerCase();
+
+		if (subtype == 'conference') {
+			return 'conferencePaper';
+		}
+		else if (subtype == 'journal' || subtype == 'magazine' || subtype == 'newsletter') {
+			return 'journalArticle';
+		}
+		else if (subtype == 'report' || subtype == 'rfc') {
+			return 'report';
+		}
+		else if (subtype == 'thesis') {
+			return 'thesis';
+		}
+		else if (subtype == 'software') {
+			return 'computerProgram';
+		}
+		else if (subtype == 'dataset') {
+			return 'document';
+		}
+		else if (subtype == 'book') {
+			let bookTypeRegex = /page:string:([\w ]+)/;
+			let bookType = extractedContext.match(bookTypeRegex)[1].toLowerCase();
+			if (bookType == 'book page') {
+				return 'book';
+			}
+			else {
+				return 'bookSection';
+			}
+		}
 	}
+	else if (getSearchResults(doc, false)) {
+		return 'multiple';
+	}
+	return false;
 }
 
 function doWeb(doc, url) {
 	if (detectWeb(doc, url) == "multiple") {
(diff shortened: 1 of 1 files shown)


ACM Digital Library.js
```diff
@@ -1,203 +1,200 @@
...
+			delete cslItem.editor;
+		}
+		
+		let item = new Zotero.Item();
+		ZU.itemFromCSLJSON(item, cslItem);
+		
+		let abstractElements = doc.querySelectorAll('div.article__abstract p, div.abstractSection p');
+		let abstract = Array.from(abstractElements).map(function (element) {
+			return element.textContent;
+		}).join(' ');
```
zuphilip: You can also use the short form
```
map(x => x.textContent)
```
Moreover, I would suggest to use `.join('\n\n')` instead, because you are dealing here with different paragraphs in `p` nodes.

ACM Digital Library.js
```diff
@@ -1,203 +1,200 @@
...
+			return 'report';
+		}
+		else if (subtype == 'thesis') {
+			return 'thesis';
+		}
+		else if (subtype == 'software') {
+			return 'computerProgram';
+		}
+		else if (subtype == 'dataset') {
+			return 'document';
```
zuphilip: Add in this case then also `itemType: data` in the extra somewhere below.

ACM Digital Library.js
```diff
@@ -1,203 +1,200 @@
...
-				item.complete();
+		}
+		
+		if (item.itemType == 'bookSection' && item.creators.length == 0) {
+			// A chapter of a book is extracted as a bookSection, but the
+			// authors are not included in the CSL so we must scrape them.
+			// e.g. https://dl.acm.org/doi/abs/10.5555/3336323.C5474411
+			let authorElements = doc.querySelectorAll('div.citation span.loa__author-name');
+			authorElements.forEach(function (element) {
+				item.creators.push(ZU.cleanAuthor(element.textContent));
```
zuphilip: :bug: This function `cleanAuthor` needs a second argument for the creatorType!

ACM Digital Library.js
```diff
@@ -1,203 +1,200 @@
...
-				}
-			} else {
-				item.complete();
+		}
+		
+		if (item.itemType == 'bookSection' && item.creators.length == 0) {
+			// A chapter of a book is extracted as a bookSection, but the
+			// authors are not included in the CSL so we must scrape them.
+			// e.g. https://dl.acm.org/doi/abs/10.5555/3336323.C5474411
+			let authorElements = doc.querySelectorAll('div.citation span.loa__author-name');
```
zuphilip: We could do this always when `item.creators` array is empty but `authorElements` not, i.e. there is no reason to restrict this to bookSections only.

ACM Digital Library.js
```diff
@@ -1,203 +1,200 @@
...
+			if (!item.extra) {
+				item.extra = '';
 			}
-			
+			let isbnLength = cslItem.ISBN.replace('-', '').length;
+			item.extra += '\nISBN-' + isbnLength + ': ' + cslItem.ISBN;
+		}
+		
+		let pagesElement = doc.querySelector('div.pages-info span');
+		if (pagesElement) {
```
zuphilip:
```suggestion
		if (pagesElement && !item.numPages) {
```

ACM Digital Library.js
```diff
@@ -1,203 +1,200 @@
...
 
+	***** END LICENSE BLOCK *****
+*/
 
 function detectWeb(doc, url) {
-	if (url.includes("/results.cfm") || url.includes("/author_page.cfm") || url.includes("/ccs/ccs.cfm")) {
-		return getSearchResults(doc, true) ? 'multiple' : false;
-	} else if (url.includes("/citation.cfm")) {
-		return getArticleType(doc);
+	if ((url.includes('/doi/') || url.includes('/do/')) && !url.includes('/doi/proceedings')) {
```
zuphilip: Can you also give an example/test case for a `/do/` url?

zuphilip:
@GuyAglionby Let me know if you can work on this within the next days. Otherwise, I can maybe try to work a little further here.

GuyAglionby:
Sorry about the delay here -- I should have time tomorrow to have a look

ACM Digital Library.js
```diff
@@ -1,203 +1,200 @@
...
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
 
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU Affero General Public License for more details.
 
-// attr()/text() v2
-function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null;}function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null;}
```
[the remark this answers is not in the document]
GuyAglionby: I misremembered the discussion in another issue re. those functions as getting rid of them entirely rather than incorporating them as globals. Thanks for the heads up!

ACM Digital Library.js
```diff
@@ -1,203 +1,200 @@
...
-		return getArticleType(doc);
+	if ((url.includes('/doi/') || url.includes('/do/')) && !url.includes('/doi/proceedings')) {
+		let extractedContext = doc.querySelector('meta[name=pbContext]').content;
+		let subtypeRegex = /csubtype:string:(\w+)/;
+		let subtype = extractedContext.match(subtypeRegex)[1].toLowerCase();
+
+		if (subtype == 'conference') {
+			return 'conferencePaper';
+		}
+		else if (subtype == 'journal' || subtype == 'magazine' || subtype == 'newsletter') {
```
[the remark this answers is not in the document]
GuyAglionby: [Here](https://dl.acm.org/doi/10.1145/2854146) is a magazine ([corresponding test](https://github.com/zotero/translators/pull/2101/commits/a853a19e0dfa55c877a5731e71378e362f9f00a1#diff-2b2fb3a06b538f7aa0f0d1f03670c891R558)); [here](https://dl.acm.org/doi/10.1145/3264631.3264634) is a newsletter ([corresponding test](https://github.com/zotero/translators/pull/2101/commits/a853a19e0dfa55c877a5731e71378e362f9f00a1#diff-2b2fb3a06b538f7aa0f0d1f03670c891R518)).

ACM Digital Library.js
```diff
@@ -1,203 +1,200 @@
...
+	for (let i = 0; i < results.length; i++) {
+		let url = results[i].href;
+		let title = ZU.trimInternal(results[i].textContent);
+		if (!title || !url) {
+			continue;
+		}
 		
-		if (checkOnly) return true;
-		found = true;
+		if (!(url.includes('/doi/') || url.includes('/do/')) || url.includes('/doi/proceedings')) {
```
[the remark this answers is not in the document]
GuyAglionby: I think this was already a negation, but to avoid any future confusion I've pulled this check out into its own function

ACM Digital Library.js
GuyAglionby: I forgot about this notation, thanks

ACM Digital Library.js
GuyAglionby: `/do/` includes [computer programs](https://dl.acm.org/do/10.1145/3159287/full/) and [datasets](https://dl.acm.org/do/10.1145/3276998/full/). However, it looks like the CSL endpoint currently gives a 500 when trying to retrieve information for the DOIs corresponding to these types of entry (I checked a few). This also breaks the 'export citation' functionality on the website itself, though I recall it working before. Not sure if it's worth disabling `/do/` in the mean time (I'm not sure how long it's been broken, but have just emailed about it).

GuyAglionby:
Thanks for the comments -- have made changes as requested. The tests started working for me in Scaffold (without any changes, not sure what happened) so I updated the already-present ones, and added new ones to cover both new item types and some of the additional scraping that is done.

One thing `ZU.itemFromCSLJSON` does is return creators with a `creatorTypeID`, which doesn't seem necessary although I'm not sure it's doing any harm. Can remove that field if necessary.

ACM Digital Library.js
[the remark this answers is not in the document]
zuphilip: No problem and you remember correctly, that this is the goal. However, AFAIK we are not yet there to drop these lines. Anyways it should be easy to do this then for a lot of translators at once, when it is ready.

ACM Digital Library.js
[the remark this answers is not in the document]
zuphilip: So, they don't have an submit area for authors and are maybe otherwise a little different. However, I see that dblp is classify them also as journals. Thus, I am okay with this now.

ACM Digital Library.js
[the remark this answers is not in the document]
zuphilip: Ah, the parentheses ...

ACM Digital Library.js
zuphilip: Thank you for the examples. No, we don't need to disable this here. As far as we know this should work on the Zotero side. Thank you also for reaching out to them (metadata matters!).

zuphilip:
Okay, great! Thank you very much! I have only some small nits pushed directly to your branch. I was not able to use Scaffold for testing. The additional field for the creators should be okay.

zuphilip:
Merged now! :rocket: Thank you very much! :bowing_man:

GuyAglionby:
Thanks!

PR #2102, opened 2019-12-31
Fix Google AI translator

GuyAglionby:
URL changed and so did the Bibtex copy link. 

I also don't know if this translator may now be more accurately called 'Google Research'?

### DIFF
--- a/Google AI.js
+++ b/Google AI.js
@@ -2,14 +2,14 @@
 	"translatorID": "5f1c4a3b-b7cf-4170-a896-e4d82c0621c9",
 	"label": "Google AI",
 	"creator": "Guy Aglionby",
-	"target": "^https://research\\.google/(pubs|people|research-areas)",
+	"target": "^https://research\\.google/(pubs|people|research-areas|teams)",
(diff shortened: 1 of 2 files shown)


Google AI.js
```diff
@@ -204,6 +205,16 @@ var testCases = [
...
+	},
+	{
+		"type": "web",
+		"url": "https://research.google/research-areas/algorithms-and-theory/",
+		"items": "multiple"
+	},
+	{
+		"type": "web",
+		"url": "https://research.google/teams/applied-science/gas/",
+		"items": "multiple"
```
zuphilip: Are the multiples now working as automatic tests? There is also an comment in the code that they might not work correctly. Anyway, you should either update the comment or delete the test cases for multiples again here.

GuyAglionby:
Only some of the multiple are -- the two in the comments are populated after the page load while the two included as tests are not, so the latter work and the former don't. I've added a comment that hopefully clarifies this

zuphilip:
Thank you very much @GuyAglionby ! Merged now. :rocket:

PR #2103, opened 2019-12-31
Semantic Scholar extend PDF extraction + fix errors when logged in

GuyAglionby:
Some relatively minor changes

* Change method of checking if a PDF is available, as they removed the `hasPDF` property
* Go through list of alternative paper URLs to find a PDF if the main link isn't one (the ones identified the old way, via 's2' and 'arxiv', all ended in .pdf already).
* More robust way of combing through the encoded data, as the previous method broke when you were logged in
* Updated the tests to reflect underlying changes on website (including new URLs). Removed one as it now redirects to a different paper. It wasn't covering anything the others don't.

### DIFF
--- a/Semantic Scholar.js
+++ b/Semantic Scholar.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2019-12-31 17:36:50"
+	"lastUpdated": "2020-01-10 01:34:44"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


Semantic Scholar.js
```diff
@@ -122,21 +129,37 @@ function parseDocument(doc, url) {
...
-				if (arxivId.length >= 1) {
-					if (item.extra) {
-						item.extra += '\narXiv: ' + arxivId[0];
+		if (rawData.primaryPaperLink) {
+			let pdfLinkElement;
+			if (rawData.primaryPaperLink.url.endsWith('.pdf')) {
+				pdfLinkElement = rawData.primaryPaperLink;
+			}
+			else if (rawData.alternatePaperLinks) {
+				for (let i = 0; i < rawData.alternatePaperLinks.length; i++) {
```
zuphilip:
```suggestion
				for (let alternateElement of rawData.alternatePaperLinks) {
```

zuphilip: Then the following line is not anymore needed (and you never use the variable `i` anyways here), so this is a simplification of your code.

Semantic Scholar.js
```diff
@@ -122,21 +129,37 @@ function parseDocument(doc, url) {
...
+					let alternateElement = rawData.alternatePaperLinks[i];
+					if (alternateElement.url.endsWith('.pdf')) {
```
zuphilip:
```suggestion
					if (!pdfLinkElement && alternateElement.url.endsWith('.pdf')) {
```

zuphilip: Then delete the `break` below and move the code about the arXiv ID up here. (This then covers the cases that there is pdf link which we use to the element, but there is also also another link following to arXiv.)

Semantic Scholar.js
```diff
@@ -122,21 +129,37 @@ function parseDocument(doc, url) {
...
+			}
+			
+			if (pdfLinkElement) {
+				item.attachments.push({
+					url: pdfLinkElement.url,
+					title: "Full Text PDF",
+					mimeType: 'application/pdf'
+				});
+				
+				if (pdfLinkElement.linkType == 'arxiv') {
```
zuphilip: Move this code block up.

Semantic Scholar.js
```diff
@@ -680,16 +639,19 @@ var testCases = [
...
+				"abstractNote": "We present a new dataset and models for comprehending paragraphs about processes (e.g., photosynthesis), an important genre of text describing a dynamic world. The new dataset, ProPara, is the first to contain natural (rather than machine-generated) text about a changing world along with a full annotation of entity states (location and existence) during those changes (81k datapoints). The end-task, tracking the location and existence of entities through the text, is challenging because the causal effects of actions are often implicit and need to be inferred. We find that previous models that have worked well on synthetic data achieve only mediocre performance on ProPara, and introduce two new neural models that exploit alternative mechanisms for state prediction, in particular using LSTM input encoding and span prediction. The new models improve accuracy by up to 19%. The dataset and models are available to the community at this http URL",
+				"extra": "arXiv: 1805.06975",
 				"itemID": "Dalvi2018TrackingSC",
 				"libraryCatalog": "Semantic Scholar",
 				"proceedingsTitle": "NAACL-HLT",
+				"publicationTitle": "NAACL-HLT",
```
zuphilip: Are both fields here come from Scaffold?

This looks strange as `proceedingsTitle` is only some sort of alias to `publicationTitle`...

GuyAglionby: Scaffold doesn't work well with updating tests for some reason -- the [element](https://github.com/zotero/translators/pull/2103/files#diff-ebdd943ca6e2a8dcc2fe2810642d2cc9L53) used to determine type in `detectWeb` isn't found for some reason, even with `defer: true`. This is despite it appearing when I wget or curl a page. Not sure why this is, but the tests run/pass as usual.

GuyAglionby:
Thanks for the review -- I don't think I understood the exact changes you suggested with the arXiv IDs, but I think the amended code should incorporate the idea

adam3smith:
I believe this is superseded by #2112 which includes PDF scraping, but haven't compared closely

GuyAglionby:
Yep, I think that's the case

PR #2105, opened 2020-01-02
Update CNKI.js

l0o0:
Hi, I have received some feedbacks after I update the CNKI translator. Now I update the CNKI translator:

1. Get attachments in the search result page 
2. The RefWorks format provided by CNKI for conference paper has an error in `RT` tag. I add a replacement, `Conference Proceeding` to `Conference Proceedings`
3. A pdf attachment will be added by default. But I add a variable `keepPDF` to switch between pdf and caj filetype if someone needs caj files.  

I can not find a better way to detect user login state by now, but the most CNKI user has access to download the pdf/caj files. So I set `var loginStatus = true` by default.

### DIFF
--- a/CNKI.js
+++ b/CNKI.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcs",
-	"lastUpdated": "2019-12-05 08:10:19"
+	"lastUpdated": "2020-01-02 08:05:38"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


l0o0:
After a few days of searching and testing, I finally can use eslint in local. In order to pass the eslint test, I add additional returns in `doWeb()` function. But I don't know why eslint gave `consistent-return`  error without these returns.  I think `doWeb()` can get its returns by `scrape()` function inside.

dstillman:
`consistent-return` just means that the return has to be the same within the same function. Nested functions don't count. So the problem there was simply been that `selectItems()` had `if (!selectedItems) return true;` but there wasn't an explicit return for the remaining code. `selectItems()` doesn't actually need a return value, so you can just change that to `if (!selectedItems) return;` and then remove the three other `return true;` lines you added.

l0o0:
@dstillman  Thank you for your detailed answer. I will update the code.

PR #2112, opened 2020-01-17
Fix Semantic Scholar Plugin

aschokking:
The way the Semantic Scholar plugin was written wasn't going to work and it's consistently throwing errors when attempted to be used today:

image

```
TypeError: rawData is undefined at moz-extension://8698ca6d-2f0d-3740-aef0-c4fcf97e2d0b/inject/translate_inject.js line 63 > eval:125
```

More background: It was trying to read data from a `DATA` variable that only appears in the html when first loading a paper page (used for sharing server side renderered data with the client). However once you load the page navigating within the Single-Page-App takes you to new papers that will not have that variable ever. Since the tests were grabbing fresh html from the server it worked fine there but in an actual browser it would fail 100% of the time that you changed pages. For example if you load a paper, click a citation and then from that new paper try to use the old plugin you'll get a mix of new/old fields mixed together.

----------------------------

In the short-run I'm proposing this simpler implementation that uses the schema-metadata section the site is providing for search engines. This will be stable and reliable. I also removed the Search functionality as it was pretty janky and didn't actually produce useful artifacts (just paper titles but no links/metadata to go along with them).

Longer term we could do something fancier to help expose more data (like the topics/tags) reliably.

### DIFF
--- a/Semantic Scholar.js
+++ b/Semantic Scholar.js
@@ -35,9 +35,17 @@
 	***** END LICENSE BLOCK *****
 */
 
+// See also https://github.com/zotero/translators/blob/master/BibTeX.js
+var bibtex2zoteroTypeMap = {
+	inproceedings: "conferencePaper",
+	conference: "conferencePaper",
+	article: "journalArticle"
+};
+
 function detectWeb(doc, url) {
-	// the regex in target will ensure we're on a paper
-	return 'journalArticle';
+	let citation = ZU.xpathText(doc, '//pre[@class="bibtex-citation"]');
+	let type = citation.split('{')[0].replace('@', '');
+	return bibtex2zoteroTypeMap[type];
 }
 
 function doWeb(doc, url) {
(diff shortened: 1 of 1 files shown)


Semantic Scholar.js
```diff
@@ -75,128 +60,55 @@ function doWeb(doc, url) {
...
-		
-		numbers[1] = baseNumber + numbers[1];
-		
-		return numbers[0] + '-' + numbers[1];
+		item.attachments.push({
+			url: paperLink,
+			title: "Publisher Link",
+			mimeType: "text/html",
+			snapshot: false
+		});
```
dstillman: If this is just a doi.org URL, it doesn't need to be saved, since the DOI field alone is sufficient to resolve it.

If it's an actual publisher URL — i.e., the thing that might be get cited when a style doesn't prefer the DOI — it should be saved to the parent item's URL field instead. Linked-URL attachments are just for secondary sources (e.g., Semantic Scholar page, Amazon product page, etc.).

Semantic Scholar.js
```diff
@@ -221,9 +133,7 @@ var testCases = [
 					}
 				],
 				"date": "2010",
-				"DOI": "10.1007/978-3-642-14770-8_33",
```
dstillman: We should figure out how to keep the DOI, one way or another, even if we have to get it via scraping rather than from the JSON-LD.

Semantic Scholar.js
aschokking: It can be a variety of things, I'm going to pull this part out for now (have added the DOI more reliably below)

Semantic Scholar.js
aschokking: Added the DOI back in by hitting the semantic scholar public api (which should have a very stable response format).

aschokking:
@dstillman I think I've addressed your feedback but let me know if there's anything else blocking.

aschokking:
@dstillman friendly weekly bump

Semantic Scholar.js
```diff
@@ -99,16 +100,16 @@ function parseDocument(doc, url) {
...
-			mimeType: "text/html",
-			snapshot: false
-		});
-	}
 
-	item.complete();
+	// use the public api to retrieve more structured data
+	const paperIdRegex = /\/(.{40})(\?|$)/;
+	const paperId = paperIdRegex.exec(url)[1];
+	const apiUrl = `http://api.semanticscholar.org/v1/paper/${paperId}?client=zotero_connect`;
```
dstillman: This should be HTTPS.

aschokking: Sure agree that would be better (looks like the api docs need to get updated to list https support), fixed.

aschokking:
@dstillman friendly weekly bump

dstillman:
Thanks! Sorry for the delay.

PR #2114, opened 2020-01-26
Project MUSE: update path detection for multi/search

owcz:
Multiples have been broken for a bit. Apparently path has changed from `results` to `search`.


adam3smith:
Thanks!

PR #2115, opened 2020-01-28
EM: Detect 'bylines' class in addition to 'byline' and 'vcard'

dstillman:
Example URL: https://www.pewresearch.org/fact-tank/2019/12/12/u-s-children-more-likely-than-children-in-other-countries-to-live-with-just-one-parent/

Reported here: https://twitter.com/sourceoress/status/1221816388706361344

(It would obviously be much better to use the JSON-LD on that page, but until then, this fixes author importing.)

It looks like there's code to detect when a byline element is within another byline (e.g., &lt;byline&gt; within &lt;bylines&gt;), though I didn't test that.


AbeJellinek:
Removed the lint commit because it causes excessive merge conflicts.

PR #2116, opened 2020-02-04
Primo: Fix tests, update regex, and lint

OhadKim:
Primo2018.js:
I moved 3 tests from primo.js to primo2018.js because all these 3 tests are failed because they are redirected to the new Primo UI (2018). I moved the tests and updated them.
**Pay attention that the first 3 tests are use 'primo-qa.hosted.exlibrisgroup.com'. This server isn't stable and Exlibris use this server only for internal utilization. Additionally, it isn't stable server (down time, deploy etc.)**
One more thing, i remove the 4th test that used 'primo-demo' which most of the time it is down and shouldn't be used outside Exlibris.

Primo.js:
Fixed the "Cannot read property 'trim' of undefined" error message by 'if (text)' command.
Fixed some fields for the tests.

Primo Normalized XML.js:
Fixed the 'tags' parsing by adding one more separator ';'
In the creator section, i added 2 things. The first, i handled the removing of year value like this (1962-...) - i added the option to remove the 3 dots.
The second change, i added one more time the removing of 'creator type' value, because it can be displayed twice like:
Gervais-Lambony, Philippe (1962-...) (Directeur de thèse) (Membre du jury)
(You can see the 5th test in Primo2018)
Last thing that i changed, is updating the tests in this page.
Thanks
Ohad.


OhadKim:
I need some help with the errors on the build. Not sure they are my errors.
Thanks.

adam3smith:
(not your error indeed; we'll fix)

Primo Normalized XML.js
```diff
@@ -342,9 +342,11 @@ function stripAuthor(str) {
 	// e.g. Wheaton, Barbara Ketcham [former owner]$$QWheaton, Barbara Ketcham
 	str = str.replace(/^(.*)\$\$Q(.*)$/, "$2");
 	return str
-		// Remove year
-		.replace(/\s*,?\s*\(?\d{4}-?(\d{4})?\)?/g, '')
-		// Remove things like (illustrator). TODO: use this to assign creator type?
+	// Remove year
+		.replace(/\s*,?\s*\(?\d{4}-?(\d{4}|.{3})?\)?/g, '')
```
adam3smith: do you mean to use the wildcard `.` there? Otherwise this should be `\.{3]`

Primo Normalized XML.js
```diff
@@ -342,9 +342,11 @@ function stripAuthor(str) {
...
+		// Remove creator type like (illustrator)
+		.replace(/\s*,?\s*[[(][^()]*[\])]$/, '')
+		// Sometimes creator type is displayed twice
```
adam3smith: Couldn't you just do `/(\s*,?\s*[[(][^()]*[\])])+$/` instead of duplicating the same line?

adam3smith:
@OhadKim  -- just making sure you've seen my questions above?

OhadKim:
@adam3smith yes sure. Will work on it. thanks.

Primo Normalized XML.js
OhadKim: No, I mean specifically for dots (author like '1962-...')

Primo Normalized XML.js
OhadKim: Yes, you right.

OhadKim:
@adam3smith -- Addressed your comments.

aniamy:
@adam3smith Hi, could you review the changes that were addressed? Or tell us if someone else can take a look? It has been more than a month since we did the changes.

aniamy:
@adam3smith Is there a chance to get this changes reviewed? We are waiting for updates since June and our clients are waiting for this improvement.

aniamy:
@zuphilip  Hi, would it be possible for you to review the changes we did to our code? We would like to have some feedback.

Primo.js
```diff
@@ -197,18 +197,20 @@ function fetchPNX(itemData) {
...
-			text = text.trim();
-			if (text.substr(0, 5) != '<?xml' || text.search(/<error\b/i) !== -1) {
-				// try a different PNX url
-				gotPNX = false;
-				return;
-			}
-			else {
-				gotPNX = true;
-				PNXUrlGenerator.confirmed = true;
+			if (text) {
```
dstillman, line 200: Can you explain a situation where this occurs?

Primo.js
```diff
@@ -391,54 +298,8 @@ var testCases = [
...
-						"tag": "Social aspects"
-					},
-					{
-						"tag": "United States"
-					},
-					{
-						"tag": "United States -- Social conditions -- 1960-"
-					}
+					"LEGGI",
+					"ITALIA - STORIA MEDIOEVALE"
```
dstillman, line 302: When I run this test in the Translator Editor, I get "LEGGI;ITALIA - STORIA MEDIOEVALE". Can you verify that this still works?

Also, how are you updating these tests? I don't believe the Translator Editor, which is what you should be using, will generate a tags array in this format.

Primo Normalized XML.js
dstillman: And it's always three literal dots, an actual ellipsis?

dstillman (changes requested):
Sorry for the delay here. Just a few quick notes, and you'll also need to rebase this against the current master branch, since there are some conflicts.

Primo Normalized XML.js
OhadKim: Yes, this specific pattern

Primo.js
OhadKim: I found that problem on test in Primo.js:
http://princeton-primo.hosted.exlibrisgroup.com/primo_library/libweb/action/dlDisplay.do?vid=PRINCETON&search_scope=All%20Princeton%20Libraries&docId=PRN_VOYAGER2778598&fn=permalink
The error message was: TypeError: Cannot read property 'trim' of undefined
This solution is for the next row, when it try trim for undefined 'text' paramter - text = text.trim();
(Primo.js throw this error but still continue to try import the PNX record.)

Primo.js
OhadKim: yes, you right, it one string. fixed it.
What do how I am updating these tests? this is my goal for this PR.

dstillman: I'm asking what method you're using to update the tests. How are you running the tests, and how are you editing them?

Primo.js
dstillman: I'm asking when `text` would be undefined — I'm not sure why that would happen, unless maybe the connection itself was interrupted. I'm not seeing that on the linked page.

Primo.js
OhadKim: manually by the chrome extension and Zotero software.

dstillman: Meaning you're looking at the fields saved to Zotero and manually editing the tests in the .js file?

For translator development, you need to be doing this in [Scaffold](https://www.zotero.org/support/dev/translators/scaffold) (Tools → Developer → Translator Editor). We can't accept manually updated tests, because they won't match our automated tests, which we depend on to make sure translators are working.

If the site runs JS on load that's necessary for the translator to work, you can add `"defer": true` to the top level of the test and click Populate in the Testing pane to reload the tests. That allows the last three tests in Primo 2018 to run, for example, though they don't currently pass. (The first three primo-qa tests should be removed or replaced with publicly accessible URLs.)

Let us know if you need help working with Scaffold.

OhadKim: @dstillman I used the Scaffold for all the tests. All of them passed successfully besides one Primo2018 test:
http://primo-prod.u-paris10.fr/primo-explore/fulldisplay?vid=UPON&docid=SCD_ALEPH000546633&context=L&search_scope=default_scope
It complained about wrong item type. I looked at the code and it should be 'thesis'.
Not sure why I got this error. Can you take a look at it?

Primo.js
OhadKim: Not sure about that, i removed it and the tests run successfully.

Primo.js
OhadKim: @dstillman - can you take a look please?

OhadKim:
@dstillman  - addressed all the comments.

OhadKim:
@dstillman - Hi, would it be possible for you to give us feedback here?

dstillman:
Sorry, I'll take a look at this later today.

Primo.js
OhadKim: @dstillman ping you about the "Detection returned wrong item type" error message.
Thanks.

adam3smith: The test checks whether the imported item type conforms to detected item type and that's never the case for non-book items since the translator always detects book. You can either see if you can improve the detect function or remove the test. It's not a critical error.

OhadKim: @adam3smith Couldn't find a solution to this error, so removed the problematic test.
The PR is ready for review.

OhadKim:
@dstillman @adam3smith - The PR is ready, thanks in advance.

OhadKim:
@dstillman @adam3smith @zuphilip - Hi, the PR is ready. Would it be possible for you to review the changes we did to our code? We would like to have some feedback.

Primo Normalized XML.js
AbeJellinek: In that case, it should be escaped, as @adam3smith mentioned: `\.{3}`. This current pattern will match any three characters.

AbeJellinek (changes requested):

AbeJellinek:
@OhadKim and @aniamy:

I would like to get these changes merged, but they're a bit hard to review. This should really be three PRs: high-priority functional changes (creator and tag cleaning), medium-priority test updates, and low-priority linting.

The high-priority parts _aren't_ ready - the regex, as was previously mentioned, isn't quite right. The medium-priority parts look good. The low-priority parts just make the rest hard to review.

As @aniamy said,

>our clients are waiting for this improvement.

I'm assuming that they're waiting for the creator/tag-cleaning improvements. Could you trim this PR down to _just_ include those? With a fixed regex, I'd be more than happy to merge that changeset. Linting and test updates can be reviewed separately. Don't worry about any CI lint failures.

OhadKim:
@AbeJellinek - fixed the regex pattern.

AbeJellinek:
OK, looks good. I updated the `lastUpdated` and removed the `attr`/`text` shims, which we no longer need, and I'll go ahead and merge. Thanks, and sorry this took so long to review!

PR #2123, opened 2020-02-17
SAGE Journals - Minor changes for URL detection and TOC pages

### LINKED ISSUE #2121: SAGE Journals detection problems
success (/doi/.../): https://journals.sagepub.com/doi/full/10.1177/1057567719896160
problem (only /doi/): https://journals.sagepub.com/doi/10.1177/1057567719896160
Also problem (/toc/): https://journals.sagepub.com/toc/SCP/current

Seems like the target pattern needs to be fixed. There has been another detection fix a short time ago, see  #2077

Any help would be appreciated!

@placardo and @adam3smith, do you have any suggestions how to extend the target pattern without breaking it?

vcarret:
Simplified the detectWeb and getSearchResults URL targets and added support for toc pages.
Fix #2121

### DIFF
--- a/SAGE Journals.js
+++ b/SAGE Journals.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2019-12-10 18:09:17"
+	"lastUpdated": "2020-02-17 01:41:54"
 }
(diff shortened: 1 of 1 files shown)


AbeJellinek:
The pages this was supposed to extend detection to are actually detected fine now even without this change.

PR #2129, opened 2020-02-21
Update translator to work with new Web Library

tnajdek:
This version can distinguish between the new and old web-library and use appropriate techniques to extract item(s). In addition I've tweaked the code style to match rules in `.eslintrc`.

### DIFF
--- a/zotero.org.js
+++ b/zotero.org.js
@@ -1,15 +1,15 @@
 {
 	"translatorID": "c82c574d-7fe8-49ca-a360-a05d6e34fec0",
 	"label": "zotero.org",
-	"creator": "Dan Stillman and Aurimas Vinckevicius",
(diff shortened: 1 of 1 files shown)


zotero.org.js
```diff
@@ -35,21 +35,20 @@
   ***** END LICENSE BLOCK *****
 */
 
-
-var sessionKey;
+/* eslint consistent-return: 0 */
```
dstillman, line 38: Better to leave this on and just return `false` if the function otherwise returns a value (e.g., no results from `detectWeb()`), so it still catches incorrect branching. (Despite the previous code, `selectItems()` doesn't actually need a return value, so that can just be `return;`.)

tnajdek:
@dstillman thank you for reviewing! I've pushed amended code that adheres to the`eslint consistent-return` rule as suggested.

I've also tweaked the web-library to preselect items for the translator based on current selection.

PR #2131, opened 2020-02-26
Stop using multi import in article main page

ncukondo:
Multi import in article main page does not work. Since Zotero can't recognize javascript result using ZU.proccessDocuments. I remove multiple import in article main page. Instead, the article main page is now import as book section as recommended as UpToDate. In addition, I add multiple import in reference page.

### DIFF
--- a/UpToDate References.js
+++ b/UpToDate References.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcv",
-	"lastUpdated": "2019-06-11 13:44:39"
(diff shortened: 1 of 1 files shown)


AbeJellinek:
Hey, sorry this took a while to get to! I don't quite understand what you mean by

>Since Zotero can't recognize javascript result using ZU.proccessDocuments.

Could you explain the purpose of this change?

adam3smith:
`Since Zotero can't recognize javascript result using ZU.proccessDocuments.`

I think this is short for the fact that running `ZU.processDocuments` on a page whose DOM is built using javascript doesn't work since the XHR request just returns the head and the javascript.

PR #2132, opened 2020-02-26
Bib(La)TeX: Add comma at the end of entry

tomh-overleaf:
See https://forums.zotero.org/discussion/81594/empty-field-in-a-bibtex-file-causing-biber-failing

Biber (one of the tools for creating bibliographfies in LaTeX) is failing when an item is empty (with no entries) and there's no comma after the citekey. As the comma can be at the end of any item (even one with entries), this simply adds the comma to fix the bug.

### DIFF
--- a/BibLaTeX.js
+++ b/BibLaTeX.js
@@ -866,6 +866,6 @@ function doExport() {
 			}
 		}
 
-		Zotero.write("\n}");
+		Zotero.write(",\n}");
 	}
 }
--- a/BibTeX.js
+++ b/BibTeX.js
@@ -1533,7 +1533,7 @@ function doExport() {
 			}
 		}
 		
-		Zotero.write("\n}");
+		Zotero.write(",\n}");
 	}
 }
 


adam3smith:
Thanks, will take a look asap. Don't worry about failing tests, those fail for other reasons.

adam3smith:
@karnesky could you help me out here? 
The issue this fixes is
```
@book{noauthor_notitle_nodate
}
```
which biber doesn't like but accepts with the comma after the citekey, i.e.
```
@book{noauthor_notitle_nodate,
}
```

However it does add a comma after the last field, i.e. 

```
@article{article,
  author  = {Peter Adams}, 
  title   = {The title of the work},
  journal = {The name of the journal},
  year    = 1993,
  number  = 2,
  pages   = {201-213},
  month   = 7,
  note    = {An optional note}, 
  volume  = 4,
}
```

(instead of 
```
...
  note    = {An optional note}, 
  volume  = 4
}
which I understand to be legal, but is this considered bad style? Any downsides you see? We can probably find a more taylored solution for the issue this fixes, but it'll be a bunch more code.

Zonkil9:
This problem still has not been fixed. 😞

karnesky:
Sorry for not replying here earlier. Day job in pandemic is keeping me swamped.

@adam3smith I'd suggest we accept the PR to add the terminal comma to all entries. While I may find it somewhat unaesthetic personally, it is valid and I think it is unlikely to break other things. While the original _BibTeXing_ does not use that style, the BibLaTeX manual uses these terminal commas.

Emitting empty items may break other tools & is of pretty questionable value. If we find ourselves addressing this edge case again, we may want to remove these empty entries.

Was this issue reported to biber? It'd seem to me that that tool should also be more accepting of questionable input like this, especially if other toolchains tolerate it.

adam3smith:
Thanks everyone!

PR #2139, opened 2020-03-09
Aosic

symac:
A quick fix for this translator which was crashing (see on https://archivesic.ccsd.cnrs.fr/sic_00439849 )


symac:
some diff are not linked to that issue.

PR #2140, opened 2020-03-09
Aosic fix

symac:
This parser was outdated and didn't work because of the way link to pdf are built on the site (see https://archivesic.ccsd.cnrs.fr/sic_00439849 for example) + lint improvements.


adam3smith:
Thanks!

PR #2142, opened 2020-03-09
allows "www" in the beginning of eLibrary.ru url

i3v:
I've noticed that currently elibrary.ru translator does not work because the page url is 
`https://www.elibrary.ru/item.asp?id=30694319`, instead of 
`https://elibrary.ru/item.asp?id=30694319` .
This PR adjusts the `target` accordingly.


dstillman:
Thanks!

i3v:
That was fast :) 
Thanks for merging!

PR #2143, opened 2020-03-13
DOI: Find DOIs in links, and other improvements

dstillman:
Search links in addition to page text, as discussed [in the forums](https://forums.zotero.org/discussion/comment/350735/#Comment_350735).

Also:

- Make the order of results match the page
- Possibly fix an issue where some DOIs could be left out if they
  appeared within the same text string (lastMatch should've been
  lastIndex; the former is read-only and non-numeric, so it's not clear
  that this ever worked)
- Simplify the code

Note: I added a test for this, but for some reason it fails via the test runner due to some site cookie issue, even though it works when run manually (and even works when populating the tests). Apparently our handling of cookies is different between tests and manual translation.

### DIFF
--- a/DOI.js
+++ b/DOI.js
@@ -9,9 +9,8 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2019-01-27 14:06:07"
+	"lastUpdated": "2020-03-13 02:38:54"
 }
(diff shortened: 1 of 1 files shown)


adam3smith:
This looks good to me and works -- the new test runs for me too, btw. 
Do you want to go ahead & merge @dstillman ?

PR #2150, opened 2020-03-23
Updating ThesesFR translator

eonm-abes:
Hi all I updated the ThesesFR.js translator.

__Change log :__

* The translator uses a RDF file to get Metadata. This RDF file is more precise and up to date than embedded metadata in the html page
* The translators is able to import multiples items from organization and person pages
* Supervisors are considered as contributors of the final work
* The mention "Thèses de doctorat" (defended thesis) or "Thèse en préparation" (thesis in preparation) has been added to the type field. This change was requested by theses.fr users.
* More info has been added in the extra field (jury members, laboratory)
* The mention "Les données de Theses.fr sont sous licence Etalab" has been added in the authorization field. (__Can this field be used for licensing purposes ?__)

It's the first time I contribute to update a translator so feedback are welcome. I had some trouble to lint the code with eslint I hope there is no more issues.

A few questions nonetheless :

* I use `const` as often as possible to avoid reassignment. Is this a good practice in the context of translators ? Should I use `let` instead ?
* I use arrow functions. Can it causes compatibility issues ?

Regards

### DIFF
--- a/ThesesFR.js
+++ b/ThesesFR.js
@@ -142,10 +142,12 @@ function scrape(doc, url) {
 		newItem.libraryCatalog = 'theses.fr';
 		newItem.rights = 'Les données de Theses.fr sont sous licence Etalab';
 
-		// Keep extra information such as laboratory, graduate schools, etc.
-		newItem.extra = Array.from(doc.getElementsByClassName('donnees-ombre')).map((description) => {
-			return Array.from(description.getElementsByTagName('p')).map(description => description.textContent.replace(/\n/g, ' '));
+		// Keep extra information such as laboratory, graduate schools, etc. in a note
+		let note = Array.from(doc.getElementsByClassName('donnees-ombre')).map((description) => {
+			return Array.from(description.getElementsByTagName('p')).map(description => description.textContent.replace(/\n/g, ' ').trim());
 		}).join(' ');
+
+		newItem.notes.push({description: note});
 
 		ZU.xpath(xmlDoc, '//dc:subject', ns).forEach((t) => {
 			let tag = t.textContent;
@@ -224,9 +226,12 @@ var testCases = [
 						"tag": "W boson mass"
 			   		}
 				],
-				"extra": "Sous la direction de  Maarten Boonekamp.  Soutenue le 19-09-2016,à l'Université Paris-Saclay (ComUE) , dans le cadre de   École doctorale Particules, Hadrons, Énergie et Noyau : Instrumentation, Imagerie, Cosmos et Simulation (Orsay, Essonne ; 2015-....) , en partenariat avec  Département de physique des particules (Gif-sur-Yvette, Essonne)   (laboratoire)  ,  Centre européen pour la recherche nucléaire   (laboratoire)   et de  Université Paris-Sud (1970-2019)   (établissement opérateur d'inscription)  .",
 				"rights": "Les données de Theses.fr sont sous licence Etalab",
-				"notes": [],
+				"notes": [
+					{
+						"description": "Sous la direction de  Maarten Boonekamp. Soutenue le 19-09-2016,à l'Université Paris-Saclay (ComUE) , dans le cadre de   École doctorale Particules, Hadrons, Énergie et Noyau : Instrumentation, Imagerie, Cosmos et Simulation (Orsay, Essonne ; 2015-....) , en partenariat avec  Département de physique des particules (Gif-sur-Yvette, Essonne)   (laboratoire)  ,  Centre européen pour la recherche nucléaire   (laboratoire)   et de  Université Paris-Sud (1970-2019)   (établissement opérateur d'inscription)  ."
+					}
+				],
 				"seeAlso": []
 			}
 		]
--- a/ThesesFR.js
+++ b/ThesesFR.js
@@ -142,12 +142,23 @@ function scrape(doc, url) {
 		newItem.libraryCatalog = 'theses.fr';
 		newItem.rights = 'Les données de Theses.fr sont sous licence Etalab';
 
-		// Keep extra information such as laboratory, graduate schools, etc. in a note
+		// Keep extra information such as laboratory, graduate schools, etc. in a note for thesis not yet defended
+		let notePrepa = Array.from(doc.getElementsByClassName('donnees-ombreprepa2')).map((description) => {
+			return Array.from(description.getElementsByTagName('p')).map(description => description.textContent.replace(/\n/g, ' ').trim());
+		}).join(' ');
+
+		if (notePrepa) {
+			newItem.notes.push({description: notePrepa});
+		}
+
+		// Keep extra information such as laboratory, graduate schools, etc. in a note for defended thesis
 		let note = Array.from(doc.getElementsByClassName('donnees-ombre')).map((description) => {
 			return Array.from(description.getElementsByTagName('p')).map(description => description.textContent.replace(/\n/g, ' ').trim());
 		}).join(' ');
 
-		newItem.notes.push({description: note});
+		if (note) {
+			newItem.notes.push({description: note});
+		}
 
 		ZU.xpath(xmlDoc, '//dc:subject', ns).forEach((t) => {
 			let tag = t.textContent;
@@ -254,7 +265,11 @@ var testCases = [
 				 		"creatorType": "contributor"
 					}
 				],
-				"notes": [],
+				"notes": [
+					{
+						"description": "Thèses en préparation à Paris 2 , dans le cadre de   Ecole doctorale Georges Vedel Droit public interne, science administrative et science politique (Paris)  depuis le 01-10-2014 ."
+					}
+				],
 				"tags": [],
 				"seeAlso": [],
 				"attachments": [],
--- a/ThesesFR.js
+++ b/ThesesFR.js
@@ -136,7 +136,7 @@ function scrape(doc, url) {
 		const fullDate = ZU.xpathText(xmlDoc, '//dcterms:dateAccepted', ns);
 		const year = ZU.xpathText(xmlDoc, '//dc:date', ns);
 
-		// Some old reccords doesn't have a full date instead we can use the defense year
+		// Some old records doesn't have a full date instead we can use the defense year
 		newItem.date = fullDate ? fullDate : year;
 
 		const permalink = ZU.xpath(xmlDoc, '//bibo:Document', ns);
--- a/ThesesFR.js
+++ b/ThesesFR.js
@@ -48,13 +48,13 @@ function detectWeb(doc, url) {
 }
 
 function getSearchResults(doc, checkOnly) {
-	const items = {};
+	let items = {};
 	let found = false;
-	const rows = ZU.xpath(doc, '//div[contains(@class, "encart arrondi-10")]//h2/a');
+	let rows = ZU.xpath(doc, '//div[contains(@class, "encart arrondi-10")]//h2/a');
 
 	rows.forEach((row) => {
-		const href = row.href;
-		const title = ZU.trimInternal(row.textContent);
+		let href = row.href;
+		let title = ZU.trimInternal(row.textContent);
 		if (checkOnly) return true;
 		found = true;
 		items[href] = title;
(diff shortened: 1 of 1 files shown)


TesteFixe:
Nice Work!

adam3smith:
@dstillman -- do you have an opinion on `const`? See e.g.
https://github.com/zotero/translators/pull/2150/files#diff-ae76dd6444254ff5be2191eeb6d4189cR51
for a respresentative example. 
I don't see a downside, but it's not something we've been doing (let alone enforcing) in translators otherwise.

ThesesFR.js
```diff
@@ -79,24 +85,80 @@ function doWeb(doc, url) {
...
+			dcterms: 'http://purl.org/dc/terms/',
+			foaf: 'http://xmlns.com/foaf/0.1/',
+			marcrel: 'http://www.loc.gov/loc.terms/relators/',
+			rdf: 'http://www.w3.org/1999/02/22-rdf-syntax-ns#'
+		};
+	
+		const title = ZU.xpathText(xmlDoc, '//dc:title', ns);
+		
+		// The case should't exists : preve;nts crashes while importing multiple records
+		if (!title) return;
```
adam3smith, line 114: Let's take this out please or replace it with a `throw` -- if an item unexpectedly doesn't import, we would _want_ an error to be thrown rather than to continue silently, letting the user assume all selected items have been imported.

ThesesFR.js
```diff
@@ -79,24 +85,80 @@ function doWeb(doc, url) {
...
+		});
+
+		// Supervisor(s) must be considered as contributor(s) for french thesis
+		ZU.xpath(xmlDoc, '//marcrel:ths//foaf:Person/foaf:name', ns).forEach((sup) => {
+			const supervisor = ZU.cleanAuthor(sup.textContent, 'contributor', true);
+			newItem.creators.push(supervisor);
+		});
+
+		const abstractNotes = ZU.xpath(xmlDoc, '//dcterms:abstract', ns);
+		newItem.abstractNote = abstractNotes.length > 0 ? abstractNotes[0].textContent : undefined;
```
adam3smith, line 132: Why no just use `ZU.xpathText()` here? (and similarly in other cases below)

ThesesFR.js
```diff
@@ -79,24 +85,80 @@ function doWeb(doc, url) {
...
+		// Some old records doesn't have a full date instead we can use the defense year
+		newItem.date = fullDate ? fullDate : year;
+
+		const permalink = ZU.xpath(xmlDoc, '//bibo:Document', ns);
+		newItem.url = permalink.length > 0 ? permalink[0].getAttribute('rdf:about') : url;
+		newItem.libraryCatalog = 'theses.fr';
+		newItem.rights = 'Les données de Theses.fr sont sous licence Etalab';
+
+		// Keep extra information such as laboratory, graduate schools, etc.
+		newItem.extra = Array.from(doc.getElementsByClassName('donnees-ombre')).map((description) => {
```
adam3smith, line 152: This is better suited for a note than `extra`

ThesesFR.js
```diff
@@ -119,114 +181,93 @@ var testCases = [
...
 				"abstractNote": "Cette thèse décrit une mesure de la masse du boson W avec le détecteur ATLAS. La mesure exploite les données enregistrées par ATLAS en 2011, a une énergie dans le centre de masse de 7 TeV et correspondant à une luminosité intégrée de 4.6 inverse femtobarn. Les mesures sont faites par ajustement aux données de distributions en énergie transverse des leptons charges et en masse transverse du boson W obtenues par simulation, dans les canaux électron et muon, et dans plusieurs catégories cinématiques. Les différentes mesures sont en bon accord et leur combinaison donne une valeur de m_W = 80371.1 ± 18.6 MeV. La valeur mesurée est compatible avec la moyenne mondiale des mesures existantes, m_W = 80385 ± 15 MeV, et l'incertitude obtenue est compétitive avec les mesures les plus précises réalisées par les collaborations CDF et D0.",
 				"libraryCatalog": "theses.fr",
-				"thesisType": "thesis",
-				"university": "Paris Saclay",
+				"thesisType": "These de doctorat",
+				"university": "Université Paris-Saclay (ComUE)",
 				"url": "http://www.theses.fr/2016SACLS590",
-				"attachments": [
-					{
-						"title": "Full Text PDF",
```
adam3smith, line 132: no more PDFs? That seems like it'd be useful

ThesesFR.js
```diff
@@ -36,41 +36,47 @@
 */
 
 function detectWeb(doc, url) {
-	if (url.includes("?q=")) {
-		return "multiple";
+	// Match against a results page or a Ph. D/supervisor/organization page which might contains multiple records e.g.
+	// http://www.theses.fr/fr/?q=zotero
+	// http://www.theses.fr/fr/154750417
```
adam3smith, line 41: cool -- please add a test for this

adam3smith:
I don't quite understand this question:

> The mention "Les données de Theses.fr sont sous licence Etalab" has been added in the authorization field. (Can this field be used for licensing purposes ?) 

are you asking whether the `rights` field is the correct place for licensing information? If so, the answer is yes.

dstillman:
> @dstillman -- do you have an opinion on const? […] I don't see a downside, but it's not something we've been doing (let alone enforcing) in translators otherwise.

We generally don't use `const` anywhere in Zotero beyond actual constants (e.g., all caps variables). It's more verbose without adding much value beyond `let` (particularly in small scopes, where it's just noise). There's no need to reject translators that use `const`, but for brevity and consistency we prefer it not be used.

(We also tend to use `var` at the top level of functions to differentiate variables shared across multiple scopes from block-level `let`, as well as to allow for quicker `try` debugging, but we certainly don't enforce that.)

ThesesFR.js
eonm-abes: @adam3smith I agree with you, It would be very useful. I tried to implement it my self but I encountered multiple issues with PDF attachments :
- Most PDF files are around 2Go or more (Theses.fr doesn't enforce PDF size). It might cause real performance issues when you have to import multiple PDF files at once. In my case my web browser slows down or freezes. Over time PDF files get bigger and bigger on theses.fr
- Finding full text PDF on theses.fr is a hard task. Currently there is a special path in the URL `/document` that redirects user to the full text PDF if it exists. But this URL redirects either to :
   - The PDF file on theses.fr (https://www.theses.fr/2011PA040275/document)
   - The PDF file on another website (University website, Open archive, ...) (https://www.theses.fr/2008ECDL0023/document). In this case we are fetching resources from 3rd party websites without user consent. Furthermore **there is no way to know if the URL is pointing to a third-party website before being redirected**. Currently half of open access PDF are stored on external websites. 
  - A web page containing the PDF file. In this case we have to parse each pages individually to figure out where the PDF is stored. It implies to make multiple HTTP calls to find and download the PDF file. (https://www.theses.fr/2020REN20023/document).
  - A restricted access web page (https://www.theses.fr/2019LIL1R054/document)

It seems like the previous implementation is also broken.

To implement it I need more time and maybe help. I'm also concerned about user consent.

adam3smith: It's certainly fine to leave this out, I was just checking.

FWIW, I'm not super concerned about user consent to retrieve PDFs from 3rd party academic websites -- Zotero's unpaywall integration does this, too. The fact that we'd only be able to successfully import PDFs from 1 and 2 of your four scenarios with reasonable effort is also fine: that's why we check the mimeType.

However, we don't have a way to check for file size limits in translators and I certainly don't want to auto-import 2GB document, so if this is a regular occurrence, that'd indeed be a problem.

ThesesFR.js
eonm-abes: My xpath skills was a bit rusty. I failed to get the first element of the abstract (and other nodes) with pure xpath.

Originally I tried to use this expression `//dcterms:abstract[1]` with `xpathText()` function, but this expression was wrong in my context. It should have been `(//dcterms:abstract)[1]`.

So I used a workaround by mixing xpath and javascript which is not the best implementation. Now that I have found the right xpath expression I'll fix the code to use `xpathText()` function.

ThesesFR.js
eonm-abes: In this case I'll use a note instead

eonm-abes:
Thank you @adam3smith  and @dstillman for your comments and feedback. I updated the code accordingly to your advice.

@adam3smith my question was indeed about the right place to put licensing information. I updated the code to keep only the license name (Etalab).

@dstillman As suggested I replaced `const` with `let`.

It should be good now.

ThesesFR.js
dstillman: (Note that you can also use `text(docOrElem, selector[, index])` or `attr(docOrElem, selector, attribute[, index])` with CSS selectors instead of using XPaths.)

eonm-abes: I already fixed the code by using xpath (2c91144) and `Z.xpathText()` but that's a good trick to know.

ThesesFR.js
```diff
@@ -230,16 +247,19 @@ var testCases = [
 						"tag": "W boson mass"
 			   		}
 				],
-				"extra": "Sous la direction de  Maarten Boonekamp.  Soutenue le 19-09-2016,à l'Université Paris-Saclay (ComUE) , dans le cadre de   École doctorale Particules, Hadrons, Énergie et Noyau : Instrumentation, Imagerie, Cosmos et Simulation (Orsay, Essonne ; 2015-....) , en partenariat avec  Département de physique des particules (Gif-sur-Yvette, Essonne)   (laboratoire)  ,  Centre européen pour la recherche nucléaire   (laboratoire)   et de  Université Paris-Sud (1970-2019)   (établissement opérateur d'inscription)  .",
 				"rights": "Les données de Theses.fr sont sous licence Etalab",
```
adam3smith, line 250: doesn't look like you updated the tests after changing rights

ThesesFR.js
```diff
@@ -108,53 +107,61 @@ function scrape(doc, url) {
...
+		// Keep extra information such as laboratory, graduate schools, etc. in a note for thesis not yet defended
+		let notePrepa = Array.from(doc.getElementsByClassName('donnees-ombreprepa2')).map((description) => {
+			return Array.from(description.getElementsByTagName('p')).map(description => description.textContent.replace(/\n/g, ' ').trim());
+		}).join(' ');
 
-		// Keep extra information such as laboratory, graduate schools, etc.
-		newItem.extra = Array.from(doc.getElementsByClassName('donnees-ombre')).map((description) => {
-			return Array.from(description.getElementsByTagName('p')).map(description => description.textContent.replace(/\n/g, ' '));
+		if (notePrepa) {
+			newItem.notes.push({ description: notePrepa });
```
adam3smith, line 151: This should be `{note: notePrepa}` as per https://www.zotero.org/support/dev/translators/coding#notes 
Same below

ThesesFR.js
```diff
@@ -108,53 +107,61 @@ function scrape(doc, url) {
...
+		}
+
+		// Keep extra information such as laboratory, graduate schools, etc. in a note for defended thesis
+		let note = Array.from(doc.getElementsByClassName('donnees-ombre')).map((description) => {
+			return Array.from(description.getElementsByTagName('p')).map(description => description.textContent.replace(/\n/g, ' ').trim());
 		}).join(' ');
 
+		if (note) {
+			newItem.notes.push({ description: note });
```
adam3smith, line 160: ditto

adam3smith:
Hi -- sorry, I know I said we'd be done, but I'm just realizing that the combination of target and detect function is significantly too loose: The translator detects any page as a thesis, including search start pages (http://theses.fr/fr/personnes/) and documentation (http://theses.fr/fr/personnes/). This isn't your bug, it's already the case in the current version, but it's... not good.

I don't think you can easily address this in the target, so will probably require something on the page itself -- one of the metatags might make most sense, e.g. DC.type?

Would you mind tightening this up a bit more?

eonm-abes:
No problem. This is my fault, I forget to check the target regex when I updated the translator.

I come up with this regex `^https?:\/\/(www\.)?theses\.fr\/([a-z]{2}\/)?((s\d+|\d{4}.{8}|\d{8}X|\d{9})(?!\.(rdf|xml)$)|(sujets\/\?q=|\?q=))(?!.*&format=(json|xml))`. What do you think of it ?

This regex should exclude search start pages, .xml and .rdf pages and API pages.

More explanation and example here : https://regex101.com/r/IIr1ak/1

adam3smith:
Thanks so much. This looks great!

PR #2153, opened 2020-04-01
Translator for clinicaltrials.gov

### LINKED ISSUE #1952: Add clinicaltrials.gov
Multiple requests over the year, including here: https://forums.zotero.org/discussion/77721/add-reference-from-clinical-trials-org#latest

https://clinicaltrials.gov/ct2/resources/download#DisplaySingle 
has API-type information for getting to the XML

rdvelazquez:
TODO:
- [x] Decide the itemType to use. ~~Currently using `journalArticle` but there have been discussions of using `dataset` or creating a new type~~ using `report` as recommended by @bwwiernik
- [x] Set up a local development environment so I can test this out using the Zotero modules and tests ~~Got it set up and everything works except for an issue with matching the "extra" field~~ **All three tests are passing now**
- [x] ~~Implement the search (if we think its a feature this should have)~~ **I don't see the need for being able to cite all the trials from a particular search of clinicaltrials.gov at this point (could always be added in later if needed**, ~~item parsing and other TODOs~~
- [x] Determine the values for `translatorType` and `browserSupport` in the metadata **I think these are correct; just following these [docs](https://www.zotero.org/support/dev/translators)**
- [x] Fix the lint errors and warnings

closes #1952 
relates to https://github.com/manubot/manubot/issues/216

### DIFF
--- a/clinicaltrials.gov.js
+++ b/clinicaltrials.gov.js
@@ -86,6 +86,20 @@ function nameToFirstAndLast(rawName) {
 	return [firstName, lastName];
 }
 
+function extrasObjToExtrasString(extrasObj) {
+	let extrasString = "";
+	for (let key in extrasObj) {
+		if (key == "collaborators") {
+			const stringifiedArray = "'" + extrasObj[key].join("','") + "'";
+			extrasString = extrasString + "\n" + key + ": " + stringifiedArray;
+		}
+		else {
+			extrasString = extrasString + "\n" + key + ": " + extrasObj[key];
+		}
+	}
+	return extrasString;
+}
+
 function scrape(doc, url) {
 	const clinicalTrialID = getClinicalTrialID(url);
 	let jsonRequestURL;
@@ -163,8 +177,8 @@ function scrape(doc, url) {
 				extras.collaborators.push(collaborator.firstName);
 			});
 		}
-
-		item.extra = extras;
+		const extrasString = extrasObjToExtrasString(extras);
+		item.extra = extrasString;
 		item.complete();
 	});
 }
@@ -185,16 +199,16 @@ var testCases = [
 						"creatorType": "author"
 					}
 				],
-				"date": "February 28, 2020",
+				"date": "March 27, 2020",
 				"accessDate": "2020-04-01",
 				"libraryCatalog": "clinicaltrials.gov",
 				"shortTitle": "Study to Evaluate the Safety and Antiviral Activity of Remdesivir (GS-5734™) in Participants With Severe Coronavirus Disease (COVID-19)",
 				"url": "https://clinicaltrials.gov/ct2/show/NCT04292899",
-				"extra": {
-					"submittedDate": "February 6, 2020",
-					"responsiblePartyInvestigator": "undefined",
-					"sponsor": "Gilead Sciences"
-				}
+				"extra": "submittedDate: February 28, 2020\nresponsiblePartyInvestigator: undefined\nsponsor: Gilead Sciences",
+				"notes": [],
+          		"tags": [],
+        		"seeAlso": [],
+         		"attachments": []
 			}
 		]
 	},
@@ -220,12 +234,11 @@ var testCases = [
 				"libraryCatalog": "clinicaltrials.gov",
 				"shortTitle": "Sleep Disorders and Gastroesophageal Reflux Disease (GERD)",
 				"url": "https://clinicaltrials.gov/ct2/show/NCT00287391",
-				"extra": {
-					"submittedDate": "February 3, 2006",
-					"responsiblePartyInvestigator": "undefined",
-					"sponsor": "University of North Carolina",
-					"collaborators": ["Janssen Pharmaceutica N.V., Belgium"]
-				}
+				"extra": "submittedDate: February 3, 2006\nresponsiblePartyInvestigator: undefined\nsponsor: University of North Carolina\ncollaborators: 'Janssen Pharmaceutica N.V., Belgium'",
+				"notes": [],
+          		"tags": [],
+        		"seeAlso": [],
+         		"attachments": []
 			}
 		]
 	},
@@ -250,13 +263,12 @@ var testCases = [
 				"date": "March 22, 2020",
 				"accessDate": "2020-04-01",
 				"libraryCatalog": "clinicaltrials.gov",
-				"shortTitle": "Efficacy and Safety of Hydroxychloroquine for Treatment of Pneumonia Caused by 2019-nCoV ( HC-nCoV )",
 				"url": "https://clinicaltrials.gov/ct2/show/NCT04261517",
-				"extra": {
-					"submittedDate": "February 6, 2020",
-					"responsiblePartyInvestigator": "Hongzhou Lu",
-					"sponsor": "Shanghai Public Health Clinical Center"
-				}
+				"extra": "submittedDate: February 6, 2020\nresponsiblePartyInvestigator: Hongzhou Lu\nsponsor: Shanghai Public Health Clinical Center",
+				"notes": [],
+          		"tags": [],
+        		"seeAlso": [],
+         		"attachments": []
 			}
 		]
 	}
--- a/clinicaltrials.gov.js
+++ b/clinicaltrials.gov.js
@@ -36,11 +36,10 @@
 */
 
 function detectWeb(doc, url) {
-	// TODO: confirm that there all searches will have ct2/results in the url
 	if (url.includes('https://clinicaltrials.gov/ct2/results')){
 		return "multiple"
 	} else {
-		return "journalArticle"
+		return "report"
 	}
 }
 
@@ -79,10 +78,10 @@ function isXmlAPIRequest(url) {
 
 function getClinicalTrialID(url) {
 	// TODO: make sure this handles all the potential URLs
-	if (isXmlAPIRequest(url)){
+	if (isXmlAPIRequest(url) || isJsonAPIRequest(url)){
 		return url.split("expr=")[1].split("&")[0]
 	} else {
-		return url.split('/show/')[1]
+		return url.split('/show/')[1].split("?")[0]
 	}
 }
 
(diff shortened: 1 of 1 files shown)


rdvelazquez:
@bwiernik and @adam3smith any input on what item type to use for clinical trials?

bwiernik:
I would use Report for this.

rdvelazquez:
@dhimmel and @agitter two quick questions that I thought you may have answers to:

1. **Date**: Clinical trials seem to have lots of dates; I'm currently using the "LastUpdateSubmitDate". Is this the date we would want in the citation or would we want "StudyFirstSubmitDate" or some other date? Here's an example of the date info available:
```
"StatusVerifiedDate":"March 2020",
              "OverallStatus":"Completed",
              "ExpandedAccessInfo":{
                "HasExpandedAccess":"No"
              },
              "StartDateStruct":{
                "StartDate":"February 6, 2020",
                "StartDateType":"Actual"
              },
              "PrimaryCompletionDateStruct":{
                "PrimaryCompletionDate":"February 25, 2020",
                "PrimaryCompletionDateType":"Actual"
              },
              "CompletionDateStruct":{
                "CompletionDate":"February 25, 2020",
                "CompletionDateType":"Actual"
              },
              "StudyFirstSubmitDate":"February 6, 2020",
              "StudyFirstSubmitQCDate":"February 6, 2020",
              "StudyFirstPostDateStruct":{
                "StudyFirstPostDate":"February 7, 2020",
                "StudyFirstPostDateType":"Actual"
              },
              "LastUpdateSubmitDate":"March 22, 2020",
              "LastUpdatePostDateStruct":{
                "LastUpdatePostDate":"March 24, 2020",
                "LastUpdatePostDateType":"Actual"
              }
```

2. **Author**: I'm using the "ResponsiblePartyInvestigatorFullName" if one exists and the "LeadSponsorName" if it doesn't. Is this standard / ok? It seems like the "LeadSponsorName" will sometimes be a company.

I looked https://www.who.int/ictrp/How_to_cite.pdf and https://blogs.uoregon.edu/annie/2017/10/25/clinical-trial-apa-format/ but they didn't seem to be conclusive.

bwiernik:
For things like preprints, Zotero translators typically save the last updated date (i.e., the date of the version of the item actually being viewed) as the date. The first submit date could be stored in Extra with the label "Original date:"

adam3smith:
Agree on the regular date, but I'd be careful with using `original date` too widely. Its most common use is for historical publication dates or reprinted works, which are often rendered in citation styles. I don't really see that that'd be true for clinical trials (or preprints, for that matter).

bwiernik:
That's a really good point. "Submitted" might be a better (and rarely used in citation styles) variable.

dhimmel:
> I'm using the "ResponsiblePartyInvestigatorFullName" if one exists and the "LeadSponsorName" if it doesn't. Is this standard / ok? It seems like the "LeadSponsorName" will sometimes be a company.

Just looking at a random record [`NCT04291053`](https://clinicaltrials.gov/api/query/full_studies?expr=NCT04291053&fmt=JSON):

```json
            "SponsorCollaboratorsModule":{
              "ResponsibleParty":{
                "ResponsiblePartyType":"Principal Investigator",
                "ResponsiblePartyInvestigatorFullName":"Chen Xiaoping",
                "ResponsiblePartyInvestigatorTitle":"Principal Investigator",
                "ResponsiblePartyInvestigatorAffiliation":"Tongji Hospital"
              },
              "LeadSponsor":{
                "LeadSponsorName":"Tongji Hospital",
                "LeadSponsorClass":"OTHER"
              }
            },
```

<details>
<summary>Some documentation</summary>

From https://prsinfo.clinicaltrials.gov/definitions.html:

> **3\. Sponsor/Collaborators**
> 
> **Responsible Party, by Official Title** *\
> Definition: An indication of whether the responsible party is the sponsor, the sponsor-investigator, or a principal investigator designated by the sponsor to be the responsible party. Select one.
> 
> -   Sponsor: The entity (for example, corporation or agency) that initiates the study
> -   Principal Investigator: The individual designated as responsible party by the sponsor (see Note)
> -   Sponsor-Investigator: The individual who both initiates and conducts the study\
> Note: The sponsor may designate a principal investigator as the responsible party if such principal investigator meets all of the following requirements: is responsible for conducting the study; has access to and control over the data from the study; has the right to publish the results of the study; and has the ability to meet all of the requirements for submitting and updating clinical study information.
> 
> **Investigator Information** [*]\
> If the Responsible Party, by Official Title is either "Principal Investigator" or "Sponsor-Investigator," the following is required:
> 
> -   **Investigator Name**: Name of the investigator, including first and last name
> -   **Investigator Official Title**: The official title of the investigator at the primary organizational affiliation\
>     Limit: 254 characters.
> -   **Investigator Affiliation**: Primary organizational affiliation of the individual;\
>     Limit: 160 characters.
> 
> **Name of the Sponsor** *\
> Definition: The name of the entity or the individual who is the sponsor of the clinical study.\
> Limit: 160 characters.
> 
> Note: When a clinical study is conducted under an investigational new drug application (IND) or investigational device exemption (IDE), the IND or IDE holder is considered the sponsor. When a clinical study is not conducted under an IND or IDE, the single person or entity who initiates the study, by preparing and/or planning the study, and who has authority and control over the study, is considered the sponsor.
> 
> **Collaborators**\
> Definition: Other organizations (if any) providing support. Support may include funding, design, implementation, data analysis or reporting. The responsible party is responsible for confirming all collaborators before listing them.\
> Limit: 160 characters.

</details>

I think perhaps we want everything: the lead investigator, the sponsor, and collaborators. Each one of these could be different authors. I don't know too much about clinical trials however, so would be interested in what others think

rdvelazquez:
Thank you all for the quick responses. Much appreciated! 

My last commit attempts to incorporate that feedback.

- I'm including the initial submission date in `extra.submittedDate`
- I'm including the lead investigator, sponsor, and collaborators as creators and, as a way to be able to tell who is who, I'm also including this info in the `extra` section explicitly stating what type of creator they were. 
_(there aren't many creatorTypes for [report](https://aurimasv.github.io/z2csl/typeMap.xml#map-report) so I'm just listing everyone as an author but including this info in the extra's will let downstream analyses be able to untangle who was a collaborator vs. sponsor, etc.)_

rdvelazquez:
When I run the tests locally it says that the `extra` field doesn't match. Here's the testing output:
```
         -   "extra": {
         -     "submittedDate": "February 6, 2020"
         -     "responsiblePartyInvestigator": "undefined"
         -     "sponsor": "Gilead Sciences"
         -   }
         +   "extra": "[object Object]"
```

i.e. it's getting `"[object Object]"`

@adam3smith or someone else experienced with Zotero, any input on this?

adam3smith:
Haven't looked at your code, but `Extra` needs to be a string with the different values newline separated. It looks like you have it as an array?

rdvelazquez:
Thanks @adam3smith! That worked. All three test are now passing. Ready for review.

bwiernik:
A few immediate comments:

1. I would enter `submittedDate` just as `submitted`. That is the actual CSL variable. This will make it accessible for citations if needed.
2. I would suggest not saving the same information in multiple places; that has a decent chance of producing unexpected results in citations. I'll defer to @adam3smith as to where Principal Investigators, other collaborators, and the lead sponsor should go. My intuition is that the Principal Investigtor should be stored as Author, any collaborators stored as Contributor, and the Sponsor stored in Extra (and only in Extra) labeled `"Sponsor:"`.
3. Use proper narrative capitalization and spacing when storing data in Extra, rather than camelCase (e.g., `Principal investigator:` rather than `nresponsiblePartyInvestigator:`).
4. Undefined or missing values should be dropped rather than being stored as `undefined`
5. "ClinicalTrials.gov" should be stored in the `institution` (publisher) field, and "Clinical trial registration" in the `reportType` field. 
6. The registration number (I think NCTId) should be stored in `reportNumber`
7. The `BriefSummary` part of the `DescriptionModule` should be stored in `abstract`

rdvelazquez:
Thank you for the review @bwiernik! I've addressed everything except point 2.

I think your recommendation makes sense just one point of clarification: We would store the sponsor in extra as "Sponsor: " no matter what, even if there was no principle investigator and the sponsor was also listed as the author? 

Pro: You could tell if the sponsor was listed as the author. (if clinical trials always have a sponsor than this is negated as you could just check to see if "Sponsor: " in extra was missing and if so assume know that the "author" was actually the "Sponsor")
Con: You would be storing the same info in two places

bwiernik:
It would be fairly clear that the sponsor was listed as an author by virtue of it being an organizational (single-field) author. Another option rather than placing in Extra at all would be to include it as a "Contributor" or "Author" depending on whether there was a personal author (PI).

rdvelazquez:
@bwiernik That makes sense to me. I implemented that and we can see if anyone has other opinions on how it should be handled. Thanks again for looking at this!

dhimmel:
Just wanted to check in here. @rdvelazquez are you good on your end and just waiting for review?

rdvelazquez:
Yep. Things are good on my end and just waiting for review.

dhimmel:
@bwiernik / @adam3smith any chance either you have time to re-review? Would be greatly appreciated.

rdvelazquez:
Just checking on the status of getting this reviewed and incorporated into Zotero. This would be helpful for the https://github.com/greenelab/covid19-review project.

clinicaltrials.gov.js
```diff
@@ -0,0 +1,273 @@
...
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+function detectWeb(doc, url) {
+	if (url.includes("https://clinicaltrials.gov/ct2/results")) {
```
adam3smith, line 39: can we not make multiples work or have you just not gotten to it?
We strongly prefer having multiples available -- especially for things like systematic reviews this would seems super useful.
In any case, we should then just exclude them via target regex - and make the target regex as well as `detectWeb()` more restrictive overall: in its current form the translator would detect e.g. the homepage and help pages as reports and then fail on import. This should never happen.

clinicaltrials.gov.js
```diff
@@ -0,0 +1,273 @@
...
+	else {
+		return url.split("/show/")[1].split("?")[0];
+	}
+}
+
+function dateTimeToDateString(dateTime) {
+	return dateTime.split(" ")[0].split(":").join("-");
+}
+
+function nameToFirstAndLast(rawName) {
```
adam3smith, line 82: Could you add a comment on what exactly this does, ideally with a data example. If I'm reading this correctly, this would also remove authors initials/second names which doesn't seem ideal. Note that Zotero has a built-in `ZU.cleanAuthor` function that might be helpful here depending on the format.

clinicaltrials.gov.js
```diff
@@ -0,0 +1,273 @@
...
+						"creatorType": "author"
+					}
+				],
+				"date": "April 1, 2020",
+				"accessDate": "2020-04-01",
+				"libraryCatalog": "clinicaltrials.gov",
+				"shortTitle": "Study to Evaluate the Safety and Antiviral Activity of Remdesivir (GS-5734™) in Participants With Severe Coronavirus Disease (COVID-19)",
+				"url": "https://clinicaltrials.gov/ct2/show/NCT04292899",
+				"institution": "clinicaltrials.gov",
+         		"reportNumber": "NCT04292899",
```
adam3smith, line 193: are the test generated with scaffold? It's odd that the indenting is off for report details. Let's fix this

clinicaltrials.gov.js
rdvelazquez: I made the target more restrictive. I wasn't sure if any additional checks should go in `detectWeb()`; it seemed like the target was the preferred place to make it more restrictive.

I thought the single case was more important than the multiples and wanted to focus my limited time on getting that done first. I know we need the single use case for the active https://github.com/greenelab/covid19-review and wasn't aware of a use case for the multiples (I assume this would be getting all the citations for what the clinicalTrials website returns from a search?). I'd be willing to help get that incorporated but I'm not sure how much time I'll have in the next few weeks.

clinicaltrials.gov.js
rdvelazquez: I felt like I was likely reinventing the wheel when I wrote that function but couldn't find a good replacement. Thanks for pointing out `ZU.cleanAuthor` that was just what I needed!

clinicaltrials.gov.js
rdvelazquez: I fixed the indenting issue. I generated the tests myself without using scaffold; I think I tried using it but had some issues getting it running if I remember correctly.

dstillman: Just to clarify, Scaffold is now just Tools → Developer → Translator Editor in Zotero, not a separate plugin. The test needs to be generated by Scaffold to be valid. (If you generated it by hand, how did you test it?)

If you're having trouble with something in Scaffold, let us know.

rdvelazquez: Thanks for clarifying that @dstillman. That's how I tested it (and now that I'm looking at it again I can see that it says "Scaffold" at the top of the window... ). That was very helpful for developing the translator and how I tested it (I also used a simple node.js script to test the parsing while developing it so that I could get more instantaneous feedback)

adam3smith:
(sorry for the linting problems -- linting is a mess on Windows, no idea how people do this. Will fix asap or feel free to fix on your end)

rdvelazquez:
@adam3smith Thanks so much for the edits! I fixed the listing issues by running `yarn lint --fix clinicaltrials.js` and it looks like it took care of it. JS development on windows is tough (I've been there).

adam3smith:
Terrific, thanks -- this looks great!

PR #2155, opened 2020-04-06
Update Legifrance.js

LucDev69:
Fix URL fill for statute. Delete jsessionid and DateText from URL to keep only relevant informations.
Tested with Scaffold before submitting.

### DIFF
--- a/Legifrance.js
+++ b/Legifrance.js
@@ -57,6 +57,7 @@ function scrapecase(doc) { //Jurisprudence
 	var newItem = new Zotero.Item("case");
 	
 	// Paramètres communs
+	var numero, date, formation, cour, publication, nature;
 	
 	var title = ZU.xpathText(doc, '//h2[@class="title"]');
 	newItem.title = title;
@@ -75,10 +76,10 @@ function scrapecase(doc) { //Jurisprudence
 	var a;// Conseil constitutionnel
 	a = title.match(/(.*) - (.*) - (.*) - (.*)/);
 	if (a) {
-		var numero = a[1];
-		var date = a[2];
-		var texteparties = a[3]
-		var formation = a[4];
+		numero = a[1];
+		date = a[2];
+		var texteparties = a[3];
+		formation = a[4];
 		newItem.court = 'Conseil constitutionnel';
 		newItem.docketNumber = numero;
 		newItem.date = date;
@@ -88,11 +89,11 @@ function scrapecase(doc) { //Jurisprudence
 	var b;// Conseil d'État avec indication de publication
 	b = title.match(/(Conseil d'État), (.*), (s*[0-9/]+), (s*[0-9]+), (.*Lebon)/);
 	if (b) {
-		var cour = b[1];
-		var formation = b[2];
-		var date = b[3];
-		var numero = b[4];
-		var publication = b[5];
+		cour = b[1];
+		formation = b[2];
+		date = b[3];
+		numero = b[4];
+		publication = b[5];
 		newItem.court = 'Conseil d\'État';
 		newItem.extra = formation;
 		newItem.date = date;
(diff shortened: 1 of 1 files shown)


LucDev69:
Have to fix old code to pass checks. Lots of spaces, double declarations and blank lines.

Legifrance.js
```diff
@@ -40,252 +40,261 @@ Thanks to Sebastian Karcher and Aurimas Vinckevicius
...
+	} else return false;
+}
 
-		}
-		newItem.complete();
+function scrapecase(doc) { //Jurisprudence
+	var newItem = new Zotero.Item("case");
+	
+	// Paramètres communs
+	var numero, date, formation, cour, publication, nature;
```
adam3smith, line 60: as per Dan's comment here: https://github.com/zotero/translators/pull/2173#discussion_r422698309 
use `let` in the various if statements instead.

Legifrance.js
```diff
@@ -40,252 +40,261 @@ Thanks to Sebastian Karcher and Aurimas Vinckevicius
...
+			url: "http://www.legifrance.gouv.fr/" + rtfurl,
+			title: "Document en RTF",
+			mimeType: "application/rtf"
+		}];
+	}
+	
+	// Situation selon les juridictions
+	
+	var a;// Conseil constitutionnel
+	a = title.match(/(.*) - (.*) - (.*) - (.*)/);
```
adam3smith, line 77: no need to separately declare. Just do `const a = title.match()` and similarly below

Legifrance.js
```diff
@@ -40,252 +40,261 @@ Thanks to Sebastian Karcher and Aurimas Vinckevicius
...
+function scrapelegislation(doc) { //Législation
+	var code, date, UnParam;
+	var newItem = new Zotero.Item("statute");
+	
+	var title = ZU.xpathText(doc, '//h2[@class="title"]');
+	newItem.title = title;
+	var MonUrl = doc.location.href;
+	var UrlParam = MonUrl.substring(MonUrl.indexOf("?")+1).split("&");
+	
+	if (MonUrl.indexOf("jsessionid") != -1) {
```
adam3smith, line 222: Use `includes()` to test for literal strings (and similarly below)

Legifrance.js
```diff
@@ -40,252 +40,261 @@ Thanks to Sebastian Karcher and Aurimas Vinckevicius
...
+		MonUrl = MonUrl.substring(0,MonUrl.indexOf("jsessionid")-1)+"?";
+	} else {
+		MonUrl = MonUrl.substring(0,MonUrl.indexOf("?")+1);
+	}
+	for (UnParam in UrlParam) {
```
adam3smith, line 227: Don't use for ... in to iterate through arrays (for ... of is fine)

Legifrance.js
```diff
@@ -40,252 +40,261 @@ Thanks to Sebastian Karcher and Aurimas Vinckevicius
...
-			newItem.nameOfAct = nameOfAct + ' de la Commission Nationale de l\'Informatique et des Libertés';
-			newItem.code = code;
-			newItem.date = date;
+function scrapelegislation(doc) { //Législation
+	var code, date, UnParam;
+	var newItem = new Zotero.Item("statute");
+	
+	var title = ZU.xpathText(doc, '//h2[@class="title"]');
+	newItem.title = title;
+	var MonUrl = doc.location.href;
```
adam3smith, line 219: One more question:
I may be reading this too quickly, but couldn't we reduce the entire thing below to:
`MonUrl = MonUrl.replace("jsessionid", "?").replace(/(dateTexte[^&]+|categorieLien[^&]+)\&/g, "");`
(might need some tweaking to deal with end of URL)

AbeJellinek:
@LucDev69, any progress on this?

PR #2158, opened 2020-04-12
Remove GMail print script

mmarvick:
The GMail print preview now has a `<script>` element that's a child of the body, used to show the print dialog. This updates the translator accordingly.

### DIFF
--- a/Gmail.js
+++ b/Gmail.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsb",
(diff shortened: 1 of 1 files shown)


dstillman:
Thanks for working on this, and sorry for the delay in reviewing it.

Did you test this in the Zotero Connector? This won't actually work. The Connector doesn't pass the HTML from the page to Zotero to create a snapshot — it passes the URL and cookies, and Zotero redownloads it.

We're hoping to make some changes to how snapshots work that will automatically strip JavaScript from the page when saving, and that will address this. Until then, there's nothing to be done here.

Issue #2164, opened 2020-04-19
AbstractNote import issue with RDF.js and Embedded Metadata.js


vancleve:
Running the `Embedded Metadata.js` translator on 
https://www.nature.com/articles/ncomms14906
returns
```
 "abstractNote": "Advanced LIGO has detected gravitational waves from two binary black hole mergers, plus a merger candidate. Here the authors use the COMPAS code to show that all three events can be explained by a single evolutionary channel via a common envelope phase, and characterize the progenitor metallicity and masses."
```
which is not the Abstract, but rather the content of `og:description`. 

From here,
https://github.com/zotero/translators/blob/5a8e310028dad8b0fd3bf26bc856cf2049d0be6f/RDF.js#L1199
it looks like the Facebook OpenGraph (`og:description`) description element takes precedence over `description`:
https://github.com/zotero/translators/blob/5a8e310028dad8b0fd3bf26bc856cf2049d0be6f/RDF.js#L1301

From the debug in `Embedded Metadata.js`, it looks like the RDF is being constructed with the correct abstract as:
```
http://purl.org/dc/elements/1.1/description=>During its first four months of taking data, Advanced LIGO has detected gravitational waves from two binary black hole mergers, GW150914 and GW151226, along with the statistically less significant binary black hole merger candidate LVT151012. Here we use the rapid binary population synthesis code COMPAS to show that all three events can be explained by a single evolutionary channel—classical isolated binary evolution via mass transfer including a common envelope phase. We show all three events could have formed in low-metallicity environments (Z=0.001) from progenitor binaries with typical total masses ≳160M⊙, ≳60M⊙ and ≳90M⊙, for GW150914, GW151226 and LVT151012, respectively. Advanced LIGO has detected gravitational waves from two binary black hole mergers, plus a merger candidate. Here the authors use the COMPAS code to show that all three events can be explained by a single evolutionary channel via a common envelope phase, and characterize the progenitor metallicity and masses.
```

So, its this a "bug" that should be fixed in `RDF.js` or `Embedded Metadata.js` or in `Nature Publishing Group.js`?

adam3smith:
I'm not sure this is a "bug" in the common sense of the word. As you note, the RDF translator is explicitly written to prefer `og:description` over `dc:description`  and that's what's happening. 

Having looked at this briefly, it looks to me that the reason for this is terribly convoluted, related to the use of dc.description in the Zotero RDF format (for the extra field).

[...]

PR #2165, opened 2020-04-19
take abstract from RIS and pages from RIS for Nat. Comm.

vancleve:
The NPG translator seems to get the wrong abstract (see https://github.com/zotero/translators/issues/2164) and Nature Communications needs the article number in the `pages` field, not the PDF pages (see https://forums.zotero.org/discussion/82431/nature-communications-article-number).

### DIFF
--- a/Nature Publishing Group.js
+++ b/Nature Publishing Group.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2020-04-19 20:52:54"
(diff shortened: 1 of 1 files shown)


vancleve:
@adam3smith, could you look at issue #2164 too? taking the abstract from the RIS in this commit works, but the problem could also be fixed by changing some of the upstream translators. I just don't know if what they're doing is intentional, in which case the fix probably should be to take the abstract from RIS here, or if its not then maybe its causing problems in other translators too?

adam3smith:
Thanks!

PR #2166, opened 2020-04-19
Russian State Library (rsl.ru) translator

pchemguy:
RSL.ru has two primary catalog interfaces:
https://search.rsl.ru (sRSL)
http://aleph.rsl.ru (aRSL)
This translator handles search results and individual records for search.rsl.ru interface, as well as favorites.
aleph.rsl.ru interface is a total mess, only basic support for a single record is implemented (parsing of MARC fields and processing the result with a MARC import translator).

### DIFF
--- a/Russian State Library.js
+++ b/Russian State Library.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2020-04-14 11:03:21"
+	"lastUpdated": "2020-04-14 16:18:29"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


Russian State Library.js
```diff
@@ -0,0 +1,1558 @@
...
+ *  Windows 7 x64, April 2020.
+ *
+ *	@param {Object} item - Zotero item
+ *	@param {String} title - Link name
+ *	@param {String} url - Link url
+ *
+ *	@return {None}
+ */
+function addLink(item, title, url) {
+	item.attachments.push({ linkMode: "linked_url", // Apparently, should be the first
```
adam3smith, line 181: I'm a bit confused by this -- `linkMode` should not be set in web translators nor should it be necessary. How are you seeing the issue you mention above manifest? We have many translators that just save links using `snapshot: false` and they save/display the title properly.

pchemguy: Basically, without `linkMode` , the specified link name was ignored, as if it was not specified at all. This behavior was not reliably reproducible though. When I tried to run tests later without `linkMode`, the specified link name was set as expected. However, I also observed the same issue when I tried to create link attachments via JS API in the built-in JS console.

AbeJellinek:
This looks very good, thanks. I removed the `linkMode` stuff; I can't reproduce the issue and, if the bug remains, it should be fixed upstream, not worked around in each individual translator.

Issue #2168, opened 2020-05-03
DBLP translator fails to identify titles of articles in multiple mood


zl8716y:
![translator](https://user-images.githubusercontent.com/64294025/80922296-bd819c00-8d41-11ea-893c-6074ad9e899a.png)

Test page: https://dblp.dagstuhl.de/pers/hd/k/Knuth:Donald_E=.html.
This translator fails to show the titles. But it succeeds to add selected items into Zotero clients.

PR #2169, opened 2020-05-03
Fix DBLP translator where the multiple titles are not recognized.

zl8716y:
Fix https://github.com/zotero/translators/issues/2168 where the DBLP translator fails to identify multiple titles.


adam3smith:
Thanks! For future reference, you need to update the last updated request on all substantive translator changes -- they don't ship out to users otherwise (also, ideally, re-run the tests in Scaffold/Translator Editor)

PR #2172, opened 2020-05-09
Update London Review of Books.js

kjtlslt:
Needs to work with URLs omitting 'www'.


dstillman:
You need to update the timestamp as well or this won't get pushed out to clients.

(If you edit via the Translator Editor (Tools → Developer), this will get updated automatically.)

dstillman:
I went ahead and updated the timestamp.

Thanks!

kjtlslt:
Thanks!

PR #2173, opened 2020-05-10
Fix author names for new(ish) layout

adam3smith:
https://forums.zotero.org/discussion/82934/connector-on-amazon-not-saving-authors-names#latest

(no tests from that thread since issue also affects existing tests)

### DIFF
--- a/Amazon.js
+++ b/Amazon.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsbv",
-	"lastUpdated": "2019-08-11 10:33:35"
+	"lastUpdated": "2020-05-10 20:30:25"
 }
 
 // attr()/text() v2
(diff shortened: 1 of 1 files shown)


dstillman:
Thanks!

Amazon.js
```diff
@@ -215,9 +216,10 @@ function scrape(doc, url) {
 	) {
 		baseNode = baseNode.parentElement;
 	}
-	
+
+	var authors, name, role, invertName;
 	if (baseNode) {
-		var authors = ZU.xpath(baseNode, './/span[@id="artistBlurb"]/a');
+		authors = ZU.xpath(baseNode, './/span[@id="artistBlurb"]/a');
```
dstillman, line 222: Just to note, unless a variable is used outside a block, we prefer to use `let` inside the block rather than `var` outside — lets you easily know that a variable's scope is limited to the current block. (We also try to avoid multi-variable initialization, since it makes for worse diffs.)

adam3smith: got it, thanks!

PR #2175, opened 2020-05-11
Update FAO Publications.js

swifterslb:
Various adjustment to make the code work, mainly:

- Change Element ID from N1 to N0 at multiple occurrences;
- Add DOI;
- Adjust textVariables (metadata names). Remove some obsolete ones.

New test cases.


dstillman:
(Note that you can always just squash and force-push to clear erroneous commits. No need to start a new PR. We're happy to walk through steps to fix an existing one.)

PR #2178, opened 2020-05-16
clinicaltrials.gov: Fix authors

adam3smith:
And use field mode; fixes 
https://forums.zotero.org/discussion/83162/saving-item-from-clinicaltrials-gov/p1


dstillman:
Thanks!

adam3smith:
@dstillman -- wasn't sure where to open an issue for that, but it'd be good to catch illegal data like this in the translator tests. I obviously should have caught this on review, but it's pretty easy to miss.

PR #2179, opened 2020-05-16
FAO Publications: language methods optimized

swifterslb:
The language section is rewritten. Now it should handle all situations well.

### DIFF
--- a/FAO Publications.js
+++ b/FAO Publications.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2020-05-12 21:41:07"
+	"lastUpdated": "2020-05-16 21:40:58"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


FAO Publications.js
```diff
@@ -112,28 +112,45 @@ function scrape(doc, url) {
...
+		var pdfFile = pdfFilename.slice(0, pdfFilename.indexOf('.'));
+		// In PDF file name, the immediate char before langCode is always a number.
+		for (let i = pdfFile.length - 1; i >= 0; i--) {
+			if (isNaN(pdfFile.charAt(i))) {
+				langCode = pdfFile.charAt(i) + langCode;
+			}
+			else {
+				break;
+			}
+		}
```
dstillman, line 129: This is a bit convoluted and non-idiomatic. You can just use a regexp:

```javascript
var langCode = '';
var matches = pdfUrl.match(/([a-z]+)\.pdf$/i);
if (matches) {
	langCode = matches[1];
}
```

(Could also be restricted further to require a number before the lang code if necessary.)

swifterslb: Thanks so much! This is elegant. I have never mastered regexp well so have generally been avoiding it... There's no need to force a number to be exist before lang code for now. I've adjusted the code in the new commit.

dstillman:
Thanks!

PR #2182, opened 2020-05-18
Embedded Metadata: Add HW Editors

adam3smith:
Not linting given number of open PRs


adam3smith:
Given how widely this is used, I'd appreciate a second set of eyes on this.

PR #2186, opened 2020-05-31
Add ABC News

Joycekwc:
This is a translator for the ABC News website (an Australian news service). It only scrapes the ABC News site, not for other ABC content. It is based on the Embedded Metadata scraper but adds the author, tags and multi-item scraping.

I'm very new to this (and coding more generally), so any tips appreciated!

### DIFF
--- a/ABC News Australia.js
+++ b/ABC News Australia.js
@@ -1,6 +1,6 @@
 {
 	"translatorID": "92d45016-5f7b-4bcf-bb63-193033f02f2b",
-	"label": "ABC News",
+	"label": "ABC News Australia",
 	"creator": "Joyce Chia",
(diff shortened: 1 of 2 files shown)


AbeJellinek:
Hey, thanks for contributing this! Sorry it took so long to get to. Big backlog, but we're getting through it. I went ahead and linted and fixed some weirdnesses with video - seems like they changed their `ABC.ContentType` schema. I'd often go back and forth making comments, etc., but in a case like this where the translator has been sitting in the queue for a while, it's easier just to make the small fixes that need to be made and merge.

Joycekwc:
Great thanks! I can’t wait to use it.

Joyce.

PR #2191, opened 2020-06-10
FAO Publications: Update DOI method

swifterslb:
DOI has become standard for new FAO publications.
Method for obtaining the DOI is optimized. Also assign DOI URL as newItem.url when DOI exists.


swifterslb:
I'm sorry but this has been so long. Could an administrator have a look at this? Thanks.

dstillman:
Can you rebase this against master? You'll need to resolve some conflicts with changes from your previous FAO patches.

PR #2192, opened 2020-06-11
Remove type mapping for Citavi

adam3smith:
See https://forums.zotero.org/discussion/83613/formatting-of-journal-articles-in-reference-list-apa-7-after-import-from-citavi-5#latest


dstillman:
Thanks!

PR #2193, opened 2020-06-19
Update Idref.js to make it more generic.

symac:
The French national archive has many subdomains of archives-ouvertes.fr that can appear on idref.fr. Rather than adding an entry for each of them, using a generic test that will translate content when on this domain.

### DIFF
--- a/Idref.js
+++ b/Idref.js
@@ -54,7 +54,7 @@ function getTranslatorFromDomain(domain) {
 	}
 
 	/* Fallback for *.archives-ouvertes.fr */
-	if (domain.includes(".archives-ouvertes.fr")) {
+	if (domain.endsWith(".archives-ouvertes.fr")) {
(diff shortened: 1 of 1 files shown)


Idref.js
```diff
@@ -37,21 +37,31 @@
...
 	'dumas.ccsd.cnrs.fr': '951c027d-74ac-47d4-a107-9c3069ab7b48',
 };
 
+function getTranslatorFromDomain(domain) {
+	if (domain2translator[domain]) {
+		return domain2translator[domain];
+	}
+
+	/* Fallback for *.archives-ouvertes.fr */
+	if (domain.includes(".archives-ouvertes.fr")) {
```
dstillman, line 57: Not that it really matters, but for precision it should be `.endsWith('.archives-ouvertes.fr')`.

symac:
Thanks @dstillman for your review. I have fixed the PR with this suggestion.

dstillman:
Thanks!

PR #2194, opened 2020-06-21
Add Argentine BNM

adam3smith:
Requested on Twitter: https://twitter.com/Ronconi/status/1274087752905166861

Easily accessible MARC XML

### DIFF
--- a/Biblioteca Nacional de Maestros.js
+++ b/Biblioteca Nacional de Maestros.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2020-06-21 20:46:20"
+	"lastUpdated": "2020-06-22 00:23:44"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


Biblioteca Nacional de Maestros.js
```diff
@@ -0,0 +1,210 @@
...
+
+function constructMARCurl(url) {
+	url = url.replace(/\/(Details|Holdings)([#?].*)?/, ""); // remove panels
+	return url + "/Export?style=MARCXML";
+}
+
+function scrape(urls) {
+	for (let url of urls) {
+	// Z.debug(url);
+	// eslint-disable-next-line loopfunc
```
adam3smith, line 85: @dstillman just making sure disabling this rule is OK in this context? My understanding is that it's intended to avoid re-defining a function n times, which this doesn't do, but wasn't 100% sure (the reason there's a loop here is to avoid unnecessarily downloading the catalog item page from search results.)

dstillman: Did you mean `no-loop-func`? `loopfunc` is from JSHint, I think.

This is OK — the function doesn't reference any external variables, so there's nothing to get confused by here — but it does technically create multiple anonymous functions for no reason. You could just define `var handler = function (text) { … };` before the loop and pass `handler` to each invocation.

Biblioteca Nacional de Maestros.js
```diff
@@ -0,0 +1,211 @@
...
+		var translator = Zotero.loadTranslator("import");
+		// Z.debug(text);
+		// MARC XML
+		translator.setTranslator("edd87d07-9194-42f8-b2ad-997c4c7deefd");
+		translator.setString(text);
+		translator.translate();
+	};
+	for (let url of urls) {
+		// Z.debug(url);
+		ZU.doGet(constructMARCurl(url), handler);
```
dstillman, line 94: Actually, there's a more fundamental problem with this. When `doGet()` is used inside a loop, it's called simultaneously as many times as there are loops, resulting in many simultaneous requests (possibly limited to 6 by the network stack). doGet() can take an array of URLs, and it will automatically serialize those rather than trying to make all the requests at once. (In the future, it (or its successor) could also gently parallelize them.) So you can just remove the loop here and call doGet() once with an array of constructed URLs, with a single function for the callback.

I suspect this mistake is made in many translators. It's actually done in the Google Scholar translator, which might be the reason for the frequent throttling we see there.

adam3smith: That makes sense -- I think I got this right in https://github.com/zotero/translators/pull/2194/commits/adff8648ceca06a648b80d96a9774dad0a04be82 ?

dstillman:
Yep, that looks good.

PR #2198, opened 2020-06-25
Do not remove words from first word of title when the word is hyphenated

kernoelpanic:
The current BibLaTex export of Zotero creates the following bibtex key for this paper:
```
@book{smith_proof--stake_2020,
  title = {Proof-of-Stake Protocols},
  url = {https://eprint.iacr.org/...,
  author = {Smith, John},
  date = {2020},
}
```

but it iin my opinion it should not replace words like:
```
var citeKeyTitleBannedRe = /\b(a|an|the|some|from|on|in|to|of|do|with|der|die|das|ein|eine|einer|eines|einem|einen|un|une|la|le|l'|el|las|los|al|uno|una|unos|unas|de|des|del|d')(\s+|\b)|(<\/?(i|b|sup|sub|sc|span style="small-caps"|span)>)/g;
```
when the first word of the title is hyphenated like in the example.
So the result should be this:
```
@book{smith_proof-of-stake_2020,
  title = {Proof-of-Stake Protocols},
  url = {https://eprint.iacr.org/...,
  author = {Smith, John},
  date = {2020},
}
```

### DIFF
--- a/BibLaTeX.js
+++ b/BibLaTeX.js
@@ -383,6 +383,10 @@ var citeKeyConversions = {
 	},
 	t: function (flags, item) {
 		if (item.title) {
+      var first_word = item.title.toLowerCase().split(/\s+/g)[0];
+      if (first_word.includes("-")) {
+        return first_word;
+      }
 			return item.title.toLowerCase().replace(citeKeyTitleBannedRe, "").split(/\s+/g)[0];
 		}
 		return "notitle";


AbeJellinek:
I agree with this change in principle, but citekeys should be stable, not necessarily pretty. This could break a lot of bibliographies on re-export, and it would be relatively hard for users to figure out what happened. [A previous breaking citekey change](https://github.com/zotero/translators/commit/0f5bff5f3f29a75e7bd1ea717e791d24fa0abcc5#diff-1ecddfba976927a78ecf7a567210e4b1e37802688fd123d096ebe0954123523d) included a check in order only to affect newly-added items. This PR's change isn't vital enough to justify building that infrastructure again IMO.

[Better BibTeX](https://retorque.re/zotero-better-bibtex/) lets you change the template used to generate citekeys and even set them manually on a per-item basis. It's a great plugin and I recommend giving it a try for this use case.

PR #2199, opened 2020-06-27
Updated Researchgate translator

step21:
Hey, with this update ResearchGate translator kind of works again. Sometimes item type still fails to be detected though, as classes are sometimes there or not... or maybe detectWeb also could use RIS directly?
Would be great to merge this especially for items that have no doi, let me know what you think and if it needs any changes.

### DIFF
--- a/ResearchGate.js
+++ b/ResearchGate.js
@@ -37,11 +37,8 @@
 
 
 // attr()/text() v2
-function attr(docOrElem, selector, attr, index) {
-	var elem = index ? docOrElem.querySelectorAll(selector).item(index) : docOrElem.querySelector(selector); return elem ? elem.getAttribute(attr) : null;
-} function text(docOrElem, selector, index) {
-	var elem = index ? docOrElem.querySelectorAll(selector).item(index) : docOrElem.querySelector(selector); return elem ? elem.textContent : null;
-}
+// eslint-disable-next-line
+function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null}function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null}
 
 function detectWeb(doc, url) {
 	if (url.includes('/publication/')) {
--- a/ResearchGate.js
+++ b/ResearchGate.js
@@ -130,12 +130,12 @@ function scrape(doc, url) {
 	ZU.doGet(risURL, function (text) {
 		// Z.debug(text);
 		// fix wrong itemType information in RIS
-		if (type == "bookSection") text = text.replace('TY	- BOOK', 'TY	- CHAP');
-		if (type == "conferencePaper") text = text.replace('TY	- BOOK', 'TY	- CONF');
-		if (type == "report") text = text.replace('TY  - BOOK', 'TY  - RPRT');
-		if (type == "presentation") text = text.replace('TY  - BOOK', 'TY  - SLIDE');
-		if (type == "journalArticle") text = text.replace('TY  - BOOK', 'TY  - JOUR');
-		if (type == "thesis") text = text.replace('TY  - BOOK', 'TY  - THES');
+		if (type == "bookSection") text = text.replace('TY - BOOK', 'TY	- CHAP');
+		if (type == "conferencePaper") text = text.replace('TY - BOOK', 'TY - CONF');
+		if (type == "report") text = text.replace('TY - BOOK', 'TY - RPRT');
+		if (type == "presentation") text = text.replace('TY - BOOK', 'TY - SLIDE');
+		if (type == "journalArticle") text = text.replace('TY - BOOK', 'TY - JOUR');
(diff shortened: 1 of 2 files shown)


adam3smith:
Thanks! I have one question in line. Beyond that, please remove all the debugging code (if there's a particularly useful step you can lead 1-2 in, but currently that's too much) and relint

ResearchGate.js
```diff
@@ -35,34 +35,52 @@
...
-		var type = text(doc, 'div.publication-meta strong');
+		if (debug) {
+			Zotero.debug('Detected something');
+			Zotero.debug('text: ' + text(doc, "span.research-detail-header-section__badge"));
+			Zotero.debug('attr: ' + attr(doc, 'span.research-detail-header-section__badge', 'content'));
+		}
+		var type = text(doc, "span.research-detail-header-section__badge");
 		if (!type) {
+			Zotero.debug('trying different way');
+			type = text(doc, "b[research-meta-type]");
```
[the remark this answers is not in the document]
step21: Yeah, I might be wrong, but hard to say with tags not always being the same. (plus I am not always sure how text() or attr() filters and couldn't find anything in the docs. It definitely is not XPaths anymore.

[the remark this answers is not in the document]
adam3smith: `text()` and `attr()` use CSS Selectors. See https://www.mediawiki.org/wiki/Citoid/Creating_Zotero_translators#Common_code_blocks_in_translators for good documentation

step21:
Proposal to make it better: Instead of trying to parse the html tree, which has almost no useful info (or at least not consistently) it would be more reliable to get and parse the RIS right away, instead of only if actually saving the item. Would that be okay, or is that not done on purpose because of adverse effects or whatever?

ResearchGate.js
```diff
@@ -37,34 +37,43 @@
 
 
 // attr()/text() v2
-function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null}function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null}
+function attr(docOrElem, selector, attr, index) {
+	var elem = index ? docOrElem.querySelectorAll(selector).item(index) : docOrElem.querySelector(selector); return elem ? elem.getAttribute(attr) : null;
+} function text(docOrElem, selector, index) {
+	var elem = index ? docOrElem.querySelectorAll(selector).item(index) : docOrElem.querySelector(selector); return elem ? elem.textContent : null;
+}
```
adam3smith, line 44: Keep the code minimized and disable eslint here:

```suggestion
// eslint-disable-next-line
function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null}function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null}
```

ResearchGate.js
```diff
@@ -37,34 +37,43 @@
...
+			case "Research":
+				return "report";
+			case "Presentation":
+				return "presentation";
+			default:
+				return "book";
 		}
-	} else if ((url.match('/search(\\?|/)?') || url.includes('/profile/') || url.includes('/scientific-contributions/')) && getSearchResults(doc, true)) {
+	}
+	else if ((url.match('/search(\\?|/)?') || url.includes('/profile/') || url.includes('/scientific-contributions/')) && getSearchResults(doc, true)) {
```
adam3smith, line 76:
```suggestion
	else if ((url.includes('/search') || url.includes('/profile/') || url.includes('/scientific-contributions/')) && getSearchResults(doc, true)) {
```

should be equivalent & much simpler. (If the regex is needed, don't use `search` or `test` to test against it)

ResearchGate.js
```diff
@@ -123,13 +133,13 @@ function scrape(doc, url) {
 	ZU.doGet(risURL, function (text) {
 		// Z.debug(text);
 		// fix wrong itemType information in RIS
-		if (type == "bookSection") text = text.replace('TY  - BOOK', 'TY  - CHAP');
-		if (type == "conferencePaper") text = text.replace('TY  - BOOK', 'TY  - CONF');
+		if (type == "bookSection") text = text.replace('TY	- BOOK', 'TY	- CHAP');
```
adam3smith, line 136: This spacing seems unusual for RIS -- did something go wrong here during pretty-printing or does RG really reliably use tabs between field code and hyphen?

step21: This is probably from me trying to make the linter happy in Vim. Then I guess the linter was hapyy but it messed it up too. Will fix tomorrow.

ResearchGate.js
step21: Fine with me. It was already like this.

adam3smith:
Selenium tests work locally -- there's a 403 error in there, so I'm suspecting RG is blocking this for some reason.

adam3smith:
Thank you and sorry for the delay!

step21:
No worries. Thank you!

PR #2200, opened 2020-06-27
Twitter new interface update

boan-anbo:
Fixed: title, date, author, language
Added: retweet, likes, abstract, blockquote, snapshot of links
Major issues: (1) cannot dynamically detect DOM changes on twitter (based on react). Have to refresh to detect. (2) haven't fixed multiple
Minor issue: cannot take snapshot of the tweet page itself. using link instead for now.

### DIFF
--- a/Twitter.js
+++ b/Twitter.js
@@ -1,7 +1,7 @@
 {
 	"translatorID": "31659710-d04e-45d0-84ba-8e3f5afc4a54",
 	"label": "Twitter",
-	"creator": "Avram Lyon, Philipp Zumstein, Tomas Fiers",
+	"creator": "Avram Lyon, Philipp Zumstein, Tomas Fiers, Bo An",
 	"target": "^https?://([^/]+\\.)?twitter\\.com/",
 	"minVersion": "4.0",
 	"maxVersion": "",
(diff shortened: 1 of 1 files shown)


dstillman:
Thanks for working on this. Using the title for the tweet content is smart — though, obviously, that's what makes `multiple` not an option. But better to just disable multiple for now.

Before we can review this fully, the existing tests need to result in more or less the same data. You're saving the data in various different ways from how it was before (e.g., "Twitter:" in the title, duplicate tweet content in Abstract, date in wrong format, handle in Extra instead of blogTitle).

Old code should also be removed, not commented out.

Re: DOM changes, the Zotero Connector should automatically detect when you click on a different tweet. Do you mean that things like likes/retweets won't change dynamically? `monitorDOMChanges()` is available, but it probably isn't necessary just for those things.

We're working on better snapshots that will properly save tweets and other JS-heavy pages.

dstillman:
Date is probably the most fragile here, since it's just an XPath with spans. If there's anything we can do to make that more reliable (but still work in localized Twitter versions), that would be good.

dstillman:
I mostly rewrote this and pushed out a new version, which thanks to our new SingleFile-based snapshots will also save stable snapshots of the page.

PR #2203, opened 2020-07-07
Updated Douban translator 

xuwd:
1. fix the abnormal data of some fields, and add the test case
2. crawl rating number and total rating people to extra field.
3. support Douban Movie
4. Add BiliBili translator(BiliBili.js & BiliBili Bangumi.js)
thanks!


dstillman:
I'm not really sure what's going on with this PR — you've added 46 commits — but for us to review it there needs to be a single file changed.

dstillman:
Commit messages and comments should also be in English, though you're welcome to include Chinese translations as well.

PR #2204, opened 2020-07-07
Travis Selenium PR tests with Chrome

adomasven:
Will run Chrome with translator tester for translator repo pull requests in travis-ci. Runs tests for all translators changed since the branch was created from master and additionally any translators that depend on that translator. A maximum of 10 translators are tested (although this limit could possibly be increased) to not timeout travis-ci, since each test can take [up to 15s](https://github.com/zotero/zotero/commit/8e4ac401742a56f420f73be61dac0ba558afc31c) to complete and many translators have multiple tests.

This is what a sample run in travis looks like:
https://travis-ci.com/github/adomasven/translators/builds/174607350

image

Fixes a couple of translators with non-JSON compliant test description.

You can run the tester yourself locally:
```
npm i
cd .ci/pull-request-check
./check-pull-request.sh
```

### DIFF
--- a/.ci/pull-request-check/check-pull-request.sh
+++ b/.ci/pull-request-check/check-pull-request.sh
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+ROOT_DIR="$( dirname "$DIR" )"
+
+# Build connector
+mkdir -p connectors
+cd connectors
+
+if [ -d .git ]; then
+	git pull
+	git submodule update
+	npm ci
+else
+	git clone https://github.com/zotero/zotero-connectors.git --depth 1 .
+	git submodule update --init --depth 1
+  npm ci
+fi
+
+export ZOTERO_REPOSITORY_URL="http://localhost:8085/"
+./build.sh -p b -d
+cd ..
+
+./selenium-test.js
+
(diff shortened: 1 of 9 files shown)


adomasven:
If you test this locally make sure you do not have Zotero open or the tester will get translators from Zotero instead of the local server.

dstillman:
I think `depth=1` won't work on its own now when pulling/updating the connector, since it [now](https://github.com/zotero/zotero-connectors/commit/56ef3428dc898b2ad1d8e4679f15f37bb544eedf) needs the client's schema submodule (at `resource/schema/global`) at build time. So we'd at least have to update that separately.

adomasven:
> I think `depth=1` won't work on its own now when pulling/updating the connector, since it [now](https://github.com/zotero/zotero-connectors/commit/56ef3428dc898b2ad1d8e4679f15f37bb544eedf) needs the client's schema submodule (at `resource/schema/global`) at build time. So we'd at least have to update that separately.

`depth=1` specifies the length of the commit history to fetch. We'd need `--recursive` for deep submodules. Although I don't think deep submodules is necessarily a good idea and the client's schema submodule perhaps should be submoduled (?) directly by the connectors instead?

dstillman:
Oh, sorry, I wasn't thinking straight re: `depth`.

The global schema submodule needs to stay aligned with the client code, so it makes sense to use that submodule directly. We can pull just that submodule, though, which I've done in a follow-up commit.

(We already use `clone --recursive` in the Zotero Connector readme, which pulls all nested submodules, but we could change that to just pull the global schema submodule if the connector doesn't need translators, styles, or style locales (which I don't think it does, unless I'm forgetting something).)

dstillman:
Merged in fac2812100a2463807797884f831d9d312f5c47d

Issue #2206, opened 2020-07-07
Washington Post translator no longer correctly detecting newspaper articles


kaldari:
Except in the cases of blog posts and very old articles (which use the /wp-dyn/ URL scheme), Zotero is not detecting most Washington Post articles as newspaper articles. It is detecting them as simple web pages instead.

kaldari:
I created a pull request to fix the bug: https://github.com/zotero/translators/pull/2207

dstillman:
Fixed in #2207

PR #2207, opened 2020-07-07
updating detectWeb() for Washington Post

kaldari:
See https://github.com/zotero/translators/issues/2206

Also adding a new test.


dstillman:
Thanks!

Note that while it probably doesn't matter for Wikipedia, the timestamp has to be updated for the update to be pushed out to Zotero users, so that should always be done as part of any translator fix. I updated it here.

PR #2209, opened 2020-07-08
sudoc new URL (cbs/ path) is captured again

### LINKED ISSUE #2202: [PICA] sudoc broken
Two years ago there had been an issue with http://www.sudoc.abes.fr/ that was not working properly ( #1529 ). @zuphilip had provided a quick fix by rewriting the base url.

The issue is with multiple results page. It works fine on a single record page, but for multiple, the list of available records is displayed, but trying to get one crashes.

But it appears that, following an update on their website, it is broken again. There is an extra cbs/ path in the url. I have tried to fix it by replacing the previous fix : 

```
newUrl = newUrl.replace(/sudoc\.abes\.fr\/cbs\/\/?DB=/, 'sudoc.abes.fr/cbs/xslt/DB=');
```

But it does not appear to work, if anybody has an idea on how to fix it.

kerphi:
fixes #2202 issue

cc @symac

### DIFF
--- a/Library Catalog (PICA).js
+++ b/Library Catalog (PICA).js
@@ -1,15 +1,15 @@
 {
 	"translatorID": "1b9ed730-69c7-40b0-8a06-517a89a3a278",
 	"label": "Library Catalog (PICA)",
-	"creator": "Sean Takats, Michael Berkowitz, Sylvain Machefert, Sebastian Karcher",
-	"target": "^https?://[^/]+(/[^/]+)?//?DB=\\d",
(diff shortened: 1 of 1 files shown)


adam3smith:
We'll take this as is, but if someone would be willing to lint it (which is a bunch of manual work) that'd be great.

adam3smith:
@symac Thanks! Could you rebase this into **two** commits: One for the first two issues that fix the problem, one for the second two that fix linting?

kerphi:
@adam3smith I think it's ok, I just rebased the lint stuff into one commit and and rebased the bug fix commits into one commit.
Tell me if it's ok?

cc @symac

adam3smith:
Perfect, thank you!

PR #2210, opened 2020-07-09
Fix authors on Washington Post

dstillman:
This works for everything we have tests for, but I didn't test it widely across the site.


kaldari:
@dstillman - Looks like you need to declare `authors` as a local var so it won't throw an undefined error in the Travis checks.

dstillman:
Oops, bad refactoring — thanks. Fixed, but there's a GitHub incident at the moment that's preventing the PR from getting updated.

Issue #2197, opened 2020-06-24
"DEPATISnet" translator largely broken


Ingenieur-K:
Hello everybody,

first of all I would like to thank all those who have so generously developed this great program and continue to maintain it: Thank you very much!
Since a couple of weeks I have noticed that the translator of the FireFox plugin for depatisnet (the german patent office site; [https://depatisnet.dpma.de/DepatisNet/depatisnet?window=1&space=menu&content=index&action=index]( https://depatisnet.dpma.de/DepatisNet/depatisnet?window=1&space=menu&content=index&action=index)), DEPATISnet.js, does not work properly in most parts:

If you try to take over the patent in Zotero on the bibliographic data page (e.g. https://depatisnet.dpma.de/DepatisNet/depatisnet?action=bibdat&docid=FR000001547439A), the only thing that is recognized is that it is a patent. The entry that is then created in Zotero will only contain the patent number in the metadata. The number is also used as the title. The link to the website is also added. The patent pdf will not be downloaded.
In the past, all bibliographic data was correctly transferred (title, inventor, publication date, ...) and the corresponding pdf was automatically downloaded.  All this does not work anymore.

I took a closer look at the Javascript of the Translator.I have determined that the if block statement from line 203 on (i.e. downloading the pdf) is not executed, because the content of the variable 'pages' does not seem to be correct (always 0 I think). Commenting out the if condition has caused the pdfs to be downloaded automatically again. 
If you try to transfer a patent to Zotero for which no pdf is stored (e.g. https://depatisnet.dpma.de/DepatisNet/depatisnet?action=bibdat&docid=FR000000346003A), the small infobox that always appears when you click on the icon of the plugin will show a hint that downloading the pdfs did not work. If I have understood the source code correctly, the if statement should prevent just that.
However, the bibliographic data is still not transferred to Zotero and all other problems as described above still exist. I suspect that the information from the web page is no longer read correctly in the translator (I also have the impression that the Patent Office web page has changed slightly).
I have determined that the if block statement from line 203 on (i.e. downloading the pdf) is not executed, because the content of the variable 'pages' does not seem to be correct (always 0 I think). Commenting out the if condition has caused the pdfs to be downloaded automatically again. 
[...]

PR #2211, opened 2020-07-10
update of DEPATISnet.js

Ingenieur-K:
With regard to the post https://github.com/zotero/translators/issues/2197 the xpath parameters have been updated:

Line 120:
var rows = ZU.xpath(doc, '//table[@class="tab_detail"]/tbody/tr');
to:
var rows = ZU.xpath(doc, '//table/tbody/tr');

Line 200:
var pages = ZU.xpathText(doc, '//div[@id="inhalt"]/h2');
to:
var pages = ZU.xpathText(doc, '//p/span');

Also the name for the downloaded pdf has been changed. Instead of 'fulltext' the file is now named after the patent number:

Line 207:
title: "Fulltext",
to:
title: pn,

### DIFF
--- a/DEPATISnet.js
+++ b/DEPATISnet.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2019-12-07 20:44:27"
+	"lastUpdated": "2020-07-10 11:53:20"
 }
 
 /*
@@ -81,7 +81,7 @@ function cleanTitle(value) {
 
 function cleanName(name, inventors) {
 	name = ZU.trimInternal(name);
-	if (name == "") return "";
+	if (name === "") return "";
 	
 	var parts = name.split(",");
 
@@ -117,8 +117,8 @@ function scrape(doc, url) {
 
 	var ipcs = [];
 
-	var rows = ZU.xpath(doc, '//table[@class="tab_detail"]/tbody/tr');
-	
+	var rows = ZU.xpath(doc, '//table/tbody/tr');			//old line: //var rows = ZU.xpath(doc, '//table[@class="tab_detail"]/tbody/tr');
+
 	for (var i = 0, n = rows.length; i < n; i++) {
 		var columns = ZU.xpath(rows[i], './td');
 
@@ -146,13 +146,13 @@ function scrape(doc, url) {
 				var creators = value.textContent.split(";");
 				for (let creator of creators) {
 					creator = cleanName(creator, true);
-					if (creator != "") {
+					if (creator !== "") {
 						newItem.creators.push(ZU.cleanAuthor(creator, "inventor", true));
 					}
 				}
 				break;
 			case "PA":
-				var assigneeNames = value.textContent.split(";").map(name => cleanName(name, false)).filter(name => name != "");
+				var assigneeNames = value.textContent.split(";").map(name => cleanName(name, false)).filter(name => name !== "");
 				newItem.assignee = assigneeNames.join("; ");
 				break;
 			case "ICM":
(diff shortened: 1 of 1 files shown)


Ingenieur-K:
Hello everybody,

i am still very new to all this and this is my first pull request. What happens now that the checks have been successfully completed? How does the changed file get into the root directory?

Many greetings

Ingenieur_K

DEPATISnet.js
```diff
@@ -197,13 +197,14 @@ function scrape(doc, url) {
 		snapshot: false
 	});
 	
-	var pages = ZU.xpathText(doc, '//div[@id="inhalt"]/h2');
+	var pages = ZU.xpathText(doc, '/html/body/div[4]/p/span');			// old line: var pages = ZU.xpathText(doc, '//div[@id="inhalt"]/h2');
```
dstillman, line 200: This is too fragile. It looks like you can perhaps just get this from the `caption` in the `table`, but if that's not on every page, you'd want to use at least the class name of the containing `p`.

For new code, you can also just use `text()` and CSS selectors instead of XPaths. Example: https://github.com/zotero/translators/blob/1d1ce6111ac17a4064c1d8836f6beec54404c32b/Stitcher.js#L83

DEPATISnet.js
```diff
@@ -117,8 +117,8 @@ function scrape(doc, url) {
 
 	var ipcs = [];
 
-	var rows = ZU.xpath(doc, '//table[@class="tab_detail"]/tbody/tr');
-	
+	var rows = ZU.xpath(doc, '//table/tbody/tr');			// old line: var rows = ZU.xpath(doc, '//table[@class="tab_detail"]/tbody/tr');
```
dstillman, line 120: No need for the old lines in comments. That's what the diff is for.

DEPATISnet.js
```diff
@@ -197,13 +197,14 @@ function scrape(doc, url) {
...
-	var pages = ZU.xpathText(doc, '//div[@id="inhalt"]/h2');
+	var pages = ZU.xpathText(doc, '/html/body/div[4]/p/span');			// old line: var pages = ZU.xpathText(doc, '//div[@id="inhalt"]/h2');
+	
 	// e.g. "Dokument   DE000004446098C2   (Seiten: 8)"
 	// but there is no PDF available when we have "Seiten: 0"
 	if (pages && /(Seiten|Pages):\s*[1-9][0-9]*/.test(pages)) {
 		var pdfurl = "https://depatisnet.dpma.de/DepatisNet/depatisnet/" + pn + "_all_pages.pdf?window=1&space=menu&content=download_doc_verify&action=download_doc&docid=" + pn;
 		newItem.attachments.push({
-			title: "Fulltext",
+			title: pn,		// old line:  title: "Fulltext",
```
dstillman, line 207: What's an example of what will be contained here, and what's the reason for the change? Generally "Full Text PDF" would be appropriate for the attachment item.

dstillman (changes requested):
Sorry for the delay in reviewing this.

In addition to these notes, you should rebase on master (which will make reviewing easier) and make sure you update the tests via the translator editor. (At least one change you made here would cause tests to fail.)

DEPATISnet.js
Ingenieur-K: I have changed this line to

`var pages = ZU.xpathText(doc, '//table/caption').replace("(", "$(").split("$").pop();`

I hope thats better.

DEPATISnet.js
Ingenieur-K: I changed this (for me) because I wanted to have the patent number somewhere in the entry in the overview. I can see that it might make more sense to label the individual elements of the entry according to their category; e.g. 'full text' or 'comment', ... .
However, in order for the patent number to still appear with the entry, I suggest naming the entry with patent number and title, e.g.:

DE1234567A1 - Propulsion of a spacecraft by means of warp drive.

Would this be a feasible way?

Ingenieur-K:
> 
> 
> Sorry for the delay in reviewing this.
> 
> In addition to these notes, you should rebase on master (which will make reviewing easier) and make sure you update the tests via the translator editor. (At least one change you made here would cause tests to fail.)

Unfortunately I have never worked with github before and have no idea how to do a 'rebase on master'. I can't make any changes in the original translator directory. I guess I don't have the rights to do that.

Also I need help with the translator editor and how to update the test. I have done the following:
1. start the translator editor
2. open the translator (here: DEPATISnet)
3. in the tab 'Code' I replaced the whole code with my changed code (was easier than searching for the changed lines and specifically replacing only those)
4. in the 'Tests' tab I left the entries as they were
5. under 'Tersting' I selected the single entries and then clicked on the button 'Run'. All tests ended with the status 'Detection failed'. Unfortunately I couldn't do much with the output in the right half of the window ('Test Frame'). But it seems to me that this is the error:
TypeError: Function.prototype.apply called on incompatible undefined
What this means and how I can fix it, I don't know.
But what I can say is: If I replace the Java Script file DEPATISnet.js on my hard disk with my modified one and then search for patents on the site of the German Patent and Trademark Office and paste it into Zotero via the Browser Addin for FireFox, everything works as expected. So I assume that I am doing something wrong in the translator editor. I would be grateful for some tips.

dstillman:
> Unfortunately I have never worked with github before and have no idea how to do a 'rebase on master'. I can't make any changes in the original translator directory. I guess I don't have the rights to do that.

A rebase is something you do locally, followed by a force-push to this same branch. There are lots of resources online explaining the process, but if you don't know how to do it, don't worry about it.

>  it seems to me that this is the error:
> TypeError: Function.prototype.apply called on incompatible undefined

Can you provide a <a href="https://www.zotero.org/support/reporting_problems#provide_a_report_id">Report ID</a>?

PR #2212, opened 2020-07-15
Add translator for citing publications in Mastodon federated services

febrezo:
A translator for some of the top Mastodon Microblogging sites as stated by [the Mastodon instances site](https://instances.social/list/advanced#lang=&allowed=&prohibited=&min-users=&max-users=).

It generates two different bibliography objects:

- `blogPost`. Since these publications are expected to be treated as microblogging articles as stated [here](https://forums.zotero.org/discussion/28672/citing-twitter-and-websites-on-standalone).
- `webpage`. For the profiles homepages.


dstillman:
I'm not sure there's much value in the profile page, but if we're going to keep support for that, some of the current data isn't correct — e.g., in the test, `date` shouldn't really be 2017 (I'm not sure there's really even an appropriate date to use there), and `firstName` has bad data in it.

Mastodon.js
```diff
@@ -0,0 +1,142 @@
...
+		tmp_date = ZU.xpathText(doc, '//div[@class="detailed-status__meta"]/data/@value');
+	} else {
+		var aux = ZU.xpathText(doc, '//div[@class="public-account-bio__extra"]/text()');
+		tmp_date = aux.substring(aux.length, aux.length-8);
+		newItem.title = ZU.xpathText(doc, "//meta[@property='profile:username']/@content");		
+	}
+
+	newItem.websiteTitle = urlParts[2];
+	newItem.date = ZU.strToISO(tmp_date);
+	var tmp_author = ZU.xpathText(doc, "//strong[@class='display-name__html p-name emojify']");
```
dstillman, line 63: This is going to be fragile this way regardless — is there more stable source for the author? — but if we're going to scrape it it'd be better to use `text()` with a CSS selector and filter for just a single class.

We apparently don't have `text()` and `attr()` documented — I thought we did — but you can see examples in other translators. Add these lines at the top of the code, below the copyright notice:

https://github.com/zotero/translators/blob/53afc969d3ea386827c627e50dfb666a41e8dfe1/ACM%20Digital%20Library.js#L37-L39

And then use it like this:

```javascript
 let reviewedTitle = text(doc, '.article-reviewed-item-title'); 
```

febrezo: Yes it is, the `<h1>` tag can be used. I have updated it.

```
<h1>
Félix Brezo
<small>
@febrezo@mastodon.social

</small>
</h1>
```

I noticed that the error in the first name was caused because it was not properly grabbed.

febrezo:
> I'm not sure there's much value in the profile page, but if we're going to keep support for that, some of the current data isn't correct — e.g., in the test, `date` shouldn't really be 2017 (I'm not sure there's really even an appropriate date to use there), and `firstName` has bad data in it.

I have updated the profile page to correctly grab the values. The problem with the date is that it is in a strange format but the last values seem to be the month and the year. See the HTML code below in Spanish:

```
<div class='public-account-bio__extra'>
Se unió el sep 2017
</div>
```
And English:

```
<div class='public-account-bio__extra'>
Joined Sep 2017
</div>
```

So I grab the last 8 characters to get the month and date. I was hoping to let it automatically guess that it was 2017/09 but only the year got grabbed.

dstillman:
I suspect most of your XPaths/selectors can be made less brittle. You should assume that the page structure — and particularly things like class names — will change as minor tweaks are made to the site and choose selectors defensively so that the translator doesn't need to be constantly updated. E.g., for the date, there's a `<time>` element with a `datetime` attribute containing an ISO date, which you should use instead of a very specific XPath.

dstillman:
My point on the profile date is that I don't think it particularly makes sense to even try to use that date. The join date of someone's account isn't particularly appropriate for a Date field on a cited profile page.

febrezo:
> My point on the profile date is that I don't think it particularly makes sense to even try to use that date. The join date of someone's account doesn't seem particularly appropriate for a Date field on a saved/cited profile page.

My idea with the profile was to reflect the moment in which the microblogging site was created (the account creation date) as I thought that I could be interesting as the original publication date of a book. But we can remove that!

> I suspect most of your XPaths/selectors can be made less brittle. You should assume that the page structure — and particularly things like class names — will change as minor tweaks are made to the site and choose selectors defensively so that the translator doesn't need to be constantly updated. E.g., for the date, there's a <time> element with a datetime attribute containing an ISO date, which you should use instead of a very specific XPath.

I'll work on that and let you know! Thanks!

dstillman:
Just to explain this more clearly, when using scraping instead of embedded metadata, the goal is to choose selectors that are likely to survive a site redesign. There's no reason to choose a class like `formatted` that could easily be changed. I would just use `<time>` and assume that the first `<time>` on a detected page was going to be the main entry's timestamp. (Even `.entry time` would be better than `formatted`, since it seems more likely that `.entry` would remain, but that would still be more brittle than necessary.)

Same with the author name, which you haven't changed — you're requiring a whole set of classes (e.g., `emojify`) that could trivially be changed at any point, which would result in authors silently being omitted going forward. That one's a bit of a toss-up between `.display-name__html` and `.p-name`, but I'd probably go with `.p-name`, which feels less likely to be tied to a particular underlying technology. (It'd still be using `display-name__account`, but not really another option there (other than `og:title`, but that seems misused).

Mastodon.js
```diff
@@ -0,0 +1,137 @@
...
+	newItem.date = ZU.strToISO(tmpDate);
+	if (tmpAuthor) {
+		newItem.creators.push(ZU.cleanAuthor(tmpAuthor, "author", false));
+	}
+	newItem.websiteType = "Microblogging (Mastodon)";
+	newItem.url = url;
+
+	// Adding the attachment
+	newItem.attachments.push({
+		title: "Mastodon Snapshot",
```
dstillman, line 69: This can just be `Snapshot`.

dstillman:
Note that you still haven't fixed the author XPath.

And I think we should remove the date for the profile page — it's mostly just a bit of trivia shown on the page, not something representative of the profile itself.

Mastodon.js
```diff
@@ -43,21 +43,19 @@ function doWeb(doc, url) {
...
-	var tmpDate, tmpAuthor;
+	var tmpAuthor;
 	if (resourceType == "blogPost") {
 		newItem.blogTitle = ZU.xpathText(doc, "(//span[@class='display-name__account'])[1]");
 		newItem.title = ZU.xpathText(doc, "//meta[@name='description']/@content");
-		tmpDate = ZU.xpathText(doc, '//time[@class="formatted"]/@datetime');
-		tmpAuthor = ZU.xpathText(doc, "(//strong[@class='display-name__html p-name emojify'])[1]");
+		var tmpDate = ZU.xpathText(doc, '(//time)[1]/@datetime');
+		newItem.date = ZU.strToISO(tmpDate);
+		tmpAuthor = ZU.xpathText(doc, "//meta[@property='og:title']/@content").split("(")[0];
```
dstillman, line 52: 1) Won't this leave the closing parenthesis? If you know there's going to be specific punctuation at the end, you should remove it manually rather than just relying on `cleanAuthor()` below.

2) I specifically said that `og:title` isn't appropriate for the author name, since they're misusing it — the title of the page for an individual post shouldn't be the author's name.

febrezo: No, since it's a `split` and only the first part of the `og:title` tag is grabbed. I can also grab it using a regexp if that's read better. Regarding the selection, I have added the space before the parenthesis to avoid relying on `cleanAuthor` to remove the trailing space. Some sample code of the grabbed content can be found below.

```tmp = "Félix Brezo (@febrezo@mastodon.social)";
"Félix Brezo (@febrezo@mastodon.social)"
tmp.split("(")[0];
"Félix Brezo "
tmp.split(" (")[0];
"Félix Brezo"
```
Regarding 2, there are only three places in which the author name appears in the HTML. The `<title>` together with some extra information, this `<meta>` and the already discarded `<strong class='display-name__html p-name emojify'>`. I don't really know where else to grab it from.

dstillman: > No, since it's a split and only the first part of the og:title tag is grabbed.

Sorry, I misread. A regexp that anchored the final `(` and `)` to the end would technically be better, since right now it would be wrong if someone included parentheses in their display name, though I don't know if that's possible on Mastodon. (Not a big deal regardless.)

> the already discarded `<strong class='display-name__html p-name emojify'>`

I didn't say not to use that — I said to only use one class, not all three, such that a simple change to any of them would break it. You [can do that with an XPath](https://stackoverflow.com/a/1604480), but it'd be better to add [these temporary lines](https://github.com/zotero/translators/blob/50aece7c7dfd4e6a57967c8aad8f50bd06c82e85/ACM%20Digital%20Library.js#L37-L39) to the top of the code and then use `text(.p-name)` (a CSS selector) to extract the author.

febrezo: I have just made an update since I noticed that the aforementioned approach was not correctly working when saving toots on threads. The aforementioned approach was assuming that the first toot was the one to quote, but it is not the case if the toot is part of a thread. I had to capture the highlighted publication using `//a[@class='detailed-status__display-name u-url']` and `//div[@class='detailed-status detailed-status--flex detailed-status-public']`. I know that there are too much classes to wrap but just choosing one of them (`//a[@class='detailed-status__display-name'` or  `//div[@class='detailed-status'`) was returning no results.

At the same time, I thought that I could use structures as `//a//strong` instead of specifying the full path `//a/span/bdi/strong`. I am finding out that I have more gaps on how Xpath works than I initially thought. Sorry for the inconvenience and thanks for the help anyway. Let's see if we can have this merged soon.

dstillman: I'm not going to approve this with such specific XPaths — there's just no reason for it. It's better to spend time understanding how to write resilient translators than to write a translator that will just break and take up people's time unnecessarily later.

I linked to instructions above for matching on specific classes with an XPath — just using `class=` will not work — but as I say, you should stop using XPaths and just use text() with CSS selectors instead.

febrezo: I have gone through this old translator which wasn't approved I have to admit that this is awkward for me since I cannot understand the function provided, whcih I have manually added at the beginning of my code as suggested:

```
// attr()/text() v2
// eslint-disable-next-line
function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null;}function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null;}
```
I have made it work for a simple example making these two lines match exactly:
```
		//newItem.title = ZU.xpathText(doc, "//meta[@name='description']/@content").replace(/\n/g, " ").replace(/ +/g, " ");
		newItem.title = attr(doc, 'meta[name=description]', 'content').replace(/\n/g, " ").replace(/ +/g, " ");
```
However, I'm struggling with more complex examples such as the following which dramatically crashes:
```
		//let tmpDate = ZU.xpathText(doc, "//div[@class='detailed-status detailed-status--flex detailed-status-public']/div/data/@value");
		let tmpDate = attr(doc, 'div[class=detailed-status detailed-status--flex detailed-status-public]/div/data', 'value');
```
Where can I find more documentation on how to use the provided function?

dstillman: `attr` and `text` use standard [CSS selectors](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors), the same as you would use in a CSS file (or `document.querySelector()`).

E.g., you would match a class with a [class selector](https://developer.mozilla.org/en-US/docs/Web/CSS/Class_selectors) like `.detailed-status-public`.

Issue #668, opened 2014-01-07
Better category support for arxiv translator


melsophos:
I was wondering how the field "publication" for arxiv entries is filled: first I was thinking that the first item to appear is the main category, while the ones after - prepended with "physics:" (in my example to follow) - were the secondary categories.

But if I take the following paper: http://arxiv.org/abs/0912.5286
The publication field is "arXiv: 0912.5286 [astro-ph, physics:gr-qc, physics:hep-th]".
On the opposite, the category list is: "**gr-qc**, astro-ph(.CO), hep-th", where gr-qc is the principal.

I need to know precisely what is the primary class because I have a script which parses the bib file exported from Zotero in order to insert the various fields for eprints (archivePrefix, eprint, primaryClass).
(Note that I don't know if it is easy to find it, since at this point the categories are all collected at the same time.)
- It would also be better to have the extension, e.g. ".CO" after astro-ph here.
- I don't know if it is a good idea to insert the "physics:" because only some categories go under physics, if we look on the [arxiv main page](http://arxiv.org).

PR #2215, opened 2020-07-19
change oai to atom xml (#668)

wuchichou:
As suggested in the discussion in #668, the atom feed instead of the oai
is used to give a clear primary category.


AbeJellinek:
Sorry, I don't think getting slightly better categories is worth switching the data source for a translator that has been stable for many years. We know that the current translator works and gives good results; I don't think it's worth swapping that out to gain three characters in `publicationTitle`. I'd love to hear others' reasoning if there are compelling reasons to make the switch, but I'm closing this for now.

PR #2219, opened 2020-08-02
Add IETF translator

febrezo:


IETF.js
```diff
@@ -0,0 +1,217 @@
...
+						"lastName": "Resnick",
+						"creatorType": "author"
+					}
+				],
+				"date": "2001-04",
+				"abstractNote": "This document specifies a syntax for text messages that are sent between computer users, within the framework of \"electronic mail\" messages.",
+				"institution": "IETF",
+				"libraryCatalog": "IETF",
+				"reportNumber": "RFC 2822",
+				"reportType": "STANDARDS-TRACK",
```
dstillman, line 202: We should normalize these to, e.g., "Standards Track" (as shown in the report itself).

IETF.js
```diff
@@ -0,0 +1,217 @@
...
+function detectWeb(doc, url) {
+	return "report";
+}
+
+function doWeb(doc, url) {
+	var resourceType = detectWeb(doc, url);
+	var newItem = new Zotero.Item(resourceType);
+	
+	var targetUri;
+	if (url.indexOf("/rfc/") > -1) {
```
dstillman, line 43: `.includes()`, not `indexOf()`, and same below, but it'd be better to just replace this whole set of conditions with a simple regexp that just extracts the RFC number, and then build a new URL with that.

IETF.js
```diff
@@ -0,0 +1,217 @@
...
+		// Start scraping
+		newItem.title = ZU.xpathText(metadataDoc, "//meta[@name='DC.Title']/@content");
+		
+		// Iterating through authors
+		var index = 0;
+		while (true) {
+			index++;
+			var tmpAuthor = ZU.xpathText(metadataDoc, "(//meta[@name='DC.Creator'])[" + index + "]/@content");
+			if (tmpAuthor) {
+				var splittedAuthor = tmpAuthor.split(" <")[0].split(", ");	// Remove references to emails that sometimes appear
```
dstillman, line 70: * `splittedAuthor` → `splitAuthor`, and same elsewhere
  * Use `let` instead of `var` inside blocks to keep variables block scope

IETF.js
```diff
@@ -0,0 +1,217 @@
...
+					newItem.creators.push({ lastName: splittedAuthor[0], firstName: splittedAuthor[1], creatorType: "author" });
+				}
+			}
+			else {
+				break;
+			}
+		}
+		newItem.reportNumber = "RFC " + ZU.xpathText(metadataDoc, "//meta[@name='DC.Identifier']/@content").split(":")[3];
+		newItem.institution = "IETF";
+		var abstractContent = ZU.xpathText(metadataDoc, "//meta[@name='DC.Description.Abstract']/@content").split("[");
```
dstillman, line 84: Same as my comment in the Mastodon PR — a regexp is just going to be much more reliable, since you can anchor the `[FOO]` to the end. It can still be optional, in case it's ever omitted, but you don't want this to break if there happens to be a `[` in the middle of the description.

IETF.js
```diff
@@ -0,0 +1,217 @@
...
+		newItem.abstractNote = abstractContent[0].replace(/\n/g, " ");
+		newItem.reportType = abstractContent[1].split("]")[0];
```
dstillman, line 86: Extract this with the same regexp match as above

IETF.js
```diff
@@ -0,0 +1,217 @@
...
+				"title": "Internet Message Format",
+				"creators": [
+					{
+						"firstName": "Peter W.",
+						"lastName": "Resnick",
+						"creatorType": "author"
+					}
+				],
+				"date": "2001-04",
+				"abstractNote": "This document specifies a syntax for text messages that are sent between computer users, within the framework of \"electronic mail\"\nmessages.",
```
dstillman, line 165: Shouldn't be a newline in the abstract

IETF.js
```diff
@@ -129,11 +133,11 @@ var testCases = [
 					}
 				],
 				"date": "1998-12",
-				"abstractNote": "This document defines the IP header field, called the DS (for differentiated services) field.",
+				"abstractNote": "This document defines the IP header field, called the DS (for differentiated services) field. [STANDARDS-TRACK]",
```
dstillman, line 136: You're no longer properly stripping "[STANDARDS-TRACK]" here.

IETF.js
```diff
@@ -81,11 +81,15 @@ function doWeb(doc, url) {
 		}
 		newItem.reportNumber = "RFC " + ZU.xpathText(metadataDoc, "//meta[@name='DC.Identifier']/@content").split(":")[3];
 		newItem.institution = "IETF";
-		var abstractContent = ZU.xpathText(metadataDoc, "//meta[@name='DC.Description.Abstract']/@content").split("[");
-		newItem.abstractNote = abstractContent[0].replace(/\n/g, " ");
-		newItem.reportType = abstractContent[1].split("]")[0];
+		let abstractContent = ZU.xpathText(metadataDoc, "//meta[@name='DC.Description.Abstract']/@content");
+		newItem.abstractNote = abstractContent.replace(/\n/g, " ");
+		
+		let regexp_type = /\[([^\]]+)\]$/;
```
dstillman, line 87: `let regexp_type` → `let re`, but you want to use a regexp with groups for both the abstract itself and the type at the end.

PMaynard:
Would it be possible to extend this to also include the datatracker: e.g. https://datatracker.ietf.org/doc/rfc8557/

IETF.js
febrezo: If I understood this, I think that I have addressed the variable change correctly.

febrezo:
@PMaynard, I have just added your request in the best way I could. Let me know if the example is right for you.

IETF.js
```diff
@@ -0,0 +1,355 @@
...
+		let allContent = ZU.xpathText(doc, "//pre");
+		try {
+			let re = /Category: (.+)  +[^\n]+?/;
+			newItem.reportType = allContent.match(re)[1];
+		} catch (catError) {
+			let re = /Category: (.+)\n/;
+			newItem.reportType = allContent.match(re)[1];
+		}
+		
+		let auxAuthors = ZU.xpathText(doc, "((//table[@class='table table-condensed']/tbody/tr)[2]/td)[2]").split(',');
```
dstillman, line 58: This isn't a good XPath to use. There's no need for the extra classes or the `tbody`, which make this unnecessarily fragile, and just going by row index is dangerous — e.g., they could easily change the 'table-condensed' class or add an extra row. In this case, it would probably be better to use something like `//table//th[text()="Authors"]/following-sibling::td/following-sibling::td`, unless the 'Authors' label is localized. Still fragile — it's dependent on a hidden `td` with class `edit`, though that could be ignored — but a bit less so.

Scraping still isn't great, though. Why not just retrieve the BibTeX for these and import it via the BibTeX translator? Even if some data had to be adjusted after ("RFC Editor" → "IETF"), that would provide a more stable and complete base. Grep`9cb7002` in the repo root for examples.

The 8557 example is missing the date and abstract, so that has to be fixed one way or another. These should also all set a DOI, as shown in the BibTeX. (That won't currently show up in Scaffold but will be saved to the Extra field for later migration to a real DOI field for Report.)

(It looks like old RFCs have a warning about incomplete BibTeX, so scraping could potentially be used for those — not sure if there's a specific numeric cutoff or we'd have to just look for the warning comment.)

febrezo: Thanks for the comments @dstillman. I have just added the date and the abstract for the new example which was missing. Anyway, the usage of the Bibtex would have simplified many of the selections made.  I'll try to find an example of another translator that deals with Bibtex links found inside the website for an easier import.

I have also considered your recommendations about the usage of XPath Axes like `following-sibling` and so on. In any case, I still needed to apply a regular expression to grab the abstract in the Datatracker since the abstract text is out of any XML node and cannot specifically grab it with decendants, siblings, parents and so on. I still feel that the Bibtex approach may help here a lot since, in fact, the abstract seems to be there.

In any case, I think that this needs another update to include the Bibtex. I'll do my best to find a good example and I'll try to add it for a more consistent approach probably during the weekend.

PR #2221, opened 2020-08-03
Fix CI translator test runner changed file logic. Closes #2220

### LINKED ISSUE #2220: Fix CI translator test commit range detection to work with force pushes
E.g., https://travis-ci.org/github/zotero/translators/builds/714134551

Follow-up to https://github.com/zotero/translators/commit/5db02f3ee354a8e4a3cb852bac49856d5eb170f7#commitcomment-40928628

@adomasven, can you see if there's a way to do this?

adomasven:
See the huge comment in `check-pull-request.sh`

### DIFF
--- a/.ci/pull-request-check/check-pull-request.sh
+++ b/.ci/pull-request-check/check-pull-request.sh
@@ -3,6 +3,34 @@
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 ROOT_DIR="$( dirname "$DIR" )"
 
+TRANSLATORS_TO_CHECK=""
+get_translators_to_check() {
+		# If a PR branch has no conflicts with the master then git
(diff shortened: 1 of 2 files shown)


dstillman:
Sorry, I didn't see this. There is zero chance this is reviewable, so let's just try it out.

What about non-PR pushes? For that, maybe we can actually use `TRAVIS_COMMIT_RANGE`? It doesn't really matter if it works reliably against merged PRs, since those will have already been tested on the PR branch, but it would be nice to test pushes that are made directly to master.

adomasven:
> What about non-PR pushes? For that, maybe we can actually use `TRAVIS_COMMIT_RANGE`? It doesn't really matter if it works reliably against merged PRs, since those will have already been tested on the PR branch, but it would be nice to test pushes that are made directly to master.

I'll look into it.

PR #2223, opened 2020-08-08
Add Ncpssd.org  translator.

018:


dstillman:
For us to review this, it needs to include a single file. Please remove the other two files.

PR #2225, opened 2020-08-08
Update MEDLINEnbib.js

### LINKED ISSUE #2224: Import NCBI/Medline abstracts where newlines start with all caps letters
Reported here: https://forums.zotero.org/discussion/83719/some-abstracts-from-pubmed-truncated#latest

Sample data here:
https://gist.github.com/michi-zuri/5750a395438492c6c32e30d667fa4174

michi-zuri:
Fixes #2224 
Whitespace gets cropped at the right moment, not too early. This way tag lines can easily be distinguished from new lines of continued tags.

### DIFF
--- a/MEDLINEnbib.js
+++ b/MEDLINEnbib.js
@@ -8,7 +8,7 @@
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 1,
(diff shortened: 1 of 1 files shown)


adam3smith:
Great, thanks!

PR #2229, opened 2020-08-28
Added unsupported CrossRef content types

diegodlh:
`proceedings-series` and `peer-review` are listed as CrossRef content types in https://api.crossref.org/v1/types. I understand these 
types do not have equivalent item types in Zotero. Therefore, here I added them to the comment about unsupported CrossRef types.


adam3smith:
Thanks!

PR #2234, opened 2020-09-07
Add Bangkok Post translator

matthewmayer:
Trying to add an improved translator for Bangkok Post. The motivation is that if you add Bangkok Post articles to Wikipedia the references get messed up, it always loads the author as "Limited, Bangkok Post Public Company" and doesn't grab the publication date.

image

This is my first attempt to write a Zotero translator so please be gentle, feedback is much appreciated!

### DIFF
--- a/Bangkok Post.js
+++ b/Bangkok Post.js
@@ -1,0 +1,172 @@
+{
+	"translatorID": "7f74d823-d2ba-481c-b717-8b12c90ed874",
+	"label": "Bangkok Post",
+	"creator": "Matt Mayer",
+	"target": "https://www.bangkokpost.com/.*",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2020-09-07 19:12:01"
+}
+
+/*
+ * ***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2020 Matt Mayer
+	
+	This file is part of Zotero.
+
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+
+function doWeb(doc, url) {
+		scrape(doc, url);
+}
+function getMetaTag(doc,attr, value, contentattr) {
+	const tag = Array.from(doc.getElementsByTagName("meta")).filter(m => m.attributes[attr] && m.attributes[attr].value==value)[0]; 
+	if (tag && tag.attributes[contentattr]) {
+		return tag.attributes[contentattr].value;
+	}
+}
+function scrape(doc, url) {
+	var translator = Zotero.loadTranslator('web');
+	// Embedded Metadata
+	translator.setTranslator('951c027d-74ac-47d4-a107-9c3069ab7b48');
+	translator.setDocument(doc);
+	
+	translator.setHandler('itemDone', function (obj, item) {
+		// Add data for fields that are not covered by Embedded Metadata
+		//author name is stored as firstname lastname
+		const authorName=getMetaTag(doc,"property","cXenseParse:author","content")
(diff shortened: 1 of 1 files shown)


Bangkok Post.js
```diff
@@ -0,0 +1,175 @@
+{
+	"translatorID": "7f74d823-d2ba-481c-b717-8b12c90ed874",
+	"label": "Bangkok Post",
+	"creator": "Matt Mayer",
+	"target": "https://www.bangkokpost.com/.*",
```
dstillman, line 5: 1) `^` at the beginning to restrict to the beginning of the URL

2) Either this or detectWeb() needs to be more restrictive to keep it from returning `newspaperArticle` for all pages on the site (e.g., https://www.bangkokpost.com/subscription). It looks like the target could potentially be limited to a URL with a `[0-9]+` as the second or third segment in the path, or you could check for something unique to articles in the page DOM from detectWeb().

Bangkok Post.js
```diff
@@ -0,0 +1,175 @@
...
+	***** END LICENSE BLOCK *****
+*/
+
+function detectWeb(doc, url) {
+	return 'newspaperArticle';
+}
+function doWeb(doc, url) {
+	scrape(doc, url);
+}
+function getMetaTag(doc,attr, value, contentattr) {
```
dstillman, line 44: Various formatting issues throughout like the missing space between `doc,` and `attr`. If possible, run `npm run lint -- Bangkok\ Post.js` on the file from the command line. (If you can't, don't worry about it — we can lint separately after.)

Bangkok Post.js
```diff
@@ -0,0 +1,175 @@
...
+		item.publicationTitle = "Bangkok Post"
+		
+		
+		item.complete();
+	});
+
+	translator.getTranslatorObject(function(trans) {
+		trans.itemType = "newspaperArticle";
+		trans.doWeb(doc, url);
+	});
```
dstillman, line 79: I know there are many examples of using `getTranslatorObject()` like this in existing translators, but I'm not actually sure why — I haven't tested it, but I don't see why it should be necessary. As noted in the [documention](https://www.zotero.org/support/dev/translators/coding#using_gettranslatorobject), that's intended for situations where you need to access objects exposed by the child translator.

I think you should just be able to run `translator.translate();` and set the item type in the `itemDone` handler if necessary, as is done in the Washington Monthly translator:

https://github.com/zotero/translators/blob/c40859799b0091f4bbaecd43ffeb28834457ceee/Washington%20Monthly.js#L90-L119

dstillman (changes requested):
Thanks for working on this!

Bangkok Post.js
matthewmayer: Thank you so much for the detailed feedback! What's the correct behavior in detectWeb if I don't want to handle a URL? Do I return undefined?

dstillman: Return false.

Bangkok Post.js
matthewmayer: I think the reason you're seeing this pattern show up frequently it's because that's how it works in the tutorial on the Wikipedia page at https://m.mediawiki.org/wiki/Zotero_translator_for_Wikimedia_blog So probably a lot of people are copying from that

Bangkok Post.js
matthewmayer: I made the regex stricter so it looks for paths like /foo/bar/1234567 or /foo/1234567 as you suggested

Bangkok Post.js
matthewmayer: lint now runs clean

Bangkok Post.js
matthewmayer: simplified as suggested

dstillman:
Thanks!

mvolz:
This is now deployed on Wikipedia, however, I notice that although the itemType and publicationTitle are now fixed, it's still getting Bangkok Post as the author and not getting the actual author. I'm guessing this is another one of those "the author is loaded with Javascript" issues?

matthewmayer:
@mvolz do you have some sample urls it's failing to get author for? It was working OK in the tests, not sure if the bangkok post changed something since I wrote this

mvolz:
> @mvolz do you have some sample urls it's failing to get author for? It was working OK in the tests, not sure if the bangkok post changed something since I wrote this

It's all of them, including all the ones in the tests, i.e. https://www.bangkokpost.com/tech/1979315/air-force-satellite-napa-1-launched

mvolz:
view-source:https://www.bangkokpost.com/tech/1979315/air-force-satellite-napa-1-launched - the metadata tag author in all of the test links seem to be Bangkok post, the only place the author is is in the text of the page... 

	<meta name="author" content="Bangkok Post Public Company Limited" />

https://github.com/zotero/translators/pull/2234/files#diff-64f88985a43353742d6b30b1d2ed20b6ab1cd0e42774ff3f63c53bb23853aa74R60 seems to just be getting the meta tags, not looking in the page itself?

matthewmayer:
it seems in the last two months since I wrote this they stopped embedding the nicely formatted "cXenseParse:author" and "cXenseParse:recs:publishtime" metatags :(

so now have to grab directly from the HTML instead

PR #2235, opened 2020-09-08
Extend Wiley regex

adam3smith:
to capture https://esj-journals.onlinelibrary.wiley.com
No new tests because currently all Wiley journal articles fail on Scaffold because of cookies. Works in browser

### DIFF
--- a/Wiley Online Library.js
+++ b/Wiley Online Library.js
@@ -31,13 +31,15 @@
  */
 
 // attr()/text() v2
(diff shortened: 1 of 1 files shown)


dstillman:
What's the cookie issue?

adam3smith:
@dstillman All tests (and regular doWeb attempts) from Scaffold fail with:
`  Error: Error retrieving BibTeX`

When I look at the output of what's supposed to be bibtex (by uncommenting the `Z.debug(text)` in `scrapeBibTex`)
I get an error webpage which includes:
> <p>Cookies are disabled for this browser. Wiley Online Library requires cookies for authentication and use of other site features; therefore, cookies must be enabled to browse the site. Detailed information on how Wiley uses cookies can be found in our <a href="https://www.wiley.com/privacy">Privacy Policy</a>.

PR #2243, opened 2020-09-14
Update CanLII.js

pmwills:
This updates 2 errors:
(1) Neutral citation (with no square brackets ([]) around the date) wasn't being recorded as a Reporter Volume. It should be; my solution was to make the square-bracketed year optional.
(2) the "Other Citations" field may sometimes have a singular "Other Citation" (and the french equivalent)

I added a case that detects both of those.


pmwills:
Determined how to test this properly, and it's not working. I'll reopen the pull request after I've fixed that!

PR #2245, opened 2020-09-14
Pmwills patch 1

pmwills:
- Make the reporterVolume, reporter, first page match what CanLII puts as those (neutral cite or CanLII cite)
- Note reporter sometimes has lower case letters (e.g. "CanLII")
- Store any additional citation (e.g., SCR cite) in the note as well
- Alter selector to include singular "Other Citation" or "Autre Citation"
- N.B. : old (e.g. 1800s) SCR cites may not have a date attached to them
- Remove extra spaces in docket
- Add CanLII as the publisher
- Add jurisdiction field that pulls from the url, and maps jurisdiction based on names from Juris-M legal-resource-registry
- NB: Courts are kept as CanLII does them, not converted into legal-resource-registry types
- Add test to ensure provincial case gets handled properly
- Add test to ensure old case gets handled properly

### DIFF
--- a/CanLII.js
+++ b/CanLII.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2019-11-25 22:05:12"
+	"lastUpdated": "2020-09-11 17:09:11"
 }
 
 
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks -- don't worry about the failing CI test for now. It'll take me a bit to properly review the code changes

pmwills:
I've been discussing with some other Canadian users about how to design a proper Canadian CSL style following the main conventions in legal use, and the restrictions on the CSL language mean this shouldn't be integrated yet. I'm going to close this pull request, and then return to it when the field mapping can be better defined

Issue #2106, opened 2020-01-05
Trove failure


dstillman:
https://forums.zotero.org/discussion/78599/downloading-from-trove

> GET http://trove-proxy.herokuapp.com/pdf/73543424 500 (INTERNAL SERVER ERROR)

zuphilip:
Okay, the PDF generation needs Python 2 but heroku now (?) uses Python 3. I tried a quickly fix for this in https://github.com/wragge/trove-proxy/pull/2 tested in https://arcane-ravine-97394.herokuapp.com/ .

Anyways, it is bad that the remaining code is depending on this part and cannot complete without it. I will try to fix that now.

dstillman:
It looks like the proxy is fixed? I went to test #2108 but it successfully used the proxy.

dstillman:
It also looks like HTTPS works for the proxy. @wragge, any reason we can't switch to that instead?

wragge:
Sorry, I've been moving house/states and been off the air for a while. I'm not sure what the problem was as the proxy seems to be fine. It's on hobby dynos (ie I pay $ to keep it going), so it's not dependent on free hours. Yes, https is fine.

I will update to Python 3 however.

PR #2246, opened 2020-09-15
Fix Trove tag detection

### LINKED ISSUE #2244: Trove newspapers failing on tags
I had a look at the urls in this forum post: https://forums.zotero.org/discussion/comment/362334/#Comment_362334

It seems that most of these articles have tags, and the layout of the tags in Trove has changed to insert a `<div>` around the `<a>`. I will have a go at fixing it.

wragge:
This fixes #2244 - a `<div>` had been wrapped around the `<a>` tag so I just updated the xpath. 

I also updated the PDF proxy to use https. This was discussed in #2106 .

I also commented out the detection of 'works' in Trove. Trove recently underwent a major redesign and has been rebuilt in React. This means that the scraping code for works and search results are very broken. I'm not sure if there's a fix at this stage, so I thought it best just to prevent detection until there was a chance for further investigation.

### DIFF
--- a/Trove.js
+++ b/Trove.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2020-09-15 01:00:33"
+	"lastUpdated": "2020-09-15 01:07:09"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks!

PR #2247, opened 2020-09-17
Update base url of direct links into RecordSearch

wragge:
The National Archives changed the location of the script that lets you build direct links into RecordSearch. This pull requests updates the urls to use the new redirect script.

### DIFF
--- a/National Archives of Australia.js
+++ b/National Archives of Australia.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2020-09-18 07:13:48"
+	"lastUpdated": "2020-09-18 08:17:27"
 }
 
 /*
@@ -201,7 +201,7 @@ function parseItemTable(table) {
 		else if (label == 'related searches') {
 			var childrens = td[1].getElementsByTagName('a');
 			data = [];
-			for (let i = 0; i < childrens.length; i++) {
+			for (let j = 0; j < childrens.length; j++) {
 				data.push(childrens[i].textContent.trim());
 			}
 		}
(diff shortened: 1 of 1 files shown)


National Archives of Australia.js
```diff
@@ -125,7 +125,7 @@ function scrape(doc, url) {
 		default:
 			throw new Error("Unknown page type: " + m[1]);
 	}
-	
+	Zotero.debug(item);
```
dstillman, line 128: Remove

National Archives of Australia.js
```diff
@@ -186,7 +186,7 @@ function scrapeItem(doc, url) {
 	item.archiveLocation = meta.citation.replace(/^NAA\s*:\s*/i, '');
 	
 	var barcode = encodeURIComponent(meta['item barcode']);
-	item.url = 'http://www.naa.gov.au/cgi-bin/Search?O=I&Number=' + barcode;
+	item.url = 'http://recordsearch.naa.gov.au/scripts/AutoSearch.asp?O=I&Number=' + barcode;
```
dstillman, line 189: It looks like this site supports (and in fact redirects to) HTTPS, so these should all be changed to that.

dstillman (changes requested):

wragge:
I've fixed those things, and checked against ESLint, but while I was doing that I found a few other things in need of fixing/improvement:

* the layout of PhotoSearch results & individual items had changed, so weren't being captured
* only digitised items were showing up when capturing item search results, and the titles weren't displaying
* series search results weren't being captured at all
* there's now a download link for PDF versions of digitised files that works! Added this in as well as some metadata enrichments for items and series

National Archives of Australia.js
wragge: Done

National Archives of Australia.js
wragge: Done

dstillman:
Thanks!

PR #2248, opened 2020-09-20
Add OSF Preprints translator

adam3smith:
This works on all preprint sites managed by OSF. Multiples need to be 
handled via the API rather than scrape because page retrieve via XHR 
doesn't contain any info.

Currently just for preprints, this can be extended to all other OSF 
objects relatively easily

### DIFF
--- a/OSF Preprints.js
+++ b/OSF Preprints.js
@@ -1,0 +1,380 @@
+{
+	"translatorID": "3b978207-5d5c-416f-b15e-2d9da4aa75e9",
+	"label": "OSF Preprints",
+	"creator": "Sebastian Karcher",
+	"target": "^https?://(osf\\.io|psyarxiv\\.com|arabixiv\\.org|biohackrxiv\\.org|eartharxiv\\.org|ecoevorxiv\\.org|ecsarxiv\\.org|edarxiv\\.org|engrxiv\\.org|frenxiv\\.org|indiarxiv\\.org|mediarxiv\\.org|paleorxiv\\.org)",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2020-09-19 18:57:13"
+}
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2020 Sebastian Karcher
+
+	This file is part of Zotero.
+
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+// attr()/text() v2
+// eslint-disable-next-line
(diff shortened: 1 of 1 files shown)


OSF Preprints.js
```diff
@@ -0,0 +1,380 @@
...
+function scrape(doc, url) {
+	var translator = Zotero.loadTranslator('web');
+	// Embedded Metadata
+	translator.setTranslator('951c027d-74ac-47d4-a107-9c3069ab7b48');
+	// translator.setDocument(doc);
+
+	translator.setHandler('itemDone', function (obj, item) {
+		// TODO adjust if needed:
+		if (item.extra) {
+			item.extra += "\ntype: article";
```
adam3smith, line 152: @dstillman running into a problem here that looks to me like a bug. Testing e.g. on https://psyarxiv.com/grwd6/ this \only imports the DOI into extra, not `type: article`. If I add a space before type, i.e. `" \ntype: article"` this works (but I think then Zotero/citeproc doesn't pick it up + it should be necessary).
This appears to be a general issue for multi-line entries in Extra -- it fails when importing multiples ([l.114](https://github.com/zotero/translators/pull/2248/files#diff-ba54fd86950c3677ea1d6afed010e23aR114)) as well. 
Tested on Chrome and Firefox. Thoughts?

dstillman: Not a bug. The item type is `document`; CSL `article` maps to Zotero `document` and is therefore removed as redundant. Why do you want to set it?

adam3smith: doh! Of course. 

The reason I want to set it is to mark the item as a preprint in the metadata. I'm assuming/hoping that Zotero will get a preprint item type in the future and this should be migrated. `article` is what CSL 1.0.2 explicitly recommend for preprints: https://github.com/citation-style-language/documentation/blob/master/specification.rst#appendix-iii---types 

Thoughts on how to best do this in a forward-looking way?

dstillman: We can add an exception to allow `type: article`, if you think we'll want to migrate such items to the Preprint item type. Presumably not many items are being created with `type: article` for other purposes?

bwiernik: Yes, I agree that anything with `type: article` should be moved to Preprint. CSL 1.0.2 has clarified that `type: article` should be used for preprints and working papers. It also adds a generic `type: document`. I would suggest that when Zotero adds a Preprint type (mapped to CSL `article`), it remap Document to CSL `document`. 

For existing Document items _without_ `type: article` in them, I wonder if they should be migrated to Preprint? That would produce consistent citations (e.g., in my library, I store preprints as Document currently because of the type mapping).

adam3smith: > For existing Document items without type: article in them, I wonder if they should be migrated to Preprint? That would produce consistent citations (e.g., in my library, I store preprints as Document currently because of the type mapping).

My sense is that most people wouldn't be happy with this. Document is used quite differently.

bwiernik: Fair enough.

adam3smith:
(Linting passed, Travis failure due to https://github.com/zotero/translators/pull/2258 )

OSF Preprints.js
adam3smith: @dstillman I don't know when that changed, but I'm now seeing reports with `type: article` in Extra being auto-converted to `Document` -- that means the above solution for marking preprints doesn't work anymore. Could the previous behavior be re-enabled (or do you have another idea)?

dstillman: @adam3smith: It was [this](https://forums.zotero.org/discussion/comment/369340/#Comment_369340), which you [signed off on](https://forums.zotero.org/discussion/comment/369350/#Comment_369350). The idea here is that an item coming in (from sync or translators) with `Type: <valid CSL type>` should be migrated to the appropriate Zotero type in a future version. This code doesn't bother to exclude the types that already exist now, since those might as well be removed if redundant. But since `article` already exists and we want to use it specifically for a new Preprint item type, we can add an exception.

dstillman: (To be clear, you're only seeing this in the Zotero Beta, right? This was changed since the latest production release.)

PR #2251, opened 2020-09-23
CSL JSON: Default to 'article' if no 'type'

dstillman:
'type' is required in CSL-JSON, but some DOI registration agencies
provide bad data, and this is better than failing. (itemFromCSLJSON()
will already default to 'article' for unknown 'type' values.)

Technically this should go in the DOI Content Negotation translator, but
it's easier to do this here after the JSON has been parsed, and it might
benefit other translators.

This is just for imports from other translators. File/clipboard imports
without 'type' still won't work, because a valid 'type' is required in
detectImport(). (This also can't be tested in Scaffold, which does a
detect first.)

https://forums.zotero.org/discussion/85273/error-importing-dois-via-add-item-by-identifier

### DIFF
--- a/CSL JSON.js
+++ b/CSL JSON.js
@@ -11,8 +11,7 @@
 	},
 	"inRepository": true,
 	"translatorType": 3,
-	"browserSupport": "gcsibv",
-	"lastUpdated": "2019-01-31 00:12:00"
+	"lastUpdated": "2020-09-23 04:40:23"
 }
 
 function parseInput() {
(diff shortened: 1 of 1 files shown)


bwiernik:
This sounds reasonable to me.

adam3smith:
This makes sense, yes, though the issue as reported would remain since Zotero throws an error in the absence of a title, right?

bwiernik:
If that's the case, seems like undesirable behavior--a title-less item isn't unreasonable if the item has a genre or medium field or if the type were personal_communication.

adam3smith:
I think there's no reason not to merge. @dstillman did you want to lint this still?

PR #2253, opened 2020-09-28
Update MODS.js

bwbohl:
add Marc relators cmd (Composer) to roleTerms

### DIFF
--- a/MODS.js
+++ b/MODS.js
@@ -944,15 +944,15 @@ function processExtent(extent, newItem) {
 					if (!rt[i]) continue;
 
 					switch (rt[i].charAt(0).toLowerCase()) {
-					case 'h':
-						hrs = rt[i - 1];
-						break;
-					case 'm':
-						mins = rt[i - 1];
-						break;
-					case 's':
-						secs = rt[i - 1];
-						break;
+						case 'h':
+							hrs = rt[i - 1];
+							break;
+						case 'm':
+							mins = rt[i - 1];
+							break;
+						case 's':
+							secs = rt[i - 1];
+							break;
 					}
 				}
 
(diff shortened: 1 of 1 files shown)


dstillman:
Thanks. For this to be pushed to users, the lastUpdated timestamp needs to be updated as well.

bwbohl:
> Thanks. For this to be pushed to users, the lastUpdated timestamp needs to be updated as well.

yeah just realised, from the travis log thanks!
Apparently indentation needs to be fixed, too…
I'm just wondering how I can test the export locally before committing or merging. Any help appreciated ;-)

dstillman:
Don't worry about the linting (unless it's in lines you changed) other than the timestamp — we can clean it up separately. (You're welcome to, in a separate commit, but it's not necessary.)

Testing locally is a little tricky because export translators aren't currently supported in Scaffold (Tools → Developer → Translator Editor), but you'd want to make this change to MODS.js in the 'translators' directory of your <a href="https://www.zotero.org/support/zotero_data">Zotero data directory</a> and then restart Zotero to test it.

dstillman:
You should actually drop that indentation commit — that changes lots of lines unnecessarily and isn't what the linter is complaining about. (We happily use indents to match surrounding indentation on blank lines.)

bwbohl:
ok I'll fix lines 947–955 indentation, but what about the travis log: "17:20 browserSupport set, but translatorType (3) does not include web (4)"

bwbohl:
just converted this to a draft as I still have to test locally

dstillman:
Yeah, you can remove the browserSupport line in the linter commit. This is an import translator, so that line isn't relevant.

Once you've tested this, can you rebase and squash (or reset and add -p) into two separate commits, one for the change and one for the linter?

PR #2254, opened 2020-09-28
Add Marc relators for soundRecording roles

bwbohl:
* new Relator Codes
  * Composer --> cmp: composer
  * Words By --> lyr: lyricist
  * Performer --> prf: performer
* remove some listing errors
* closes #2253

### DIFF
--- a/MODS.js
+++ b/MODS.js
@@ -14,8 +14,7 @@
 	},
 	"inRepository": true,
 	"translatorType": 3,
-	"browserSupport": "gcsibv",
-	"lastUpdated": "2019-05-20 09:05:38"
+	"lastUpdated": "2020-05-29 00:25:03"
 }
 
 
@@ -953,15 +952,15 @@ function processExtent(extent, newItem) {
 					if (!rt[i]) continue;
 
 					switch (rt[i].charAt(0).toLowerCase()) {
-					case 'h':
-						hrs = rt[i - 1];
-						break;
-					case 'm':
-						mins = rt[i - 1];
-						break;
-					case 's':
-						secs = rt[i - 1];
-						break;
+						case 'h':
+							hrs = rt[i - 1];
+							break;
+						case 'm':
+							mins = rt[i - 1];
+							break;
+						case 's':
+							secs = rt[i - 1];
+							break;
 					}
 				}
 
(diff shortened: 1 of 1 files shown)


dstillman:
Not sure why you opened this as a second PR? It's never necessary to open another PR when you already have one open against the same upstream branch — you can (and should) always rebase, squash, and force-push. It also doesn't make sense to have one PR closing another that you also control — you would have one closing an open issue with no code attached, or superseding another open PR that you can't close yourself.

In any case, if you only want one of these merged, you should close the other one.

Note that you're mixing tabs and spaces in this PR, so that will need to be corrected. There should only be tabs.

dstillman:
Thanks, but you'll need to squash these into just two commits — the fix and the linting changes. If you need help doing that, I can point you in the right direction.

(We can squash into a single commit when merging via GitHub, but these should remain separate, and the alternative is our checking things out and squashing manually. This is why I said the linting changes weren't necessary.)

bwbohl:
ok, sorry, was failing on squashing the commits and thought this'd be the easier way. (Also with closing the other PR from this one). Now again I seem to have failed on squashing commits after correcting the whitespaces to tabs (must have been some editor defaults going against me).
Help appreciated in getting it the way you want

dstillman:
Two ways to do it:

1) `git rebase -i upstream master`, which should show just these four commits, and do an interactive rebase. You can read up on how to do that — it's incredibly useful — but here, basically, you'd want to move the last two commits after the commits they should be squashed with and change `pick` to `s` (for squash) on those lines. (You could use `f` for `fixup`, which squashes and throws away the commit messages, but in this case you'll want to reword the original commits so that they both begin with "MARC: ". When you're done, you'll end up with just two nicely descriptive commits.

2) Look at `git log`, make a note of the last commit on the branch before your changes, `git reset <that hash>` to unstage all your changes, and then use `git add -p` to pick the specific parts for each commit, committing them separately with appropriate descriptions. In a small commit like this where it's trivial to separate the changes and rewrite the commit messages, this way is a bit simpler.

bwbohl:
thanks, for the tutorial!
hope we're there now
closed #2253 manually
wondering what travis-ci is stumbling over

dstillman:
Yeah, Travis doesn't particularly like force-pushes.

Looks good. Can you just reword the second commit to begin with "MARC: ", so it's clear what translator it applies to? (Easy to do with an interactive rebase and `r`.)

bwbohl:
done

dstillman:
Er, right. Merged — thanks!

PR #2257, opened 2020-10-07
Add LingBuzz

cadadr:
This patch adds a translator for LingBuzz. It's pretty simple given the spartan nature of the website. Works on the article details page. The website doesn't have search or a meaningful listing feature (apart from the homepage), so I only implemented saving a single article.

I'm not sure if I'm using the correct field on line 74, but LingBuzz requests that the article number in the form `lingbuzz/\d+` be used when citing LingBuzz articles, and that's the field that looked the most sensible for me. They don't seem to use DOIs.

### DIFF
--- a/LingBuzz.js
+++ b/LingBuzz.js
@@ -67,16 +67,16 @@ function doWeb(doc, url) {
 		});
 	}
 	else {
-		if(url.match(/semanticsArchive/)) {
-			scrapeSA(doc, url);
-		}
-		else {
-			scrape(doc, url);
-		}
+		scrape(doc, url);
 	}
 }
 
-function scrape(doc, _url) {
+function scrape(doc, url) {
+	if (url.match(/semanticsArchive/)) {
+		scrapeSA(doc, url);
+		return;
+	}
+	
 	var newItem = new Zotero.Item("report");
 	newItem.extra = "type: article\n"; // will map to preprint
 
--- a/LingBuzz.js
+++ b/LingBuzz.js
@@ -1,21 +1,21 @@
 {
 	"translatorID": "e048e70e-8fea-43e9-ac8e-940bc3d71b0b",
 	"label": "LingBuzz",
-	"creator": "Göktuğ Kayaalp",
-	"target": "^https://ling.auf.net/lingbuzz/\\d+",
+	"creator": "Göktuğ Kayaalp and Abe Jellinek",
+	"target": "^https://ling\\.auf\\.net/lingbuzz/(\\d+|_search)",
 	"minVersion": "3.0",
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2020-10-07 15:14:37"
+	"lastUpdated": "2021-06-03 21:54:52"
 }
 
 /*
 	***** BEGIN LICENSE BLOCK *****
 
-	Copyright © 2020 Göktuğ Kayaalp <self at gkayaalp dot com>
+	Copyright © 2021 Göktuğ Kayaalp <self at gkayaalp dot com> and Abe Jellinek
 
 	This file is part of Zotero.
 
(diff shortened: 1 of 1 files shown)


cadadr:
After my last force-push linter seems to be happy with the newly added file but the CI build fails because of some timeout, AFAIU as it tries to install Selenium or a related package.

adam3smith:
Yes, you're fine -- you can see the successful lint in line 249 of the Travis detail. 

@adomasven is currently working on CI that automatically runs the translator tests using Chrome drive & Selenium, but the build for that is failing here and hence the Travis failure.

LingBuzz.js
```diff
@@ -0,0 +1,81 @@
+{
+	"translatorID": "e048e70e-8fea-43e9-ac8e-940bc3d71b0b",
+	"label": "LingBuzz",
+	"creator": "Göktuğ Kayaalp",
+	"target": "^https://ling.auf.net/lingbuzz/\\d+",
```
adam3smith, line 5: escape periods in the regex please

LingBuzz.js
```diff
@@ -0,0 +1,81 @@
...
+	var title = idBlock[0];
+	var author = idBlock[1];
+	var date = idBlock[2];
+	var abstract = doc.querySelector(
+		"center"
+	).nextElementSibling.nextSibling.textContent;
+	var table = doc.querySelector("tbody").innerText.split(/\n|\t/);
+	var reference = table[3];
+	var keywords = table[8].split(/[;,] /);
+	var pdfURL = doc.querySelector(
```
adam3smith, line 56: yeah, that's no good. Why not just match on the .pdf in the link, something like `a[href*=".pdf"]`

more generally speaking though, this seems very fragile. I'd instead use xpath where you can use `constains(text(), "string")` as a selector and then just use `following-sibling` to get the value for keyword, format, etc.

LingBuzz.js
```diff
@@ -0,0 +1,81 @@
...
+		"body > table:nth-child(4) > tbody:nth-child(1) > tr:nth-child(1) "
+		+ "> td:nth-child(2) > a:nth-child(1)"
+	).href;
+	// Phew.  That was... disgusting, if I'm polite.
+
+	newItem.title = title;
+	newItem.creators.push(
+		Zotero.Utilities.cleanAuthor(author, "author"));
```
adam3smith, line 64: I don't think this will work for co-authored papers

LingBuzz.js
```diff
@@ -0,0 +1,81 @@
...
+	newItem.abstractNote = abstract;
+	newItem.date = date;
+	newItem.language = "English"; // haven't seen an article in other language.
```
adam3smith, line 67: should be `en` instead, but I'd be inclined to leave this out if we don't know for sure going forward.

LingBuzz.js
```diff
@@ -0,0 +1,81 @@
...
+
+	newItem.attachments = [
+		{ document: doc, title: "Snapshot", mimeType: "text/html" },
```
adam3smith, line 70: no need for `mimeType` when grabbing `doc` directly

LingBuzz.js
```diff
@@ -0,0 +1,81 @@
...
+		{ url: pdfURL, title: "LingBuzz Full Text PDF", mimeType: "application/pdf" }
+	];
+
+	newItem.tags = keywords;
```
adam3smith, line 74: I'm not 100% sure here, but I think this will work better with tags pushed into the tag array rather than replacing it.

LingBuzz.js
```diff
@@ -0,0 +1,81 @@
...
+	newItem.callNumber = reference; // XXX: IDK if this is the apt field.
```
adam3smith, line 75: no, probably not call Number. I don't quite understand how they want this cited given that it's in the URL (which would surely be part of any citation)?

LingBuzz.js
```diff
@@ -0,0 +1,81 @@
...
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+function detectWeb(_doc, _url) {
+	return "journalArticle";
```
adam3smith, line 39: Since these are preprints, I'd save them as `report` and then add `type: article` to the Extra field so that they'll get mapped to preprint once Zotero has that available.

LingBuzz.js
```diff
@@ -0,0 +1,81 @@
...
+
+	newItem.attachments = [
+		{ document: doc, title: "Snapshot", mimeType: "text/html" },
+		{ url: pdfURL, title: "LingBuzz Full Text PDF", mimeType: "application/pdf" }
+	];
+
+	newItem.tags = keywords;
+	newItem.callNumber = reference; // XXX: IDK if this is the apt field.
+
+	newItem.publicationTitle = "LingBuzz";
```
adam3smith, line 77: Once you've switched this to report, this should be publisher rather than publication

adam3smith (commented):
Thanks -- see a bunch of comments in line. We'd also definitely want tests and I'd much prefere if you'd get multiples (in particular search results) to work -- that shouldn't be hard.

cadadr:
On 2020-10-12 06:33 +03, Sebastian Karcher <notifications@github.com> wrote:
> @adam3smith commented on this pull request.
>
> Thanks -- see a bunch of comments in line. We'd also definitely want tests and I'd much prefere if you'd get multiples (in particular search results) to work -- that shouldn't be hard.

You’re welcome! I couldn’t find much guidance on how to do tests and
docs in general, so if there’s some place I can consult I can do add
that stuff (tho my lessons just resumed this week so might take some
time).

I notice today that LingBuzz has a search feature---stupid of me
really.  I’ll update the trivial stuff soon and work on multiples and
tests after that.

Also, it’d be helpful if I can run at least the linter locally.  Would
really appreciate any instructions or links to documents on how to do
that as I’m totally OOTL on how to do JS dev these days with npm and the
like. Couldn’t find a README in this repo and info was limited on the
wiki (or more probably I missed something somewhere).

Find some replies and questions below; I’m in agreement regarding the
stuff that I omitted from the quote.

> yeah, that's no good. Why not just match on the .pdf in the link, something like `a[href*=".pdf"]`
>
> more generally speaking though, this seems very fragile. I'd instead use xpath where you can use `constains(text(), "string")` as a selector and then just use `following-sibling` to get the value for keyword, format, etc.

I’ll look into xpath (hadn’t used it before), but the trouble here is
LingBuzz’s HTML has zero structure and makes _no_ use of any semantic
HTML tags. No ‘h*’, not even ‘p’ in a proper manner.  But if there’s a
way to do it in a more robust manner I’m ready to give it a try. Is
xpath better at handling totally unstructured HTML?

> should be `en` instead, but I'd be inclined to leave this out if we don't know for sure going forward.

I just went over the first few pages and didn’t encounter anything other
than English.  I don’t mind leaving this out if you say so but I can
also look throught the entire thing (it’s less than 6k articles) and
make sure too (tho ofc there’s no reason someone will upload something
non-English in future even if 100% of what’s already available is
English, so there’s that).

> I'm not 100% sure here, but I think this will work better with tags pushed into the tag array rather than replacing it.

I could do that just to be on the safe side.

> no, probably not call Number. I don't quite understand how they want this cited given that it's in the URL (which would surely be part of any citation)?

I’ll mail the maintainer of the site about that, because I don’t really
know either.  But this platform is something of a cross between a
squabbling space for linguist in addition to occasional preprints so it
might be more relevant for the former use.

--
İ. Göktuğ Kayaalp / @cadadr / <https://www.gkayaalp.com/>
pgp:   024C 30DD 597D 142B 49AC 40EB 465C D949 B101 2427

adam3smith:
> Is xpath better at handling totally unstructured HTML?

Not necessarily, but it allows you to match on text, which CSS selectors dont:
`//tr/td[contains(text(), "Format:")]/following-sibling::td/a/@href` e.g. matches the link to the PDF by looking in the sibling to the field that contains "Format:".

Yes, contributor documentation isn't as good as I'd like it to be currently, apologies. The Wikimedia docs are pretty good, especially on using Scaffold and creating tests:
https://www.mediawiki.org/wiki/Citoid/Creating_Zotero_translators#Run_on_Scaffold

For linting: you'll need node/npm installed (current version, 14+ preferred, don't know how far back you can go), but then it's pretty simple:

`npm run lint -- LingBuzz.js` will run the linter and
`npm run lint -- --fix LingBuzz.js` will make all automated fixes (such as spacing, semicolons), which saves a huge amount of time.

AbeJellinek:
Hey, this is great! I'd love to get it merged. Would you be able to rebase on `master` and address @adam3smith's comments?

cadadr:
@AbeJellinek I would love to, this is still on my mind, but I haven't been able to find the free time for a while... I may have the chance in a few weeks, but if someone else wants to complete in the mean time, that's okay with me.

AbeJellinek:
@cadadr Awesome. No worries, I'll check back in in a couple weeks.

AbeJellinek:
@cadadr Just kidding, I went ahead and finished it, just for fun. Most of the changes I made were to switch to `report` and get multiples working, but I also tried to find selectors that might hold up when the site gets its big Y2K redesign. (I actually really like the minimalism, but wow, it's hard to parse in code.) Check it out and let me know what you think.

cadadr:
@AbeJellinek Oh, thanks! I was feeling guilty about this for a while, this is a relief. I'll test this later today and get back to you, and I'll leave a couple inline comments on the patch in a bit.

LingBuzz.js
```diff
@@ -35,47 +35,151 @@
...
 	newItem.title = title;
-	newItem.creators.push(
-		Zotero.Utilities.cleanAuthor(author, "author"));
+	for (let authorLink of authors) {
+		newItem.creators.push(
+			Zotero.Utilities.cleanAuthor(authorLink.innerText, "author"));
+	}
 	newItem.abstractNote = abstract;
-	newItem.date = date;
-	newItem.language = "English"; // haven't seen an article in other language.
```
cadadr: I did encounter rare non-English documents since writing this IIRC, but I guess avoiding setting `.language` is the right call as there's no cue in the page. Just noting in case it's relevant.

AbeJellinek: Yeah, I think it's better to store less data than to store data we aren't really sure about.

LingBuzz.js
```diff
@@ -35,47 +35,151 @@
...
-
-	newItem.attachments = [
-		{ document: doc, title: "Snapshot", mimeType: "text/html" },
-		{ url: pdfURL, title: "LingBuzz Full Text PDF", mimeType: "application/pdf" }
-	];
-
-	newItem.tags = keywords;
-	newItem.callNumber = reference; // XXX: IDK if this is the apt field.
```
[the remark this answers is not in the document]
AbeJellinek: An [archived version](https://web.archive.org/web/20121101011545/https://ling.auf.net/buzzdocs/) of the site's About page mentions that citing the full URL is equivalent to citing the `lingbuzz/*` fragment, so I think we're good.

AbeJellinek:
Don't worry about it!

cadadr:
I discovered, while testing the new patch, that there's a decent subset of articles (about 1/6 of the whole, if I'm not mistaken) that are marked as "semanticsArchive". These can pop up in search results along with other items, so need to be handled. I modified the patch as such, and added tests. It's somewhat nastier than the normal pages, and the last test with `q=semanticsarchive` fails. The other multiple test passes, but when I try to save the articles it may fail. Individual articles don't. I couldn't spot where the error is.

An alternative way to handle these is to ignore this subset of articles of course, which can be done via an url match in `doWeb`:

```js
if(url.match(/semanticsArchive/) { continue; }
```

P.S.: the `first-child` trick in there is needed because the title in these pages does not link to a `current.pdf`; the link is similar in form to the author URLs in the normal pages.

cadadr:
If you think skipping these articles is the better option, then I think reverting my patch and adding the `continue` snippet from above to `doWeb` should suffice, and then I guess this can be merged as such, because when I tested your version of the code with that little modification it works perfectly.

AbeJellinek:
Weird! We should definitely handle those, even though it looks like it's been a while since any have been added.

It was failing because `doWeb` calls `scrape` directly for each search result, so `scrapeSA` would never run for search results. I moved it into `scrape` and everything works now.

AbeJellinek:
Alright, I've gone ahead and merged. Thanks so much for contributing this great work!

cadadr:
@AbeJellinek Thank you so much as well and you're welcome!

Dellu:
Thank you so much. 
This is extremely useful. 

One suggestion (question): is it impossible to put the snapshot into the url fields?
The url fields is currently empty. That  important number which Lingbuzz wants us to put as a citation number  (like DOI) is inside the Snapshot. 
https://take.ms/w5DlS

AbeJellinek:
Oh, yeah, why aren't we setting the URL? That's a mistake. I'll fix it.

AbeJellinek:
@Dellu fixed (77977ea199ac1827704dc4a604cc0ed98d93127a)

Dellu:
@AbeJellinek: 
Thank you so much. This is very useful.

Dellu:
@AbeJellinek  

This is translator is unable to download PDF files in the beta version. 

I was trying Zotero-5.0.97-beta.38 (https://www.zotero.org/support/dev_builds) because it came with a nice pdf reader. Unfortunately, this translator failed to import the pdf files when I try to import. 
- for now, I have downgraded. But, it would be nice if u guys can fix it, as the new build  is pretty wonderful.

dstillman:
@Dellu: Are you saying the release version of Zotero can download PDFs but the beta version can't? There shouldn't be any difference in that regard. Can you provide an example URL where that happens, along with a Debug ID from Zotero for the save attempt?

Dellu:
I cannot provide Debug ID  because I already downgraded my zotero to the release version. 

I was trying to scrap the references of the top 5 articles [here](https://ling.auf.net/lingbuzz/_listing?community=Syntax&start=931). 

I was getting this error message when I click on the Chrome addon: 
`An error occurred saving with LingBuzz. Attempting to save using Save as Webpage instead.`

The reference data was correctly inserted, however. It was only the attachment pdf that was missing. 

Could the issue be with Zotfile?

dstillman:
That just means it didn't save with the LingBuzz translator at all — you won't get PDFs if it's just saving it as a webpage. Nothing to do with ZotFile either.

There shouldn't be any difference with the beta version here, and those articles save fine for me. If you'd like to debug this, please post to the Zotero Forums with a <a href="https://www.zotero.org/support/debug_output#debug_output_logging">Debug ID</a> from the Zotero Connector (not Zotero) for a save attempt that fails.

Dellu:
> those articles save fine for me

Are you using the beta version?

Probably, a change in the database make it harder for the translator to save the file into the database---was thinking along these lines. 
But, thank you. 
I will do so if I upgrade to  the beta again.

dstillman:
Yes, I'm saying it's fine in the beta. Again, the beta should make absolutely no difference here.

dstillman:
> Probably, a change in the database make it harder for the translator to save the file into the database

No, again, this isn't about the file. You were getting an error saving with the translator — you weren't having a problem saving files.

cadadr:
It saves as a single webpage for me as well. I have the connector v5.0.91 on Firefox and 5.0.86 on Ungoogled Chromium, along with the latest beta of Zotero today, 5.0.97-beta.38+f27432347. Tried with the bookmarklet too.

@dstillman If you're trying with the linked page, the regexp for LingBuzz translator shouldn't match its url. This updated regexp matches:

```
^https://ling\.auf\.net/lingbuzz/(repo/semanticsArchive/article/)?(\d+|_search|_listing)
```

`detectWeb` works after that but the other three fail, testing from the Translator Editor. I attached their outputs. Zotero's debug logs don't have anything relevant for me, just some indexing stuff.

[detectImport.txt](https://github.com/zotero/translators/files/7041513/detectImport.txt)
[detectWeb.txt](https://github.com/zotero/translators/files/7041514/detectWeb.txt)
[doImport.txt](https://github.com/zotero/translators/files/7041515/doImport.txt)
[doWeb.txt](https://github.com/zotero/translators/files/7041516/doWeb.txt)
[import-diff.txt](https://github.com/zotero/translators/files/7041518/import-diff.txt)

Summary for above files:

```
==> detectImport.txt <==
21:54:25 Detect using LingBuzz failed:
         TypeError: Function.prototype.apply called on incompatible undefined


==> detectWeb.txt <==
21:52:47 detectWeb returned type "report"

==> doImport.txt <==
21:55:00 Translation using LingBuzz failed:
         TypeError: Function.prototype.apply called on incompatible undefined


==> doWeb.txt <==
21:53:38 Translation using LingBuzz failed:
         TypeError: right.querySelector(...) is null
```

The errors from `*Import` are virtually identical, but there's some little differences, which is why I include a diff.

dstillman:
> @dstillman If you're trying with the linked page, the regexp for LingBuzz translator shouldn't match its url.

I'm not — that page isn't currently supported, so it can't be what @Dellu was been referring to, since it couldn't produce a LingBuzz translator error. I'm assuming @Dellu's report was about saving from the individual article pages linked from there, which, again, work for me.

We're happy to take a patch that adds support for the `_listing` page, but the translator appears to be working fine.

Dellu:
Can you guys map "Published in" field to one of the Zotero's fields?
Currently, the "Published in" field is left out. It would be great if it can be included to one of the standard fields: such as Extra or Note. 
- The Extra field is currently containing "type: article"; which is useless as most of the items are not really articles. Some of them are dissertations; and others are books.

PR #2259, opened 2020-10-11
Pub med journal abbr

adam3smith:
https://forums.zotero.org/discussion/85572/pubmed-translator-missing-journal-abbreviations#latest


adam3smith:
Just to flag for @adomasven : the CI translator tests now work and it's *awesome*
However, the test is marked as passing in Travis even though 2 of the tests fail (in this case because the other translators are broken) -- is that on purpose?

PR #2267, opened 2020-10-15
Add superlib.net translator.

018:


018:
It has been submitted for a long time, please handle it. Thank you very much!

Superlib.js
```diff
@@ -0,0 +1,633 @@
...
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+// eslint-disable-next-line
+function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null;}function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null;}function trim(content){return content.replace(/^[\xA0\s]+/gm, '').replace(/[\xA0\s]+$/gm, '').replace(/\n+/g, '\n').replace(/:\n+/g, ': ').replace(/]\n/g, ']').replace(/】\n/g, '】').replace(/\n\/\n/g, '/')}
```
AbeJellinek, line 39: `attr`/`text` shims are no longer required.

Superlib.js
```diff
@@ -0,0 +1,633 @@
...
+};
+
+function doPerson(item, data) {
+	// （日）本田晃一著；朱运程译 translator author
+	if (!data || data.length <= 0) return;
+	const persons = data.split(/；|;/g);
+	for (var person of persons) {
+		if (person.endsWith('著')) {
+			item.creators.push({
+				lastName: person.trim().replace(/著/g, ''),
```
AbeJellinek, line 61: The regex should be `/著$/`, since you're removing it when it appears at the end.

Superlib.js
```diff
@@ -0,0 +1,633 @@
...
+				creatorType: 'author',
+				fieldMode: 1
+			});
+		} else if (person.endsWith('译')) {
+			item.creators.push({
+				lastName: person.trim().replace(/译/g, ''),
```
AbeJellinek, line 67: same comment here

Superlib.js
```diff
@@ -0,0 +1,633 @@
...
+	图书: 'book',
+	会议论文: 'conferencePaper',
+	学位论文: 'thesis',
+	期刊: 'journalArticle',
+	专利: 'patent'
+};
+
+function doPerson(item, data) {
+	// （日）本田晃一著；朱运程译 translator author
+	if (!data || data.length <= 0) return;
```
AbeJellinek, line 56: This check (and the other ones like it) can, I would think, just be simplified to `if (!data)`.

Superlib.js
```diff
@@ -0,0 +1,633 @@
...
+				creatorType: 'author',
+				fieldMode: 1
+			});
+		}
+	}
+}
+
+function doInventor(item, data) {
+	if (!data || data.length <= 0) return;
+	const persons = data.split(/，/g);
```
AbeJellinek, line 83: Does this need to be a regex?

Superlib.js
```diff
@@ -0,0 +1,633 @@
...
+function doTag(item, data) {
+	if (!data || data.length <= 0) return;
+	const tags = data.split('；');
+	for (var tag of tags) {
+		item.tags.push(tag.trim());
+	}
+}
+
+function sourceType(doc) {
+	var type = doc.querySelector('.on span').textContent;
```
AbeJellinek, line 102: Can just be `text('.on span')`, but not a big deal.

Superlib.js
```diff
@@ -0,0 +1,633 @@
...
+
+		if (checkOnly) return true;
+		
+		let url = a.href;
+
+		// Z.debug(url);
+		let title = ZU.trimInternal(a.textContent);
+		
+		found = true;
+		if (!found['']) {
```
AbeJellinek, line 144: Not sure what's happening here! Indexing into a boolean?

Superlib.js
```diff
@@ -0,0 +1,633 @@
...
+			items[''] = '【提醒：存在失败的可能，请隔几个小时后试试，或单个抓取。】';
```
AbeJellinek, line 145: Don't really understand this either. You can log an error, maybe, but adding it under an empty key in the object doesn't make sense to me.

Superlib.js
```diff
@@ -0,0 +1,633 @@
...
+		if (itemInfo) {
+			var download = row.querySelector('.get a');
+			if (download && download.textContent == 'PDF下载') {
+				itemInfo[url] = {
+					pdfurl: download.href
+				};
+			}
+		}
+		if (title.startsWith('《') && title.endsWith('》')) {
+			title = title.replace(/《|》/g, '');
```
AbeJellinek, line 157: This whole `if` can be simplified to something like

```js
title = title.replace(/^《(.*)》$/, '$1');
```

Superlib.js
```diff
@@ -0,0 +1,633 @@
...
+	}
+	return found ? items : false;
+}
+
+function doWeb(doc, url) {
+	if (detectWeb(doc, url) == "multiple") {
+		var itemInfo = {};
+		Zotero.selectItems(getSearchResults(doc, false, itemInfo), function (items) {
+			if (items) {
+				if (items['']) {
```
AbeJellinek, line 170: same issue as above

Superlib.js
```diff
@@ -0,0 +1,633 @@
...
+	var itemType = detectType(doc);
+	var item = new Zotero.Item(itemType);
+
+	item.url = url;
+	switch (sourceType(doc)) {
+		case '图书':
+			item.title = text(doc, '.tutilte');
+			var infos = text(doc, '.tubox dl');
+			infos = trim(infos);
+			for (var section0 of Object.values(infos.split('\n'))) {
```
AbeJellinek, line 201: I don't think `Object.values` is doing anything here - is it?

Superlib.js
```diff
@@ -0,0 +1,633 @@
...
+				if (!section0 || section0.trim().length <= 0) continue;
```
AbeJellinek, line 202: Can be `if (!section0.trim())`.

Superlib.js
```diff
@@ -0,0 +1,633 @@
...
+		
+				let index = section0.indexOf('】');
+				if (index <= -1) continue;
```
AbeJellinek, line 205: It'll never be _less than_ -1

Superlib.js
```diff
@@ -0,0 +1,633 @@
...
+		
+				let key = section0.substr(0, index + 1).trim();
+				let value = section0.substr(index + 1).trim();
+				// Z.debug(key + ':' + value);
+				switch (key) {
+					case "【作　者】":
+						doPerson(item, value);
+						break;
+					case "【出版项】":
+						var vals = Object.values(value.split(/：|,/g));
```
AbeJellinek, line 215: Again, not sure why we need `Object.values` here.

Superlib.js
```diff
@@ -0,0 +1,633 @@
...
+						item.archiveLocation = value.replace(/\(.*\)/, '');
+						break;
+						
+					default:
+						break;
+				}
+			}
+		
+			// 摘要
+			item.abstractNote = text(doc, '.tu_content').replace(/内容提要:\n*/g, '');
```
AbeJellinek, line 235: Will the regex always match the beginning? If so, should probably add a `^`.

Superlib.js
```diff
@@ -0,0 +1,633 @@
...
+			break;
+		case '期刊':
+			item.title = ZU.trimInternal(text(doc, 'h1.title'));
+			infos = doc.querySelectorAll('#m_top li');
+			for (var section1 of infos) {
+				let content = ZU.trimInternal(section1.textContent);
+				if (!content || content.length <= 0) continue;
```
AbeJellinek, line 242: Can simplify this one like the others, I think

Superlib.js
```diff
@@ -0,0 +1,633 @@
...
+	
+	var itemType = detectType(doc);
+	var item = new Zotero.Item(itemType);
+
+	item.url = url;
+	switch (sourceType(doc)) {
+		case '图书':
+			item.title = text(doc, '.tutilte');
+			var infos = text(doc, '.tubox dl');
+			infos = trim(infos);
```
AbeJellinek, line 200: If you remove the shims from the top of the script, replace `trim(s)` with `ZU.trim(s)`. (`s.trim()` will still work.)

Superlib.js
```diff
@@ -0,0 +1,633 @@
...
+						"creatorType": "author",
+						"fieldMode": 1
+					},
+					{
+						"lastName": "朱运程",
+						"creatorType": "translator",
+						"fieldMode": 1
+					}
+				],
+				"date": "2019.04",
```
AbeJellinek, line 452: Didn't want to note this everywhere you set the date, but these should be normalized with `ZU.strToISO`.

Superlib.js
```diff
@@ -0,0 +1,633 @@
...
+						item.date = value;
+						break;
+					case "【摘 要】":
+						item.abstractNote = value;
+						break;
+					case "【关键词】":
+						doTag(item, value);
+						break;
+					case "【影响因子】":
+						item.extra = '影响因子: ' + value;
```
AbeJellinek, line 270: What gets stored in this field? The unfortunate fact of Extra is that keys can't be localized, but I'll leave it up to others - @dstillman? - whether the key should be in English or Chinese.

Superlib.js
```diff
@@ -0,0 +1,633 @@
...
+						break;
+					default:
```
AbeJellinek, line 272: this isn't needed

Superlib.js
```diff
@@ -0,0 +1,633 @@
...
+							lastName: value,
+							creatorType: 'contributor',
+							fieldMode: 1
+						});
+						break;
+					case "【学位年度】":
+						item.date = value;
+						break;
+					case "【摘 要】":
+						item.abstractNote = value.replace(/隐藏更多/g, '');
```
AbeJellinek, line 308: same note about regex targeting the beginning

Superlib.js
```diff
@@ -0,0 +1,633 @@
...
+						item.conferenceName = value;
+						break;
+					case "【会议录名称】":
+						item.series = value;
+						break;
+					case "【日 期】":
+						item.date = value;
+						break;
+					case "【摘 要】":
+						item.abstractNote = value.replace(/隐藏更多/g, '');
```
AbeJellinek, line 348: ^

Superlib.js
```diff
@@ -0,0 +1,633 @@
...
+						item.filingDate = value;
+						break;
+					case "【地 址】":
+						item.place = value;
+						break;
+					case "【专利类型】":
+						item.extra = '专利类型: ' + value;
+						break;
+					case "【IPC号】":
+						item.extra += '; IPC号: ' + value;
```
AbeJellinek, line 394: use `\n`, not a semicolon

Superlib.js
```diff
@@ -0,0 +1,633 @@
...
+						break;
+					default:
+						break;
+				}
+			}
+			break;
+	}
+	
+	if (item.title.startsWith('《') && item.title.endsWith('》')) {
+		item.title = item.title.replace(/《|》/g, '');
```
AbeJellinek, line 407: same suggestion as above with this replacement (could some of this duplicated processing logic get spun out into a separate function?)

AbeJellinek (changes requested):

AbeJellinek:
OK, requested some changes. Let me know what you think.

Superlib.js
dstillman: Generally we'd want English, since some keys in Extra will be converted to real Zotero values that get added later, and some are even already automatically used in citations (e.g., "Original Date:"). But if we think there's no chance of that for a given key, and we think the audience is going to be entirely Chinese-speaking, Chinese is fine. The translator is already saving all other values in Chinese.

018:
> OK, requested some changes. Let me know what you think.

Thank you very much. I should have written it more carefully. Now the modification has been completed. Please review it again.

PR #2268, opened 2020-10-15
Fix bugs caused by website upgrades

018:


CNKI.js
```diff
@@ -136,6 +137,11 @@ function getItemsFromSearchResults(doc, url, itemInfo) {
 	}
 
 	if (!links.length) {
+		// 018<lyb018@gmail.com>: 20200909
```
AbeJellinek, line 140: No need to leave a signature (this duplicates `git blame`!). Applies for similar comments below.

CNKI.js
```diff
@@ -217,9 +228,8 @@ function scrape(ids, doc, url, itemInfo) {
 					creator.lastName = creator.lastName.substr(lastSpace + 1);
 				}
 				else {
-					// Chinese name. first character is last name, the rest are first name
-					creator.firstName = creator.lastName.substr(1);
-					creator.lastName = creator.lastName.charAt(0);
+					// 018<lyb018@gmail.com>: 20200909
+					creator.firstName = '';
```
AbeJellinek, line 232: Why this change?

CNKI.js
```diff
@@ -252,6 +262,26 @@ function scrape(ids, doc, url, itemInfo) {
...
+					e = e.parentElement;
+				}
+				if (e) {
+					e = e.parentElement;
+				}
+				if (e) {
+					e = e.querySelector('.quote a');
+				}
+				if (e) {
+					num  = e
```
AbeJellinek, line 279: This is a bit of a DOM roller-coaster ride! Is there a more reliable way to get to the element you want? It just feels likely to break quickly.

CNKI.js
```diff
@@ -252,6 +262,26 @@ function scrape(ids, doc, url, itemInfo) {
...
+				}
+			}
+			if (num) {
+				newItem.extra = num.textContent.replace(/[^\d*]/g, '');
```
AbeJellinek, line 283: Can you add a label before this? Items in `Extra` are formatted as `Key: Value`.

AbeJellinek (changes requested):

CNKI.js
018: OK, I'll fix it.

CNKI.js
018: Some people reported that Chinese names are divided into first name and last name, and only the last name is displayed in the creator column of Zotero's list. It's not easy to tell.

CNKI.js
018: I have fixed it and resubmitted it. Please review it again. Thank you.

CNKI.js
018: I have fixed it and resubmitted it. Please review it again. Thank you.

CNKI.js
AbeJellinek: >only the last name is displayed in the creator column of Zotero's list.

That's just how Zotero displays creator names in the list, no matter the language. I could see the case for an optional `Creator (full name)` column in future releases of the desktop app, but I don't think it's something we can or should address in a single translator. [See the approach we take for Baidu Scholar.](https://github.com/zotero/translators/blob/d9f2915f78f9f8c5c51c9e88bbeee94f08ad3aa5/Baidu%20Scholar.js#L133)

CNKI.js
```diff
@@ -262,25 +258,10 @@ function scrape(ids, doc, url, itemInfo) {
...
-				}
-				if (e) {
-					num  = e
-				}
-			}
-			if (num) {
-				newItem.extra = num.textContent.replace(/[^\d*]/g, '');
+			var numEle = doc.querySelector('#func3 .num');
+			if (numEle) {
+				Z.debug(numEle.textContent.replace(/[^\d*]/g, ''))
```
AbeJellinek, line 263: I don't think the `Z.debug` line needs to be here, but otherwise this looks much better, thanks!

AbeJellinek (changes requested):

CNKI.js
```diff
@@ -262,25 +258,10 @@ function scrape(ids, doc, url, itemInfo) {
 			//	newItem.extra = 'CN ' + newItem.callNumber;
 				newItem.callNumber = "";
```
AbeJellinek, line 259: Here you remove the call number...

CNKI.js
```diff
@@ -262,25 +258,10 @@ function scrape(ids, doc, url, itemInfo) {
...
-				if (e) {
-					num  = e
-				}
-			}
-			if (num) {
-				newItem.extra = num.textContent.replace(/[^\d*]/g, '');
+			var numEle = doc.querySelector('#func3 .num');
+			if (numEle) {
+				Z.debug(numEle.textContent.replace(/[^\d*]/g, ''))
+				newItem.callNumber = numEle.textContent.replace(/[^\d*]/g, '');
```
AbeJellinek, line 264: And here you add a different one? I'm not quite sure what this number is and if it should take precedence over whatever the RefWorks translator returns!

AbeJellinek (changes requested):

018:
> @018 any progress on this?

Recently, I have been a little busy. I haven't received your reply yet. I will spare some time to repair it. thank you!

018:
The code has been restored and fixed as recommended. thank you!

PR #2270, opened 2020-10-15
Fix translator test check exiting with 0 despite failing on Travis

### LINKED ISSUE #2269: Failing translator tests don't result in failed CI run
https://travis-ci.org/github/zotero/translators/builds/735994172

@adomasven: Is this not supposed to show as failed?

adomasven:
Closes #2269


adam3smith:
We'll continue to have tests that fail, e.g. because of paywalls. We're OK with still including them in the translators, right? They're super useful to have for manual testing.

dstillman:
Yes, we should definitely still include paywalled tests. We could add a flag to mark those as optional, but that might be more trouble than it's worth — not a big deal to look at the CI output when there's a failure.

PR #2272, opened 2020-10-21
Add Blacklight Discovery Layer

adam3smith:
Currently restricting this to a subset of libraries via target. Blacklight allows for a lot of customization, some places don't have alternate metadata links, other only have basic DC metadata in there.


adam3smith:
@adomasven -- Selenium tests are not running/failing to build again

PR #2273, opened 2020-10-21
Fix chromedriver detection in CI

adomasven:
As noted by @adam3smith in https://github.com/zotero/translators/pull/2272#issuecomment-713261312 it's stopped working in CI.


adam3smith:
@adomasven sorry, still not working I think. See
https://github.com/zotero/translators/pull/2274 which was started after dstillman merged this PR

PR #2274, opened 2020-10-21
Updates to Scholars Portal Journals

bkawula:
This PR addresses new interface changes and issues caused by proxied URLs.

### DIFF
--- a/Scholars Portal Journals.js
+++ b/Scholars Portal Journals.js
@@ -1,5 +1,5 @@
 {
-	"translatorID": "0a8fec76-6ddf-4574-b2a8-bcef8a7f0a4c",
+	"translatorID": "5ac0fd37-5578-4f82-8340-0e135b6336ee",
 	"label": "Scholars Portal Journals",
 	"creator": "Bartek Kawula",
 	"target": "^https?://journals\\.scholarsportal\\.info/",
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2020-10-21 22:57:41"
+	"lastUpdated": "2020-10-22 18:59:55"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


Scholars Portal Journals.js
```diff
@@ -1,21 +1,21 @@
 {
-	"translatorID": "5ac0fd37-5578-4f82-8340-0e135b6336ee",
+	"translatorID": "0a8fec76-6ddf-4574-b2a8-bcef8a7f0a4c",
```
adam3smith, line 2: don't change the id -- this would leave people with both translators

Scholars Portal Journals.js
```diff
@@ -35,23 +35,26 @@
...
 
-function detectWeb(doc, url) { 
-	var myArticles = doc.getElementById('my-articles');
-	if (myArticles) {
-  		Zotero.monitorDOMChanges(myArticles.parentElement, {attributes: true, attributeFilter: ["style"]});
-	}
-	// see if saved list is toggled open
-	if (doc.getElementsByClassName('inner-wrap-open') [0]) {
+function detectWeb(doc, url) {
+	if (url.indexOf("/my-articles") != -1) {
```
adam3smith, line 39: We prefer -- and now strictly enforce -- `string.includes()` for testing literal matches -- please fix throughout

Scholars Portal Journals.js
```diff
@@ -75,18 +78,24 @@ function doWeb(doc, url) {
 }
 
 function getURI(url){
-	if (url.indexOf("xml") != -1){
+	if (url.indexOf("/details/") != -1){
 		var a = url.indexOf("details");
```
adam3smith, line 82: you'll want to use `let` rather than `var` for these, both as a matter of clean code and to not upset the linter.

Scholars Portal Journals.js
```diff
@@ -159,7 +167,7 @@ var testCases = [
...
...ere investigated by the X-ray scattering method. Comparison of the X-ray scattering data revealed that only one kind of intermolecular water-organic molecule interaction is commonly involved throughout all mole fractions of these liquid mixtures, in addition to the intermolecular water-water and organic molecule-organic molecule interactions, which are present in neat water and organic liquids, respectively. On the basis of this finding, we proposed a new analytical method for studying liquid mixtures. By this method the structural information on the intermolecular water-organic molecule interaction as well as the concentrations of the intermolecular water-water, water-organic molecules, and organic molecule-organic molecule interactions were obtained. Combining the concentrations of the intermolecular water-water interaction with the concentrations of water in the liquid mixtures, the number of water-water hydrogen bonds at various mole fractions was experimentally determined for the first time. From the dependence of the number of water-water hydrogen bonds on the composition of the liquid mixtures, the change of the size of the self-associated water-water clusters was deduced.",
```
adam3smith, line 170: Any way we can strip the "Abstract" at the start?

adam3smith (commented):
Thanks -- some small things that would be good to fix, also it'd be great if you could fix linting as well, shouldn't be too complicated.

adam3smith:
Thanks -- sorry you did all the linting fixes manually, I didn't realize you wouldn't use `lint`. (Ran the tests locally since they're not running in CI)

PR #2278, opened 2020-10-28
Add Reddit translator WIP

lukaskawerau:
This is a very WIP version of a Reddit translator.
I followed [this guide at mediawiki.org](https://www.mediawiki.org/wiki/Citoid/Creating_Zotero_translators), so I don't have deep understanding of what's necessary and what isn't.

Some issues:

- [adding the author/creator of a reddit post](https://github.com/zotero/translators/compare/master...lukaskawerau:master#diff-5c1a4e80bf64eef22e7ff5583c68750d4fd2fb525210a5c25e79d3e495bed711R79) adds their username as a last name, even though it should arguably be a single field
- I'm not entirely sure how well this works for comments instead of top-level posts
- no multiple selection: it's currently not possible to go to a subreddit and select multiple posts, or to a post and select multiple comments

### DIFF
--- a/reddit.js
+++ b/reddit.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2020-10-28 15:28:00"
+	"lastUpdated": "2020-10-28 15:35:26"
 }
 
 /*
@@ -107,7 +107,8 @@ function scrape(doc, url) {
 		Zotero.done();
 	}, function() {});
 
-}/** BEGIN TEST CASES **/
+}
+/** BEGIN TEST CASES **/
 var testCases = [
 	{
 		"type": "web",
@@ -116,16 +117,12 @@ var testCases = [
 			{
 				"itemType": "forumPost",
 				"creators": [
-					"0": {
+					{
 						"firstName": "",
 						"lastName": "SpikePlayz",
 						"creatorType": "author"
 					}
 				],
-				"notes": [],
-            	"tags": [],
-            	"seeAlso": [],
-        	    "attachments": [],
 				"title": "SpaceX Partners with LeoLabs to Track Starlink Satellites",
 				"url": "www.reddit.com/r/spacex/comments/jj7arf/spacex_partners_with_leolabs_to_track_starlink/",
 				"date": "2020-10-27T18:54:37.000Z",
--- a/reddit.js
+++ b/reddit.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2020-10-29 08:25:52"
+	"lastUpdated": "2020-10-29 08:53:37"
 }
 
 /*
@@ -62,7 +62,7 @@ function getSearchResults(doc, checkOnly) {
 }
 
 function doWeb(doc, url) {
-	var json_url = url + '.json';
+	var json_url = url.split("?")[0] + '.json';
 	if (detectWeb(doc, url) == "multiple") {
 		Zotero.selectItems(getSearchResults(doc, false), function (items) {
 			Zotero.debug(items);
@@ -96,7 +96,7 @@ function scrape(text) {
 	newItem.forumTitle = 'r/'+reddit_data["subreddit"];
 	newItem.websiteTitle = "reddit.com";
 	newItem.attachments.push({
-    	url : 'https://www.reddit.com' + reddit_data["permalink"],
+		url : 'https://www.reddit.com' + reddit_data["permalink"],
 		title : "Reddit Post Snapshot",
 		type : "text/html"
 	});
@@ -118,7 +118,7 @@ function scrape_comment(text) {
 	newItem.websiteTitle = "reddit.com";
 	newItem.extra = 'Post URL: www.reddit.com' + parent_data["permalink"]; 
 	newItem.attachments.push({
-    	url : 'https://www.reddit.com' + reddit_data["permalink"],
+		url : 'https://www.reddit.com' + reddit_data["permalink"],
 		title : "Reddit Comment Snapshot",
 		type : "text/html"
 	});
--- a/reddit.js
+++ b/reddit.js
@@ -9,24 +9,24 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2020-10-29 14:43:51"
+	"lastUpdated": "2020-10-29 16:05:25"
 }
 
 /*
-***** BEGIN LICENSE BLOCK *****
-Reddit Translator
-Copyright (C) 2020 Lukas Kawerau, lukas@kawerau.org
-This program is free software: you can redistribute it and/or modify
-it under the terms of the GNU General Public License as published by
-the Free Software Foundation, either version 3 of the License, or
-(at your option) any later version.
-This program is distributed in the hope that it will be useful,
-but WITHOUT ANY WARRANTY; without even the implied warranty of
-MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-GNU General Public License for more details.
-You should have received a copy of the GNU General Public License
-along with this program.  If not, see <http://www.gnu.org/licenses/>.
-   ***** END LICENSE BLOCK *****
+	***** BEGIN LICENSE BLOCK *****
+	Copyright © 2020 Lukas Kawerau
+	This file is part of Zotero.
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+	GNU Affero General Public License for more details.
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+	***** END LICENSE BLOCK *****
 */
 
 
@@ -35,7 +35,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null}function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null}
 
 function detectWeb(doc, url) {
-	var regex = new RegExp("\/r\/[a-z\d]+\/comments\/");
+	var regex = new RegExp("\\/r\\/[a-z\\d]+\\/comments\\/");
 	if (url.includes('search') && getSearchResults(doc, true)) {
 		return 'multiple';
 	} 
@@ -64,10 +64,9 @@ function getSearchResults(doc, checkOnly) {
 
 function doWeb(doc, url) {
 	var jsonUrl = url.split("?")[0] + '.json';
-	var commentRegex = new RegExp("\/r\/[a-z\d]+\/comments\/[a-z\d]+\/[a-z\d_]+\/[a-z\d]+\/");
+	var commentRegex = new RegExp("\\/r\\/[a-z\\d]+\\/comments\\/[a-z\\d]+\\/[a-z\\d_]+\\/[a-z\\d]+\\/");
 	if (detectWeb(doc, url) == "multiple") {
 		Zotero.selectItems(getSearchResults(doc, false), function (items) {
-			Zotero.debug(items);
 			if (!items) {
 				return true;
 			}
--- a/reddit.js
+++ b/reddit.js
@@ -2,14 +2,14 @@
 	"translatorID": "23bacc11-98e3-4b78-b1ef-cc2c9a04b893",
 	"label": "reddit",
 	"creator": "Lukas Kawerau",
-	"target": "^https?://(www\\.)?reddit.com",
+	"target": "^https?://www\\.reddit\\.com",
 	"minVersion": "3.0",
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2020-10-28 13:27:20"
+	"lastUpdated": "2020-10-28 15:28:00"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


adam3smith:
Could you add some tests? (using Zotero's Translator Editor)

lukaskawerau:
Yes, working on that.  
The test I added fails, though - but I can't figure out why.
It says it's because of `unknown (Item 0 does not match)`.

adam3smith:
You're seeing the automated test creation ability in the Translator Editor, right? You really shouldn't be doing any of this by hand -- it's just clicking "New Web" "Run" and "Save" in the Testing panel

lukaskawerau:
Aah, that makes sense! Done.

reddit.js
```diff
@@ -0,0 +1,139 @@
+{
+	"translatorID": "23bacc11-98e3-4b78-b1ef-cc2c9a04b893",
+	"label": "reddit",
+	"creator": "Lukas Kawerau",
+	"target": "^https?://www\\.reddit\\.com",
```
adam3smith, line 5: Could you make this stricter? I think everything you're looking for is either in `r/comments` or `/search` isn't it?

reddit.js
```diff
@@ -0,0 +1,139 @@
...
+
+function attr(docOrElem, selector, attr, index) {
+	var elem = index ? docOrElem.querySelectorAll(selector).item(index) : docOrElem.querySelector(selector);
+	return elem ? elem.getAttribute(attr) : null;
+}
+
+function text(docOrElem, selector, index) {
+	var elem = index ? docOrElem.querySelectorAll(selector).item(index) : docOrElem.querySelector(selector);
+	return elem ? elem.textContent : null;
+}
```
adam3smith, line 41: Please use the minimized version preceded by eslint diable, see e.g. here: https://github.com/zotero/translators/blob/master/Library%20Catalog%20(Blacklight).js#L39

reddit.js
```diff
@@ -0,0 +1,139 @@
...
+
+function detectWeb(doc, url) {
+	// Adjust the inspection of url as required
+	if (url.indexOf('search') != -1 && getSearchResults(doc, true)) {
```
adam3smith, line 45:
```suggestion
	if (url.includes('search') && getSearchResults(doc, true)) {
```

(always use `includes` for literal matches)

reddit.js
```diff
@@ -0,0 +1,139 @@
...
+		return 'multiple';
+	}
+	// Adjust the inspection of url as required
```
adam3smith, line 48: remove comments (here and throughout)

reddit.js
```diff
@@ -0,0 +1,139 @@
...
+	else {
+		return 'forumPost';
```
adam3smith, line 50: you'll likely want to restrict this further e.g. by checking for a specific CSS selector

reddit.js
```diff
@@ -0,0 +1,139 @@
...
+	}
+	// Add other cases if needed
+}
+
+function getSearchResults(doc, checkOnly) {
+	var items = {};
+	var found = false;
+	// Adjust the CSS Selectors
+	var rows = doc.querySelectorAll('.mw-search-result-heading a');
```
adam3smith, line 59: I think you should be able to make search results work, no?

reddit.js
```diff
@@ -0,0 +1,139 @@
...
+	for (var i=0; i<rows.length; i++) {
```
adam3smith, line 60: better: `for (row of rows) {` (and adjust the rest accordingly. Note that Zotero's translator actually has updated templates built-in that you can use. The ones on Wikimedia are fine but outdated in some minor ways

reddit.js
```diff
@@ -0,0 +1,139 @@
...
+		newItem.url = 'www.reddit.com' + reddit_data[0]["data"]["children"][0]["data"]["permalink"];
+		var post_date = new Date(reddit_data[0]["data"]["children"][0]["data"]["created_utc"]*1000);
+		newItem.date = post_date.toISOString();
+		// Zotero.debug(reddit_data);
+		newItem.postType = "Reddit Post";
+		newItem.forumTitle = 'r/'+reddit_data[0]["data"]["children"][0]["data"]["subreddit"];
+		newItem.websiteTitle = "reddit.com";
+		newItem.complete();
+		//Zotero.debug(newItem);
+		Zotero.done();
```
adam3smith, line 107:
```suggestion
```
(That's no longer needed/used)

reddit.js
```diff
@@ -0,0 +1,139 @@
...
+	}
+}
+
+
+function scrape(doc, url) {
+	var json_url = url + '.json'
+	Zotero.Utilities.HTTP.doGet(json_url, function(text) {
+		var newItem = new Zotero.Item("forumPost");
+		var reddit_data = JSON.parse(text);
+		newItem.title = reddit_data[0]["data"]["children"][0]["data"]["title"];
```
adam3smith, line 96: I'd recommend assigning `reddit_data[0]["data"]["children"][0]["data"]` to a new variable and then just use that. Much easier to read (and shorter)

reddit.js
```diff
@@ -0,0 +1,139 @@
...
+		newItem.creators.push(ZU.cleanAuthor(reddit_data[0]["data"]["children"][0]["data"]["author"], "author", false));
```
adam3smith, line 97: Pretty sure if you set `hasComma` to true here, this will automatically import usernames as single field. Otherwise, you can create the creators object by specifying the elements individually

reddit.js
```diff
@@ -0,0 +1,139 @@
...
+		});
+	} else {
+		scrape(doc, url);
+	}
+}
+
+
+function scrape(doc, url) {
+	var json_url = url + '.json'
+	Zotero.Utilities.HTTP.doGet(json_url, function(text) {
```
adam3smith, line 93: Use the equivalent but updated `ZU.doGet(url, function())`

reddit.js
```diff
@@ -0,0 +1,139 @@
...
+		var post_date = new Date(reddit_data[0]["data"]["children"][0]["data"]["created_utc"]*1000);
+		newItem.date = post_date.toISOString();
+		// Zotero.debug(reddit_data);
+		newItem.postType = "Reddit Post";
+		newItem.forumTitle = 'r/'+reddit_data[0]["data"]["children"][0]["data"]["subreddit"];
+		newItem.websiteTitle = "reddit.com";
+		newItem.complete();
+		//Zotero.debug(newItem);
+		Zotero.done();
+	}, function() {});
```
adam3smith, line 108:
```suggestion
	});
```

If there's no on done function, no need to specify an empty one

reddit.js
```diff
@@ -0,0 +1,139 @@
...
+	if (detectWeb(doc, url) == "multiple") {
+		Zotero.selectItems(getSearchResults(doc, false), function (items) {
+			if (!items) {
+				return true;
+			}
+			var articles = [];
+			for (var i in items) {
+				articles.push(i);
+			}
+			ZU.processDocuments(articles, scrape);
```
adam3smith, line 83: If you're going to make multiples work, then rather than calling the reddit pages, Id go straight to the JSON from here, i.e. add the JSON url in `getSearchResults` and then call `ZU.doGet(articles, scrape)` here and move the call to the JSON out of the scrape function.

reddit.js
```diff
@@ -0,0 +1,139 @@
...
+
+
+function scrape(doc, url) {
+	var json_url = url + '.json'
+	Zotero.Utilities.HTTP.doGet(json_url, function(text) {
+		var newItem = new Zotero.Item("forumPost");
+		var reddit_data = JSON.parse(text);
+		newItem.title = reddit_data[0]["data"]["children"][0]["data"]["title"];
+		newItem.creators.push(ZU.cleanAuthor(reddit_data[0]["data"]["children"][0]["data"]["author"], "author", false));
+		newItem.url = 'www.reddit.com' + reddit_data[0]["data"]["children"][0]["data"]["permalink"];
```
adam3smith, line 98: we'd also want a Snapshot (using the permalink)

adam3smith (changes requested):
Thanks for submitting this! Some comments in line. Let me know if anything isn't clear

reddit.js
dstillman: \* `for (let row of rows) {`

reddit.js
lukaskawerau: I've left this as is for now, because this way we can select multiple posts from the main feed on reddit.com

reddit.js
lukaskawerau: Great suggestions, thank you!

reddit.js
```diff
@@ -0,0 +1,204 @@
...
+	} else {
+		ZU.doGet(json_url, scrape);
+	}
+}
+
+
+function scrape(text) {
+	var newItem = new Zotero.Item("forumPost");
+	var reddit_json = JSON.parse(text);
+	var reddit_data = reddit_json[0]["data"]["children"][0]["data"];
```
adam3smith, line 89: also as per styleguide (and general JS conventions) use dot notation (`redditJSON[0].data.children[0].data`) here and below

reddit.js
```diff
@@ -0,0 +1,204 @@
...
+	newItem.url = 'www.reddit.com' + reddit_data["permalink"];
+	var post_date = new Date(reddit_data["created_utc"]*1000);
+	newItem.date = post_date.toISOString();
+	newItem.postType = "Reddit Post";
+	newItem.forumTitle = 'r/'+reddit_data["subreddit"];
+	newItem.websiteTitle = "reddit.com";
+	newItem.attachments.push({
+		url : 'https://www.reddit.com' + reddit_data["permalink"],
+		title : "Reddit Post Snapshot",
+		type : "text/html"
```
adam3smith, line 101:
```suggestion
		mimeType : "text/html"
```

ditto below

adam3smith (changes requested):
This is coming together nicely! If you don't have eslint installed locally (and unless you're motivated to do so) don't start cleaning up linting manually. Most of this can be done automatically (`npm run lint -- --fix reddit.js` if you're curious) and I'm happy to run this for you if you're otherwise ready to go.

reddit.js
```diff
@@ -0,0 +1,204 @@
...
+		return 'forumPost';
+	}  else {
+		return 'multiple';
+	}
+}
+
+function getSearchResults(doc, checkOnly) {
+	var items = {};
+	var found = false;
+	var rows = doc.querySelectorAll('.SQnoC3ObvgnGjWt90zD9Z._2INHSNB8V5eaWp4P0rY_mE');
```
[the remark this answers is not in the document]
lukaskawerau: Yes - they seem to be stable across searches and browsers. I've tested across Chrome and Safari, including private windows and different searches on different subreddits - always stable.

reddit.js
```diff
@@ -0,0 +1,204 @@
...
+			if (!items) {
+				return true;
+			}
+			var articles = [];
+			for (var i in items) {
+				articles.push(i);
+			}
+			ZU.doGet(articles, scrape);
+		});
+	} else if (detectWeb(doc, url) == "forumPost" && url.match(/\/r\/[a-z\d]+\/comments\/[a-z\d]{6}\/[a-z\d_]+\/[a-z\d]{7}\//g)) {
```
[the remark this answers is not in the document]
lukaskawerau: You are right, that was too restrictive - I've generalized this.

reddit.js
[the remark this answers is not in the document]
dstillman: They still might be generated at build time, meaning a site update could break them. I haven't looked at the page to see what else is available, but something less random would certainly be preferable.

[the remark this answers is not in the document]
lukaskawerau: Hm, I see. I'm not skilled enough to find a less random identifier, so if this is blocking the PR I'd appreciate any help in finding a more suitable way of going about this.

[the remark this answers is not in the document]
adam3smith: I think this would only work with xpaths since you need a parent selector. `h3` is the only clearly distinguishing feature (no useful attributes anywhere) and the link is in `a` two levels up. I'd have to test but something like
`var rows = ZU.xpath(doc, '//a[div/h3]`); looks like it should work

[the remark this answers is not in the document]
lukaskawerau: Seems to work!

adam3smith:
Great! Here are the linting errors that `lint --fix` doesn't fix automatically. Happy to fix them for you, but figured you'd want the satisfaction of finalizing this yourself. Let me know if any of them are unclear (ignore the znc at the end of the line)

15:1   warning  Copyright preferred to be GNU Affero General Public License  z
otero-translator/license
 38:26  error    Unnecessary escape character: \\/                             n
o-useless-escape
 38:29  error    Unnecessary escape character: \\/                             n
o-useless-escape
 38:35  error    Unnecessary escape character: \d                             n
o-useless-escape
 38:39  error    Unnecessary escape character: \\/                             n
o-useless-escape
 38:49  error    Unnecessary escape character: \\/                             n
o-useless-escape
 68:33  error    Unnecessary escape character: \\/                             n
o-useless-escape
 68:36  error    Unnecessary escape character: \\/                             n
o-useless-escape
 68:42  error    Unnecessary escape character: \d                             n
o-useless-escape
 68:46  error    Unnecessary escape character: \\/                             n
o-useless-escape
 68:56  error    Unnecessary escape character: \\/                             n
o-useless-escape
 68:62  error    Unnecessary escape character: \d                             n
o-useless-escape
 68:66  error    Unnecessary escape character: \\/                             n
o-useless-escape
 68:72  error    Unnecessary escape character: \d                             n
o-useless-escape
 68:77  error    Unnecessary escape character: \\/                             n
o-useless-escape
 68:83  error    Unnecessary escape character: \d                             n
o-useless-escape
 68:87  error    Unnecessary escape character: \\/                             n
o-useless-escape
 70:52  error    Expected to return a value at the end of function            c
onsistent-return

lukaskawerau:
Thanks!

I'm confused by the `38:35 error Unnecessary escape character: \d` - does `new RegExp("/r/[a-zd]+/comments/");` really do what `new RegExp("/r/[a-z\d]+/comments/");` does? I'm not that firm with RegEx, so this was confusing to read.

adam3smith:
No, it's a bit misleading: you actually need double backslashes in javascript regular expressions -- rule of thumb is if you put a regular expression in "" you need double backslashes, if it's between // use single. So it should be `\\d` here. On the other hand, because / isn't a delimiter, you don't need to escape that.

lukaskawerau:
Aaah, okay, that makes much more sense! 😁

reddit.js
```diff
@@ -0,0 +1,205 @@
...
+	return found ? items : false;
+}
+
+function doWeb(doc, url) {
+	var jsonUrl = url.split("?")[0] + '.json';
+	var commentRegex = new RegExp("\\/r\\/[a-z\\d]+\\/comments\\/[a-z\\d]+\\/[a-z\\d_]+\\/[a-z\\d]+\\/");
+	if (detectWeb(doc, url) == "multiple") {
+		Zotero.selectItems(getSearchResults(doc, false), function (items) {
+			if (!items) {
+				return true;
```
adam3smith, line 71:
```suggestion
				return;
```

for consistent return

adam3smith:
Awesome, thanks! (The tests are still failing because of a problem with the automated test suite, so no actual problem there)

PR #2281, opened 2020-11-12
YouTube: Fix authors & date

adam3smith:
https://forums.zotero.org/discussion/comment/368044#Comment_368044
https://forums.zotero.org/discussion/86077/pulling-channel-name-from-youtube-videos/p1

I'm disabling multiples which are currently broken and would require a massive amount of work to get right (page isn't loaded fully before scrape triggers, so this would require rebuilding this from unofficial API calls)

### DIFF
--- a/YouTube.js
+++ b/YouTube.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsbv",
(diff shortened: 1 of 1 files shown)


adam3smith:
@dstillman let me know if you have any objections against disabling multiples -- I don't have time to work on it,seems like it'd take some time & it's currently broken

dstillman:
Fine with me

PR #2286, opened 2020-11-22
new translator for the Computer History Museum Collections

boan-anbo:
https://www.computerhistory.org/collections


Computer History Museum Archive.js
```diff
@@ -0,0 +1,589 @@
...
+		"items": [
+			{
+				"itemType": "book",
+				"title": "Knuth, Donald oral history",
+				"creators": [
+					{
+						"firstName": "Edward",
+						"lastName": "Feigenbaum",
+						"creatorType": "contributor",
+						"fieldMode": true
```
dstillman, line 247: How are you creating these tests? `fieldMode: true` is not correct and shouldn't be created by any automated tools. Are you using the Translator Editor (Scaffold) to create these?

Computer History Museum Archive.js
```diff
@@ -0,0 +1,589 @@
...
+	},
+	{
+		"type": "web",
+		"url": "https://www.computerhistory.org/collections/catalog/102765924",
+		"items": [
+			{
+				"itemType": "book",
+				"title": "Jacquard programming cards",
+				"creators": [],
+				"abstractNote": "Category: I/O/punched card device\nDescription: The object is a series of cardboard punch cards connected by string. The cards are hand numbered sequentially from 6 to 44.\nType: Physical Object",
```
dstillman, line 391: Some of this looks like it should be in Extra.

Computer History Museum Archive.js
```diff
@@ -0,0 +1,589 @@
...
+						"fieldMode": true
+					}
+				],
+				"date": "2007-03-14; 2007-03-21",
+				"abstractNote": "Collection Title: Oral history collection\nSubject: Fellow Award Honoree; Knuth, Donald; IBM 650 (Computer); Combinatorial analysis--Data processing; Combinatorics; Analysis of algorithms; The Art of Computer Programming; Stanford University; TeX; METAFONT; Religion; Literate programming\nCategory: Transcription\nDescription: In this wide-ranging interview conducted by Edward Feigenbaum, Donald Knuth talks about the progression of his life and career.  Topics include his family background and early interest in music, physics and mathematics, his first exposure to programming,  finding a mentor, and writing a doctoral thesis.   He describes how \"The Art of Computer Programming\" became \"the story of my life\", and why it was put on hold for the TeX and METAFONT projects.  He also talks about personal work habits, programming style, analysis of algorithms, the influence of religion in his life, and his advice to the next generation of scientists.\nType: Document",
+				"archive": "Computer History Museum",
+				"archiveLocation": "X3926.2007",
+				"libraryCatalog": "102658053",
```
dstillman, line 266: What is this number? This shouldn't be in `libraryCatalog`.

Computer History Museum Archive.js
```diff
@@ -0,0 +1,589 @@
...
+		"items": [
+			{
+				"itemType": "book",
+				"title": "A guide to Fortran IV programming",
+				"creators": [],
+				"date": "1972",
+				"abstractNote": "Category: Book\nDescription: Second edition.  Signed by McCracken on title page.\nType: Document",
+				"archive": "Computer History Museum",
+				"archiveLocation": "X3682.2007",
+				"extra": "Identifying Numbers: ISBN10: 0471582816\nLCCN\n72-4745\nLOC call num\nQA76.73.F25 M3 1972\nDimensions: 28 cm.",
```
dstillman, line 325: - We should parse out the things we understand. E.g., the ISBN should go in the ISBN field, and we have a Call Number field.
- Extra values should generally be `key: val`, so having "Identifying Numbers:" and then additional lines isn't ideal. It would be better to parse this if possible and create separate `key: val` lines.

Computer History Museum Archive.js
```diff
@@ -0,0 +1,589 @@
...
+		const spans = personDiv.querySelectorAll('span')
+		const name = spans[0].textContent
+		const [lastName, firstName] = name.split(', ')
+		Zotero.debug(lastName)
+		Zotero.debug(firstName)
+		contributors.push({
+			firstName,
+			lastName,
+			creatorType: "contributor",
+			fieldMode: true
```
dstillman, line 227: Oh, I see, you're setting this manually. That's not correct. If you have first and last name, you can just omit `fieldMode` (which would be `0` for names with first and last names).

Computer History Museum Archive.js
boan-anbo: thx. moved to Extra.

Computer History Museum Archive.js
boan-anbo: fixed

Computer History Museum Archive.js
boan-anbo: fixed. all moved to Extra/

Computer History Museum Archive.js
boan-anbo: fixed. fieldMode omitted now.

Computer History Museum Archive.js
boan-anbo: fixed. thanks!

boan-anbo:
all previous problems fixed. added a few fields and audio attachments with new tests.

Computer History Museum Archive.js
```diff
@@ -0,0 +1,1000 @@
...
+				break;
+			case 'Participants':
+			case 'Contributor':
+				newItem.creators = getContributors(node)
+				break;
+			case 'Publisher':
+				newItem.publisher = getContent(node, fieldTitle)
+				break;
+			case 'Place of Publication':
+				newItem.publisher += '; ' + getContent(node, fieldTitle)
```
AbeJellinek, line 112: This should just go in `newItem.place`.

Computer History Museum Archive.js
```diff
@@ -0,0 +1,1000 @@
...
+				const numberDivs = node.querySelectorAll('tr')
+				numberDivs.forEach(div => {
+					const numberKey = div.querySelector('td.col1').textContent
+					const numberVal = div.querySelector('td.col2').textContent
+					switch(numberKey) {
+						case 'ISBN10':
+							newItem.ISBN = numberVal
+							break;
+						case 'LOC call num':
+							newItem.callNumber = numberVal
```
AbeJellinek, line 158: `callNumber` refers to the call number of the item in its `libraryCatalog`, but in this case the library catalog is the CHM and the call number is for the LOC. This should go in Extra, if anywhere. I'm not sure we really need to keep it.

Computer History Museum Archive.js
```diff
@@ -0,0 +1,1000 @@
...
+	// add pdf documents
+	let pdfPath = undefined;
+	const pdfDiv = doc.querySelectorAll('div.mediaDocument li a');
+	if (pdfDiv) {
+		pdfDiv.forEach((div, pdfIndex) => {
+			const fileName = div.textContent;
+			pdfPath = div.href
+			if (pdfPath) {
+				newItem.attachments.push({
+					url: pdfPath.replace('https', 'http'),
```
AbeJellinek, line 181: why get rid of https?

Computer History Museum Archive.js
```diff
@@ -0,0 +1,1000 @@
...
+	// add audio recordings
+	let audioPath = undefined;
+	const audioDiv = doc.querySelectorAll('div.mediaAudio li a');
+	if (audioDiv) {
+		audioDiv.forEach((div, audioIndex) => {
+			audioPath = div.href
+			const fileName = div.textContent;
+			if (audioPath) {
+				newItem.attachments.push({
+					url: audioPath.replace('https', 'http'),
```
AbeJellinek, line 199: as above

Computer History Museum Archive.js
```diff
@@ -0,0 +1,1000 @@
...
+	}
+	else if (isArtifact) {
+	//todo it'd be better to use different items types according to item types in the collection, i.e. interview, video, audio etc, but for now 'book' has most fields needed.
+		return 'book';
+	}
+	return false;
+}
+
+function doWeb(doc, url) {
+	if (detectWeb(doc, url) === "multiple") {
```
AbeJellinek, line 50: `==` is safer than `===` for strings in general.

Computer History Museum Archive.js
```diff
@@ -0,0 +1,1000 @@
...
+
+function scrape(doc, url) {
+	const fields = [...doc.querySelectorAll("div.field")];
+
+	let newItem = new Zotero.Item("book");
+	newItem.archive = 'Computer History Museum';
+
+	fields.forEach((nodeOriginal) => {
+		// clone the node to avoid altering the document.
+		const node = nodeOriginal.cloneNode(true)
```
AbeJellinek, line 94: Please lint and particularly add semicolons!

AbeJellinek (changes requested):

Computer History Museum Archive.js
boan-anbo: Hi Abe, Thanks for the comments. I have fixed others but this. For some reason, the repo ESLINT config is not working on IDEs (Intellij, VS Code) as they show ESLINT no header error. Do you know how to fix this? If not, don't bother, I'll try lint by other means. Thx.

dstillman: It's working for me in VS Code, but you need to have installed the repo dependencies with `npm i`, and you need to give VS Code permission to load the local installation when you first click on ESLINT in the bottom right:

eslint-validation

If you previously installed dependencies but don't remember what you chose for that, you can run "ESLint: Reset Library Execution Decisions" to force it to ask you again.

boan-anbo: Thanks @dstillman! I just got it working but in a very hacky way. It turned out to be an weird bug. I still don't know the exact cause.

When I run `eslint "computer history museum archive.js"`, the "no header" error appears because of these two lines:

https://github.com/zotero/translators/blob/809e3973697036a2d0ccb3fbfa7b96c585c4168e/.ci/eslint-plugin-zotero-translator/lib/translators.js#L109

```
	let m = source.match(re.undecorated);
	if (!m) throw new Error('no header');
```

And then the script is run NOT on the "computer history museum archive.js", but on the very first translator file, i.e. A Contra Corriente.js (if you remove it, the next will be `Ab Impreio.js.` with the same error.), in the folder by alphabetic order. In other words, the "no header" error is raised when checking all the translator files and it failed on the very first one.

However, t**he strange thing is** that when I checked both `source` (the text of the linted translator file) and the regular expression `re.undecorated`, **they both have the correct values.** And if I match them in an empty node script  by calling `source.match(re)` or in a regular expression tester, it returns the correct matched array. 

- Regex Tester
image


But **when ESLink calls it, it returns "Null" as the result**, which raises the exception in the second line.


- Test Code:
image


- Console output:
image


So the only way for me to run lint on my file is to bypass this function by `return decorated` in all cases, and  two other steps. 1. rename the file into one without empty spaces (otherwise it will raise a separate `TypeError: Cannot read property 'source' of undefined` error). 2. Comment out the beginning meta translator information section. 

 
image

# In short

**For unknown reasons, `let m = source.match(re.undecorated);` returned "null" for m even though both `source` and `re.undecorated` have the correct text and pattern.** And it caused the parser for the translator files and eventually ESLint to fail.

boan-anbo: A possible explanation: https://stackoverflow.com/questions/54607777/string-match-function-returns-null-even-when-there-is-a-match

Computer History Museum Archive.js
boan-anbo: fixed.

Computer History Museum Archive.js
boan-anbo: For some reasons, the files provided by CHM can only be fetched by Zotero with http.

Computer History Museum Archive.js
boan-anbo: Thanks! I moved it to extra. Considering that most users of the translator would probably like more archival location information than less especially when using Zotero for on-site research, I kindly ask to preserve this information.

Computer History Museum Archive.js
boan-anbo: Fixed. Thanks!

Computer History Museum Archive.js
dstillman: All the files in your tests appear to be accessible via HTTPS, and are linked that way from the pages. If you can't access them that way, that's a problem with your internet connection.

dstillman: The files download fine to Zotero for me over HTTPS. This will need to be removed. We can't accept a translator that converts HTTPS URLs into HTTP ones.

If you're having trouble with your Zotero downloading files, we can debug that separately in the Zotero Forums. (It's likely some sort of issue with your internet connection, a proxy server, or security software on your computer.)

Computer History Museum Archive.js
AbeJellinek: OK, linting looks good now. (The comment on line 42 isn't quite indented enough, but that doesn't really matter and doesn't cause a CI failure.)

Computer History Museum Archive.js
AbeJellinek: Agreed. It's odd that it doesn't automatically redirect to HTTPS, for that matter.

boan-anbo: Thanks @dstillman and @AbeJellinek ! I’ve removed the conversions. Please review or advise for further improvement.

Computer History Museum Archive.js
```diff
@@ -0,0 +1,988 @@
...
+		"items": [
+			{
+				"itemType": "book",
+				"title": "People's Computer Company; People's Computers; Recreational Computing",
+				"creators": [],
+				"date": "1972-1981",
+				"abstractNote": "Founded in Oct 1972 as a large-format bimonthly newsprint publication called \"People's Computer Company.\" Name changed to \"People's Computers\" beginning with the May-June 1977 issue and the format changed to a more conventional magazine style, albeit with uncoated paper. Name changed again to \"Recreational Computing\" with Jan-Feb 1979 issue with slicker covers and paper. \n\nNote: The print collection has been augmented with scans provided by Bob Zeidman under lot number X6691.2013. Some issues in the collection exist in digital form only.",
+				"archive": "Computer History Museum",
+				"archiveLocation": "X2595.2004",
+				"extra": "Catalog Number: 102661095\nType: Document\nCategory: Periodical\nCredit: Gift of Jim Warren",
```
dstillman, line 756: `Type` shouldn't be in any of these. If these items aren't books, they shouldn't have the Book item type. They should be saved as appropriate item types. This may require a mapping between the CHM types and Zotero item types.

Computer History Museum Archive.js
```diff
@@ -0,0 +1,988 @@
...
+				"archive": "Computer History Museum",
+				"archiveLocation": "X8201.2017",
+				"extra": "Catalog Number: 102738261\nType: Moving image\nFormat: MOV\nCategory: Oral history\nSubject: ITRI; Apple; iPhone; Camera\nCredit: Computer History Museum",
```
dstillman, line 596: `Subject` should become tags.

dstillman: And the `Video Recording` item type, which this should be, has a `Format` field (though that doesn't add a lot of value).

Computer History Museum Archive.js
```diff
@@ -0,0 +1,988 @@
...
+				"archive": "Computer History Museum",
+				"archiveLocation": "X2595.2004",
+				"extra": "Catalog Number: 102706809\nType: Audio\nOther number: camvchm_000022\nOther number: Tape #8\nFormat: Standard audio cassette\nSeries Title: The First West Coast Computer Faire\nCredit: Gift of Jim Warren",
```
AbeJellinek, line 724: `Other number` at least needs a better name. Or to be removed? If there's a lot of data that you think users of the translator will want but that doesn't fit in a Zotero item field, maybe that's reason enough to save a page snapshot.

boan-anbo: Good point! I’ll try add a snapshot and remove Other number.

Computer History Museum Archive.js
boan-anbo: I see! Can’t believe I forgot about this. Let me restructure the program to tell the Zotero item type first.

Computer History Museum Archive.js
AbeJellinek: Sounds great! Thanks.

boan-anbo: Fixed!

Computer History Museum Archive.js
boan-anbo: Fixed.

Computer History Museum Archive.js
boan-anbo: fixed.

Computer History Museum Archive.js
boan-anbo: Fixed!

boan-anbo:
All previous questions fixed. Please review and advise.

Computer History Museum Archive.js
```diff
@@ -0,0 +1,1291 @@
...
+				"seeAlso": []
+			}
+		]
+	},
+	{
+		"type": "web",
+		"url": "https://www.computerhistory.org/collections/catalog/102706809",
+		"items": [
+			{
+				"itemType": "book",
```
dstillman, line 987: This is `Type: Audio` on the page, and in Zotero should be an `audioRecording` item, which has `format` field for the `Standard audio cassette` value.

Computer History Museum Archive.js
```diff
@@ -0,0 +1,1291 @@
...
+		"url": "https://www.computerhistory.org/collections/catalog/102630889",
+		"items": [
+			{
+				"itemType": "artwork",
+				"title": "Apple IIe",
+				"creators": [],
+				"abstractNote": "Black and white identification photograph of the Apple IIe main terminal including monitor, disk drive and keyboard. appriximately 1/2 inch white border surrounds main image. Background is gray. Computer is sitting on a ledge.",
+				"archive": "Computer History Museum",
+				"archiveLocation": "X2870.2005",
+				"extra": "Catalog Number: 102630889\nDimensions: 8 x 10 in.\nFormat: Photographic print\nCategory: Identification photograph; Publicity photograph\nCredit: Gift of CHM AppleLore",
```
dstillman, line 1268: - `Dimensions` is `artworkSize`
- `Format` is `medium`

Computer History Museum Archive.js
```diff
@@ -0,0 +1,1291 @@
...
+					}
+				],
+				"date": "2015-03-31",
+				"abstractNote": "C. Richard Kramlich co-founded the venture capital firm New Enterprise Associates (NEA) in 1978, serving as its managing general partner for two decades. In this oral history, Kramlich discusses his education at the Harvard Business School, his early career in finance, and his entry into venture capital in a partnership with Arthur Rock starting in 1969. He details the establishment of NEA with his co-founders, his personal investment in Apple Computer, and one of NEA’s earliest investments, 3Com. After noting the rise of the graphical in computing, Kramlich discusses at length his investment and involvement with the Apple Computer spinoff, Forethought Inc., its history, and its development of PowerPoint. The interview concludes with some of his thoughts about his investment in Silicon Graphics, the software industry, NEA’s approach, and collecting video art.",
+				"archive": "Computer History Museum",
+				"archiveLocation": "X7447.2015",
+				"extra": "Catalog Number: 102740039\nDuration: 01:31:00\nCategory: Oral history\nCollection Title: CHM Oral History Collection\nCredit: Computer History Museum",
```
dstillman, line 891: - `Duration` is `runningTime`

dstillman (changes requested):

Computer History Museum Archive.js
boan-anbo: Thanks! Fixed. BTW, Scafford doesn't have the format for the audio- & video-recording formats fields. Perhaps needs an update?

Computer History Museum Archive.js
boan-anbo: Fixed. Thanks!

Computer History Museum Archive.js
boan-anbo: Fixed. Thanks!

Computer History Museum Archive.js
AbeJellinek: It does on my end, at least. The format field for `videoRecording` is called `videoRecordingFormat`, and for `audioRecording` it's called `audioRecordingFormat`.

boan-anbo: > It does on my end, at least. The format field for `videoRecording` is called `videoRecordingFormat`, and for `audioRecording` it's called `audioRecordingFormat`.

Oh I see! Thanks! Fixed. Is it ready to merge?

AbeJellinek: I still see the two `replace('https', 'http')` lines, but after that's gone, yes!

boan-anbo: ahh, sorry about that! The earlier change was made directly on Github and I forgot to pull. Just updated! Thanks so much for all the helps.

AbeJellinek:
Great. Merged your changes. Thanks so much for your contribution, and for bearing with so many code reviews! A complex translator that can produce so many item types will also have lots of edge cases, unfortunately.

boan-anbo:
> Great. Merged your changes. Thanks so much for your contribution, and for bearing with so many code reviews! A complex translator that can produce so many item types will also have lots of edge cases, unfortunately.

Thanks Abe! A quick question though, if I make further changes to this translator, do I need to reopen this or create another pull request?

dstillman:
Once a PR is merged, any further updates would go in a new PR.

boan-anbo:
> Once a PR is merged, any further updates would go in a new PR.

I see. Thanks! I'll do that once there are a sizeable number of changes to it.

PR #2289, opened 2020-11-24
PQ-71894: fix results list error in zotero ProQuest translator

pmusson:


pmusson:
Should fix the title undefined error we are seeing. Probably caused by a change in the proquest webapp markup

Also added to this fix the change to the new proquest url www.proquest.com  (search.proquest.com also still required)

adam3smith:
Cool thanks -- you'd need to bump the timestamp, please, otherwise this doesn't actually update for users.

pmusson:
Thanks, I've updated the timestamp

adam3smith:
Thanks!

PR #2290, opened 2020-11-24
improve "title" field for the "Library Catalog (PICA)" translator

kerphi:
The current behavior is problematic for French records because a space is needed before a colon.
The current code removes ALL spaces before a colon and breaks the French typo rule.

This patch remove all extra spaces before a colon in the title field and leave one space if the cleanup occurs.
It also add few comments to explain what theses regexp are doing on the title field.

### DIFF
--- a/Library Catalog (PICA).js
+++ b/Library Catalog (PICA).js
@@ -381,7 +381,10 @@ function scrape(doc, url) {
 				if (!newItem.title) {
 					newItem.title = title[0];
 				}
-				newItem.title = newItem.title.replace(/\s+:/, ":").replace(/\s*\[[^\]]+\]/g, "");
+				// replace(/\s+:/, " :") => cleanup extra spaces before a colon
+				//                          and respects the english (no space) or french typo (with a space)
+				// replace(/\s*\[[^\]]+\]/g, "") => cleanup "[ xxx ]" patterns in the title
+				newItem.title = newItem.title.replace(/\s+:/, " :").replace(/\s*\[[^\]]+\]/g, "");
 				break;
 
 			case 'periodical':


Library Catalog (PICA).js
```diff
@@ -381,7 +381,10 @@ function scrape(doc, url) {
 				if (!newItem.title) {
 					newItem.title = title[0];
 				}
-				newItem.title = newItem.title.replace(/\s+:/, ":").replace(/\s*\[[^\]]+\]/g, "");
+				// replace(/\s+:/, " :") => cleanup extra spaces before a colon
+				//                          and respects the english (no space) or french typo (with a space)
+				// replace(/\s*\[[^\]]+\]/g, "") => cleanup "[ xxx ]" patterns in the title
+				newItem.title = newItem.title.replace(/\s+:/, " :").replace(/\s*\[[^\]]+\]/g, "");
```
adam3smith, line 387: I don't understand how that doesn't break things for German & English catalogs? We removed the space before the colon based on user feedback.
Not sure if this can even be handled in the data in any reasonable way (especially since the convention in France is technically a nbsp before the colon)

kerphi: I think it does not break German/English catalogs because this patch will not add any space if there is no space before the colon in the title coming from the source.
Does German or English PICA catalogs add spaces before the colon in the title field into there catalogs? If yes, I agree this patch will break German or English catalogs.

adam3smith: > Does German or English PICA catalogs add spaces before the colon in the title field into there catalogs? If yes, I agree this patch will break German or English catalogs.

Yes -- the space before the colon is an international cataloging convention, all library catalogs using MARC (i.e. almost all of them) use it to delimit title and subtitle.

adam3smith: My inclination would be to say that this might be best handled by the citation processor, but I'm open for other solutions, including in the Zotero data. But it seems like someone who publishes in English and French shouldn't have to constantly change around their data and even citing a French title in English you wouldn't want the spaces before the colon (at least according to Chicago Manual, and I assume most other styles).

kerphi: Thanks for your quick reply.
I'm wondering why zotero decided to truncate the space before the colon if having a space before the colon is an international cataloging convention? maybe because end-user are not aware with these conventions?

adam3smith: Because it's a convention for library catalogs, not one for referencing. It doesn't make a lot of sense for Zotero to import data in a way that's not actually useful on export to either other metadata format or to citation styles. It's also not how a lot of metadata outside of professionally cataloged libraries is displayed -- e.g. most journal metadata and DOI metadata doesn't follow that convention.

kerphi: I have an idea (not very clean ...) but what do you think about something like that: 

```js
if (the domain == sudoc.fr) { // <=== to adjust
  newItem.title = newItem.title.replace(/\s+:/, ":").replace(/\s*\[[^\]]+\]/g, "");
} else {
  newItem.title = newItem.title.replace(/\s+:/, " :").replace(/\s*\[[^\]]+\]/g, "");
}
```

adam3smith: I'd like to hear what @dstillman thinks about this. I thought about making this dependent on the catalog or the language of an item but
a) That still leaves French users with most titles incorrect -- e.g. anything imported via DOI
b) It doesn't solve the problem for people authoring in multiple languages
c) It's still not actually correct French typography, which should use a non-breaking space (or, I think, ideally a narrow non-breaking space), not just a space before the colon. Having the : dangle on a new line isn't great...

AbeJellinek:
I'm going to close this, sorry. It would be somewhat nice to save French users a step, yeah, but in the end it's not in the purview of Zotero translators, which are supposed to produce neutral data.

PR #2291, opened 2020-11-28
Add lesoir.be translators

Alexanb:
Here is my first trial to write a translator, for a Belgian newspaper.


Le Soir.js
```diff
@@ -0,0 +1,262 @@
...
+		item.publicationTitle = "Le Soir";
+		item.ISSN = "1375-5668";
+		var authorString = ZU.xpathText(doc, '//div[@class="gr-article-infos"]/p[@class="gr-meta gr-meta-author"]|//div[@class="required-fields gr-infos-author"]/p[@class="gr-meta gr-meta-author"]');
+		if (authorString) {
+			authorString = authorString.replace('Par ', '');
+			item.creators = [];
+			var authors = authorString.split(/,|et/);
+			for (var i = 0; i < authors.length; i++) {
+				if (i == authors.length - 1) {
+					authors[i] = authors[i].split(",")[0];
```
AbeJellinek, line 90: Could you add an example of this behavior in the tests?

Le Soir.js
```diff
@@ -0,0 +1,262 @@
...
+	}
+}
+
+function scrape(doc, url) {
+	var translator = Zotero.loadTranslator('web');
+	// Embedded Metadata
+	translator.setTranslator('951c027d-74ac-47d4-a107-9c3069ab7b48');
+	// translator.setDocument(doc);
+	translator.setHandler('itemDone', function (obj, item) {
+		item.date = ZU.strToISO(ZU.xpathText(doc, '//time[@class=" pubdate updated"]'));
```
AbeJellinek, line 80: Use a CSS selector here for more flexibility:

```js
text(doc, 'time.pubdate.updated')
```

Le Soir.js
```diff
@@ -0,0 +1,262 @@
...
+		item.publicationTitle = "Le Soir";
+		item.ISSN = "1375-5668";
+		var authorString = ZU.xpathText(doc, '//div[@class="gr-article-infos"]/p[@class="gr-meta gr-meta-author"]|//div[@class="required-fields gr-infos-author"]/p[@class="gr-meta gr-meta-author"]');
```
AbeJellinek, line 83: Do we need all these specifiers? Could we get away with just `p[@class="gr-meta gr-meta-author"]`? (Same note about `text` applies here, too.)

Le Soir.js
```diff
@@ -0,0 +1,262 @@
...
+		if (authorString) {
+			authorString = authorString.replace('Par ', '');
```
AbeJellinek, line 85: Should use a regex to make sure someone with `Par ` in their name doesn't give us problems :)

```js
authorString = authorString.replace(/^\s*Par\s*/, '');
```

AbeJellinek (changes requested):
Hey, thanks for contributing this and sorry for the delay in reviewing it. I left a few comments - let me know what you think.

PR #2294, opened 2020-12-01
Another attempt to fix wrong chromedriver version CI issues

### LINKED ISSUE #2283: Automated Selenium Tests broken
The Selenium builds in Travis are broken and now show every PR as not passing (see e.g. https://travis-ci.org/github/zotero/translators/builds/743092324?utm_source=github_status&utm_medium=notification ). Having them work would be super cool, but if this is currently too fragile and requires too much maintenance, could we disable them?
CC @adomasven

adomasven:
Closes #2283

Sorry @adam3smith, I missed the original issue you've created in the email stream, would have got around to this sooner.

For the record there's been some changes to the way dependency install scripts work with npm 7/node 15, to which Travis must have switched recently. They claim they've only disabled logging, which may be true and after this PR the test checker will still fail, but at least we will have better logging about why the correct chromedriver version is not being installed.

### DIFF
--- a/.ci/pull-request-check/check-pull-request.sh
+++ b/.ci/pull-request-check/check-pull-request.sh
@@ -57,7 +57,7 @@ export ZOTERO_REPOSITORY_URL="http://localhost:8085/"
 ./build.sh -p b -d
 cd ..
 
-npm install chromedriver --detect_chromedriver_version
+npm explore chromedriver -- npm run install --detect_chromedriver_version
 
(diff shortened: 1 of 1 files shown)


adam3smith:
That Worked: https://travis-ci.com/github/zotero/translators/builds/205700884
Thanks and of course no worries -- I figured it may have gotten lost, hence the bump on the ticket.

PR #2296, opened 2020-12-05
SAGE fix detect for new article format

adam3smith:
https://forums.zotero.org/discussion/86489/zotero-connector-issue-with-sage-pub/p1

### DIFF
--- a/SAGE Journals.js
+++ b/SAGE Journals.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2020-12-05 02:59:46"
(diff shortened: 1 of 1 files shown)


adam3smith:
Tests don't run in Scaffold, so can't update reasonably, so merging with failing CI

dstillman:
> Tests don't run in Scaffold

Sorry about that — it looks like we weren't supporting cookies properly in Scaffold, so translator HTTP requests that required cookies sent with the original webpage didn't work properly. Should be fixed now in the latest Zotero beta. I think you reported a similar problem with another translator recently, and I'm guessing that's fixed now too.

(FWIW when I was testing this it looked like tags were broken on SAGE, even though they're still available on the page.)

PR #2302, opened 2020-12-10
Proquest detect fix

### LINKED ISSUE #2301: ProQuest: Not detecting on "Full Text" page
https://forums.zotero.org/discussion/86609/fields-not-inputting-correctly-for-proquest-articles-incomplete-bibliographies/p1

> One thing I noticed - when I saved the page with the Full Text tab, but not the full pdf, it didn't capture all of the metadata, but when I clicked on the Full Text - PDF tab, it did.

And Abstract page works.

I haven't tested this.

adam3smith:
closes #2301

### DIFF
--- a/ProQuest.js
+++ b/ProQuest.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2020-11-25 09:30:01"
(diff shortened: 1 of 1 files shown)


adam3smith:
No test updates -- I only have web-proxy access

dstillman:
(I'm currently without proxy access, so I can't test this.)

PR #2304, opened 2020-12-12
JSTOR: Fix errant comma in book review titles

owcz:
Prior code left extra ", " in book reviews with multiple titles reviewed


dstillman:
Thanks!

Issue #2303, opened 2020-12-12
Project MUSE returns degraded result from EM when authenticated


owcz:
When _logged out_, this translator performs as expected.

When _logged in or proxied_, it looks like the Project MUSE translator sends the article ([for example](https://muse.jhu.edu/article/558726)) to Embedded Metadata (EM), which now returns a degraded result with no metadata (uses the document title tag as title, etc.) Replicated the issue across different Zotero instances. Not sure what I should be looking for when logged in to troubleshoot this.

Logged out
```
02:48:17 Returned item:
           {
             "itemType": "journalArticle"
             "creators": [
               {
                 "firstName": "Alban"
                 "lastName": "Bargain-Villéger"
                 "creatorType": "author"
               }
             ]
             "notes": []
             "tags": []
             "seeAlso": []
             "attachments": [
               {
                 "title": "Full Text PDF"
                 "mimeType": "application/pdf"
               }
             ]
             "title": "The Scarecrow on the Other Side of the Pond: The Paris Commune of 1871 in the Canadian Press"
             "publicationTitle": "Labour / Le Travail"
             "volume": "74"
             "issue": "1"
         -   "publisher": "The Canadian Committee on Labour History"
             "pages": "179-198"
             "date": "2014-11-08"
             "ISSN": "1911-4842"
             "url": "https://muse.jhu.edu/article/558726"
             "language": "en"
             "libraryCatalog": "Project MUSE"
         -   "accessDate": "CURRENT_TIMESTAMP"
             "shortTitle": "The Scarecrow on the Other Side of the Pond"
           }
02:48:17 Translation successful
```

When logged in (with Muse account)
```
02:52:35 Returned item:
           {
             "itemType": "webpage"
             "creators": []
             "notes": []
             "tags": []
             "seeAlso": []
             "attachments": [
               {
                 "title": "Full Text PDF"
                 "mimeType": "application/pdf"
               }
             ]
             "url": "https://muse.jhu.edu/article/558726"
             "title": "Project MUSE - The Scarecrow on the Other Side of the Pond: The Paris Commune of 1871 in the Canadian Press"
             "language": "en"
         -   "accessDate": "CURRENT_TIMESTAMP"
             "shortTitle": "Project MUSE - The Scarecrow on the Other Side of the Pond"
           }
02:52:35 Translation successful
```

adam3smith:
[...]

PR #2307, opened 2020-12-14
updated Project MUSE.js new approach as suggested in https://github.com/zotero/translators/issues/2303

socheres:


Project MUSE.js
```diff
@@ -188,75 +194,144 @@ var testCases = [
...
 				"url": "https://muse.jhu.edu/article/551992",
 				"volume": "49",
-				"attachments": [
+				"attachments": [],
+				"tags": [],
+				"notes": [
 					{
-						"title": "Full Text PDF",
-						"mimeType": "application/pdf"
+						"note": "<p>Volume 49, Number 2, 2014</p>"
```
adam3smith, line 259: Discard the notes -- they don't contain additional information

Project MUSE.js
```diff
@@ -86,100 +86,106 @@ function doWeb(doc, url) {
...
-	if (!abstract) abstract = ZU.xpathText(doc, '//div[@class="description"][1]');
-	if (!abstract) abstract = ZU.xpathText(doc, '//div[contains(@class, "card_summary") and contains(@class, "no_border")]');
-	var translator = Zotero.loadTranslator('web');
-	// Embedded Metadata
-	translator.setTranslator('951c027d-74ac-47d4-a107-9c3069ab7b48');
-	translator.setHandler('itemDone', function (obj, item) {
-		if (abstract) {
-			item.abstractNote = abstract.replace(/^\s*Abstract/, "").replace(/show (less|more)$/, "").replace(/,\s*$/, "")
-				.trim();
+	let citationURL = ZU.xpath(doc, '//*[contains(concat( " ", @class, " " ), concat( " ", "view_citation", " " ))]//a');
```
adam3smith, line 89: why not just 
`ZU.xpathText(doc, '//li[@class="view-citation"]/a/@href')`
? And then you can just plug this right into the ZU.processDocuments function

adam3smith (commented):
Thanks! A couple of small tweaks/requests

Project MUSE.js
socheres: thanks for the code review. yes, i discard it.

Project MUSE.js
socheres: yes. this is much simpler.

adam3smith:
Thanks!

adam3smith:
Remaining test discrepancy is authenticated vs. non-authenticated. The test represented the (better) authenticated data

PR #2309, opened 2020-12-15
Add translator for Westlaw

georgegebbett:
This is the beginning of a rough and ready WestLaw translator. Due to the complete lack of metadata provided by WestLaw, it relies on elements of the page being in the right places. It is aimed at UK users and will add UK cases and legislation to Zotero with the information needed to cite them in Oscola filled into the fields.


WestLaw UK.js
```diff
@@ -0,0 +1,163 @@
...
+
+
+function scrapeCase(doc, url) {
+	var translator = Zotero.loadTranslator('web');
+	// Embedded Metadata
+	translator.setTranslator('951c027d-74ac-47d4-a107-9c3069ab7b48');
+	// translator.setDocument(doc);
+	
+	translator.setHandler('itemDone', function (obj, item) {
+		// TODO adjust if needed:
```
AbeJellinek, line 78: can remove

WestLaw UK.js
```diff
@@ -0,0 +1,163 @@
...
+		
+		var citationArray = parseCitationList(doc.getElementsByClassName("co_docContentMetaField")[3].innerText);
```
AbeJellinek, line 80: Is there a better selector for this? It's pretty fragile to assume that it'll always be the fourth element with that class on the page. You could select on an attribute with

```js
text(doc, ".co_docContentMetaField[some-attr='some-value']")
```

WestLaw UK.js
```diff
@@ -0,0 +1,163 @@
...
+}
+
+function scrapeStatute(doc, url) {
+	var translator = Zotero.loadTranslator('web');
+	// Embedded Metadata
+	translator.setTranslator('951c027d-74ac-47d4-a107-9c3069ab7b48');
+	// translator.setDocument(doc);
+	
+	translator.setHandler('itemDone', function (obj, item) {
+		// TODO adjust if needed:
```
AbeJellinek, line 108: as above, you can remove these placeholder comments

WestLaw UK.js
```diff
@@ -0,0 +1,163 @@
...
+		trans.itemType = "statute";
+		// TODO map additional meta tags here, or delete completely
+
+		trans.doWeb(doc, url);
+	});
+}
+
+
+function parseCitationList(citList){
+	var allCitations = citList.match(/^[\[|\(](\d+)[\]|\)] (\d*) ?([A-z\. ]+) ([\w]+)$/gm); //annoyingly matchAll is not supported
```
AbeJellinek, line 149: Will it always always take this form? Feels like there *must* be a better way. Check if the page has any relevant metadata in `<meta>` tags or hidden fields?

AbeJellinek (approved):
Thanks, this looks great! I left a few comments. The biggest issue is just making sure that selectors are stable so that we don't have to update the translator constantly.

WestLaw UK.js
```diff
@@ -0,0 +1,163 @@
+{
```
AbeJellinek, line 1: That was meant to be a "request changes," not an approval, but there's apparently no way to change it. Other small issue is that we shouldn't have .DS_Store files checked into the repo.

AbeJellinek (changes requested):

WestLaw UK.js
```diff
@@ -0,0 +1,163 @@
...
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2021-01-06 11:36:52"
+}
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2020 YOUR_NAME <- TODO
```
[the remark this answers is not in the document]
georgegebbett: Done!

WestLaw UK.js
georgegebbett: I have found a much better way of doing this - we now select on element ID!

WestLaw UK.js
georgegebbett: All gone

WestLaw UK.js
georgegebbett: I hate this as much as you do, I promise. Luckily law case citations take a pretty standard form, but there really is no way to get this from the page 😓

WestLaw UK.js
georgegebbett: All changes _should_ now be made - between writing this originally and now I did the FreeCodeCamp full stack qualification and always felt like I didn't learn much but looking back at what my JS was like before has really made me see the progress I've made 😂

WestLaw UK.js
```diff
@@ -0,0 +1,139 @@
+{
+	"translatorID": "e6d9cf77-53e4-47f5-98ae-a69eba6d985f",
+	"label": "WestLaw UK",
+	"creator": "George Gebbett",
+	"target": "^https://uk\\.westlaw\\.com/Document/.*",
```
AbeJellinek, line 5: We don't need the `.*` at the end - since the pattern starts with `^` (matches the beginning of the string) but doesn't end in `$` (matches the end of the string), we implicitly accept trailing characters not matched by the regex.

WestLaw UK.js
```diff
@@ -0,0 +1,139 @@
...
+		case "statuteSection":
+			return scrapeStatuteSection(doc, url);
+		default:
+			return false;
+	}
+}
+
+function scrapeCase(doc, url) {
+	const translator = Zotero.loadTranslator('web');
+	translator.setTranslator('951c027d-74ac-47d4-a107-9c3069ab7b48');
```
AbeJellinek, line 71: `setTranslator` calls should include a comment specifying the name of the translator being loaded, for ease of understanding. So:

```suggestion
	// Embedded Metadata
	translator.setTranslator('951c027d-74ac-47d4-a107-9c3069ab7b48');
```

WestLaw UK.js
```diff
@@ -0,0 +1,139 @@
...
+	translator.setHandler('itemDone', function (obj, item) {
+		const citationArray = parseCitationList(text(doc, "#co_docContentWhereReported", 0));
+
+		item.reporter = citationArray[3];
+		item.reporterVolume = citationArray[2];
+		item.firstPage = citationArray[4];
+		item.dateDecided = citationArray[1]
+		item.court = /Court(.+)/.exec(text(doc, "#co_docContentCourt"))[1];
```
AbeJellinek, line 79: Is there ever a case where this won't match? The array access will cause an exception if so, which means translation will fail. If the format of `#co_docContentCourt` is expected always to look like this unless the site changes significantly, that's fine - we'd probably _want_ translation to fail in that situation.

WestLaw UK.js
```diff
@@ -0,0 +1,139 @@
...
+	if (allCitations.length === 1){ //if there is only one citation we'll have to use it
+		return /[\[|(](\d+)[\]|)] (\d*) ?([A-z. ]+) (\w[^Judgment])+/g.exec(allCitations[0]);
+	} else { //if there's more than one we will find the first one that isn't WLUK
+		for (const citation of allCitations) {
+			const citationAsArray = /[\[|(](\d+)[\]|)] (\d*) ?([A-z. ]+) (\w[^Judgment])+/g.exec(citation);
+			if (citationAsArray[3] !== "WLUK"){
+				return citationAsArray;
+			}
+		}
+	}
```
AbeJellinek, line 138: Not having `matchAll` is indeed annoying, but there's a better way!

```suggestion
	let citationRe = /[\[|(](\d+)[\]|)] (\d*) ?([A-z. ]+) (\w[^Judgment])+/g;
	let citationAsArray;
	while ((citationAsArray = citationRe.exec(citation))) {
		if (citationAsArray[3] !== "WLUK") {
			return citationAsArray;
		}
	}
	return citationAsArray;
```

If there's no non-WLUK citation, do we want not to return anything at all? Or should we fall back on, say, the last citation? I assumed the latter in the code above.

AbeJellinek (changes requested):
Thanks! This is looking great. A few more comments.

WestLaw UK.js
georgegebbett: That element has looked the same for the whole two years I've been using Westlaw - if it changed I would be surprised but this would probably be part of a large redesign where we would want translation to fail and to redo the translator

WestLaw UK.js
georgegebbett: I had no idea you could do that with `exec()`! How fun is that!!

This approach would work well - the WLUK citations are the least ideal and cases that have only a WLUK citation are luckily few and far between, but a WLUK citation is better than no citation at all.

WestLaw UK.js
georgegebbett: Got rid of that

georgegebbett:
@AbeJellinek 

image

AbeJellinek:
OK, lastly, we need tests! Just load a page in the Scaffold browser and click *New Web* in the Testing tab, then *Save*. It's fine if tests don't run on CI; it seems like WestLaw is pretty heavily paywalled. As long as they pass on your end, that's good enough.

georgegebbett:
The paywalls and login screens seem to be wreaking havoc with Scaffold's ability to actually run the tests, however it is able to fully populate a test case with all the required data and items also pull through to Zotero correctly

image

georgegebbett:
@AbeJellinek - whenever you're ready, I think (hope) this is ready to go

AbeJellinek:
Sorry for the delay in reviewing! Could you actually save the test in the translator and then push? Otherwise I think this is ready to merge.

georgegebbett:
All done!

AbeJellinek:
Thank you!

PR #2319, opened 2020-12-29
Add The Straits Times

robertsky:
Primarily to fix detection of authors from the newspaper website. My first contribution to the project, do advise if there are places to improve.


The Straits Times.js
```diff
@@ -0,0 +1,286 @@
...
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+function detectWeb(doc, url) {
+	if (url.match(/(advertise-with-the-straits-times|newsletter-signup)/)) {
```
AbeJellinek, line 39: It feels like there must be a better heuristic than this to exclude article-like non-articles. The "Advertise" page has the URL `https://www.straitstimes.com/advertise-with-the-straits-times`, while a normal article takes the form `https://www.straitstimes.com/asia/east-asia/asean-china-foreign-ministers-meet-discuss-trade-covid-19-myanmar` - could you match on the number of path segments, or are there exceptions?

The Straits Times.js
```diff
@@ -0,0 +1,286 @@
...
+
+function doWeb(doc, url) {
+	if (detectWeb(doc, url) === "multiple") {
+		getMultipleItems(doc, url);
+	} else {
+		scrape(doc, url, detectWeb(doc, url));
+	}
+}
+
+function scrape(doc, url, type) {
```
AbeJellinek, line 61: Taking `type` as an argument to `scrape` can be problematic, because `processDocuments` (called in `getMultipleItems`) won't pass it. Will search results always be `newspaperArticle`s?

The Straits Times.js
```diff
@@ -0,0 +1,286 @@
...
+	var newItem = new Zotero.Item(type || 'newspaperArticle');
+	newItem.ISSN = '0585-3923';
+	newItem.url = url;
+	newItem.publicationTitle = 'The Straits Times';
+	newItem.title = ZU.xpathText(doc, '//meta[@name="dcterms.title"]/@content');
+	newItem.abstractNote = ZU.xpathText(doc, '//meta[@name="dcterms.description"]/@content') ? ZU.xpathText(doc, '//meta[@name="dcterms.description"]/@content').replace('Read more at straitstimes.com.', '').trim() : '';
```
AbeJellinek, line 67: This could be simplified to

```js
newItem.abstractNote = (ZU.xpathText(doc, '//meta[@name="dcterms.description"]/@content') || '')
	.replace('Read more at straitstimes.com.', '').trim();
```

The Straits Times.js
```diff
@@ -0,0 +1,286 @@
...
+	newItem.date = ZU.xpathText(doc, '//meta[@name="dcterms.date"]/@content');
+	newItem.place = 'Singapore';
+	newItem.language = 'en';
+	var authors = ZU.xpath(doc, '//meta[@property="article:author"]');
+	for (var i = 0; i < authors.length; i++) {
+		insertCreator(authors[i].getAttribute('content'), newItem);
+	}
+	newItem.attachments = [{
+		document: doc,
+		title: "The Straits Times Snapshot",
```
AbeJellinek, line 77: We're trying to move to just calling these `"Snapshot"`.

The Straits Times.js
```diff
@@ -0,0 +1,286 @@
...
+	}];
+	if (doc.evaluate('//div[@class="paid-premium st-flag-1"]', doc, null, XPathResult.ANY_TYPE, null).iterateNext()) {
+		newItem.extra = "(Subscription only)";
```
AbeJellinek, line 80: Items in `extra` should be in `Key: Value` format, so this could be something like `"Straits Times Access: Subscription only"` (or it could just be omitted).

The Straits Times.js
```diff
@@ -0,0 +1,286 @@
...
+
+function getMultipleItems(doc, url) {
+	var items = [];
+	var rows;
+	if (url.includes('/search?') && url.includes('searchKey')) {
+		rows = ZU.xpath(doc, '//div[@class="queryly_item_row"]');
+		if (rows.length) {
+			for (var i = 0; i < rows.length; i++) {
+				var searchItem = rows[i];
+				var searchItemUrl = searchItem.querySelector('a') ? searchItem.querySelector('a').href : '';
```
AbeJellinek, line 93: `attr` is useful for selectors like this -

```js
var searchItemUrl = attr(searchItem, 'a', 'href'); 
```

`searchItemUrl` will be `''` if the selector doesn't match.

AbeJellinek: What you have works totally fine, it's just less idiomatic.

The Straits Times.js
```diff
@@ -0,0 +1,286 @@
...
+				items.push(searchItemUrl);
+			}
+		}
+	} else {
+		rows = ZU.xpath(doc, '//a[@class="block-link"]|//span[@class="story-headline"]/a');
+		if (rows.length) {
+			for (var k = 0; k < rows.length; k++) {
+				var headlineItem = rows[k];
+				var headlineItemUrl = headlineItem.href ? headlineItem.href : '';
```
AbeJellinek, line 102: Will an `a` ever not have an `href`?

The Straits Times.js
```diff
@@ -0,0 +1,286 @@
...
+			return (!!item.match(/^https:\/\/www\.straitstimes.com/));
+		});
+		if (items.length) {
+			ZU.processDocuments(items, scrape);
+		}
+	}
+}
+
+function insertCreator(authorName, newItem) {
+	// to account for mostly Chinese names in formats of: <last> <first>, <first (typically English)> <last> <first (typically Chinese)>
```
AbeJellinek, line 118: It's really unfortunate that they don't give us a better way to figure out name order in the metadata! It feels wrong to have to embed a list of authors that will become out of date over time. But it might be our only option.

The Straits Times.js
```diff
@@ -0,0 +1,286 @@
...
+	// list derived from searching in Google with the follow search pattern: authors site:straitstimes.com/authors
+	var authorList = {
+		'Alison de Souza': { first: 'Alison', last: 'de Souza' },
+		'Aw Cheng Wei': { first: 'Cheng Wei', last: 'Aw' },
+		'Benjamin Kang Lim': { first: 'Kang', last: 'Benjamin, Lim' },
```
AbeJellinek, line 123: His English given name is Benjamin, his Chinese given name is Lim, and his surname is Kang, right? In that case, we want to store this as `{ first: 'Benjamin (Lim)', last: 'Kang' }` (and similarly for other names in the list). I'm not sure how citation processors will handle a name with a comma in it, but it's less standard than parentheses, and we definitely shouldn't reverse the order and put first names in `last`.

The Straits Times.js
```diff
@@ -0,0 +1,286 @@
...
+		]
+	},
+	{
+		"type": "web",
+		"url": "https://www.straitstimes.com/singapore/fast-and-furious-can-we-trust-the-speedy-development-of-covid-19-vaccines",
+		"items": [
+			{
+				"itemType": "newspaperArticle",
+				"title": "Fast and Furious. Can we trust the speedy development of Covid-19 vaccines?",
+				"creators": [],
```
AbeJellinek, line 224: The page lists the author as Danny Soon, so we should be scraping that.

AbeJellinek (changes requested):
Hey, thanks for contributing this and sorry it's taken me a while to get through! We have a big backlog, unfortunately. I left some comments - would you be able to make the requested changes? I could be misunderstanding the way you're handling names with Chinese and English parts. Please let me know if so!

robertsky:
opps. I didn't notice this until now. will work on this soon.

The Straits Times.js
robertsky: Nope. English given name: Benjamin, Chinese given name is Kang, Surname is Lim. It took awhile for me to figure that out, but the hint is in the email of the journalist. Now looking at this, it might have been a mistake here. I should have put 'Benjamin, Kang' in the first parameter, and Lim in the last parameter. I think I will just simply put his first as 'Benjamin Kang'.

The Straits Times.js
robertsky: yeah, unfortunately. This is an artefact of Singaporean society. After growing up here, we would know the general race of a person from the name, and which is the first name and which is the surname/family/patronymic name. I couldn't think of a better way unless we forgo breaking down into first and last name.

The Straits Times.js
robertsky: When I built this, it was of the aim that maybe the translation layer over at Wikipedia will pick this up when detecting urls for citations. Looking into this to see if it can be dropped.

The Straits Times.js
robertsky: Yes newspaperArticle will always be the result. There are some opinion/forum pieces from guest writers, but those are mostly published in print as well.

The Straits Times.js
robertsky: Ah. This is one of the guest article which doesn't have article:author meta tag.

The Straits Times.js
robertsky: updated

The Straits Times.js
robertsky: updated

The Straits Times.js
robertsky: Updated

The Straits Times.js
robertsky: updated

The Straits Times.js
robertsky: updated

The Straits Times.js
robertsky: updated

The Straits Times.js
robertsky: @AbeJellinek what do you propose to resolve this?

AbeJellinek: I would suggest that in cases where `detectWeb` would return `webpage`, it should just return false and let the EM translator handle it. Then we can remove the `type` argument from `scrape` and avoid an unnecessary extra call to `detectWeb`.

AbeJellinek:
Thanks for the updates! Looks good. See my reply above.

The Straits Times.js
robertsky: @AbeJellinek  apologies. EM translator?

robertsky: how about the pages that return 'multiple'?

AbeJellinek: Embedded Metadata, sorry! Essentially just let it fall back to a generic translator. You don't need to handle non-article pages in this.

`multiple` is handled separately in `doWeb`, so that's fine.

robertsky: done

robertsky:
@AbeJellinek completed the changes as requested. thanks!

AbeJellinek:
Great. Thanks for bearing with all the reviews! This will be really nice to have.

robertsky:
lovely. thanks for the process! now, to write translators for the other Singaporean newspapers. 😂

PR #2325, opened 2021-01-19
Add Rechtspraak.nl translator

pjrwees:
This translator translates the Dutch judiciary's case law database at uitspraken.rechtspraak.nl. Please excuse the messy commit history, I'm new to git.

### DIFF
--- a/Rechtspraak.nl.js
+++ b/Rechtspraak.nl.js
@@ -2,7 +2,7 @@
 	"translatorID": "aede3fda-1894-4dfc-8bca-1c3463f11076",
 	"label": "Rechtspraak.nl",
 	"creator": "Pieter van der Wees",
-	"target": "^https?://*uitspraken\\.rechtspraak\\.nl/",
+	"target": "^https?://uitspraken\\.rechtspraak\\.nl/",
 	"minVersion": "3.0",
 	"maxVersion": "",
 	"priority": 100,
--- a/Rechtspraak.nl.js
+++ b/Rechtspraak.nl.js
@@ -171,7 +171,8 @@ var testCases = [
 				"abstractNote": "Klimaatzaak Urgenda. Onrechtmatige daad. Schending zorgplicht ex artikelen 2 en 8 EVRM. Staat moet broeikasgassen nu verder terugdringen. Vonnis bekrachtigd\n\nVindplaatsen:\nRechtspraak.nl \nJM 2018/128 met annotatie van W.Th. Douma \nAB 2018/417 met annotatie van G.A. van der Veen, Ch.W. Backes \nO&A 2018/66 \nO&A 2018/51 met annotatie van G.A. van der Veen, T.G. Oztürk \nSEW 2019, afl. 1, p. 35 \nJB 2019/10 met annotatie van Sanderink, D.G.J. \nOGR-Updates.nl 2018-0234 \nPS-Updates.nl 2018-0814 \nJOM 2018/1182 \nJOM 2018/1154 \nJA 2019/37 \nJIN 2019/78 met annotatie van Sanderink, D.G.J.",
 				"court": "Hof Den Haag",
 				"docketNumber": "ECLI:NL:GHDHA:2018:2591",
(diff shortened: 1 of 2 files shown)


Rechtspraak.nl.js
```diff
@@ -0,0 +1,277 @@
...
+	}
+
+	// Attachments: PDF and Snapshot
+	var pdfurl = "https://uitspraken.rechtspraak.nl" + attr(doc, 'a.pdfUitspraak', 'href');
+	newItem.attachments = [{
+		"url": pdfurl,
+		"title": "Rechtspraak.nl PDF",
+		"mimeType": "application/pdf",
+	}];
+	newItem.attachments.push({
```
AbeJellinek, line 160: If we already have a PDF, we don't necessarily need a snapshot. (Looks like the PDF is just a nicer print layout for the same content.)

Rechtspraak.nl.js
```diff
@@ -0,0 +1,277 @@
...
+		"title": "Snapshot",
+		"document": doc
+	});
+
+	newItem.complete();
+}/** BEGIN TEST CASES **/
```
AbeJellinek, line 166: Add two line breaks at the end of your code ("Code" tab in Scaffold) to make CI happy. (This should be automatic, yeah.)

Rechtspraak.nl.js
```diff
@@ -0,0 +1,277 @@
...
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+// attr() v2
+function attr (docOrElem, selector, attr, index) { var elem = index ? docOrElem.querySelectorAll(selector).item(index) : docOrElem.querySelector(selector); return elem ? elem.getAttribute(attr) : null; }
```
AbeJellinek, line 39: This `attr()` shim is no longer necessary.

AbeJellinek:
Hey, thank you! This is great work. Sorry I'm just getting to reviewing it now.

I left you a few comments. Don't worry about the long commit history - we can squash when merging.

AbeJellinek:
OK, I took a stab at getting this ready to merge.

pjrwees:
Thanks for your comments and revision. I think we're ready to merge - let me know if I have to change anything else.

AbeJellinek:
The only thing I'm still concerned about is the `Soort:` in `Extra`. Is there a CSL field that it maps to? If so, it should have its canonical English name.

pjrwees:
Soort outputs the dcterms:type of the Rechtspraak specification and can either be 'Uitspraak' (judgment) or 'Conclusie' (advisory opinion). It does not map to a CSL field -- the closest would be the [document-name](https://citeproc-js.readthedocs.io/en/latest/csl-m/index.html#document-name-extension) field in CSL-M. Importing it makes sense though, as all legal cases before the Hoge Raad have a [conclusie](https://uitspraken.rechtspraak.nl/inziendocument?id=ECLI:NL:PHR:2019:1026) as well as a [judgment](https://uitspraken.rechtspraak.nl/inziendocument?id=ECLI:NL:HR:2019:2007).

Rechtspraak.nl.js
```diff
@@ -0,0 +1,259 @@
+{
+	"translatorID": "aede3fda-1894-4dfc-8bca-1c3463f11076",
+	"label": "Rechtspraak.nl",
+	"creator": "Pieter van der Wees",
+	"target": "^https?://*uitspraken\\.rechtspraak\\.nl/",
```
AbeJellinek, line 5: I don't think this regex is right. The `*` causes it to match zero or more of the previous character, i.e., zero or more slashes.

Will the URL ever be anything other than `uitspraken.rechtspraak.nl` or `deeplink.uitspraken.rechtspraak.nl`? If not, you can just do

```suggestion
	"target": "^https?://(deeplink\\.)?uitspraken\\.rechtspraak\\.nl/",
```

AbeJellinek (changes requested):

AbeJellinek:
OK, makes sense. What we have right now works, then. See a suggestion for the target regex above.

Rechtspraak.nl.js
pjrwees: deeplink.rechtspraak.nl just redirects to uitspraken.rechtspraak.nl, so we do not need to add that to the regex. I have simply removed the `*`.

AbeJellinek:
Thanks!

PR #2326, opened 2021-01-22
Trim off the (Published XXXX) suffix from NYTimes.com titles

matthewlw:
This is currently causing most of the NYTimes.com tests to fail; the suffix is indeed present in the HTML title, but isn't properly speaking part of the article title.


dstillman:
Thanks for bringing this to our attention. I've fixed it in a slightly cleaner way.

PR #2328, opened 2021-01-28
eLife: Deal with campaign links

### LINKED ISSUE #2133: eLife - campaign links break translator
When using Chrome [Version 80.0.3987.122 (Official Build) (64-bit)], and Zotero 5.0.82, I can't import papers from eLife. The save to Zotero button works, I get a saving to [some collection], but then nothing appears below and nothing is saved. I believe the translator needs to be updated.

mstimberg:
This fixes the eLife translator for links directly taken from the eLife newsletter, e.g.
https://elifesciences.org/articles/54967?utm_source=content_alert&utm_medium=email&utm_content=fulltext&utm_campaign=24-August-20-elife-alert

Closes #2133

### DIFF
--- a/eLife.js
+++ b/eLife.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsbv",
-	"lastUpdated": "2018-09-27 13:36:20"
+	"lastUpdated": "2021-01-28 14:40:53"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


eLife.js
```diff
@@ -84,7 +84,7 @@ function doWeb(doc, url) {
 
 
 function scrape(doc, url) {
-
+	url = url.split('?')[0] // Remove URL parameters like ?utm_source
 	var risURL = url.replace(/#.+/, "") + ".ris";
```
adam3smith, line 88: why not just add this here instead, so `url.replace(/([#?].+/, "") + ".ris";`?

mstimberg: Ah, sure why not. I only had a cursory look and thought that the url was also used to construct the PDF url, but that isn't the case.

adam3smith:
Thanks!
@adomasven @dstillman Looks like not getting any CI on this? This was small enough to just merge, but would be very helpful for larger ones.

PR #2335, opened 2021-02-13
Fix for National Archives of Australia barcode change

wragge:
The NAA database recently changed one of their labels from 'Barcode' to 'Item ID'. This fixes the affected scrapers.

### DIFF
--- a/National Archives of Australia.js
+++ b/National Archives of Australia.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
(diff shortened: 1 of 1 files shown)


wragge:
I have a lot of trouble getting tests to work for this because of the NAA's weird session and redirect stuff. I've deleted a few of the tests that we're having problems and left the one that doesn't have to go through the session manager.

wragge:
The tests were out of date as well, so I added a new batch. Can't get them to work on Scaffold (because of sessions or redirects on the NAA site I think), but they worked here.

dstillman:
Thanks!

PR #2336, opened 2021-02-13
Adding Victoria & Albert museum translator

atiro:
New translator for new collections site for the V&A Museum (https://collections.vam.ac.uk/)


Victoria & Albert Museum.js
```diff
@@ -0,0 +1,198 @@
...
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+function detectWeb(doc, url) {
+	if (url.includes("/item") != -1) {
```
AbeJellinek, line 39: Should just be `if (url.includes("/item")) {` - `includes` returns a boolean.

Victoria & Albert Museum.js
```diff
@@ -0,0 +1,198 @@
...
+		return "artwork";
+	}
+	else if (url.includes("/search") != -1) {
```
AbeJellinek, line 42: as above

Victoria & Albert Museum.js
AbeJellinek: Also make this `"/item/"`, because the regex match at the top of `scrape` will throw an error if there's no trailing slash in the URL. Not sure that can ever happen, but consistency is good.

Victoria & Albert Museum.js
```diff
@@ -0,0 +1,198 @@
...
+	var apiUrl = "https://api.vam.ac.uk/v2/museumobject/" + id;
+	Zotero.Utilities.doGet(apiUrl, function (text) {
+		var museumobject = JSON.parse(text);
+		var data = museumobject.record;
+		var item = new Zotero.Item("artwork");
+		if (data.titles.length > 0) {
+			item.title = data.titles[0].title;
+			item.shortTitle = data.objectType;
+		}
+		else if (data.objectType !== "") {
```
AbeJellinek, line 60: Can just be `if (data.objectType)`. Also might want to have an `else` case where you log an error early, because otherwise translation will only fail at the very end when you call `item.complete()` without setting a title.

Victoria & Albert Museum.js
```diff
@@ -0,0 +1,198 @@
...
+			item.title = data.objectType;
+			item.shortTitle = data.objectType;
+		}
+
+		item.abstractNote = data.briefDescription;
+		item.medium = data.materialsAndTechniques;
+		item.artworkSize = data.dimensionsNote;
+		if (data.productionDates.length > 1) {
+			item.date = data.productionDates[0].date.text + " and further dates";
```
AbeJellinek, line 69: The "and further dates" suffix shouldn't be there, and dates should be cleaned with `ZU.strToISO()`. That method removes prefixes like "ca.", but citations [will export correctly](https://forums.zotero.org/discussion/65723/how-works-is-uncertain-date) if you add the uncertain date to `item.extra` in the form `issued: circa 1994`. `circa` has to be spelled out.

Victoria & Albert Museum.js
```diff
@@ -0,0 +1,198 @@
...
+		item.callNumber = data.accessionNumber;
+		item.rights = data.creditLine;
+		item.url = url;
+
+
+		var artistMakerPerson = data.artistMakerPerson;
+		for (var i = 0; i < artistMakerPerson.length; i++) {
+			// This is not ideal and assume surname, firstnames convention is correct
+			var fullName = artistMakerPerson[i].name.text;
+			var lastComma = fullName.lastIndexOf(",");
```
AbeJellinek, line 86: This logic can be replaced by `ZU.cleanAuthor(fullName, "author", fullName.includes(","))`.

Victoria & Albert Museum.js
```diff
@@ -0,0 +1,198 @@
...
+				fieldMode: "1",
+				creatorType: "author"
+			});
+		}
+
+		var artistMakerOrganisations = data.artistMakerOrganisations;
+		for (var k = 0; k < artistMakerOrganisations.length; k++) {
+			item.creators.push({
+				lastName: artistMakerOrganisations[k].name.text,
+				fieldMode: "1",
```
AbeJellinek, line 118: as above

Victoria & Albert Museum.js
```diff
@@ -0,0 +1,198 @@
...
+				firstName: firstName,
+				creatorType: "author"
+			});
+		}
+
+		var artistMakerPeople = data.artistMakerPeople;
+		for (var j = 0; j < artistMakerPeople.length; j++) {
+			item.creators.push({
+				lastName: artistMakerPeople[j].name.text,
+				fieldMode: "1",
```
AbeJellinek, line 109: `fieldMode` should just be the integer `1`

Victoria & Albert Museum.js
```diff
@@ -0,0 +1,198 @@
...
+			item.creators.push({
+				lastName: artistMakerOrganisations[k].name.text,
+				fieldMode: "1",
+				creatorType: "author"
+			});
+		}
+
+		item.attachments.push({
+			url: url,
+			title: "Explore the Collections (snapshot)",
```
AbeJellinek, line 125: Should just be called "Snapshot".

Victoria & Albert Museum.js
```diff
@@ -0,0 +1,198 @@
...
+	var id = url.match(/item\/(O[0-9]+)/i)[1];
+
+	var apiUrl = "https://api.vam.ac.uk/v2/museumobject/" + id;
+	Zotero.Utilities.doGet(apiUrl, function (text) {
+		var museumobject = JSON.parse(text);
+		var data = museumobject.record;
+		var item = new Zotero.Item("artwork");
+		if (data.titles.length > 0) {
+			item.title = data.titles[0].title;
+			item.shortTitle = data.objectType;
```
AbeJellinek, line 58: I don't think `objectType` should be the `shortTitle` when it rarely seems to have much to do with the full `title`.

AbeJellinek (changes requested):
Hey, thanks for contributing this, and sorry I'm just getting to review it now! I left some comments - let me know what you think.

AbeJellinek:
@atiro, any progress? If you're not available, I can work on getting this ready to merge, but you'll need to [give me access](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).

Victoria & Albert Museum.js
atiro: I was trying to indicate with this that the first date isn't necessarily "the date", especially for copies of other objects when the first date is the date of the original. Did you mean others dates could be added to the item.extra field ?

Victoria & Albert Museum.js
atiro: Many objects don't have an artist assigned title, so object type (Vase, Chair, etc) is what is used as an alternative. Maybe should work through that logic in the title field instead

Victoria & Albert Museum.js
atiro: Not sure about this one, organisation names can't really be tidied up following any rules ?

Victoria & Albert Museum.js
atiro: Don't quite follow how much this handles

atiro:
Cheers for feedback @AbeJellinek hadn't noticed your response! Have updated obvious ones but uncertain around the dates/author/title fields, have replied with comments on those.

atiro:
Hmm, trying to give you access but not seeing that option on my PR

dstillman:
You might not be able to give access when you use your `master` branch instead of a feature branch. (Always use a feature branch for PRs.)

atiro:
Ah cheers @dstillman , I'll close this one then and create a new PR from a branch

PR #2337, opened 2021-02-18
Add translator for Art Institute of Chicago (artic.edu) artwork pages

nikhiltri:


Art Institute of Chicago.js
```diff
@@ -0,0 +1,222 @@
+{
+	"translatorID": "8e98b11a-5648-42b2-8542-5f366cb953f6",
+	"label": "Art Institute of Chicago",
+	"creator": "nikhil trivedi, Illya Moskvin",
+	"target": "^https?://(www.)?artic.edu/artworks/.*",
```
adam3smith, line 5: This doesn't need the .* at the end but you should escape the other `.`

Art Institute of Chicago.js
```diff
@@ -0,0 +1,222 @@
...
+
+function detectWeb(doc) {
+	if (ZU.xpathText(doc, '//html[contains(@class, "p-artwork-show")]')) {
+		return 'artwork';
+	}
+	return false;
+}
+
+function doWeb(doc) {
+	scrape(doc);
```
adam3smith, line 46: I think it's be great to add multiple import from the artwork search. Doesn't look like that'd be hard.

Art Institute of Chicago.js
```diff
@@ -0,0 +1,222 @@
...
+
+function scrape(doc) {
+	var item = new Zotero.Item('artwork');
+	item.title = ZU.xpathText(doc, '//h1');
+
+	var artists = ZU.xpath(doc, '//meta[contains(@name, "citation_author")]');
+	for (var i = 0; i < artists.length; i++) {
+		var cleaned = artists[i].content.replace(/\(.*\)$/, '').trim();
+		if (cleaned.split(' ').length > 2) {
+			item.creators.push({ lastName: cleaned, creatorType: 'artist', fieldMode: 1 });
```
adam3smith, line 57: Is that really the best approach here? Fair number of artists would seem to have  more than 2 names. How common are corporate creators like `Herman Miller Inc.` below and is there any other way we could detect them?

Art Institute of Chicago.js
```diff
@@ -0,0 +1,222 @@
...
+		}
+		else {
+			item.creators.push(ZU.cleanAuthor(cleaned, 'artist'));
+		}
+	}
+
```
adam3smith, line 63: I think adding a snapshot would be appropriate here

adam3smith (changes requested):
Looks good, thanks! A couple of questions and requests for improvement above

Art Institute of Chicago.js
nikhiltri: Got it.

Art Institute of Chicago.js
nikhiltri: Got it.

Art Institute of Chicago.js
nikhiltri: We struggled a bit with this one. Arguably, forcing names into two distinct "first" and "last" names only really works in a western context: https://www.w3.org/International/questions/qa-personal-names. With a number of international artists in our collection, not to mention organizations as you mentioned, I wonder if it would actually be best to handle all artists the way our `else` block does? I'll do that in the next commit, happy to revisit this if you have other ideas!

Art Institute of Chicago.js
nikhiltri: Sorry, I don't know what a snapshot is!

Art Institute of Chicago.js
AbeJellinek: Let's go with the `else` block logic. `cleanAuthor` tries to be clever and doesn't always succeed, but users are used to the pitfalls of our name-splitting logic. Introducing a more complicated heuristic for this translator alone doesn't really make the user experience better; it'll only lead to more confusion when it fails.

And if `cleanAuthor` gets it wrong, it's one button click in the Zotero client to convert a two-field name to one-field.

Art Institute of Chicago.js
```diff
@@ -0,0 +1,254 @@
...
+	item.url = ZU.xpathText(doc, '//link[@rel="canonical"]/@href');
+
+	item.complete();
+}
+
+function getSearchResults(doc, checkOnly) {
+	var items = {};
+	var found = false;
+	var rows = ZU.xpath(doc, '//ul[@id="artworksList"]/li/a');
+	for (var i=0; i<rows.length; i++) {
```
AbeJellinek, line 101: Add spacing here to make the linter happy! (I think this is the old template's fault.)

AbeJellinek (changes requested):
This looks great and I'd love to get it merged. Left a couple comments inline - let me know what you think.

Art Institute of Chicago.js
nikhiltri: I think I addressed the suggestions you raised. Let me know if you want me to square the commit, rebase, or make any other changes!

AbeJellinek: Great, thanks! It looks like all that's left is to update the tests in Scaffold. It'd be helpful if you could rebase, too. I'll squash when I merge everything in.

AbeJellinek: CI will probably keep failing, because it doesn't deal well with the branch point being too far in the past, but that's fine as long as tests pass locally.

AbeJellinek:
Awesome, thank you. Merged this in! Thanks for bearing with all the changes.

nikhiltri:
Great, thanks so much!

PR #2340, opened 2021-02-20
Added translator for K10plus pica-json data (import) and Staatsbibliothek zu Berlin stabikat.de (web)

Marcel-Tronco:


adam3smith:
Interesting approach to go through the API -- that will always be up to date? And could we extend this to other PICA catalogs? We currently have two existing PICA translators that don't seem to work for the stabikat -- they parse what looks to me like much less structured data, so if we could move them towards picajson, that's be great.

We definitely would want some tests here if that's possible -- since the catalog does have permalinks I would hope it is.

Marcel-Tronco:
Usability of the approach for other pica data sets:
The unapi used is an API provided by the association of libraries the Staatsbibliothek is part in (gbv.de) . At least for the catalogs of those libraries the json-parseing will be very similar. For other libraries I'd need to check that. As the parsing is directed according to a field naming convention (k10plus)  entries there chances are good that libraries that accord to that will have similiar picajson data. 

is the unapi data up to date: generally yes.
I don't know how database updating in stabikat catalogs is organized and if there might be some divergences once in a while, but generally they rely on one unified data set. 

I'll provide some test cases soon.

Marcel-Tronco:
Hi there,

I looked into your question concerning the pica translator and abstracted an import translator that uses the same parsing strategy and can be applied for every catalogue that has pica-json data according to k10plus classification available. The range is broad, but is still limited to german libraries, so it can't replace the website strategy used in the recent pica translator.

Check's seem to have failed but I couldn't figure out why exactly, it works in my ide. I guess, it might be connected to the one of the two translators added is using the other one and calling it, while it is not yet in the repository...

Best m

adam3smith:
(I think the Selenium process currently fails; ignore the CI)

Marcel-Tronco:
Hi there, I just looked back on this pull requests because I wondered if it was merged already -- also because I was looking for maybe contributing some other code. Is there some timetable for that? I just ask to get my expectations straight.

Stabikat.js
```diff
@@ -0,0 +1,255 @@
...
+		});
+	}
+	else {
+		scrape(doc);
+	}
+}
+
+function scrape(doc) {
+	// get the identifaction number necessary for the api call
+	var permalink = attr(doc, ".cnt > table:nth-child(1) > tbody:nth-child(1) > tr:nth-child(4) > td:nth-child(1) > table:nth-child(1) > tbody:nth-child(1) > tr:nth-child(1) > td:nth-child(2) > a:nth-child(1)", "href");
```
AbeJellinek, line 113: This is almost certainly going to break. (It probably already has.) See if there's an ID, class name, or attribute we could match on.

Stabikat.js
```diff
@@ -0,0 +1,255 @@
...
+	var ppn = permalink.slice(permalink.indexOf("PPN=") + 4);
+	var translator = Zotero.loadTranslator('import');
+	translator.setTranslator('041335e4-6984-4540-b683-494bc923057a');
```
AbeJellinek, line 116: Indicate which translator this is calling in a comment.

Stabikat.js
```diff
@@ -0,0 +1,255 @@
...
+	translator.setString(JSON.stringify({ ppn: [ppn]}));
```
AbeJellinek, line 117: Won't be necessary if PICA JSON is a search translator.

Stabikat.js
```diff
@@ -0,0 +1,255 @@
...
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+function attr(docOrElem, selector, attr, index) {
```
AbeJellinek, line 38: This shim should be removed - it's no longer needed.

Stabikat.js
```diff
@@ -0,0 +1,255 @@
+{
+	"translatorID": "599ff9de-2049-4b99-ad67-691bde0df74a",
+	"label": "Stabikat",
+	"creator": "Marcel Klotz",
+	"target": "^https?://(www\\\\.)?stabikat.de/",
```
AbeJellinek, line 5: Remove extra escaped backslash.

Stabikat.js
```diff
@@ -0,0 +1,255 @@
...
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcs",
```
AbeJellinek, line 11: Support all browsers.

K10plus pica-json.js
```diff
@@ -0,0 +1,764 @@
...
+	item.title = item.title ? item.title : ' ';
+	
+	return item;
+}
+
+/** BEGIN TEST CASES **/
+var testCases = [
+	{
+		"type": "import",
+		"input": "{\"ppn\":[\"88511759X\"], \"catalogid\": \"stabikat\"}",
```
AbeJellinek, line 655: This test no longer works on my end.

AbeJellinek (changes requested):
Sorry, big backlog but I'm trying to get through it! For the most part, this looks good to me. A few issues, though.

1. Should `K10plus pica-json` really be classified as an import translator? An import translator generally reads a file with all or most of the data needed to create Zotero items and then does so; only in exceptional cases would one need to make HTTP requests. A search translator, though, takes an object with some kind of identifier (like the PICA JSON `ppn`) and fetches the data necessary to create Zotero items from some online repository. For an example of that behavior, check out [ADS Bibcode](https://github.com/zotero/translators/blob/master/ADS%20Bibcode.js) or even [K10plus ISBN](https://github.com/zotero/translators/blob/master/K10plus%20ISBN.js).
2. Should probably be called `K10plus PICA JSON`, since that seems to be the canonical name for the format.

Additional comments inline.

Stabikat.js
```diff
@@ -0,0 +1,255 @@
...
+	}
+	else {
+		scrape(doc);
+	}
+}
+
+function scrape(doc) {
+	// get the identifaction number necessary for the api call
+	var permalink = attr(doc, ".cnt > table:nth-child(1) > tbody:nth-child(1) > tr:nth-child(4) > td:nth-child(1) > table:nth-child(1) > tbody:nth-child(1) > tr:nth-child(1) > td:nth-child(2) > a:nth-child(1)", "href");
+	var ppn = permalink.slice(permalink.indexOf("PPN=") + 4);
```
AbeJellinek, line 114: What if something comes after the PPN in the URL?

Marcel-Tronco:
Hi there! I hope this version resolves the issues.

K10plus PICA JSON.js
```diff
@@ -651,8 +610,8 @@ function generalParser(picaObj, item, hasSuperiorWork, superiorWorkPicaObj, isSu
 /** BEGIN TEST CASES **/
 var testCases = [
 	{
-		"type": "import",
-		"input": "{\"ppn\":[\"88511759X\"], \"catalogid\": \"stabikat\"}",
+		"type": "search",
+		"input": "{\"ppn\": \"88511759X\"}",
```
AbeJellinek, line 614: Will this actually work? I'm not sure it should be a string, since detectSearch/doSearch expect an object.

K10plus PICA JSON.js
```diff
@@ -699,8 +658,8 @@ var testCases = [
 		]
 	},
 	{
-		"type": "import",
-		"input": "{\"ppn\":[\"1701681854\", \"364349824\"]}",
+		"type": "search",
+		"input": "{\"ppn\":\"1701681854\"}",
```
AbeJellinek, line 662: Same here.

AbeJellinek (changes requested):
Thanks! More comments below.

K10plus PICA JSON.js
dstillman: Scaffold doesn't support search translators at all, so these tests wouldn't be doing anything now.

AbeJellinek: Yeah, true, although you can run them with `testTranslators.html` in the browser connector (and they run automatically on CI). I think it's worthwhile to keep them - we want to have some kind of indicator of whether the translator is still working in the future.

dstillman: Ah, OK, yeah, I was about to ask that. We definitely want them to work, then.

Marcel-Tronco:
Yeah, so I fixed the inputs. Sadly I couldn't test them myself as the testTranslators.html has not worked. Using `chrome://zotero/content/tools/testTranslators/testTranslators.html` as indicated on the website didn't resolve and using the path to the local repo threw an error `testResults.json could not be loaded (XMLHttpRequest returned 0).`

AbeJellinek:
Yeah, you have to have a development version of the connector installed. Kind of a pain for one-off translator development particularly - it's something we're working on.

All looking good to me. I guess on second thought this could be considered both a search translator _and_ an import translator, but for now I think it's reasonable to assume that it'll only ever be called by other translators, as a search translator. It's rare, I would imagine, to have PICA JSON lying around on your hard drive that you'd like to import into Zotero.

I'm pretty much ready to merge this. I am curious about one test thing, though: what's with Jutta Georg-Lauer being listed twice (as an editor and an author) in the second test? If it's an unavoidable problem with their metadata, that's fine, but it might be a sign of a bug in the conversion.

Marcel-Tronco:
It's specific to the entry. This is a book section authored by J. G-L that is part of a book of different authors where J.G-L and one other person are both the editors. You can see the library perma link here: [https://opac.k10plus.de/DB=2.299/PPNSET?PPN=1701681854&PRS=HOL&HILN=888&INDEXSET=21](url)
Best M.T.

PR #2341, opened 2021-02-21
Add online document link to beck-online.js

meteoric122:
@zuphilip
Would it be possible to add the link to the online document to the beck-online.js tranlator?

### DIFF
--- a/beck-online.js
+++ b/beck-online.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcs",
-	"lastUpdated": "2019-06-03 15:12:35"
+	"lastUpdated": "2021-02-21 16:18:09"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


adam3smith:
If this is a direct link to the document, would it make sense to put it into URL rather than attaching as a link -- possible that I'm missing something, I don't have access to the site, just making sure this goes in the right place. 
Either way, if you could please also update tests?

meteoric122:
For some reason, the Juris translator (Juris is a very similar site on german law resources) adds the link as an attachment. But I agree, it's more sensible to put it into the URL.

Tests are updated, except for one which I don't have access to:
https://beck-online.beck.de/default.aspx?vpath=bibdata/ents/lsk/2014/3500/lsk.2014.35.0537.htm&pos=1

AbeJellinek:
Looks good. I don't have access to the service in general, or to that out-of-date test's target document in particular, but it's not really worth holding this up over. Pretty clean, simple, and useful change, so I'm going to go ahead and merge. Sorry this took so long to get to!

AbeJellinek:
Thanks so much for your contribution!

PR #2342, opened 2021-02-26
Update dLibra translator.

kgronowski:
Added support for dLibra 6.

### DIFF
--- a/dLibra.js
+++ b/dLibra.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2021-02-26 11:45:18"
+	"lastUpdated": "2021-07-21 11:39:27"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


AbeJellinek:
Hey, thanks for contributing this! Sorry I'm just getting to it now.

As the site's HTML says:

```html
<!--type is "document" for import all tags into zotero-->
```

But it's kind of a cop-out to import every item as `document`, and I think we can do better :) It looks like the dLibra 5 translator used the RDF - why aren't we doing that with dLibra 6? [The RDF they give us](https://demo.dl.psnc.pl/dlibra/rdf.xml?type=e&id=1243) has the correct item type. It's in Polish, so the translator will have to have a Polish -> Zotero item type mapper, but that's not a big deal.

dLibra.js
```diff
@@ -1,135 +1,228 @@
...
+		translator.setHandler("itemDone", function (obj, item) {
+			if (item.extra) item.notes.push(item.extra);
+			item.extra = "";
+			item.itemID = "";
+			item.complete();
 		});
-		
-	})
+		translator.getTranslatorObject(function (trans) {
+			trans.defaultUnknownType = 'document';
```
AbeJellinek, line 117: map item types instead of doing this

AbeJellinek (changes requested):

AbeJellinek:
Trying to test but the demo site is returning an internal server error. The first site in the tests works but looks to be dLibra 5.

kgronowski:
Thanks for the notice. dLibra 6 demo should be working fine now.

dLibra.js
```diff
@@ -1,135 +1,282 @@
...
+					}
+				],
+				"notes": [],
+				"seeAlso": [],
+				"url": "http://cyfrowa.bibliotekakolbuszowa.pl/dlibra/docmetadata?id=945"
+			}
+		]
+	},
+	{
+		"type": "web",
```
AbeJellinek, line 223: These three tests for multiples aren't working.

AbeJellinek (changes requested):
Thanks! A few more comments.

AbeJellinek:
Thanks so much! Linted and merged.

PR #2343, opened 2021-03-01
New Translator: SMW SemanticCite Exporter

rubenvandeven:
Dear maintainers,

I created an exporter to the [SemanticCite](https://github.com/SemanticMediaWiki/SemanticCite) format for Semantic Media Wiki, allowing me to expand my Zotero library through linked data.

The exporter is largely indebted to the work of @simonster @karnesky and @retorquere on the default BibTex translator. It integrates with SemanticCite's BibTex support.

Please also let me know what would be necessary to have this translator pulled into your repo. Furthermore, what would be needed to have it supported through the Zotero API? It would be really great to be able to update the Wiki in an automated way.

Many thanks for all the work on Zotero.


dstillman:
I'm not familiar with SemanticCite, but could you say more about the desired format here? If you simply need to post-process Zotero's native BibTeX export, there would be better ways to do that than duplicating most of the stock BibTeX translator, which would require a huge amount of duplicated code and wouldn't benefit from any future BibTeX fixes.

rubenvandeven:
Dear Dan, good point. I agree with the sentiment that it is not too elegant. Let me explain how I came to this: Semantic Cite is a way to use bibliographic references in Semantic Mediawiki. Aside from its own format it supports wrapped Bibtex imports, which is great. However, the Mediawiki platform uses the curly brackets `{{}}` as tags to call functions. So any occurence of  curly brackets in the  output screws up the parsing of the bibliography... Bibtex tends to use these a lot, instead of quotation marks. As this is not configurable in the Bibtex exported, I adapted the script to use quotation marks  instead of curly brackets.

I could have written a whole new exporter that uses the native SemanticCite format, but that seemed more work _and_ more error prone than re-using the existing Bibtex codebase. Let me know if you feel another approach to this exporter would be preferred.

AbeJellinek:
>As this is not configurable in the Bibtex exported, I adapted the script to use quotation marks instead of curly brackets.

Hmm. Are Semantic Cite quotes equivalent to BibTeX braces?

For example, one of the BibTeX tests includes `title = {{The physical volcanology of the 1600 eruption of Huaynaputina, southern Peru}}`. The braces around the value ensure that it will be displayed with its original capitalization. They won't be displayed in the output.

If you instead outputted that line as `title = {"The physical volcanology of the 1600 eruption of Huaynaputina, southern Peru"}`, how would Semantic Cite import it?

If the quotes would be displayed, I'm not sure you want that. You might just be able to get away with loading the `BibTeX` translator, calling `getTranslatorObject()`, and setting `trans.caseProtectedFields = []`. I believe (please check me on this if I'm wrong) that that should eliminate the double braces.

retorquere:
Eh -- what am I saying. You can just apply https://retorque.re/zotero-better-bibtex/installation/preferences/advanced/#apply-case-protection-to-capitalized-words-by-enclosing-them-in-braces

dstillman:
@retorquere: Not really sure what you're saying here. This is a contribution to the main translators repo, so it wouldn't have anything to do with BBT.

In any case, we really don't want to bundle an extra 130KB translator that's mostly a duplicate of the built-in BibTeX translator, so it would be very good to find a way to do this by post-processing the stock BibTeX output.

retorquere:
Sorry -- I just meant to say that what was attempted to be achieved here was possible with stock BBT. Or is this meant to be used in the translation server?

I got a mail notification so I assumed I was asked to chime in, but I don't see a mention here.

AbeJellinek:
I think it's just because you're mentioned in the original PR description!

You have a good point - maybe Semantic Cite should just recommend Better BibTeX with that setting enabled rather than adding a whole new translator.

AbeJellinek:
Going to close this, sorry. Better methods than adding a new translator have been suggested in this thread.

PR #2345, opened 2021-03-04
PICA: Adding links to the online document

eonm-abes:
This PR adds links to online document for PICA based catalogs. Links are stored in the url field of Zotero.

__About tests :__

* For links extraction : I added tests for all PICA catalogs already cited in the tests.
* When possible I updated the tests. I keep three of them untouched because of errors (these error are not related to link extraction but to UI changes) :
http://cbsopac.rz.uni-frankfurt.de/DB=2.1/PPNSET?PPN=318490412
http://gso.gbv.de/DB=2.1/PPNSET?PPN=768059798
* I deleted tests for catalogue.rug.nl which seems not  to be active anymore. It's looks like they're using a new catalog system (https://rug.on.worldcat.org/)

Best regards

### DIFF
--- a/Library Catalog (PICA).js
+++ b/Library Catalog (PICA).js
@@ -1,7 +1,7 @@
 {
 	"translatorID": "1b9ed730-69c7-40b0-8a06-517a89a3a278",
 	"label": "Library Catalog (PICA)",
-	"creator": "Sean Takats, Michael Berkowitz, Sylvain Machefert, Sebastian Karcher, Stéphane Gully",
+	"creator": "Sean Takats, Michael Berkowitz, Sylvain Machefert, Sebastian Karcher, Stéphane Gully, Mathis Eon",
 	"target": "^https?://[^/]+(/[^/]+)*//?DB=\\d",
 	"minVersion": "3.0",
 	"maxVersion": "",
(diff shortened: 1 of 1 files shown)


AbeJellinek:
This looks good to me. In the future, please prefix commits with the translator's name (like you did in the PR title)! Not a big issue, though.

AbeJellinek:
Actually, on second thought - changes look technically good, but is there precedent for this? A glance through other library translators isn't turning up any that store online resource URLs in the Zotero item. Could you describe why it's important here?

adam3smith:
Yes, we do this where we know a URL is an actual document/full text, see e.g.
https://github.com/zotero/translators/blob/master/Primo%20Normalized%20XML.js#L280

AbeJellinek:
Ah, OK, great. Thanks for clarifying.

In that case, I'll merge this in and look into fixing the two now-incompatible catalogs in the tests.

@eonm-abes, thanks for your contribution!

eonm-abes:
Hi @AbeJellinek,

Sorry for the late reply, I didn't see GitHub notifications.

> This looks good to me. In the future, please prefix commits with the translator's name (like you did in the PR title)! Not a big issue, though.

I will try to remember that!

Thank you

PR #2346, opened 2021-03-06
Fix Wired translator

danielzgtg:


dstillman:
Thanks! And sorry for the delay.

For future translator updates, be sure to use the Translator Editor, which can run/update tests and will also automatically update the timestamp (which is necessary for the change to be pushed to clients).

danielzgtg:
> be sure to use the Translator Editor

Oh, I didn't realize there was one. I found it now, and it looks like it would have made things much easier for me

PR #2351, opened 2021-03-11
Map Bibtex "@misc" to Document type

buddhist-uni:
Book is certainly the wrong type for a "misc" bibtex item, as the Bibtex item (whatever it actually is) would have had "book" or "booklet" or "manual" type if it were truly a book. Since Zotero has a "miscellaneous" type, why not use it?

### DIFF
--- a/BibTeX.js
+++ b/BibTeX.js
@@ -226,7 +226,7 @@ var bibtex2zoteroTypeMap = {
 	"booklet":"book",
 	"manual":"book",
 	"mastersthesis":"thesis",
-	"misc":"book",
+	"misc":"document",
 	"proceedings":"book",
 	"online":"webpage",
 	// alias for online from BibLaTeX:


adam3smith:
@retorquere -- I'm sure you're doing more sophisticated things with `@misc` in BBT, but do you have a default? Makes sense to me to switch this away from book, which is indeed a rather puzzling choice

retorquere:
My default for `@misc` is `journalArticle`.

retorquere:
Zotero has a miscellaneous type?

adam3smith:
I think it's fair to describe "Document" as miscellaneous, yes. We used to discourage its use because it wasn't mapped in CSL, but it now goes to CSL `article` and will likely go to CSL `document` in CSL 1.0.2, so that no longer applies.

AbeJellinek:
Merged! Thanks for contributing this. `document` is definitely a better mapping than `book`.

PR #2353, opened 2021-03-11
EBSCO Hotfix for discovery layer

adam3smith:
https://forums.zotero.org/discussion/88250/zotero-connector-does-not-offer-all-sources-in-database-results#latest

### DIFF
--- a/EBSCOhost.js
+++ b/EBSCOhost.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsib",
-	"lastUpdated": "2018-12-26 11:18:52"
+	"lastUpdated": "2021-03-11 13:47:13"
 }
 
 function detectWeb(doc, url) {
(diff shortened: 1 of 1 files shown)


adam3smith:
Question @adomasven @dstillman  -- is it possible to disable the push test, since that will always fail?

dstillman:
For specific translators, you mean?

dstillman:
We could add something to the translator metadata and have CI look for it.

adam3smith:
No, I'm always seeing two tests:
image

The (pull_request) one is the real test and that should always run. The (push) one always fails with a version of

```
fatal: ambiguous argument '0a03ebbf4cbf1a1b8140f11f65b2337fc5632eca^^!': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
Error: Process completed with exit code 128.
```

(and, emails that to me). I don't see any reason for that to ever run unless I'm missing something

adam3smith:
See also the above -- the relevant test now passes, but PR is still marked as failing in the overview

PR #2365, opened 2021-04-08
Added GameStar and GamePro

Mailaender:
They are sister magazines from the same publisher and their website layout is pretty much identical. I only found differences between older articles and newer ones, so there are different ways to scrape the correct author.

### DIFF
--- a/GameStar GamePro.js
+++ b/GameStar GamePro.js
@@ -1,0 +1,199 @@
+{
+	"translatorID": "c87a910d-70aa-4f33-9bec-f703d63ba84f",
+	"label": "GameStar/GamePro",
+	"creator": "Matthias Mailänder",
+	"target": "^https?://(www\\.)?game(star|pro)\\.de",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2021-06-01 18:22:07"
+}
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2021 Matthias Mailänder
+
+	This file is part of Zotero.
+
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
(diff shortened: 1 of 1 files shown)


GameStar GamePro.js
```diff
@@ -0,0 +1,160 @@
...
+		return "magazineArticle";
+	}
+	return false;
+}
+
+function doWeb(doc, url) {
+	if (detectWeb(doc, url) == "magazineArticle") {
+		scrape(doc, url);
+	}
+}
```
AbeJellinek, line 50: Could you add a `multiple` case? The "web translator template" generated by Scaffold will give you 95% of the code you need - you'll probably just have to tweak the selectors.

Mailaender: If `multiple` means grabbing search results, then I don't see the purpose of those.

AbeJellinek: A lot of the time, neither do I. But almost every Zotero web translator supports them, unless its target site doesn't have useful search results. That's not the case here, so they should be supported. It'll seriously be ten simple lines of code from the template.

Mailaender: So essentially, you want me to write a parser for stuff like https://www.gamestar.de/suche/?query=grand+theft+auto just like
https://github.com/zotero/translators/blob/2b2de18fad702214f00c292d394c83f644f4e7e5/PC%20Gamer.js#L76-L90
Sadly the HTML/CSS on that site is a lot messier.

Mailaender: I used 
https://github.com/zotero/translators/blob/1d1ce6111ac17a4064c1d8836f6beec54404c32b/PC%20Gamer.js#L76-L90
as a reference and copied the selectors using the Chrome developer tools.

Mailaender:
Also cleared the warnings.

AbeJellinek:
@Mailaender Awesome. Thanks for doing that. I think I found a selector that's a bit more resilient to changes in the site's layout (or different ad blockers, etc.). Pushed a commit. Check it out and let me know what you think.

Mailaender:
Yes, the CSS was a bit messy. Thanks.

AbeJellinek:
Merged! Thanks so much for your contribution.

PR #2371, opened 2021-05-17
Add a translator for womennews.co.kr

saschanaz:


AbeJellinek:
Tests all pass locally (won't work on CI without a rebase) and code is good. Merging. Thanks for your contribution!

PR #2372, opened 2021-05-18
Add translator for Daum News

saschanaz:


AbeJellinek:
Looks good, thanks! Merged.

It would be nice to support multiples, but since the search box just sends you to Daum's search engine on another domain, it would be out of the scope of this translator.

PR #2373, opened 2021-05-18
Add Jewish Life in America, fix Gale, pull PDFs from Project MUSE

AbeJellinek:
Some initial translator forays.

- I added a translator for Jewish Life in America, an Adam Matthew Digital archive digitization project. Other AM Digital databases have a similar structure, but they're not identical, and I want to keep the scope of this translator small to start.
- The Gale translator wasn't working anymore; it seems like it relied upon some now-defunct metadata fields. I rewrote it to lean more heavily upon the RIS they give us. This generally works well, although `detectWeb` has to do a bit more guesswork than I'd like in order to figure out the item type.
- Project MUSE gives an individual PDF for each chapter, the files are pretty small (200-300 kB), and it's nice not to have to download each one manually, so the translator now adds them as attachments.


adam3smith:
Could you split these into individual PRs? Much easier to review that way.
(Also, are you working on translators for Zotero now? If so, welcome!

AbeJellinek:
Sure, I'll split it up. And yes, I am! Thank you!

AbeJellinek:
Alright, spun Gale and MUSE out into #2374 and #2375.

Adam Matthew Digital.js
```diff
@@ -142,11 +142,14 @@ function scrapeNew(doc, url) {
...
-		url: attr(doc, 'a[href*="/FullDownload"]', 'href')
-	});
+	let pdfUrl = attr(doc, 'a[href*="/FullDownload"]', 'href');
+	if (pdfUrl) {
+		item.attachments.push({
+			title: "Full Text PDF",
+			mimeType: 'application/pdf',
+			url: pdfUrl
+		});
+	}
```
dstillman, line 152: This is actually worth debating. I'm not sure to what extent blindly assigning download URLs and letting them fail if inaccessible vs. detecting their availability ahead of time was a conscious policy decision.

If there's a PDF but you don't have access to it, is it more user-friendly to show it with a red X or not show it at all?

(Curious what @adam3smith thinks of this.)

dstillman: Just to lay out a few obvious arguments:

1. If you get an X, it shows that you're missing something that's available, and you might be able to take steps (e.g., reconnect through your proxy) to remedy that.
1. If you never have access, it's annoying to get what looks like an error all the time.
1. It's sort of nice to show that Zotero could access something if you had access.

adam3smith: I don't think we have consistent behavior from a user perspective in translators and it doesn't seem to have ever caused any confusion. Basically, we don't go out of our way to determine whether a user has access to a PDF (which wouldn't always be possible reasonably anyway), but we also don't try to retrieve a PDF where we know it's going to fail (e.g., when we'd have to make up a URL that's only available via DOM). 

I haven't thought this through in detail, but one nice feature of this approach is that the red x is technically meaningful, i.e. it gives us some indication of where things fail (and which debug we'd need) for troubleshooting.

PR #2374, opened 2021-05-18
Fix Gale translator

AbeJellinek:
The Gale translator wasn't working anymore; it seems like it relied upon some now-defunct metadata fields. I rewrote it to lean more heavily upon the RIS they give us. This generally works well, although `detectWeb` has to do a bit more guesswork than I'd like in order to figure out the item type.

### DIFF
--- a/Gale Databases.js
+++ b/Gale Databases.js
@@ -1,7 +1,7 @@
 {
 	"translatorID": "e3748cf3-36dc-4816-bf86-95a0b63feb03",
 	"label": "Gale Databases",
-	"creator": "Jim Miazek",
+	"creator": "Abe Jellinek and Jim Miazek",
 	"target": "^https?://[^?&]*(?:gale|galegroup|galetesting|ggtest)\\.com(?:\\:\\d+)?/ps/",
 	"minVersion": "3.0",
 	"maxVersion": "",
@@ -9,13 +9,14 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2019-12-25 09:35:52"
+	"lastUpdated": "2021-05-19 17:08:22"
 }
 
 /*
 	***** BEGIN LICENSE BLOCK *****
 
-	Gale Databases Translator - Copyright © 2019
+	Copyright © 2021 Abe Jellinek and Jim Miazek
+	
 	This file is part of Zotero.
 
 	Zotero is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


Gale Databases.js
```diff
@@ -25,93 +26,222 @@
...
-		.replace(/^SP\s+-\s+NA\n/gm, '') // Remove missing page numbers
-		.replace(/^N1(?=\s+-\s+copyright)/igm, 'CR');
-}
-
-function getCitableDocuments(doc) {
-	return doc.getElementsByClassName('zotero');
+	else if (url.includes('Search.do') && getSearchResults(doc, true)) {
+		return "multiple";
+	}
+	else if (doc.querySelector('a[data-gtm-feature="bookView"]')) {
```
dstillman, line 46: Can remove the `else` after `return`

Gale Databases.js
```diff
@@ -25,93 +26,222 @@
...
-	}
-	var entries = getCitableDocuments(doc);
-	switch (entries.length) {
-		case 0: return false;
-		case 1: return entries[0].getAttribute('data-zoterolabel');
-		default: return 'multiple';
+function getSearchResults(doc, checkOnly) {
+	let items = {};
+	let found = false;
+	let rows = doc.querySelectorAll('h3.title>a.documentLink');
```
dstillman, line 55: Add whitespace around `>` (and we should fix the Scaffold template if it does that)

Gale Databases.js
```diff
@@ -25,93 +26,222 @@
...
+	let mcode = citeData.dataset.mcode; // undefined for bookSections
+	let productName = citeData.dataset.productname;
+	let docId = mcode ? undefined : decodeURIComponent(url.match(/(?:docId|id)=([^&]+)/)[1]);
+	let risPostBody = JSON.stringify([{
+		documentUrl: `<span class="docUrl">${documentUrl}</span>`,
+		mcode,
+		docId,
+		productName
+	}]);
+	let pdfURL = attr(doc, 'button[data-gtm-feature="download"]', 'data-url');
```
dstillman, line 90: Rebase on current master and the lint error should go away

Gale Databases.js
```diff
@@ -25,93 +26,222 @@
...
+						"creatorType": "editor"
+					}
+				],
+				"date": "2005",
+				"archive": "Gale eBooks",
+				"libraryCatalog": "Gale",
+				"place": "Detroit, MI",
+				"publisher": "Gale",
+				"series": "Ancient Egypt 2675-332 B.C.E.",
+				"url": "https://link.gale.com/apps/pub/5BBU/GVRL?u=ucberkeley&sid=GVRL",
```
dstillman, line 160: You said if you remove the `u=` you get a login page. Did you mean a Gale login page or a Berkeley proxy login page?

I know the old translator did this, but it doesn't seem like it makes sense to includes these in citations if they're not usable to others, so we'd only want to save a URL that someone at a different institution could actually use.

(I don't currently have access to test this.)

Gale Databases.js
dstillman: (I'm not sure we don't have this in our eslint config. I can add it.)

Gale Databases.js
AbeJellinek: A Gale login page. It doesn't seem to matter whether the `u=` value matches the institution you're connecting from; I can access the link in the last test with `u=nysl_ce_syr`. (I need to hunt down some links from other institutions to make sure it isn't just that one.) And the links the translator is saving are the same ones that the `Get Link` button on the page gives you, which are supposedly permalinks that you can cite/share.

AbeJellinek: Hmm, now I see that the tests with Berkeley URLs fail on CI but the New York URL doesn't. Odd. I'll find more cases to test with.

PR #2375, opened 2021-05-18
Project MUSE: Download full text PDFs

AbeJellinek:
Project MUSE gives an individual PDF for each chapter, the files are pretty small (200-300 kB), and it's nice not to have to download each one manually, so the translator now adds them as attachments.

### DIFF
--- a/Project MUSE.js
+++ b/Project MUSE.js
@@ -1,7 +1,7 @@
 {
 	"translatorID": "c54d1932-73ce-dfd4-a943-109380e06574",
 	"label": "Project MUSE",
-	"creator": "Sebastian Karcher",
+	"creator": "Sebastian Karcher and Abe Jellinek",
 	"target": "^https?://[^/]*muse\\.jhu\\.edu/(book/|article/|issue/|search\\?)",
 	"minVersion": "3.0",
 	"maxVersion": "",
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2020-12-15 03:03:06"
+	"lastUpdated": "2021-05-18 20:04:42"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


Project MUSE.js
```diff
@@ -50,12 +50,12 @@ function detectWeb(doc, url) {
...
 function getSearchResults(doc, checkOnly) {
-	var items = {};
-	var found = false;
-	var rows = ZU.xpath(doc, '//li[@class="title"]//a[contains(@href, "/article/") or contains(@href, "/book/")]');
-	for (var i = 0; i < rows.length; i++) {
-		var href = rows[i].href;
-		var title = ZU.trimInternal(rows[i].textContent);
+	const items = {};
+	let found = false;
+	const rows = ZU.xpath(doc, '//li[@class="title"]//a[contains(@href, "/article/") or contains(@href, "/book/")]');
```
dstillman, line 55: Revert if just `const`/`let` changes

dstillman (commented):
If you rebase this on master now, the lint errors should go away. (You can also squash the extra commits. We can do for people via GitHub but for an initial PR it's always best to start with a clean commit.)

AbeJellinek:
Lint errors gone! Thanks for making that change.

Tests are failing on the server but not when I run locally. The elements that the translator is querying for are there in the HTML, not injected on the client side, so my best guess is that the CI server is getting served a slightly different page because of different institutional access. Translation still *works*, so it's not a total deal-breaker, but I don't like the inconsistency.

dstillman:
The PDFs not matching makes sense, since the CI wouldn't have access to those, but the tag and abstract changes are weird.

AbeJellinek:
> the tag and abstract changes are weird.

The translator isn't pulling the tags/abstract from the RIS (which doesn't contain them) but from the HTML, and the version that gets served to unauthenticated users has much less information.

Authenticated:

Screen Shot 2021-05-19 at 10 53 13 AM

Unauthenticated:

Screen Shot 2021-05-19 at 10 53 42 AM

For some reason, they give us an excerpt but not an abstract.

I think I prefer not pulling an abstract at all in that case.

PR #2376, opened 2021-05-19
BiblioCommons: Support v2 catalogs

### LINKED ISSUE #766: Expand Library Catalog (BiblioCommons) to work with other websites
See http://www.bibliocommons.com/about/participating-libraries for a list of participating libraries. Might be as easy as tweaking target regexp.

AbeJellinek:
BiblioCommons overhauled the layout of their catalog software at some point in the past few years, and the translator broke. There's a new URL format for records and searches and MARC data is displayed a bit differently. I imagine that there might be some libraries still running an older version, so the translator just checks for a /v2/ URL.

Closes #766: I tried every site that BiblioCommons lists as a partner and could find none that still actually use its catalog software but don't have a *.bibliocommons.com catalog URL.

eslint complains about the V1 translator.

### DIFF
--- a/Library Catalog (BiblioCommons).js
+++ b/Library Catalog (BiblioCommons).js
@@ -1,7 +1,7 @@
 {
 	"translatorID": "5d506fe3-dbde-4424-90e8-d219c63faf72",
 	"label": "Library Catalog (BiblioCommons)",
-	"creator": "Avram Lyon",
+	"creator": "Avram Lyon and Abe Jellinek",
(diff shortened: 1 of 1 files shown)


AbeJellinek:
Now that I think about it, the fact that I couldn't find any non-vendor-hosted BiblioCommons catalogs means that there are unlikely to be many V1 installs still out there. I'll remove the V1 translator if there are no objections!

AbeJellinek:
Merged this, don't ask me why the PR shows as closed (my fault for trying to do the merge locally).

PR #2377, opened 2021-05-20
Wikipedia: Detect and parse mobile articles

### LINKED ISSUE #2366: Recognize mobile wikipedia links as wikipedia
Mobile wikipedia links don't default to being saved via the Wikipedia translator when they should be treated identically.

Desktop link (correctly recognized by the Zotero Firefox Connector as a Wikipedia article):
https://en.wikipedia.org/wiki/1%25_rule_(Internet_culture)

Mobile link (defaults to COinS instead)
https://en.m.wikipedia.org/wiki/1%25_rule_(Internet_culture)

I'm guessing this is just an issue with the target regex over at https://github.com/zotero/translators/blob/master/Wikipedia.js#L5

Anyone with more knowledge able to confirm this?

AbeJellinek:
Fixes #2366

### DIFF
--- a/Wikipedia.js
+++ b/Wikipedia.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2021-05-20 00:12:40"
+	"lastUpdated": "2021-05-24 15:30:53"
 }
 
 /**
@@ -31,6 +31,8 @@
 */
 
 function detectWeb(doc, url) {
+	// on desktop, the article title is in #firstHeading.
+	// on mobile, it's #section_0.
 	if (doc.getElementById('firstHeading') || doc.getElementById('section_0')) {
 		return 'encyclopediaArticle';
 	}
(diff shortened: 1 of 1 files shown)


Wikipedia.js
```diff
@@ -210,6 +223,35 @@ var testCases = [
...
+				"title": "1% rule (Internet culture)",
+				"creators": [],
+				"date": "2021-01-11T20:19:42Z",
+				"abstractNote": "In Internet culture, the 1% rule is a rule of thumb pertaining to participation in an internet community, stating that only 1% of the users of a website add content, while the other 99% of the participants only lurk. Variants include the 1–9–90 rule (sometimes 90–9–1 principle or the 89:10:1 ratio), which states that in a collaborative website such as a wiki, 90% of the participants of a community only consume content, 9% of the participants change or update content, and 1% of the participants add content. This also applies, approximately, to Wikipedia.Similar rules are known in information science; for instance, the 80/20 rule known as the Pareto principle states that 20 percent of a group will produce 80 percent of the activity, however the activity is defined.",
+				"encyclopediaTitle": "Wikipedia",
+				"extra": "Page Version ID: 999756024",
+				"language": "en",
+				"libraryCatalog": "Wikipedia",
+				"rights": "Creative Commons Attribution-ShareAlike License",
+				"url": "/w/index.php?title=1%25_rule_(Internet_culture)&diff=prev&oldid=999756024",
```
dstillman, line 242: Shouldn't be a relative URL

Wikipedia.js
```diff
@@ -31,14 +31,14 @@
 */
 
 function detectWeb(doc, url) {
-	if (doc.getElementById('firstHeading')) {
+	if (doc.getElementById('firstHeading') || doc.getElementById('section_0')) {
```
dstillman, line 34: Let's document when these two cases are used

Wikipedia.js
```diff
@@ -69,14 +71,24 @@ function doWeb(doc, url) {
...
-	if (item.url) {
-		revID = item.url.match(/[&?]oldid=(\d+)/)[1];
+	if (permalink) {
+		revID = permalink.match(/[&?]oldid=(\d+)/)[1];
 		item.extra = 'Page Version ID: ' + revID;
-		item.url = doc.location.protocol + '//' + doc.location.hostname
-					+ item.url;
-	} else {
+		if (permalink.startsWith('/')) {
+			item.url = doc.location.protocol + '//' + doc.location.hostname
```
dstillman, line 83: Wikipedia is always HTTPS, so it won't make a difference, but should just hard-code this to `https://` for cleanliness. (Should always just do that unless a site is known not to support it, which is rare (and inexcusable).)

PR #2378, opened 2021-05-20
Aleph: Allow missing slash when detecting MAB2 URL; linting

### LINKED ISSUE #2255: Kubikat (Aleph) Fails to Import
User reports this used to work and MARC is present on the page:
Reported: https://forums.zotero.org/discussion/85358/issue-with-translator-kubikat#latest
Example: https://forums.zotero.org/discussion/85358/issue-with-translator-kubikat#latest

AbeJellinek:
Fixes #2255.

### DIFF
--- a/Library Catalog (Aleph).js
+++ b/Library Catalog (Aleph).js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsb",
-	"lastUpdated": "2016-12-17 23:52:01"
+	"lastUpdated": "2021-05-21 03:54:32"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


Library Catalog (Aleph).js
```diff
@@ -302,76 +302,6 @@ function scrape(newDoc, marc, url) {
...
-				],
-				"notes": [
-					{
-						"note": "Gesehen am 12.09.17"
-					}
-				],
-				"seeAlso": []
-			}
-		]
-	},
```
dstillman: Are there permalinks for these that we can use instead? Not sure if there's anything different in these worth testing (looks like these are non-English, for what that's worth), but best to assume that tests were added to cover specific situations.

dstillman: Ah, and I guess the French one is from a different library catalog, so would be good to keep that if possible.

AbeJellinek: I added all the tests! It had none before. I’d like to keep them, but I couldn’t find any way to get a permalink. I’m sure I’m just missing something, so I’ll look back at it in a bit.

dstillman: Ah, sorry, was looking at the later commit in isolation.

I can get a permalink if I tweak the URL manually: http://naude.bibliotheque-mazarine.fr/F/?func=direct&doc_number=000319644

I don't know if there's any way for a user to get to that directly, but I guess no harm in adding that, just to have a different installation (assuming that's how Aleph works) tested.

AbeJellinek: Added! Thanks. I wasn't having success making the link myself.

dstillman:
Looks good. You can squash the last two commits into the first and then merge.

PR #2379, opened 2021-05-21
Add translator for Taylor & Francis eBooks

### LINKED ISSUE #2136: Translator Request: taylorandfrancis.com books and chapters
Requested: https://forums.zotero.org/discussion/81741/translator-request-www-taylorfrancis-com#latest

Book example: https://www.taylorfrancis.com/books/9781315120621
Chapter example: https://www.taylorfrancis.com/books/9781315120621/chapters/10.1201/9781315120621-1

AbeJellinek:
Closes #2136

### DIFF
--- a/Taylor & Francis eBooks.js
+++ b/Taylor & Francis eBooks.js
@@ -1,0 +1,334 @@
+{
+	"translatorID": "71f28f90-764a-42a8-8107-d87252da9b50",
+	"label": "Taylor & Francis eBooks",
+	"creator": "Abe Jellinek",
+	"target": "^https?://(www\\.)?taylorfrancis\\.com/(books|chapters|search)",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2021-05-28 02:42:40"
+}
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2021 Abe Jellinek
+	
+	This file is part of Zotero.
+
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+const FIELDS = {
+	Edition: 'edition',
+	'First Published': 'date',
+	'eBook Published': 'date',
+	Imprint: 'publisher',
+	Pages: 'numPages',
+	DOI: 'extra',
+	'eBook ISBN': 'ISBN',
+	'Pub. Location': 'place'
+};
+
+function detectWeb(doc, url) {
+	if (url.includes('/chapters/')) {
+		return "bookSection";
+	}
+	if (url.includes('/books/')) {
+		return "book";
+	}
+	if (getSearchResults(doc, true)) {
+		return "multiple";
+	}
+	return false;
+}
+
+function getSearchResults(doc, checkOnly) {
+	var items = {};
+	var found = false;
+	var rows = doc.querySelectorAll('search-results-product > a');
+	for (let row of rows) {
+		let href = row.href;
+		let title = ZU.trimInternal(text(row, 'h1'));
+		if (!href || !title) continue;
+		if (checkOnly) return true;
+		found = true;
+		items[href] = title;
+	}
+	return found ? items : false;
+}
+
+function doWeb(doc, url) {
+	let itemType = detectWeb(doc, url);
+	if (itemType == "multiple") {
+		Zotero.selectItems(getSearchResults(doc, false), function (items) {
+			if (items) ZU.processDocuments(Object.keys(items), scrape);
+		});
+	}
+	else {
+		scrape(doc, url, itemType);
+	}
+}
+
+function scrape(doc, url, itemType) {
+	if (!itemType) {
+		itemType = detectWeb(doc, url);
+	}
+
+	let item = new Zotero.Item(itemType);
+	item.title = text(doc, 'h1.product-banner-title');
+	item.libraryCatalog = '';
+	item.extra = '';
+	let subtitle = text(doc, '.product-banner-subtitle');
+	if (subtitle) {
+		item.title += ": " + subtitle;
+	}
+	if (itemType == 'bookSection') {
+		item.bookTitle = text(doc, 'a.product-book-link');
+	}
+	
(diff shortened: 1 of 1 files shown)


Taylor and Francis Group.js
```diff
@@ -0,0 +1,287 @@
...
+			}
+			else {
+				item[field] = value;
+			}
+		}
+	}
+	
+	if (doc.cookie) {
+		let apiId = doc.getElementById('pyramid-website-ssr-state').textContent
+			.match(/&q;_id&q;:&q;([^&]+)&q;/)[1];
```
dstillman, line 133: This is `type="application/json"`, so no need to use a regex to parse. It's not quite JSON on its own, though — you have to replace the `&q;` stuff (which maybe is [an Angular thing](https://stackoverflow.com/a/67047950)?).

This gives me the id in the console:

```javascript
JSON.parse(document.getElementById('pyramid-website-ssr-state').textContent.replace(/&q;/g, '"')).product._id
```

Taylor and Francis Group.js
```diff
@@ -0,0 +1,287 @@
...
+	}
+	
+	for (let creator of text(doc, '.product-banner-author-name').split(", ")) {
+		let creatorType = text(doc, '.product-banner-author').includes('Edited')
+			? 'editor'
+			: 'author';
+		item.creators.push(ZU.cleanAuthor(creator, creatorType, false));
+	}
+	
+	for (let row of doc.querySelectorAll('product-more-details .display-row')) {
```
dstillman, line 113: So they're just…making up tag names? Seems like there are classes, too. Not sure which is more likely to remain stable.

Taylor and Francis Group.js
```diff
@@ -0,0 +1,287 @@
...
+				"creators": [
+					{
+						"firstName": "Leslie R.",
+						"lastName": "Rudnick",
+						"creatorType": "editor"
+					}
+				],
+				"date": "29 June 2017",
+				"ISBN": "9781315120621",
+				"edition": "3rd Edition",
```
dstillman, line 189: Check me on this, but I think the edition needs to be just a number to be properly handled by the citation processor, and otherwise it's passed through as a literal (meaning it won't be localized). If that's the case, when we can parse it cleanly here, we'd want to add it as just an integer.

Taylor and Francis Group.js
```diff
@@ -0,0 +1,287 @@
...
+				"itemType": "book",
+				"title": "Lubricant Additives: Chemistry and Applications",
+				"creators": [
+					{
+						"firstName": "Leslie R.",
+						"lastName": "Rudnick",
+						"creatorType": "editor"
+					}
+				],
+				"date": "29 June 2017",
```
dstillman, line 187: Our date handling/localization is complicated and imperfect, but we generally use `strToISO()` to save parse dates into YYYY-MM-DD.

(Basically, `strToDate()`, which that uses, can parse English month names in any user locale, so that will work here and the user won't have to see English month names and British date order. The ultimate plan is to display cleanly parsed Date values in the user's own locale, but we haven't gotten around to that.)

Taylor and Francis Group.js
```diff
@@ -0,0 +1,287 @@
...
+	
+	for (let row of doc.querySelectorAll('product-more-details .display-row')) {
+		let label = text(row, 'span:first-child');
+		let value = text(row, 'span:last-child');
+
+		let field = FIELDS[label];
+		if (field) {
+			if (field == 'extra') {
+				if (label == "DOI") {
+					value = value.match(/(?:doi\.org\/)?([^/]+\/[^/]+)/)[1];
```
dstillman, line 121: There's a `cleanDOI` utility function for this.

Taylor and Francis Group.js
```diff
@@ -0,0 +1,287 @@
+{
+	"translatorID": "71f28f90-764a-42a8-8107-d87252da9b50",
+	"label": "Taylor and Francis Group",
```
dstillman, line 3: "Taylor & Francis eBooks"? I think "Group" is just the company, and "eBooks" is the relevant product here. It's an ampersand regardless.

Taylor and Francis Group.js
```diff
@@ -0,0 +1,287 @@
...
+
+function scrape(doc, url, itemType) {
+	if (!itemType) {
+		itemType = detectWeb(doc, url);
+	}
+
+	let item = new Zotero.Item(itemType);
+	item.title = text(doc, 'h1.product-banner-title');
+	item.libraryCatalog = '';
+	item.extra = '';
```
dstillman, line 97: Can we get the abstract? Seems like it's present on individual pages, at least. Clicking the Abstract button on search results also seems to show it without a relevant network request, though I'm not sure where the data is coming from.

Taylor and Francis Group.js
```diff
@@ -0,0 +1,287 @@
...
+							});
+							break;
+						}
+					}
+				}
+				
+				item.complete();
+			}, null, null, {
+				Authorization: 'idtoken ' + token,
+				Referrer: 'https://www.taylorfrancis.com/'
```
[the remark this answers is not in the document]
AbeJellinek: Oh, yes. A typo on a typo. It isn't used - I was getting unexplained errors from their server and decided to try setting that header. Shouldn't have been left in.

Taylor and Francis Group.js
AbeJellinek: I was a bit confused by the `&q;` stuff. I'll use that approach.

Taylor and Francis Group.js
AbeJellinek: Angular is [like that](https://angular.io/guide/elements). I think the tag names are more likely to remain stable, just because they describe the semantics rather than style.

Taylor and Francis Group.js
AbeJellinek: Oh, good to know. Will add some code to try to parse it (`cleanEdition` could be a good `Zotero.Utilities` function...).

Taylor and Francis Group.js
AbeJellinek: Lack of ampersand is to match [the other T&F translator](https://github.com/zotero/translators/blob/master/Taylor%20and%20Francis%2BNEJM.js), which also calls it "Taylor and Francis." Is it more important to be consistent or correct? Either way, you're right, "eBooks" would be better than "Group."

Taylor and Francis Group.js
adam3smith: >  I think the edition needs to be just a number to be properly handled by the citation processor, and otherwise it's passed through as a literal

Basically, yes. The tricky part in parsing this is to not remove useful information: e.g. "2nd edition with a forword by the author" should be left as literal, so you don't just want to match on a number.
You also want to typically remove 1st editions (those are often in library catalogs) entirely.

AbeJellinek: Oh, not removing useful information is a good point. I *think* I have it right as of the latest commit: I trim the edition text, discard it if it looks like a first edition, and then try to pull out just the number as long as it's followed only by a suffix and `ed.`/`edition`.

Now that I think about it, maybe I want to reverse the order so that we keep stuff like "1st edition with a foreword by the author." Not a common edge case, but it's possible.

dstillman: Yeah, a `cleanEdition()` seems like a good idea. If you write it as a separate function, we could bundle it and start including it as a shim in other translators.

Taylor and Francis Group.js
dstillman: Correct is better.

AbeJellinek: Done.

PR #2381, opened 2021-05-24
Add translator for Publications Office of the EU

### LINKED ISSUE #768: Add bookshop.europa.eu translator
Both as web and search (for DOI). See #767

Can't find any API, so maybe use http://bookshop.europa.eu/en/search/advanced/ to search for DOI (as Identifier)

AbeJellinek:
Closes #768; bookshop.europa.eu is now op.europa.eu.

### DIFF
--- a/Publications Office of the European Union.js
+++ b/Publications Office of the European Union.js
@@ -1,0 +1,217 @@
+{
+	"translatorID": "7d8e6337-3f52-4e8c-8915-95a2ec755b6c",
+	"label": "Publications Office of the European Union",
+	"creator": "Abe Jellinek",
+	"target": "^https?://op\\.europa\\.eu/",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2021-06-01 17:01:22"
+}
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2021 Abe Jellinek
+	
+	This file is part of Zotero.
+
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	Zotero is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


Publications Office of the European Union.js
```diff
@@ -0,0 +1,222 @@
...
+function scrape(doc, url) {
+	// we would ideally just grab the "Metadata RDF" file here, but it doesn't
+	// give us RDF that Zotero can do anything with. the official metadata
+	// service (opac.publications.europa.eu) doesn't seem especially reliable.
+	// so we'll instead try ISBN and DOI.
+	
+	let pdfLinks = doc.querySelectorAll('a.download');
+	let search = Zotero.loadTranslator("search");
+	
+	// ISBN gives the best data, but we can fall back on DOI if needed
```
dstillman, line 85: What are the differences? Generally speaking, we always want to use a DOI over an ISBN, because a DOI offers canonical metadata from the publisher available via stable APIs, whereas ISBN metadata varies and we often have to scrape websites to retrieve it.

AbeJellinek: For this catalog, DOI returns incorrect/very incomplete author lists and almost every field has some weird formatting (extraneous punctuation, etc). ISBN results are all spot on.

dstillman: Well, it seems like the main difference is that they don't register personal authors with the DOI, so the first test case only has the institution. The actual presentation of the institutional author is more correct in the DOI, though — e.g., `European Commission. Representation in Luxemburg.` (correct-ish) vs. "European Commission" and "Representation in Luxemburg" as separate authors (less correct, arguably). But if we think these are cited by the individual authors, including those makes sense.

DOI doesn't have abstract (as is generally the case with DOIs) or ISBN, but it does have `Place: LU` and `Publisher: Publications Office`.

In a case like this, where there's a lot of uniformity, it's probably worth determining how these things should generally be cited and hard-coding values/transformations. E.g., if "Publications Office of the European Union" should just always be in a citation, we could just put that in the Publisher field. Same with the institutional author, which may be limited to a relatively small number of options and could be normalized (and maybe that value makes more sense as a Publisher anyway when there are individual authors?). While we can't exactly be faulted for just saving the metadata they provide, whenever we have a chance to save something such that every single Zotero user who cites it doesn't have to tweak the metadata, we should take the opportunity to do so.

PR #2383, opened 2021-05-25
Add translator for News Corp Australia sites

### LINKED ISSUE #163: The Australian - moved to theaustralian.com.au, site updated; needs complete rewrite
It's likely that new.com.au will be able to use a very similar translator as the two sites look alike.

AbeJellinek:
Closes #163


News Corp AU.js
```diff
@@ -0,0 +1,216 @@
+{
+	"translatorID": "393afc28-212d-47dd-be87-ec51bc7a58a4",
+	"label": "News Corp AU",
```
dstillman, line 3: Let's call it `News Corp Australia`

PR #2384, opened 2021-05-25
Financial Times: Remove FW, update for current site

### LINKED ISSUE #1075: Financial Times translator issues
- Pulls author as "Review by Demetri Sevastopulo" in [Review — A bleak outlook on an unequal America](http://www.ft.com/intl/cms/s/0/5c416bca-0233-11e6-99cb-83242733f755.html)
- Pulls a paywalled screenshot when using the translator, even if the page isn't paywalled on view. Embedded Metadata's screenshot pulls what's currently on the page.

Should be easy to fix, but tracking here in case someone gets to it before me

### LINKED ISSUE #1391: Financial Times broken after redesign
e.g. https://www.ft.com/content/d733c466-5eab-11e1-a04d-00144feabdc0
will need a free account to view articles & will quickly run out of free articles to view
Reported here: https://forums.zotero.org/discussion/38752/financial-times-translator-giving-an-error-message#latest

AbeJellinek:
The search test will fail on CI. Fixes #1391, fixes #1075.

### DIFF
--- a/Financial Times.js
+++ b/Financial Times.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2021-05-27 16:04:49"
+	"lastUpdated": "2021-05-27 16:17:23"
 }
 
 /*
@@ -88,18 +88,15 @@ function scrapeAmp(doc, url) {
 	
 	item.title = meta.headline;
 	item.date = ZU.strToISO(meta.datePublished);
-	// abstracts and authors usually won't show up in tests - they're rendered
-	// client-side and only the first author makes it into the JSON-LD.
-	// https://github.com/Financial-Times/next-json-ld#example-markup-on-article-page-behind-paywall-as-seen-by-google-bot
 	item.abstractNote = meta.description
-		|| text('.o-topper__standfirst');
+		|| text('.article-standfirst');
 	// something funky is going on with the JSON-LD authors, so we'll just
 	// parse from the HTML
 	item.creators = [...doc.querySelectorAll('a.article-author-byline__author')]
 		.map(link => ZU.cleanAuthor(link.innerText, 'author', false));
 	item.publicationTitle = 'Financial Times';
-	item.section = text('a[data-trackable="primary-brand"]')
-		|| text('a[data-trackable="primary-theme"]');
+	item.section = text('h2.primary-brand a')
+		|| text('h2.primary-theme a');
 	item.url = meta.mainEntityofPage;
 	item.libraryCatalog = '';
 	item.attachments.push({


Financial Times.js
```diff
@@ -42,105 +36,94 @@
...
-creators         : FW.Xpath('//span[@class="author_byline"]/a').text().replace(/^\s*by\s*/, "").cleanAuthor("author"),
-date             : FW.Xpath('//span[contains(@class, "entry-date")]').text(),
-ISSN			 : "0307-1766",
-publicationTitle : "Financial Times"
-}); 
- 
+function getSearchResults(doc, checkOnly) {
+	var items = {};
+	var found = false;
+	var rows = doc.querySelectorAll('.search-item a.js-teaser-heading-link[href*="/content/"]');
```
dstillman, line 52: Can we omit that class? Seems pretty fragile.

Financial Times.js
```diff
@@ -42,105 +36,94 @@
...
+function scrape(doc, url) {
+	let item = new Zotero.Item('newspaperArticle');
+	let meta = JSON.parse(text('script[type="application/ld+json"]'));
+	if (meta['@type'] == 'WebSite') {
+		throw new Error("No article metadata (probably hit paywall)");
+	}
+	
+	item.title = meta.headline;
+	item.date = ZU.strToISO(meta.datePublished);
+	// abstracts and authors usually won't show up in tests - they're rendered
```
dstillman, line 84: What about with `defer: true`?

Financial Times.js
```diff
@@ -42,105 +36,94 @@
...
+	// client-side and only the first author makes it into the JSON-LD.
+	// https://github.com/Financial-Times/next-json-ld#example-markup-on-article-page-behind-paywall-as-seen-by-google-bot
+	item.abstractNote = meta.description
+		|| text('.o-topper__standfirst');
```
dstillman, line 88: I've sure you've checked, but something like this just feels almost worthless, since it's almost certainly going to break. Better than nothing, but this is the best we can do now?

Financial Times.js
```diff
@@ -42,105 +36,94 @@
...
+	// something funky is going on with the JSON-LD authors, so we'll just
+	// parse from the HTML
+	item.creators = [...doc.querySelectorAll('a[data-trackable="author"]')]
+		.map(link => ZU.cleanAuthor(link.innerText, 'author', false));
+	if (meta.publisher) {
+		item.publicationTitle = meta.publisher.name;
```
dstillman, line 94: Is this ever not "Financial Times"? If missing, should it still be "Financial Times"? Any reason not to just hard-code?

Financial Times.js
```diff
@@ -42,105 +36,94 @@
...
-						"lastName": "Hill",
-						"creatorType": "author"
-					}
-				],
-				"date": "November 25, 2011 9:32 pm",
-				"ISSN": "0307-1766",
-				"abstractNote": "When 1,200 partners of McKinsey&Company – the elite of global consulting – arrived at the Gaylord National Hotel & Convention Center, outside Washington DC, early on the morning of March 15 this year, they found themselves where they least wanted to",
-				"libraryCatalog": "Financial Times",
+				"title": "Vonovia and Deutsche Wohnen to combine in €18bn real estate deal",
+				"creators": [],
```
dstillman, line 118: Missing authors for these seems bad

Financial Times.js
AbeJellinek: Probably, yes. I'll give it a try.

Financial Times.js
AbeJellinek: True. Done.

Financial Times.js
AbeJellinek: It seems like the AMP site gives us `meta.description` reliably, so that should do most of the time, but yeah, it's a fragile selector.

Financial Times.js
AbeJellinek: Fixed this!

dstillman:
Looks good. (Not sure we've ever used AMP for anything before. I guess we'll see how stable it is.)

AbeJellinek:
Meant to squash, but oh well. Next time.

PR #2386, opened 2021-05-26
RIS: Defer processing deprecated tags

### LINKED ISSUE #2125: RIS should prefer standard over "degenerate" fields for import
The below RIS (from https://journals.aom.org/doi/abs/10.5465/amj.2011.0651 ) imports the date as May 28, 2013, i.e. the `Y1` field. The actual publication date is April 2014, i.e. the content of the `DA` field. According to the RIS specifications, `DA` is the correct field for the date and Y1 is deprecated. The translator actually lists Y1 as "degenerate". When there are competing values, I think we should prefer standard fields. Currently, the order of tags determines which is used.

Related bug report: https://forums.zotero.org/discussion/81506/item-selectors-for-academy-of-management-is-something-wrong#latest
```
TY  - JOUR
T1  - Reputation and Decision Making under Ambiguity: A Study of U.S. Venture Capital Firms' Investments in the Emerging Clean Energy Sector
AU  - Petkova, Antoaneta P.
AU  - Wadhwa, Anu
AU  - Yao, Xin
AU  - Jain, Sanjay
Y1  - 2013/05/28
PY  - 2013
DA  - 2014/04/01
N1  - doi: 10.5465/amj.2011.0651
DO  - 10.5465/amj.2011.0651
T2  - Academy of Management Journal
JF  - Academy of Management Journal
JO  - AMJ
SP  - 422
EP  - 448
VL  - 57
IS  - 2
PB  - Academy of Management
SN  - 0001-4273
M3  - doi: 10.5465/amj.2011.0651
UR  - https://doi.org/10.5465/amj.2011.0651
Y2  - 2020/02/19
ER  - 
```

AbeJellinek:
Deprecated tags get processed in a second pass, and they'll end up in Extra if the Zotero item field they correspond to has already been filled. This is a pretty important translator - sanity checks/criticism appreciated.

Fixes #2125.

### DIFF
--- a/RIS.js
+++ b/RIS.js
@@ -17,7 +17,7 @@
 	},
 	"inRepository": true,
 	"translatorType": 3,
-	"lastUpdated": "2019-10-19 17:04:49"
+	"lastUpdated": "2021-05-25 22:40:15"
 }
 
 function detectImport() {
@@ -461,12 +461,21 @@ var degenerateImportFieldMap = {
  *   not present in the list, the next list is checked. If the RIS tag is
  *   present, but an item type does not match (no __default) or is explicit
  *   excluded from matching (__exclude), the next list is checked.
+ *   The `deprecatedMap` should be a subset of the maps in this list.
+ * @param {Tag <-> zotero field map} deprecatedMap A map, in the same format as
+ *   the entries in the `mapList`, containing deprecated tags that should only
+ *   be used if no other tag maps to the same Zotero item field.
  */
-var TagMapper = function(mapList) {
+var TagMapper = function(mapList, deprecatedMap) {
 	this.cache = {};
 	this.reverseCache = {};
 	this.mapList = mapList;
+	this.deprecatedMap = deprecatedMap;
 };
+
+TagMapper.prototype.isDeprecated = function (tag) {
+	return this.deprecatedMap.hasOwnProperty(tag);
+}
 
 /**
  * Given an item type and a RIS tag, return Zotero field data should be mapped to.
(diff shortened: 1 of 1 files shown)


RIS.js
```diff
@@ -6393,6 +6420,55 @@ var testCases = [
 				"seeAlso": []
 			}
 		]
+	},
+	{
+		"type": "import",
+		"input": "TY  - JOUR\nT1  - Reputation and Decision Making under Ambiguity: A Study of U.S. Venture Capital Firms' Investments in the Emerging Clean Energy Sector\nAU  - Petkova, Antoaneta P.\nAU  - Wadhwa, Anu\nAU  - Yao, Xin\nAU  - Jain, Sanjay\nY1  - 2013/05/28\nPY  - 2013\nDA  - 2014/04/01\nN1  - doi: 10.5465/amj.2011.0651\nDO  - 10.5465/amj.2011.0651\nT2  - Academy of Management Journal\nJF  - Academy of Management Journal\nJO  - AMJ\nSP  - 422\nEP  - 448\nVL  - 57\nIS  - 2\nPB  - Academy of Management\nSN  - 0001-4273\nM3  - doi: 10.5465/amj.2011.0651\nUR  - https://doi.org/10.5465/amj.2011.0651\nY2  - 2020/02/19\nER  - ",
```
dstillman, line 6426: To clarify the behavior under test here, maybe use a test case with just a regular tag and a deprecated tag? (I'm not even clear on what's being tested in the input. The `Y2`? That doesn't seem to have been added to Extra.)

AbeJellinek: Oh, true. It looks like it actually goes in `item.unsupportedFields`, which ends up nowhere, unless the true-by-default `RIS.import.ignoreUnknown` preference is set to false (in which case it goes into a note). I could add special handling for deprecated tags, but I'm unsure about whether the information would be meaningful - the RIS spec is clear that deprecated tags are always subordinate to their non-deprecated equivalents.

dstillman: I'm OK with discarding those, then. But let's still clarify the test.

AbeJellinek: OK, done.

PR #2391, opened 2021-05-26
Wikipedia: Strip tags in page titles, update tests

AbeJellinek:
This fixes the `Wikipedia:Article_wizard` test, which was previously returning a title with a bunch of markup in it.


dstillman:
We definitely want to get rid of a `<span>` like that, but we actually don't want to get rid of tags in titles completely. We technically support [rich-text formatting in titles](https://www.zotero.org/support/kb/rich_text_bibliography) (inherited from citeproc-js, but we'll support/render it properly in the future), and it seems like it would be correct for a page like [this](https://en.wikipedia.org/wiki/List_of_Sex_and_the_City_episodes). So we kind of want a version of `cleanTags()` that removes everything except the subset of HTML tags that we do support.

AbeJellinek:
Oh, good point. OK, I'll redo this with a potentially reusable filter function.

AbeJellinek:
Done. Added that page as a test, too.

Wikipedia.js
```diff
@@ -201,16 +235,16 @@ var testCases = [
 		"items": [
 			{
 				"itemType": "encyclopediaArticle",
-				"title": "Wikipedia:Article wizard",
+				"title": "<span style=\"color:white;width:0;font-size:0;\" class=\"hidden_title hidden-title\">Wikipedia:Article wizard</span>",
```
dstillman, line 238: This is now showing the bad stuff again.

Wikipedia.js
```diff
@@ -201,16 +235,16 @@ var testCases = [
...
+				"date": "2020-11-06T04:29:05Z",
 				"encyclopediaTitle": "Wikipedia",
-				"extra": "Page Version ID: 746008393",
+				"extra": "Page Version ID: 987303078",
 				"language": "en",
 				"libraryCatalog": "Wikipedia",
 				"rights": "Creative Commons Attribution-ShareAlike License",
-				"shortTitle": "Wikipedia",
-				"url": "https://en.wikipedia.org/w/index.php?title=Wikipedia:Article_wizard&oldid=746008393",
+				"shortTitle": "<span style=\"color",
```
dstillman, line 246: And here

Wikipedia.js
```diff
@@ -120,7 +120,14 @@ function doWeb(doc, url) {
 					item.date = pages[i].touched;
 				}
 				
-				item.title = pages[i].displaytitle;
+				// https://github.com/zotero/zotero/blob/master/chrome/content/zotero/bindings/styled-textbox.xml
```
dstillman, line 123: styled-textbox.xml isn't relevant here — these are getting saved to a plain text field. All that matters is the [subset of HTML](https://www.zotero.org/support/kb/rich_text_bibliography) allowed in citeproc-js. We could allow other things if we convert them to those (e.g., `<strong>` → `<b>`, `<em>` → `<i>`), but otherwise we don't want to save them.

AbeJellinek: Thanks, fixed this + the test regressions. I was confused between the two lists of allowed tags and figured the superset was correct.

PR #2393, opened 2021-05-26
GitHub: Cover all repo pages, un-break URL, license, and search

### LINKED ISSUE #2390: GitHub: Fix incorrect URL in embedded metadata on some repo pages
https://forums.zotero.org/discussion/89795/bug-report-zotero-connector-chrome-extension-saves-wrong-url-for-many-github-pages

AbeJellinek:
Fixes #2390

### DIFF
--- a/GitHub.js
+++ b/GitHub.js
@@ -3,6 +3,7 @@
 	"label": "GitHub",
 	"creator": "Martin Fenner, Philipp Zumstein",
 	"target": "^https?://(www\\.)?github\\.com/[^/]+/[^/]+",
+	"target": "^https?://(www\\.)?github\\.com/([^/]+/[^/]+|search\\?)",
 	"minVersion": "3.0",
 	"maxVersion": "",
 	"priority": 100,
@@ -47,7 +48,7 @@ function detectWeb(doc, url) {
 function getSearchResults(doc, checkOnly) {
 	var items = {};
 	var found = false;
-	var rows = ZU.xpath(doc, '//*[contains(@class, "repo-list-item")]//h3/a');
+	var rows = doc.querySelectorAll('.repo-list-item .f4 a');
 	for (var i = 0; i < rows.length; i++) {
 		var href = rows[i].href;
 		var title = ZU.trimInternal(rows[i].textContent);
--- a/GitHub.js
+++ b/GitHub.js
@@ -151,9 +151,9 @@ var testCases = [
 				"itemType": "computerProgram",
 				"title": "zotero/zotero",
 				"creators": [],
-				"date": "2019-08-29T02:15:36Z",
+				"date": "2021-05-25T16:50:57Z",
 				"abstractNote": "Zotero is a free, easy-to-use tool to help you collect, organize, cite, and share your research sources.",
-				"company": "zotero",
+				"company": "Zotero",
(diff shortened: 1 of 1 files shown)


GitHub.js
```diff
@@ -79,13 +80,24 @@ function doWeb(doc, url) {
...
 	item.url = ZU.xpathText(doc, '//meta[@property="og:url"]/@content');
+	if (url.includes('/blob/') && !item.url.includes('/blob/')) {
+		// github is doing something weird with the og:url meta tag right now -
+		// it always points to the repo root (e.g. zotero/translators), even
+		// when we're in a specific directory/file (e.g. zotero/translators/
+		// blob/master/GitHub.js). this fix (hopefully) won't stick around
+		// long-term, but for now, let's just grab the user-facing permalink
+		let permalink = attr(doc, '.js-permalink-shortcut', 'href');
+		if (permalink) {
+			item.url = 'https://' + doc.location.host + permalink;
```
dstillman, line 98: `doc.location.host` → `github.com` — no reason not to hard-code, is there?

GitHub.js
```diff
@@ -252,6 +262,28 @@ var testCases = [
...
+	{
+		"type": "web",
+		"url": "https://github.com/zotero/translators/blob/master/GitHub.js",
+		"items": [
+			{
+				"itemType": "computerProgram",
+				"title": "zotero/translators",
+				"creators": [],
+				"date": "2021-05-26T17:17:38Z",
+				"abstractNote": "Zotero Translators. Contribute to zotero/translators development by creating an account on GitHub.",
```
dstillman, line 275: While we're in here, can we strip the "Contribute to…" part (which may or may not be localized)? Really should only include the description entered by the repo owner.

GitHub.js
AbeJellinek: Hmm, my reasoning was that it's one less thing to change if the translator is extended to support Enterprise installs / [an alternate domain](https://en.wikipedia.org/wiki/Censorship_of_GitHub) / some other weird case. But not *really*.

GitHub.js
AbeJellinek: Sure, I'll give it a try.

GitHub.js
dstillman: I guess that's fine.

Maybe worth including the `clipboard-copy` element, which has the absolute URL in its `value` attribute, as a backup to `.js-permalink-shortcut`?

GitHub.js
```diff
@@ -94,7 +94,13 @@ function scrape(doc, url) {
 		// long-term, but for now, let's just grab the user-facing permalink
 		let permalink = attr(doc, '.js-permalink-shortcut', 'href');
 		if (permalink) {
-			item.url = 'https://' + doc.location.host + permalink;
+			item.url = 'https://github.com' + permalink;
+		}
+		else {
+			let clipboardCopyPermalink = attr(doc, '#blob-more-options-details clipboard-copy', 'value');
```
dstillman, line 100: Seems like it could just be `'clipboard-copy'`

AbeJellinek: There are a lot of `clipboard-copy`s on some pages - this issue discussion page has seven. The more specific selector makes sure that it only catches the URL one.

dstillman: Ah, OK

PR #2395, opened 2021-05-27
Update ACL translator for new domain (fix #2380)

### LINKED ISSUE #2380: ACLWeb: new URL
https://aclanthology.org/ is now the canonical URL for the site (though old URLs still work). So the following URL:

https://github.com/zotero/translators/blob/db0648bb7a411eff9841d8144a442d69fd0aa999/ACLWeb.js#L5

should be extended to `^https?://(www\\.)?(aclanthology.org|aclweb\\.org/anthology)/[^#]+`

and tests should be updated or added

GuyAglionby:
* Fix renaming of SemEval in the Anthology
* Update to new URL schema + update xpath which relied on the old schema
* Update tests

### DIFF
--- a/ACLWeb.js
+++ b/ACLWeb.js
@@ -2,14 +2,14 @@
 	"translatorID": "f4a5876a-3e53-40e2-9032-d99a30d7a6fc",
 	"label": "ACLWeb",
 	"creator": "Guy Aglionby",
-	"target": "^https?://(www\\.)?aclweb\\.org/anthology/[^#]+",
+	"target": "^https?://(www\\.)?(aclanthology\\.org|aclweb\\.org/anthology)/",
 	"minVersion": "3.0",
 	"maxVersion": "",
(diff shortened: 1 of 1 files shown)


ACLWeb.js
```diff
@@ -210,7 +210,7 @@ function getSearchResults(doc, url) {
 		}
 	}
 	else {
-		let papers = ZU.xpath(doc, '//strong/a[contains(@href, "/anthology/")]');
+		let papers = ZU.xpath(doc, '//section[@id="main"]//strong/a[@class="align-middle"]');
```
AbeJellinek, line 213: Is there a better selector than that `align-middle` class? If not, that's fine. It just feels a little fragile.

GuyAglionby: Just `a` will work - thanks!

AbeJellinek:
Merged, thanks!

PR #2398, opened 2021-05-29
Wikidata: Map Q13433827 to encyclopediaArticle item type

diegodlh:
Wikidata translator maps Wikidata's Q17329259 to Zotero's item type encyclopediaArticle.

However, Q17329259 has been merged into Q13433827 in 2019.

Wikidata items that are an instance of Q13433827 (instead of Q17329259; for example [Q15892061](https://www.wikidata.org/wiki/Q15892061)) are not recognized as encyclopedia articles by the Wikidata translator.

This PR fixes that.

### DIFF
--- a/Wikidata.js
+++ b/Wikidata.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2021-04-20 14:18:00"
+	"lastUpdated": "2021-06-03 02:33:07"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


AbeJellinek:
This looks good to me! Could you add a relevant test? After that, good to merge IMO.

diegodlh:
Thanks, @AbeJellinek. I've added the test case requested and squashed both commits into one. I hope I've done that right.

AbeJellinek:
Yes, looks perfect! Thanks so much for fixing this.

PR #2404, opened 2021-06-01
COinS: Optimize detectWeb and lint

AbeJellinek:
Since this translator targets everything, we want a fast `detectWeb`. This removes an unneeded query for every span tag on the page and replaces an XPath matching on spans' class attribute with a CSS selector that should be faster in modern browsers.

Also linted, added copyright, and updated tests.

### DIFF
--- a/COinS.js
+++ b/COinS.js
@@ -9,25 +9,14 @@
 	"inRepository": true,
 	"translatorType": 6,
 	"browserSupport": "gcsv",
-	"lastUpdated": "2015-06-04 03:25:10"
+	"lastUpdated": "2021-06-01 17:17:55"
 }
 
 function detectWeb(doc, url) {
-	var spanTags = doc.getElementsByTagName("span");
-
 	var encounteredType = false;
 	
-	// This and the x: prefix in the XPath are to work around an issue with pages
-	// served as application/xhtml+xml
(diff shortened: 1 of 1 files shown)


COinS.js
```diff
@@ -302,34 +303,32 @@ var testCases = [
...
-				"itemType": "journalArticle",
+				"itemType": "book",
+				"title": "The Genealogy of the Royal Family of Great Britain, shewing the descent of ... Queen Victoria, through five of the Saxon Kings who were crowned ... at Kingston-upon-Thames.",
 				"creators": [
 					{
-						"creatorType": "author",
-						"firstName": "Hui-Wen Vivian",
-						"lastName": "Tang"
+						"lastName": "Royal Families (England)",
+						"isInstitution": true
```
dstillman, line 314: Looks like this might just be an bug in openurl.js (which has probably been present for 15 years), but `isInstitution` isn't a valid property here. It should be `fieldMode: 1`.

dstillman:
Other than the `isInstitution` thing, which we have to fix in zotero/zotero, this looks great.

AbeJellinek:
Thanks! The `isInstitution` thing is odd. I could take a stab at submitting a PR for that.

PR #2409, opened 2021-06-02
HeinOnline: Parse DOM to get PDF URL

### LINKED ISSUE #2406: Hein Online: PDF URL is parsed with fragile regex
This needs to change:

https://github.com/zotero/translators/blob/18fb071ef237f24e72dd3386cc96a823401c0aaa/HeinOnline.js#L140

This could be fixed by replacing the call to `doGet` with a call to `processDocuments` and the regex with a DOM query. We should eventually make the connector HTTP functions follow meta redirects [like the client equivalents do](https://github.com/zotero/translators/blob/18fb071ef237f24e72dd3386cc96a823401c0aaa/HeinOnline.js#L140), but that will be a breaking change for some translators (at least this one).

AbeJellinek:
This replaces the regex that broke recently due to [a change in the meta refresh tag's syntax](https://forums.zotero.org/discussion/comment/383014/#Comment_383014).

Fixes #2406.

### DIFF
--- a/HeinOnline.js
+++ b/HeinOnline.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsbv",
-	"lastUpdated": "2019-10-26 13:45:08"
+	"lastUpdated": "2021-06-02 19:00:15"
 }
 
 /*
@@ -35,10 +35,6 @@
 	** Utilities **
 	***************
 */
-
-// attr()/text() v2
-// eslint-disable-next-line
-function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null}function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null}
 
 // Get any search results from current page
 // Used in detectWeb() and doWeb()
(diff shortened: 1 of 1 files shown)


HeinOnline.js
```diff
@@ -134,17 +130,17 @@ function scrapePage(doc, url) {
...
+				ZU.processDocuments(pdfPageURL, function (pdfDoc) {
 					// Call to pdfPageURL prepares PDF for download via META refresh URL
 					var pdfURL = null;
-					var m = pdfPage.match(/<META.*URL="([^"]+)/);
+					var m = pdfDoc.querySelector('meta[http-equiv="Refresh"]');
 					// Z.debug(pdfPage)
 					// Z.debug(m)
 					if (m) {
-						pdfURL = docParams.base + m[1];
+						pdfURL = docParams.base + m.getAttribute('url');
```
dstillman, line 140: This isn't right, is it? `URL` isn't an attribute — it's a parameter within the `CONTENT` attribute. That's why we [split it](https://github.com/zotero/zotero/blob/a5fe49b86669a59bd65a20219b86e8a5c7e8e511/chrome/content/zotero/xpcom/http.js#L800) in the client.

dstillman: (And trim single quotes if present)

AbeJellinek: From the source code of the redirect page:

```html
<META HTTP-EQUIV="Refresh" CONTENT="0" URL="PDFsearchable?handle=hein.journals/alterlj18&collection=journals&section=22&id=&print=section&sectioncount=3&ext=.pdf&nocover=&display=0">
```

Not technically correct HTML, but the site uses it (and browsers seem to handle it fine).

dstillman: Are we sure browsers handle it? There's also a JS redirect on that page.

dstillman: I don't think we should code to a spec violation. I can tell their dev that it's wrong.

dstillman: I think they just changed that based on our translator looking for `"`, but I assume that dates from a previous version of their site that used single quotes for the `CONTENT` parameter and double-quotes for the `URL` parameter. But making `URL` an attribute certainly isn't correct.

AbeJellinek: Ah, right, good point. The browser does follow it, but it just reloads the same page rather than redirecting anywhere. Sure, if they fix it then I'll update the translator to parse spec-compliant HTML.

dstillman: Just to avoid temporary breakage when they fix it, we should probably push a version that can handle either a correct URL parameter within CONTENT or the URL attribute. And we can remove the latter once they've fixed it.

AbeJellinek: OK, did that. Let me know if everything looks good to you.

dstillman: Looks good

PR #2412, opened 2021-06-03
YouTube: Scrape correct creator and description

AbeJellinek:
The YouTube translator was sometimes grabbing the wrong channel name. How is this possible when we're selecting by ID? Well, on a YouTube video page like [this one](https://www.youtube.com/watch?v=pq94aBrc0pY), `document.querySelectorAll('#text-container')` returns 22 nodes.

Whatever.

The other issue was with the description, which started pulling surrounding UI elements and metadata.

Whatever the case, these selectors are a bit better, and it'll still fall back on the less specific originals.

### DIFF
--- a/YouTube.js
+++ b/YouTube.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsbv",
-	"lastUpdated": "2020-11-11 22:15:02"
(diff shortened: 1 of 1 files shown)


AbeJellinek:
The test failure is because the duration it's pulling is that of the preroll ad, not the video itself. I'm going to try to find a fix for that separately, because it might lead to incorrect translation on the server.

PR #2436, opened 2021-06-09
Remove attr()/text() shims from almost all translators

### LINKED ISSUE #1913: Remove attr()/text() lines

AbeJellinek:
I've left alone ones that seem not to work currently or that have big pending PRs/lots of dependencies (Embedded Metadata).

When ready to merge, will fix #1913.

### DIFF
--- a/ACM Digital Library.js
+++ b/ACM Digital Library.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2020-04-20 05:31:35"
+	"lastUpdated": "2021-06-09 14:47:21"
 }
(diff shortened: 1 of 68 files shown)


dstillman:
We can do this, but we'd want to keep all the dates the same. We definitely don't want to trigger pushing of 68 translator files to all users.

AbeJellinek:
Just going to do this individually when updating old translators.

PR #2437, opened 2021-06-09
RIS: Un-deprecate A1

AbeJellinek:
See the added test for an example of an item that would be imported wrong if we treated A1 as deprecated (and is it even deprecated?). This was breaking the APS translator.

### DIFF
--- a/RIS.js
+++ b/RIS.js
@@ -17,7 +17,7 @@
 	},
 	"inRepository": true,
 	"translatorType": 3,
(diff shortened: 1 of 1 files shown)


AbeJellinek:
Tests for the first 10 dependent translators show no related changes to creator metadata, and this is functionally just a partial reversion of my last set of RIS changes. Merging as a hotfix for the [APS author ordering issue](https://forums.zotero.org/discussion/comment/384940).

PR #2440, opened 2021-06-09
Louvre: New domain, use JSON-LD, try to clean names

### LINKED ISSUE #2422: Louvre translator is out of date, errors on all pages

AbeJellinek:
Fixes #2422

### DIFF
--- a/Musee du Louvre.js
+++ b/Musee du Louvre.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2021-06-09 20:15:40"
+	"lastUpdated": "2021-06-11 13:07:46"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


Musee du Louvre.js
```diff
@@ -69,45 +67,81 @@ function doWeb(doc, url) {
...
 	}
-	item.title = text(doc, 'h1>span');
-	if (!item.title) {
-		item.title = text(box, 'li p', 1);
+	
+	if (json.dateCreated) {
+		let bceDateMatch = json.dateCreated.match(/([0-9]+) av\./);
+		if (bceDateMatch) {
+			// https://docs.citationstyles.org/en/1.0.1/specification.html#ad-and-bc
+			item.date = '-' + bceDateMatch[1];
```
AbeJellinek, line 127: @adam3smith is this the right approach for BCE dates? CSL recommends it, but I'm not having a lot of success generating actual BCE citations in Zotero.

dstillman: I'm not sure what citeproc-js supports currently, but we'll be supporting EDTF both in Zotero and in citeproc-rs in the not-too-distant future, so that's theoretically what we'd want to use. I believe that EDTF does use `-` for BCE dates ([as in ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Years)), but depending on the source data, the year may need to be shifted by one — e.g., `-0379` is 380 BCE.

adam3smith: citeproc/CSL does support the - sign for BCE, but Zotero currently strips it out in date parsing, so it will only work in Extra (`issued: -379`). Not sure what you want to do about this while the current Zotero behavior exists. Seems weird to write this into Extra, so probably just leave and accept that it won't work in citation styles until the EDTF update?

dstillman: Oh, we can probably just fix that now (at least for the beta). [Issue](https://github.com/zotero/zotero/issues/2086) created.

PR #2441, opened 2021-06-10
Update Library Genesis mirrors

boan-anbo:
Fixed the extension for the new main mirror: libgen.rs. Updated tests.

### DIFF
--- a/Library Genesis.js
+++ b/Library Genesis.js
@@ -2,14 +2,14 @@
 	"translatorID": "11614156-f421-4e89-8ce0-a5e69ce3ebed",
 	"label": "Library Genesis",
 	"creator": "Reverend Wicks Cherrycoke",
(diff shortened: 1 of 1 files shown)


AbeJellinek:
Thanks for catching this and updating the translator!

It looks like `libgen.st` is also a valid mirror. Could you add that?

We could just replace the TLD part of the regex with `[a-z]{2}` if keeping up with domain changes becomes too much of a chore, but this is fine for now.

boan-anbo:
> Thanks for catching this and updating the translator!
> 
> It looks like `libgen.st` is also a valid mirror. Could you add that?
> 
> We could just replace the TLD part of the regex with `[a-z]{2}` if keeping up with domain changes becomes too much of a chore, but this is fine for now.

Good point! I just made it so `[a-z]{2}` and added a test for `.st`.

AbeJellinek:
OK, great. Thanks, all these changes look good.

PR #2453, opened 2021-06-12
New Translator: New Left Review

boan-anbo:
https://newleftreview.org/

### DIFF
--- a/New Left Review.js
+++ b/New Left Review.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2021-06-16 09:54:20"
+	"lastUpdated": "2021-06-16 16:25:48"
 }
 
 /*
@@ -151,7 +151,11 @@ var testCases = [
 				"libraryCatalog": "New Left Review",
 				"pages": "24-46",
 				"publicationTitle": "New Left Review",
-				"attachments": [],
+				"attachments": [
+					{
+						"mimeType": "application/pdf"
+					}
+				],
 				"tags": [],
 				"notes": [],
 				"seeAlso": []
@@ -184,7 +188,11 @@ var testCases = [
 				"libraryCatalog": "New Left Review",
 				"pages": "77-92",
 				"publicationTitle": "New Left Review",
-				"attachments": [],
+				"attachments": [
+					{
+						"mimeType": "application/pdf"
+					}
+				],
 				"tags": [],
 				"notes": [],
 				"seeAlso": []
@@ -217,7 +225,11 @@ var testCases = [
 				"libraryCatalog": "New Left Review",
 				"pages": "153-160",
 				"publicationTitle": "New Left Review",
-				"attachments": [],
+				"attachments": [
+					{
+						"mimeType": "application/pdf"
+					}
+				],
 				"tags": [],
 				"notes": [],
 				"seeAlso": []
--- a/New Left Review.js
+++ b/New Left Review.js
@@ -1,0 +1,257 @@
+{
+	"translatorID": "b97af278-52c1-4fc4-9c8b-5175c83941f2",
+	"label": "New Left Review",
+	"creator": "Bo An",
+	"target": "^https?://(www.)?newleftreview.org/issues/",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2021-06-12 07:59:12"
+}
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+	Copyright © 2021 Bo An
+	This file is part of Zotero.
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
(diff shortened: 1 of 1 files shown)


New Left Review.js
```diff
@@ -0,0 +1,257 @@
...
+	else {
+		scrape(doc, url);
+	}
+}
+
+function getSearchResults(doc) {
+	let items = {};
+	let found = false;
+
+	const articleLinks = ZU.xpath(doc, '//a[@class="article-link"]');
```
AbeJellinek, line 67: `@class="article-link"` will only match elements with that exact string as their `class` attribute - if another class gets added, the XPath will break. These would be better as CSS selectors like `doc.querySelectorAll('a.article-link')`.

New Left Review.js
```diff
@@ -0,0 +1,257 @@
...
+	const titles = ZU.xpath(doc, '//span[@class="programme-note__title"]');
+	const authors = ZU.xpath(doc, '//span[@class="programme-note__author"]');
+
+	articleLinks.forEach((articleLink, articleIndex) => {
```
AbeJellinek, line 71: It looks to me like the `.programme-note__title` and `.programme-note__author` elements are children of the `a.article-link` for each article. You could remove lines 68-69...

New Left Review.js
```diff
@@ -0,0 +1,257 @@
...
+		const href = articleLink.href;
+		let title = titles[articleIndex].innerText;
```
AbeJellinek, line 73: And then replace this and the following line with this:

```js
let title = text(articleLink, '.programme-note__title');
let author = text(articleLink, '.programme-note__author');
```

It's a bit less fragile - even if, in the future, some page has an extra `.programme-note__title` that *isn't* inside a `.article-link`, or a `.article-link` has two authors, or some other weird mishap like that, restricting the search to children of the article link you're looking at will make sure that the arrays don't get out of alignment.

New Left Review.js
```diff
@@ -0,0 +1,257 @@
...
+	});
+
+	return found ? items : false;
+}
+
+function scrape(doc, _) {
+	const newItem = new Zotero.Item('journalArticle');
+
+	newItem.publicationTitle = "New Left Review";
+	newItem.journalAbbreviation = "NLR";
```
AbeJellinek, line 92: Looks like its ISO 4 abbreviation is `New Left Rev.`.

New Left Review.js
```diff
@@ -0,0 +1,257 @@
...
+	if (description) {
+		newItem.abstractNote = description;
+	}
+
+	const authors = ZU.xpath(doc, '//meta[@itemprop="author"]/@content');
+
+	const authorFullNames = authors.map(author => author.value);
+
+	authorFullNames.forEach((authorFullName) => {
+		newItem.creators.push({
```
AbeJellinek, line 109: Is there a reason we should use `fieldMode: 1` and not use `ZU.cleanAuthor` here?

AbeJellinek (changes requested):
Thanks, this looks really good. Left a few comments.

New Left Review.js
boan-anbo: Thanks! I wasn't aware of this utility class. Super helpful. 

I was wondering if there is a similar utility class for titles too (to convert all caps to the title case for example)?

New Left Review.js
boan-anbo: thx so much! btw, what's this text() function?

New Left Review.js
AbeJellinek: Yes! There's [`ZU.capitalizeTitle`](https://github.com/zotero/zotero/blob/f1414731daa749c1de7b5946e9c08b571c710d1e/chrome/content/zotero/xpcom/utilities.js#L903).

AbeJellinek: In a lot of cases, when you *know* the titles are screwed up, you'll want to pass `true` as the second argument to that function (`force`). Otherwise it'll only perform the transformation if the user has manually enabled the `capitalizeTitles` preference.

New Left Review.js
```diff
@@ -225,23 +205,19 @@ var testCases = [
...
 				"abstractNote": "Michael Lipkin on Klaus Vieweg, Hegel: Der Philosoph der Freiheit. The author of Philosophy of Right as roistering liberal.",
 				"issue": "128",
-				"journalAbbreviation": "NLR",
+				"journalAbbreviation": "New Left Rev",
 				"libraryCatalog": "New Left Review",
 				"pages": "153-160",
 				"publicationTitle": "New Left Review",
-				"attachments": [
-					{
-						"mimeType": "application/pdf"
```
AbeJellinek: Where'd these PDFs go?

AbeJellinek (changes requested):

New Left Review.js
boan-anbo: Thanks for the tips Abe! I'll try using it next time.

New Left Review.js
boan-anbo: The download is paywalled. Do we want tests for unpaywalled or paywalled conditions?

dstillman: Paywalled is fine. Better to test that we're getting the right data, even if only some people can test it.

boan-anbo: I see! Thx. Retested under both conditions and PDFs are being downloaded when unpaywalled.

AbeJellinek:
OK, this is pretty much ready. Add the paywalled PDF attachments back to the tests and I'll merge it in.

AbeJellinek:
Merged, thanks!

PR #2462, opened 2021-06-17
RSC: Pull supplementary information PDFs

### LINKED ISSUE #2385: Add support for extracting supplementary information from RSC journals
For example this article (OA) has some SI:
[https://pubs.rsc.org/en/content/articlelanding/2020/cp/d0cp03745a#!divAbstract](https://pubs.rsc.org/en/content/articlelanding/2020/cp/d0cp03745a#!divAbstract)

AbeJellinek:
Fixes #2385

### DIFF
--- a/RSC Publishing.js
+++ b/RSC Publishing.js
@@ -35,23 +35,22 @@
 
 	***** END LICENSE BLOCK *****
 */
+
 function getResults(doc) {
-	/**Both search result and book ToC pages use javascript to load content, so
+	/** Both search result and book ToC pages use javascript to load content, so
 	 * this actually doesn't work as intended. Search results will work, but
 	 * will also trigger on empty result set. detectWeb for book ToC does not
 	 * work, but doWeb does,
 	 */
-	return ZU.xpath(doc, '//div[@id="all" or @id="chapterList"]\
-						//div[contains(@class,"title_text")]\
-						//a[not(contains(@href,"/database/"))]');
+	return ZU.xpath(doc, '//div[@id="all" or @id="chapterList"]//div[contains(@class,"title_text")]//a[not(contains(@href,"/database/"))]');
 }
 
 function detectWeb(doc, url) {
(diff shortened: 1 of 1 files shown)


RSC Publishing.js
```diff
@@ -35,52 +35,64 @@
...
-	return ZU.xpath(doc, '//div[@id="all" or @id="chapterList"]\
-						//div[contains(@class,"title_text")]\
-						//a[not(contains(@href,"/database/"))]');
+	return ZU.xpath(doc, '//div[@id="all" or @id="chapterList"]//div[contains(@class,"title_text")]//a[not(contains(@href,"/database/"))]');
 }
 
 function detectWeb(doc, url) {
-	if (url.search(/\/results[?\/]/i) != -1 || url.indexOf('/ebook/') != -1  &&
-		getResults(doc).length) {
+	if (url.match(/\/results[?/]/i) != -1 || url.includes('/ebook/')
```
adam3smith, line 49: @AbeJellinek not sure why you changed this from `search` to `match`, but as written, the first expression will always evaluate to true (`match` returns an array or null, never -1). Generally I've used `search` or `test` when evaluating regular expressions for T/F since they're a good bit more efficient.

adam3smith:
I think this introduced a regression as reported here: https://forums.zotero.org/discussion/90441/translator-error-for-rsc-publishing#latest

RSC Publishing.js
AbeJellinek: Oh no, I meant to remove the `!= -1`. I'll fix that. We have a rule in the style checker disallowing `s.indexOf(x) != -1`-style constructions, and I think using `search` that way has the same weaknesses.

adam3smith:
But let's then switch it to test. .match really isn't meant for testing

AbeJellinek:
Sorry, thanks for pointing this out.

(And yeah, switched to `test`.)

PR #2463, opened 2021-06-17
EM: Convert relative URLs, add a test for double NS

AbeJellinek:
Using a relative URL in `<link rel="canonical">` is apparently legal, but we weren't handling it. I added a function to make relative URLs absolute if necessary and a test for the double og: namespace changes (which depends on the relative -> absolute conversion).

### DIFF
--- a/Embedded Metadata.js
+++ b/Embedded Metadata.js
@@ -38,11 +38,6 @@
 */
 
 
-// attr()/text() v2
-// eslint-disable-next-line
-function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null;}function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null;}
-
-
 /* eslint-disable camelcase */
 var HIGHWIRE_MAPPINGS = {
 	citation_title: "title",
--- a/Embedded Metadata.js
+++ b/Embedded Metadata.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2021-06-17 20:30:35"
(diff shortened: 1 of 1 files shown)


dstillman:
Hmm, I guess we have no existing function for this?

We can do this for now, but seems like we should move [`Translate.resolveURL()`](https://github.com/zotero/zotero/blob/b5edc0d9c18bdf0c11cef348c3cac2161e0f7219/chrome/content/zotero/xpcom/translation/translate.js#L1393) into utilities_translate.js (which is mostly where it's used anyway) so that it's available to translators, while figuring out how we can still call it from translate.js in the one place where we need it.

AbeJellinek:
Yeah, I was sure that we *had to* but I looked everywhere and couldn't find one. I agree that `resolveURL` should be available in `Zotero.Utilities`. I can work on a PR for that.

dstillman:
Yeah, that'd be great. Moving it to utilities_translate.js, and adjusting the existing calls in utilities_translate.js, should be easy. I'm less sure how we'll access that from `_itemDone()` in translate.js. There might be some way to do that without duplicating the function, but doing so wouldn't be the end of the world with an appropriate comment linking the two.

Oh, but probably we want to hold off on this until @adomasven finishes modularizing the extracting the translator architecture. I think that's almost done.

AbeJellinek:
We could also keep it in `translate.js` and add a wrapper in `utilities_translate.js`, like

```js
Zotero.Utilities.Translate.prototype.resolveURL = function(url, dontUseProxy) {
	return this._translate.resolveURL(url, dontUseProxy);
}
```

That's how `doGet`, etc. call it already.

dstillman:
Oh, right.

adomasven:
Yeah, we actually discussed exposing some proxy functions to translators at some point and this might be good enough. Wrapper in `utilities_translate.js` seems fine.

AbeJellinek:
@adomasven, will a PR adding those lines in `utilities_translate.js` conflict with what you're working on?

adomasven:
It will but it's not a big deal. I thought I would be finished already but I want to have changes ready so we can start using my work in at least the connector and the zotero client, otherwise we'll just end up with more fractured codebases for translation, i.e. https://xkcd.com/927/

PR #2465, opened 2021-06-18
Pubmed/PMC: Add eid to page number

### LINKED ISSUE #875: PubMed translator finds results for 404 PMC results
i.e. http://www.ncbi.nlm.nih.gov/pmc/articles/PMC9999999/

Originally opened this ticket in the wrong spot, see thread here:

https://github.com/zotero/translation-server/issues/14#issuecomment-87019273

### LINKED ISSUE #2370: Import electronic identifiers from Pubmed
Lots of details here:
https://forums.zotero.org/discussion/89583/pubmed-translator-not-saving-some-e-pages#latest
We need to think whether these should be imported as pages or number -- we mostly do pages and I'd probably keep doing this, but some styles differentiate between them.

adam3smith:
Closes #2370
Also tightens PMCID detect (that part would benefit from another set of eyes; not sure there)
Closes #875


adam3smith:
The failing tests are in two translators that are otherwise broken

PubMed Central.js
```diff
@@ -37,7 +36,11 @@
 */
 
 function detectWeb(doc, url) {
-	if (getPMCID(url)) {
+	// Make sure the page have a PMCID and we're on a valid item page,
+	// PDF, or report page
+	if (getPMCID(url) && (url.includes(".pdf")
+	|| url.includes("/?report"))
```
AbeJellinek, line 42: I'm not sure this is a useful check - `?report` doesn't seem to do anything, i.e., https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3139813/?report=classic and https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3139813/ are the exact same page.

https://www.ncbi.nlm.nih.gov/pmc/articles/PMC999999/?report=classic would pass this check but it 404s.

Can we get away with removing this line? Is there a particular case that it handles, like a page without `.fm-ids` but with `?report` in the URL?

AbeJellinek: Even the `?report=reader` ebook-style view has a `.fm-ids` element.

adam3smith: You're completely right -- it was for the new reader layout and I had tests failing there and assumed it didn't have the fm-ids, but I think something else was going on. Removed

AbeJellinek: OK, great.

PR #2473, opened 2021-06-21
OJS: Update selectors for detection and attachments

### LINKED ISSUE #2334: Update identifier for an OJS journal article
Dear all
both of these methods to reliably identify an OJS journal article seem to be outdated
https://github.com/zotero/translators/blob/573a963d31c439b600f122a1123aced4b9c6c040/Open%20Journal%20Systems.js#L39-L41

I would suggest to simply add/replace something like
`	if (ZU.xpathText(doc, '//meta[@name="generator"]/@content') == 'Open Journal Systems 3.2.1.3'` 
I am not sure how to wildcard the version number. So please add to this.

All the best
Klaus

AbeJellinek:
Fixes #2334.

### DIFF
--- a/Open Journal Systems.js
+++ b/Open Journal Systems.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2021-06-21 16:51:49"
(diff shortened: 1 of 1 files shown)


AbeJellinek:
Not sure where the odd `journalAbbreviation`s are coming from. I'll look into that before merging.

PR #2474, opened 2021-06-21
EM: Don't use `location.toString()`

AbeJellinek:
Doesn't work when EM is called from another translator in Scaffold and maybe elsewhere.

### DIFF
--- a/Embedded Metadata.js
+++ b/Embedded Metadata.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2021-06-18 15:50:04"
+	"lastUpdated": "2021-06-21 17:09:36"
 }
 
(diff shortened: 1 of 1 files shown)


Embedded Metadata.js
```diff
@@ -934,12 +934,11 @@ function relativeToAbsolute(doc, url) {
 	if (url[0] == '/') {
 		// relative to root
 		return doc.location.protocol + '//' + doc.location.host
-			+ (doc.location.port.length ? ':' + doc.location.port : '')
```
dstillman, line 937: Was this part on purpose?

AbeJellinek: Yeah. `Location.host` [already includes the port if necessary](https://developer.mozilla.org/en-US/docs/Web/API/Location/host). `Location.hostname` doesn't. I'm not sure why the code I ripped for this method was adding it again.

PR #2481, opened 2021-06-22
Adding V&A translator

atiro:
Created new PR to allow edits from @AbeJellinek

### DIFF
--- a/Victoria & Albert Museum.js
+++ b/Victoria & Albert Museum.js
@@ -1,0 +1,274 @@
+{
+	"translatorID": "56fc13bf-411f-4409-9a61-3d334d76763b",
+	"label": "Victoria & Albert Museum",
+	"creator": "Richard Palmer - V&A Digital Media & Publishing",
+	"target": "^https?://collections\\.vam\\.ac\\.uk",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2022-04-03 16:45:24"
+}
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2022 V&A Museum
+
+	This file is part of Zotero.
+
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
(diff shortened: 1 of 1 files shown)


Victoria & Albert Museum.js
```diff
@@ -0,0 +1,198 @@
...
+		for (var k = 0; k < artistMakerOrganisations.length; k++) {
+			item.creators.push({
+				lastName: artistMakerOrganisations[k].name.text,
+				fieldMode: 1,
+				creatorType: "author"
+			});
+		}
+
+		item.attachments.push({
+			url: url,
```
AbeJellinek, line 124: You can just do `document: doc` and omit the `snapshot: true` (because it's implied with `document`). `document` is much faster (and more accurate) than `url` because it can use the already-loaded page.

Victoria & Albert Museum.js
```diff
@@ -0,0 +1,198 @@
...
+
+function doWeb(doc, url) {
+	if (detectWeb(doc, url) == "multiple") {
+		var artworks = {};
+
+		var items = ZU.xpath(doc, '//figure[contains(@class, "b-object-card--etc")]//a');
+
+		if (items.length < 1) {
+			// TODO - other multiples
+			return false;
```
AbeJellinek, line 141: You should move this `length` check into `detectWeb` for (much) better UX - as it stands, a search results page with no results will show the folder icon in the toolbar and then do absolutely nothing when the user clicks it.

See the web translator template (green plus menu in Scaffold) for an example of the pattern.

Victoria & Albert Museum.js
```diff
@@ -0,0 +1,198 @@
...
+
+			if (!artworks) {
+				return true;
+			}
+			for (var i in artworks) {
+				articles.push(i);
+			}
+			Zotero.Utilities.processDocuments(articles, scrape);
+			return false;
+		}
```
AbeJellinek, line 159: indentation

Victoria & Albert Museum.js
```diff
@@ -0,0 +1,198 @@
...
+		);
+	}
+	else {
+		scrape(doc, url);
+	}
+	return false;
+}
+
+/** BEGIN TEST CASES **/
+var testCases = [
```
AbeJellinek, line 169: Update tests (and add one or two more? test `artworkSize`?) when done.

Victoria & Albert Museum.js
```diff
@@ -0,0 +1,198 @@
...
+		else if (data.objectType) {
+			item.title = data.objectType;
+			item.shortTitle = data.objectType;
+		}
+
+		item.abstractNote = data.briefDescription;
+		item.medium = data.materialsAndTechniques;
+		item.artworkSize = data.dimensionsNote;
+		if (data.productionDates.length > 1) {
+			item.date = data.productionDates[0].date.text + " and further dates";
```
AbeJellinek, line 69: What I said on the last PR about "and further dates" still applies. I don't think we should have it. If you'd like to keep extra dates around, put them in `extra` like `V&A Additional Dates: 1905, 1910`.

Approximate dates (`ca. 1900`) should be added to extra in the form `issued: circa 1900` in order to be displayed correctly in citations.

AbeJellinek (changes requested):
Thanks @atiro. Left a few comments! Let me know what you think.

atiro:
Sorry @AbeJellinek , life got in the way. Updates as requested (I hope)

AbeJellinek:
Thank you!

PR #2483, opened 2021-06-22
Add translator for NYPL Menus

AbeJellinek:
A cool repository of historical menus. There's no way to get a PDF of all the pages, so I just attach a link to the cover in the interactive viewer.


dstillman:
I think we want to just put the main URL (e.g., `http://menus.nypl.org/menu_pages/35364`) in the URL field, with no link attachment. For something like this, I believe the URL would generally be cited, since it provides full reproductions, it's the canonical source online for these menus, and there's no other identifier.

dstillman:
Also, what was the thinking behind `manuscript`? That doesn't really seem right here, though I'm not sure what else (other than Document) would be any better.

AbeJellinek:
`document` has no equivalent for `manuscript`'s `place` and `manuscriptType`. `place` helps to disambiguate business names, `manuscriptType` lets us add a type specifier, which I think some styles ask for.

NYPL Menus.js
```diff
@@ -186,13 +180,8 @@ var testCases = [
...
 				"manuscriptType": "Restaurant menu",
 				"numPages": "6",
-				"attachments": [
-					{
-						"title": "Cover Page",
-						"mimeType": "text/html",
-						"snapshot": false
-					}
-				],
+				"url": "http://menus.nypl.org/menu_pages/54352/explore",
```
dstillman, line 183: Should remove `explore` for all of these

dstillman:
Place and type can just be added to Extra. `document` will gain [both `place` and `type`](https://github.com/zotero/zotero-bits/issues/18) with the fields update, so the fields will migrate automatically later. Better not to use a misleading item type.

I believe `type` currently has to be added to Extra as `genre`, which is the CSL field, since otherwise there's some confusion with the CSL `type`, which means item type.

PR #2492, opened 2021-06-24
GitHub: Pickier detectWeb, safer selectors

### LINKED ISSUE #2490: GitHub Discussions: Translator error, perhaps needs to be filtered out
GitHub discussions pages (such as https://github.com/MobileNativeFoundation/discussions/discussions/6) have a quite a different structure from a repo or issue. At the moment, they cause an error in the translator and it falls back to EM.

The existing GitHub translator seems meant for repo root pages, so perhaps the target URL needs to be more specific to avoid things like discussions...?

AbeJellinek:
- Test `og:title` in `detectWeb` and ensure that it contains a repo name (`abc/xyz`). `doWeb` will fail if it doesn't.
- Use `attr`, not `xpathText`, because `attr` doesn't return null (and thereby cause translator errors) when it doesn't find a match.

Fixes #2490.


AbeJellinek:
(I didn't end up rolling the scope back completely, because the `og:url` issue is still present.)

PR #2494, opened 2021-06-28
New Translator for Verso Books. 

boan-anbo:
https://www.versobooks.com/


Verso Books.js
```diff
@@ -0,0 +1,313 @@
...
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+function detectWeb(doc, url) {
+	const isSubject = url.includes("subjects/");
+	if (isSubject) {
```
AbeJellinek, line 40: We don't want to return `multiple` when there are no results to scrape, even if we're theoretically on a search results page based on the URL, because `selectItems` will throw an error. Try this:

```suggestion
	if (isSubject && getSearchResults(doc, true)) {
```

Verso Books.js
```diff
@@ -0,0 +1,313 @@
...
+						"lastName": "Tchen",
+						"creatorType": "author"
+					},
+					{
+						"firstName": "Dylan",
+						"lastName": "Yeats",
+						"creatorType": "author"
+					}
+				],
+				"date": "2014-01",
```
AbeJellinek, line 243: This date changed for some reason, so the test fails now.

AbeJellinek (changes requested):
Hey, thanks! Looks great. Left a couple comments.

Verso Books.js
```diff
@@ -0,0 +1,313 @@
...
+	***** END LICENSE BLOCK *****
+*/
+
+function detectWeb(doc, url) {
+	const isSubject = url.includes("subjects/");
+	if (isSubject) {
+		return 'multiple';
+	}
+	else {
+		return 'book';
```
AbeJellinek, line 44: Maybe predicate returning `book` on `doc.querySelector('.edition-single--book-title')` matching, so that we return `false` on 404 pages, etc.

Verso Books.js
boan-anbo: Thanks for pointing it out! It might be caused by `Zotero.Utilities.strToDate`. I am guessing that function was based on JS native date, with January starting from 0, cause it gave me 5 for June, so I later switched to `strToISO`. I updated the test.

boan-anbo:
Thanks Abe for the suggestions!

Verso Books.js
AbeJellinek: Thanks! Yeah, `strToISO` is almost always what you want to use in a translator.

Verso Books.js
```diff
@@ -0,0 +1,315 @@
...
+	return found ? items : false;
+}
+
+function scrape(doc, _) {
+	const newItem = new Zotero.Item('book');
+
+	newItem.publisher = "Verso Books";
+
+	const title = doc.getElementsByClassName('edition-single--book-title')[0].textContent;
+	const subtitle = doc.getElementsByClassName('edition-single--book-subtitle')[0].textContent;
```
AbeJellinek, line 98: When you're just pulling the `textContent` of a selector, it's best to use `text` so that the translator doesn't error out if the selector doesn't match. Like this:

```suggestion
	const title = text(doc, '.edition-single--book-title');
	const subtitle = text(doc, '.edition-single--book-subtitle');
```

Verso Books.js
```diff
@@ -0,0 +1,315 @@
...
+	newItem.title = title + (subtitle ? `: ${subtitle}` : '');
+
+	const contributors = doc.getElementsByClassName('edition-single--book-contributors')[0];
+	const contributorsEls = contributors.querySelectorAll('span');
```
AbeJellinek, line 102: In a similar vein, this can be

```suggestion
	const contributorsEls = doc.querySelectorAll('.edition-single--book-contributors span');
```

so that it just returns an empty array if there is no `.edition-single--book-contributors` rather than causing an error.

Verso Books.js
```diff
@@ -0,0 +1,315 @@
...
+	contributorsEls.forEach((contributorEl) => {
+		const authors = contributorEl.querySelectorAll('a');
+		if (authors.length > 0) {
```
AbeJellinek, line 105: Don't need this check. Your `forEach` function just won't be called if the array is empty.

Verso Books.js
```diff
@@ -0,0 +1,315 @@
...
+		// check if translator
+		const spanText = contributorEl.textContent;
+		const isTranslator = spanText.toLowerCase().includes('translated by');
+		if (isTranslator) {
+			const translatorName = spanText.replace('Translated by', '');
+			newItem.creators.push(ZU.cleanAuthor(translatorName, "translator", false));
+		}
+	});
+
+	const description = doc.getElementsByClassName('edition-single--book-description')[0].textContent;
```
AbeJellinek, line 128: Same note about `text` applies here.

Verso Books.js
```diff
@@ -0,0 +1,315 @@
...
+		if (pages) {
+			newItem.pages = pages;
+		}
+
+		if (date) {
+			newItem.date = ZU.strToISO(date);
+		}
+
+		if (ISBN) {
+			newItem.ISBN = ISBN;
```
AbeJellinek, line 150: Just in case:

```suggestion
			newItem.ISBN = ZU.cleanISBN(ISBN);
```

AbeJellinek (changes requested):
OK, a few more suggestions! Thanks for making those changes.

Verso Books.js
boan-anbo: Got it! Thanks Abe!

AbeJellinek:
Looks great, thanks!

PR #2495, opened 2021-06-28
Idref: Support five more sources

symac:
The idref website now offers records from more website. This PR adds 5 new translators to improve compatibility.

### DIFF
--- a/Idref.js
+++ b/Idref.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2020-06-19 12:54:28"
+	"lastUpdated": "2021-06-28 12:08:46"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


symac:
The test failed, but when I try to visit [the test record](https://www.idref.fr/199676100) it works fine if I wait for the record list to load. As people developing idref.fr are adding more sources to it it seems to be slower than before and the loading spin needs more time to disappear, that might explain why it now fails when last time it was fine.

AbeJellinek:
The test will work if you add `"defer" true` below `"url": "https://www.idref.fr/199676100"`.

Any chance you could find an Idref page with items that use one or two of the new translators? If not, no big deal.

symac:
Hi @AbeJellinek ,
thanks for the tip with defer. I have added two new tests containing some links to the newly added websites. The only thing is that I still have a timeout with one of the test.

AbeJellinek:
Yeah, it does spin for a while but everything seems to work fine once the page has loaded. Thanks for adding more tests!

AbeJellinek:
Also @symac, not sure what the relationship between Abes and Arabesques and Prairial is, but if you're able to, might want to update the article page metadata a bit. [This article from one of the tests](https://publications-prairial.fr/arabesques/index.php?id=145), for example, is translated as a `webpage` and not a `journalArticle` because it has two `DC.type`s:

```html
<meta name="DC.type" content="text" />
<meta name="DC.type" content="journalArticle" />
```

It would be translated as a `journalArticle` if the first tag weren't there. (Verified by testing with a local copy of the page.)

symac:
Thanks @AbeJellinek for testing further. I have contacted the person managing prairial website asking if it was possible to remove the dc.type:text tag.

PR #2501, opened 2021-06-29
Add translator for University of California Press books

AbeJellinek:
Journals are already covered by Silverchair.

### DIFF
--- a/University of California Press Books.js
+++ b/University of California Press Books.js
@@ -1,0 +1,403 @@
+{
+	"translatorID": "f27b5031-01fb-4f14-8829-6748f819eac4",
+	"label": "University of California Press Books",
+	"creator": "Abe Jellinek",
+	"target": "^https://www\\.ucpress\\.edu/(book/|search.php\\?)",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2021-06-29 06:45:51"
+}
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2021 Abe Jellinek
+	
+	This file is part of Zotero.
+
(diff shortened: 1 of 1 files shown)


University of California Press Books.js
```diff
@@ -0,0 +1,403 @@
...
+	}
+	else {
+		scrape(doc, url);
+	}
+}
+
+function scrape(doc, url) {
+	var translator = Zotero.loadTranslator('web');
+	// Embedded Metadata
+	translator.setTranslator('951c027d-74ac-47d4-a107-9c3069ab7b48');
```
boan-anbo, line 78: Hi Abe, sorry for non-review-related comment--I was just browsing to learn how to write better translators. Please feel free to ignore my question: I see this line here and there but never understood what function it does. Could you explain shortly or point to documentation? Thx!

AbeJellinek: For sure! `Zotero.loadTranslator` returns a `Translate` object, which is essentially a driver for the translation system. Passing `'web'` sets it up to run web translators, but it doesn't know *which* translator to run yet. That's what `Translate.setTranslator` is for. You pass in a translator's unique ID, as in the metadata block:

https://github.com/zotero/translators/blob/7996f2ede84b514023d2f55d0ec6343e3160e4a3/Embedded%20Metadata.js#L2

In this case we want EM, so we pass its ID. Calling `Translate.getTranslatorObject` gives us an instance of EM with `detectWeb` and `doWeb` methods that we can call.

Loading search translators looks a bit different:

https://github.com/zotero/translators/blob/d147639248c81830f727928d9b7256edee05d306/COinS.js#L95-L120

AbeJellinek: Take a look at the [type definition file](https://github.com/zotero/translators/blob/master/index.d.ts) in this repo if you're ever wondering what returns what!

boan-anbo: I see! I didn't know there are these great generic translators we could call. Thanks so much!

PR #2503, opened 2021-06-30
ScienceDirect: Construct PDF URL from JSON metadata

abroenni:
When using a partner institution network access or via an institution vpn, the PDF URL can sometimes be constructed directly.


ScienceDirect.js
```diff
@@ -78,6 +78,15 @@ function getPDFLink(doc, onDone) {
 		return;
 	}
 	
+	// when using a partnered institution network access, construct the the PDF URL directly
+	var pdfURLdirect = doc.querySelector('link[rel="canonical"]').href;
```
dstillman, line 82: This can be `attr(doc, 'link[rel="canonical"]', 'href')` — cleaner, and it won't error out if there's no rel="canonical" link on the page.

And can just reuse `pdfURL` here.

ScienceDirect.js
```diff
@@ -78,6 +78,15 @@ function getPDFLink(doc, onDone) {
...
+	if (pdfURLdirect) {
+		pdfURLdirect = pdfURLdirect + '/pdfft?isDTMRedir=true&download=true';
```
dstillman, line 84: Can you explain the query string? Does the canonical link take you to an intermediate page? If so, is the intermediate page different from the intermediate page that `citation_pdf_url` takes you to when present?

ScienceDirect.js
```diff
@@ -78,6 +78,15 @@ function getPDFLink(doc, onDone) {
 		return;
 	}
 	
+	// when using a partnered institution network access, construct the the PDF URL directly
```
dstillman, line 81: → `// When using partnered institution network access, construct the PDF URL directly`

dstillman (changes requested):
Thanks! A few comments.

ScienceDirect.js
abroenni: I had noticed when clicking the  "View PDF" or "Download PDF" (sometimes this one expands into a dropdown and "Download PDF" needs to be clicked again), the pdf opens in a new tab in an online pdf reader titled Elsevier Enhanced Reader. At the top is a blue arrow-down download button. The link behind this button always follows the same pattern: the base url + the parameters. This base url appears in a `<link rel="canonical" href="...">` element on the landing page with the View or Download PDF buttons.

It is a little hacky with the string concatting, but seems to be the easiest and shortest way to a pdfURL.

abroenni: The canonical URL appears to be just another URL to the same landing page with the View or Download PDF buttons.

dstillman: Is this PR a fix or an optimization? The existing translator should pick up PDF links, including in the drop-down (though that one only works in the Connector, where we can trigger JS, and there are technical reasons why avoiding that is actually a current priority). Did you find cases where it didn't but `rel="canonical"` was available?

An optimization would be good, particularly to avoid the drop-down link, but a concern obviously would be that it would be susceptible to changes on the site (e.g., different required parameters added to the canonical URL), and we don't currently have a way of detecting if a constructed download link is going to fail and falling back to another — we can only build our best guess for a functioning download link based on the information on the page.

So, e.g., we still might want to put this below the `.PdfEmbed` block, which is at least based on what's on the page, even if we put it above the drop-down condition.

dstillman: Well, and also, you didn't answer my question about the intermediate page. Basically I'm asking if `parseIntermediatePDFPage()` works on the canonical URL. If so, I'd be inclined to just add this below the `citation_pdf_url` line and let it still use `parseIntermediatePDFPage()` rather than hard-coding the parameters, for the reasons I give above.

But I'm also just not clear on whether the canonical URL is ever actually present when `#pdfLink` and/or `citation_pdf_url` aren't and what value that URL is actually providing, separate from the hard-coded parameters.

abroenni: > Is this PR a fix or an optimization? 
An optimization, I guess. PDF retrieval fails without it for some of links.
> Did you find cases where it didn't but rel="canonical" was available? 
Yes. 

I understand want you want to do. Let me debug with a few tests.

abroenni: Commenting my code block, I locally added comments to 
```
// Some pages still have the PDF link available
	var pdfURL = attr(doc, '#pdfLink', 'href');
	Zotero.debug("If #pdfLink existed this is the pdfURL: " + pdfURL);
	if (!pdfURL) pdfURL = attr(doc, '[name="citation_pdf_url"]', 'content');
	Zotero.debug("If citation_pdf_url existed this is the pdfURL: " + pdfURL);
	if (pdfURL && pdfURL != '#') {
		parseIntermediatePDFPage(pdfURL, onDone);
		return;
	}
...
var intermediateURL = attr(doc, '.PdfEmbed > object', 'data');
	Zotero.debug("If intermediateURL existed this is the intermediateURL: " + intermediateURL);
```
and for the DOI I am using, I get the following in the debug output:
```
(3)(+0000015): Translate: Beginning translation with RIS
(3)(+0000008): Translate: Saving item
(5)(+0000001): Translate: Running handler 0 for itemDone
(3)(+0000003): Translate: If #pdfLink existed this is the pdfURL:
(3)(+0000001): Translate: If citation_pdf_url existed this is the pdfURL:
(3)(+0000001): Translate: If intermediateURL existed this is the intermediateURL:
(3)(+0000000): Translate: Refetching HTML for PDF link
```
pdfLink, citation_pdf_url, intermediateURL did not exit and parseIntermediatePDFPage is never called.

dstillman: What's the DOI?

abroenni: [https://doi.org/10.1016/S0963-8687(96)80004-7](https://doi.org/10.1016/S0963-8687(96)80004-7) (Currie & Willcocks)

abroenni:
If have been trying to understand why these values remain empty:
```
(3)(+0000003): Translate: If #pdfLink existed this is the pdfURL:
(3)(+0000001): Translate: If citation_pdf_url existed this is the pdfURL:
(3)(+0000001): Translate: If intermediateURL existed this is the intermediateURL:
```
After adding the following two lines at the top of the `getPDFLink` function:  
```
Zotero.debug("The website outer content: " + doc.body.outerHTML);
Zotero.debug("The website inner content: " + doc.body.innerHTML);
```
the log entry started with 
```
(3)(+0000013): Translate: The website inner content: 
<body> <noscript> JavaScript is disabled on your browser. Please enable JavaScript to use all the features on this page.
        <img
            src="https://smetrics.elsevie..
...
```
It seems to be a reduced website, that hasn't finished loading yet. My config prefs for javascript.enabled is set to true.
When I use Scaffold to debug ScienceDirect.js I can check that Javascript enabled(just load the url [JSTest](https://www.whatismybrowser.com/detect/is-javascript-enabled) in the Scaffold's Browser tab). Then, some of the values get populated but not in a currently usable way:
```
11:49:32 ...trying #pdfLink. pdfURL is: 
11:49:32 ...trying citation_pdf_url. pdfURL is: 
11:49:32 ...trying intermediateURL directly using .PdfEmbed object. intermediateURL is: /science/article/pii/S0963868796800047/pdf?md5={md5here}&pid={pidnumberhere}.pdf&isDTMRedir=Y
```
Is it a loading problem with too much delay, because the embedded pdf object can be quite large?

abroenni:
I had also noticed that the loaded DOM contains a `script` object with JSON content that has all the details needed to construct a direct pdfURL.
```
<script type="application/json">
...
  "pdfDownload": {
        "linkType": "DOWNLOAD",
        "isPdfFullText": true,
        "urlMetadata": {
            "queryParams": {
                "md5": "{md5sumher}",
                "pid": "{documentpidhere}.pdf"
            },
            "pii": "S{pii}",
            "pdfExtension": "/pdf",
            "path": "science/article/pii"
        }
... 
```
As long as the direct PDF URL pattern does not change, a pdfURL could be created. If this is acceptable, given that risk that the pattern can change any time.

AbeJellinek:
Re: your first comment, are you running via the Testing pane in Scaffold or via the `Run doWeb`/`Run detectWeb` buttons with the page already loaded in the browser tab?

If you're running via the Testing pane, page JavaScript isn't given time to run by default. You'll need to add the `"defer": true` attribute to the tests in question.

That said, it would be a better approach overall to use the JSON approach you just described. Yeah, it might break, but make it break gracefully (don't fail translation entirely if the JSON is missing/its schema has changed) and it'll be fine.

abroenni:
Hi, I have added the option to create a pdfURL when the main page contains a JSON script element. 
Using var `var jsondata = attr(doc,'script[type="application/json"]', text);` doesn't store the json content in `jsondata`, but when I use
`var jsondata = doc.querySelector('script[type="application/json"]').innerHTML;` it works. Am I using the attr() function wrong?

The script section looks like so:
```
<script type="application/json" data-iso-key="_0">{"abstracts":{"content":[{"$$":[{"#name"
...
SpecialIssueLink":false}}</script>
```

dstillman:
> when I use `var jsondata = doc.querySelector('script[type="application/json"]').innerHTML;` it works. Am I using the attr() function wrong?

You want `text()` for this (which is equivalent to `.textContent`, but that's what's appropriate here). `attr()` is for attribute values.

ScienceDirect.js
```diff
@@ -117,6 +108,46 @@ function getPDFLink(doc, onDone) {
 		}
 	}
 	
+	// Attempting to use json data, which might be available to construct a direct pdf link.
+	// A json data element might be available when using partnered institution network access.
+	// The same pdf url is created that would be available from the embedded pdf object,
+	// however the embedded object code may not always be available for parsing.
+	var jsondata = doc.querySelector('script[type="application/json"]').innerHTML;
```
dstillman, line 115: `jsondata` → `json`

ScienceDirect.js
```diff
@@ -117,6 +108,46 @@ function getPDFLink(doc, onDone) {
...
+	if (jsondata) {
+		try {
+			jsondata = JSON.parse(jsondata);
+			Zotero.debug("Found json data");
+		}
+		catch {
```
dstillman, line 121: This `catch` needs to cover the whole block, not just the `JSON.parse()`. If the current catch triggered, `jsondata` wouldn't be set and the below lines would cause the entire translation to fail.

ScienceDirect.js
```diff
@@ -117,6 +108,46 @@ function getPDFLink(doc, onDone) {
...
+			Zotero.debug(e, 2);
+		}
+		var path = jsondata.article.pdfDownload.urlMetadata.path;
```
dstillman, line 124: `var` → `let` for all variables within blocks

ScienceDirect.js
```diff
@@ -117,6 +108,46 @@ function getPDFLink(doc, onDone) {
...
+		var pdfExtension = jsondata.article.pdfDownload.urlMetadata.pdfExtension;
+		var pii = jsondata.article.pdfDownload.urlMetadata.pii;
+		var md5 = jsondata.article.pdfDownload.urlMetadata.queryParams.md5;
+		var pid = jsondata.article.pdfDownload.urlMetadata.queryParams.pid;
+		if (path && pdfExtension && pii && md5 && pid){
+			pdfURL = doc.location.protocol + '//' + doc.location.hostname + '/' + path + '/' + pii + pdfExtension + '?md5=' + md5 + '&pid=' + pid + '&isDTMRedir=Y';
```
dstillman, line 130: This can be a bit cleaner with string interpolation (`` `foo ${bar} foo` ``).

ScienceDirect.js
```diff
@@ -117,6 +108,46 @@ function getPDFLink(doc, onDone) {
...
 	
+	// Attempting to use json data, which might be available to construct a direct pdf link.
+	// A json data element might be available when using partnered institution network access.
+	// The same pdf url is created that would be available from the embedded pdf object,
+	// however the embedded object code may not always be available for parsing.
+	var jsondata = doc.querySelector('script[type="application/json"]').innerHTML;
+	if (jsondata) {
+		try {
+			jsondata = JSON.parse(jsondata);
+			Zotero.debug("Found json data");
```
dstillman, line 119: `json` → `JSON`

ScienceDirect.js
```diff
@@ -117,6 +108,46 @@ function getPDFLink(doc, onDone) {
...
+			Zotero.debug(e, 2);
+		}
+		var path = jsondata.article.pdfDownload.urlMetadata.path;
+		var pdfExtension = jsondata.article.pdfDownload.urlMetadata.pdfExtension;
+		var pii = jsondata.article.pdfDownload.urlMetadata.pii;
+		var md5 = jsondata.article.pdfDownload.urlMetadata.queryParams.md5;
+		var pid = jsondata.article.pdfDownload.urlMetadata.queryParams.pid;
+		if (path && pdfExtension && pii && md5 && pid){
+			pdfURL = doc.location.protocol + '//' + doc.location.hostname + '/' + path + '/' + pii + pdfExtension + '?md5=' + md5 + '&pid=' + pid + '&isDTMRedir=Y';
+			Zotero.debug("Created pdfURL from json data: " + pdfURL);
```
dstillman, line 131: - `pdfURL` → `PDF URL`
- `json` → `JSON`

ScienceDirect.js
```diff
@@ -117,6 +108,46 @@ function getPDFLink(doc, onDone) {
...
+			return;
+		}
+		else {
+			Zotero.debug("Missing elements in json data required for url creation.");
+		}
+	}
+
+	// The intermediate page provides a download button that uses a canonical link to download the pdf.
+	// In case, the intermediate page fails to load, the pdf can still be downloaded directly without loading
+	// the intermediate page. By creating the same canonical url link format, the pdf can be retrieved directly.
```
dstillman, line 142: - Capitalize acronyms (PDF, URL)
- `In case, the intermediate` → `In case the intermediate`

But I'm not following this comment. What does this have to do with the intermediate page? This block is only relevant if we didn't actually generate a URL via any method above. This isn't something that's relevant if "the intermediate page fails to load".

dstillman (changes requested):
I can't test the site at the moment, but a few comments. @AbeJellinek may have others.

ScienceDirect.js
```diff
@@ -108,15 +112,58 @@ function getPDFLink(doc, onDone) {
...
 	}
 	
-	// If none of that worked for some reason, get the URL from the initial HTML, where it is present,
-	// by fetching the page source again. Hopefully this is never actually used.
+	// On some institutional networks with access to ScienceDirect, the site
+	// serves JSON metadata probably used to preload dynamic content without a
+	// separate network request. If we find it, we can take advantage of it to
+	// grab the PDF url's parts directly, constructing it in the same way that
+	// the site's frontend JavaScript would.
+	var json = doc.querySelector('script[type="application/json"]').innerHTML;
```
dstillman, line 120: We should be a bit more defensive just in case there's ever not a `<script type="application/json">` block at all, which would cause this to throw.

ScienceDirect.js
```diff
@@ -74,22 +69,31 @@ function getPDFLink(doc, onDone) {
...
 		return;
 	}
 	
 	// If intermediate page URL is available, use that directly
 	var intermediateURL = attr(doc, '.PdfEmbed > object', 'data');
 	if (intermediateURL) {
-		// Zotero.debug("Embedded intermediate PDF URL: " + intermediateURL);
-		parseIntermediatePDFPage(intermediateURL, onDone);
+		Zotero.debug("Found embedded PDF URL: " + intermediateURL);
+		if (intermediateURL.includes('isDTMRedir=Y')) {
```
dstillman, line 81: Might this be `isDTMRedir=true` too?

PR #2513, opened 2021-07-05
RIS: Fix parsing error when attachment URL is not encoded correctly

adomasven:
See report https://forums.zotero.org/discussion/90606/error-importing-from-paperpile-report-id-506998060

Report log contains
> [JavaScript Error: "URIError: malformed URI sequence" {file: "chrome://zotero/content/xpcom/translation/translate_firefox.js line 425 > eval" line: 1398}]"

### DIFF
--- a/RIS.js
+++ b/RIS.js
@@ -1395,9 +1395,16 @@ function processTag(item, tagValue, risEntry, allowDeprecated) {
 					
 					//get title from file name
 					title = url.match(/([^\/\\]+)(?:\.\w{1,8})$/);
-					if (title) title = decodeURIComponent(title[1]);
-					else title = "Attachment";
-					
+					if (title) {
+						try {
+							title = decodeURIComponent(title[1]);
+						}
+						catch (e) {}
+					}
+					else {
+						title = "Attachment";
+					}
+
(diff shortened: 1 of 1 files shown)


AbeJellinek:
Should the title also be set to `Attachment` in the `catch` block? Otherwise we end up with an attachment with an empty title.

RIS.js
```diff
@@ -1395,9 +1395,16 @@ function processTag(item, tagValue, risEntry, allowDeprecated) {
...
 					//get title from file name
 					title = url.match(/([^\/\\]+)(?:\.\w{1,8})$/);
-					if (title) title = decodeURIComponent(title[1]);
-					else title = "Attachment";
-					
+					if (title) {
+						try {
+							title = decodeURIComponent(title[1]);
+						}
+						catch (e) {}
```
AbeJellinek, line 1402: I initially misread this and thought that an error would result in an empty title, but it looks like the issue is actually that an error will result in gibberish, since it'll set the attachment title to the regex match array. Try this?

```suggestion
						catch (e) {
							title = title[1];
						}
```

AbeJellinek (changes requested):

PR #2520, opened 2021-07-06
CI: Don't limit fetch depth at all

AbeJellinek:
I'm not so sure that limiting fetch depth is actually gaining us anything, and it makes reviewing old (you know, > 2 weeks) PRs difficult. I'm not entirely sure that this tweak will fix the problem, but it's worth a try.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
       # and get more history on the current branch so we can find the branch point
       - run: git fetch origin master:master --depth=1
         if: github.ref != 'refs/heads/master'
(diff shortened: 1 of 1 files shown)


dstillman:
That failure might not be meaningful. Let's wait for another translator commit and evaluate after that.

dstillman:
I guess this won't work because it's already a shallow repo, so a fetch without an explicit depth won't retrieve past the shallow boundary. I suppose `fetch --unshallow` would probably work, but we could probably also just set `fetch-depth: 0` for the [`actions/checkout` action](https://github.com/actions/checkout). Or we could play with something like `--shallow-since=`, but I guess there's no real reason to do that. We can always change it back to a shallow repo after the backlog is cleared.

PR #2535, opened 2021-07-08
Add translator for Woocommerce bookstore sites

AbeJellinek:
I'm not completely sure whether this is a good idea compared to writing individual translators for Woocommerce sites that would actually be useful with Zotero, but it seemed worth a try.


AbeJellinek:
Closing this; it essentially amounts to a domain-specific ISBN detector and it isn't very useful.

PR #2541, opened 2021-07-12
Add translator for the Library of Congress's METS format

AbeJellinek:
Really just a wrapper for MODS, MARC, etc. Supports a lot of formats that we haven't dealt with yet, but all the test cases I can find in the wild are MODS or MARC.

@adam3smith, how should I handle unsupported metadata formats? Should an item with metadata in EAD format cause translation to fail, or will it be better just to log a debug message and skip it?


adam3smith:
I think I would be inclined to throw an error and fail, ideally logging something that allows us to infer the format. My thinking is that people may use this for large imports, and then having a couple of items missing is bad.

PR #2553, opened 2021-07-14
MODS: Various detection and translation improvements

AbeJellinek:
- Support `letter`s, detecting them based on the presence of an `rcp` (recipient) MARC relator. It's possible that there's a genre heuristic we could use, but I don't have a good example of an appropriate item. (The letter item I'm working with has all its genre information, etc., in German.)
- Support `hearing`s, and don't treat all government publications as books.
- Improve detection: `tagName` includes the tag's namespace, so in order to detect namespaced MODS documents we need to use `endsWith` instead of `===`.
- When possible, fill the `language` field using language codes instead of localized names.
- Allow non-inverted single-part names, exclude publishers from `creators`, and use appropriate fallback creator types for item types that don't support the usual one (e.g. `hearing` doesn't support `author`).
- Actually process all attachments, not just the first. (That was a bug.)


AbeJellinek:
- Support `letter`s, detecting them based on the presence of an `rcp` (recipient) MARC relator. It's possible that there's a genre heuristic we could use, but I don't have a good example of an appropriate item. (The letter item I'm working with has all its genre information, etc., in German.)
- Support `hearing`s, and don't treat all government publications as books.
- Improve detection: `tagName` includes the tag's namespace, so in order to detect namespaced MODS documents we need to use `endsWith` instead of `===`.
- When possible, fill the `language` field using language codes instead of localized names.
- Allow non-inverted single-part names, exclude publishers from `creators`, and use appropriate fallback creator types for item types that don't support the usual one (e.g. `hearing` doesn't support `author`).
- Actually process all attachments, not just the first. (That was a bug.)

PR #2560, opened 2021-07-18
zotero.org: Don't try to scrape multiples when empty

AbeJellinek:
Previously, opening an empty collection would display the folder icon in the connector. When clicked, the translator would try to call `selectItems` with empty input and the connector would display an error. This fixes detection so that we no longer detect anything if we're in an empty collection.


AbeJellinek:
oops. third time's the charm, they say

PR #2566, opened 2021-07-20
Add translator for Substack

AbeJellinek:


Substack.js
```diff
@@ -0,0 +1,164 @@
...
+	item.abstractNote = json.description;
+	item.date = ZU.strToISO(json.dateModified || json.datePublished);
+	item.creators.push(ZU.cleanAuthor(json.author.name, 'author'));
+	item.blogTitle = json.publisher.name;
+	item.url = attr(doc, 'link[rel="canonical"]', 'href');
+	item.attachments.push({
+		title: 'Snapshot',
+		document: doc
+	});
+	item.websiteType = 'Substack newsletter';
```
dstillman, line 89: Should this just be "Newsletter"? "Substack" seems a bit redundant here with the URL.

(I've never been clear on how exactly the type field is used by citation styles. @adam3smith, @bwiernik?)

adam3smith: It's a description of the type of resource like "tweet" or "Infographic" or "Facebook page". See e.g. this sample entry by @bwiernik for APA:
https://www.zotero.org/groups/2205533/items/7AYFWZNP
So I guess the question is if substack is generic like Twitter and Facebook by now. I'd be inclined to say it is, so to keep Substack, in particular because it's nowhere else in the citation info (other than the URL).

Related question is if the item type should be blog or webpage. The two are obviously closely related and I think substack has basically replaced blog, but at least for Chicago and APA style, we mainly use webpage for "all sorts of online things"

bwiernik: Yes, since the Substack name is taking up the Website Title field, let's keep "Substack" in the type field. I agree that "Substack newsletter" is similar at this point to "Facebook post" or "Tweet". 

I think making this a Blog post makes sense considering that these are serialized. I suggest we do that.

AbeJellinek: OK, I'll leave it as `blogPost` with "Substack" in the type field. I would definitely argue that "Substack" is generic and conveys a particular type of publication that there isn't a better adjective for: an email newsletter that's also published online, often behind a paid subscription.

PR #2567, opened 2021-07-20
Twitter: Don't detect untitled tweet views, fix link attachment regressions

AbeJellinek:
This primarily affects when a user opens a tweet by clicking on the image, not the text, to get the full-screen lightbox view. Twitter doesn't update the page title in that case, so it's very difficult for us to scrape anything useful. Previously the translator would detect 'blogPost' but error out in `doWeb`; now we'll just not detect anything at all. Better than a confusing error.

### DIFF
--- a/Twitter.js
+++ b/Twitter.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2021-07-20 02:25:24"
+	"lastUpdated": "2021-07-22 20:11:59"
 }
 
 /*
@@ -61,7 +61,7 @@ function unshortenURLs(doc, str) {
 
 function unshortenURL(doc, tCoURL) {
 	var a = doc.querySelector('a[href*="' + tCoURL + '"]');
-	return (a ? a.title : false) || tCoURL;
+	return (a ? a.textContent.replace(/…$/, '') : false) || tCoURL;
 }
 
 function extractURLs(doc, str) {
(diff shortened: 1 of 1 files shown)


Twitter.js
```diff
@@ -271,7 +270,7 @@ var testCases = [
 		"items": [
 			{
 				"itemType": "blogPost",
-				"title": "Zotero 3.0 beta is now available with duplicate detection and tons more. Runs outside Firefox with Chrome or Safari! http://www.zotero.org/blog/announcing-zotero-3-0-beta-release/",
+				"title": "Zotero 3.0 beta is now available with duplicate detection and tons more. Runs outside Firefox with Chrome or Safari! http://t.co/bRvJgRy",
```
dstillman, line 273: I like to leave tests broken when they're actually regressions rather than incorporating the worse data. This may not actually be fixable at the moment, but it's not desirable behavior, and the tests help us remember that.

AbeJellinek: OK, I restored the original tests and more or less fixed the regressions that cropped up. The only remaining problem is that `www` subdomains are stripped before the URL is inserted into the page; `https://www.zotero.org/example` becomes `https://zotero.org/example`. For almost all sites, that isn't an issue, so I'm not too concerned, but it is odd. What do you think?

dstillman: Oh, weird — because they show it as `zotero.org/blog/announcin…`, but the hidden part before only has `http://` rather than `http://www`.

It does show the full URL with `www` in a JS tooltip on hover. We could try to figure out where that's coming from…

AbeJellinek: The internal API, and then it gets stored in a state object somewhere. We could try to make API calls like the frontend does - I looked into it earlier and it only seems to require one random-looking key that appears(?) to be consistent between pages and sessions - but I think that might be too fragile. There's no way they've left their internal API as open as it looks at first glance.

dstillman: OK, well we can just live with this for now.

(We probably should've switched zotero.org over to the apex domain years ago when we switched to AWS and got Alias functionality.)

dstillman:
`detectWeb()` change reverted in https://github.com/zotero/translators/commit/971a137c022d2e7fd7fb2adf24fb3cf5ae878305

Issue #1512, opened 2018-01-05
Update PubMed import translator for 5.0


dstillman:
https://forums.zotero.org/discussion/comment/297360/#Comment_297360

dstillman:
Example: https://github.com/zotero/translators/commit/e38e8f403116b050586df382bc9ad5d8d0a91191?w=1

PR #2575, opened 2021-07-21
Fix importing PubMed XML with only a non-English title

AbeJellinek:
I believe that this fixes the issue that spawned #1512 by using `<VernacularTitle>` as a fallback when an English title can't be found. Other than that, though, I am not sure what would benefit from an upgrade to 5.0 translator APIs.

### DIFF
--- a/PubMed XML.js
+++ b/PubMed XML.js
@@ -11,7 +11,7 @@
 	},
 	"inRepository": true,
 	"translatorType": 1,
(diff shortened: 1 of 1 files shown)


dstillman:
Can merge this now, but 5.0 thing is explained in https://github.com/zotero/translators/issues/1353

AbeJellinek:
Thanks.

PR #2576, opened 2021-07-21
[WIP] MARC translator modernization

AbeJellinek:
Some efforts to modernize and modularize the MARC translator.

- [x] Pseudo-class -> class
- [x] Call Zotero utilities directly
- [ ] Use modern JS syntax (partially done)
- [ ] Allow callers to add custom field mappings
- [ ] Add tests for UNIMARC

### DIFF
--- a/MARC.js
+++ b/MARC.js
@@ -8,7 +8,7 @@
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 1,
-	"lastUpdated": "2021-07-13 23:18:07"
+	"lastUpdated": "2021-07-13 23:35:05"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


AbeJellinek:
@adam3smith Do you have a binary UNIMARC record that I could use as a test case? Surprisingly hard to find test data sometimes.

adam3smith:
None that I know of. Generally relatively little raw MARC out there that I'm aware of. Not sure how hard it is to convert the human readable format into machine-readable? (e.g. the French National Library has human-readable UNIMARC)

AbeJellinek:
CI times out, so I'll test all this locally.

PR #2577, opened 2021-07-21
Add translator for NTSB Accident Reports

AbeJellinek:
Prepared after transportation crashes, disasters, etc. Relatively structured (thanks, SharePoint!) and worth having support for.

### DIFF
--- a/NTSB Accident Reports.js
+++ b/NTSB Accident Reports.js
@@ -1,0 +1,243 @@
+{
+	"translatorID": "fef8575b-da80-40ec-8084-631a08cae2bd",
+	"label": "NTSB Accident Reports",
+	"creator": "Abe Jellinek",
+	"target": "^https?://(www\\.)?ntsb\\.gov/investigations/AccidentReports/Pages/",
+	"minVersion": "3.0",
+	"maxVersion": "",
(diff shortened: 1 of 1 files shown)


NTSB Accident Reports.js
```diff
@@ -0,0 +1,243 @@
...
+				"url": "https://www.ntsb.gov/investigations/AccidentReports/Pages/HWY21MH006-preliminary-report.aspx",
+				"attachments": [
+					{
+						"title": "Snapshot",
+						"mimeType": "text/html"
+					},
+					{
+						"title": "Full Text PDF",
+						"mimeType": "application/pdf"
+					}
```
dstillman, line 229: Do we really need both here? Seems like the PDF has the full report.

(Also, this crash, my god.)

AbeJellinek: True. For some reason I thought that the snapshot had images that the PDF lacks, but that's not the case.

And yeah. 23 passengers in an SUV. It's horrible. [The Times' story about it](https://www.nytimes.com/2021/04/04/us/migrants-crash-california-imperial-valley.html) is disturbing and definitely worth a read.

PR #2578, opened 2021-07-21
EM: Add heuristic for two-word Spanish surnames

### LINKED ISSUE #1333: Spanish author names break splitting multiple authors heuristic in EM translators
See the lines https://github.com/zotero/translators/blob/master/Embedded%20Metadata.js#L384-L390 with the example
```html
<meta content="Ahumada Sempoal, Miguel Angel" name="citation_author">
```
from http://upcommons.upc.edu/handle/2117/14979

(occured during rewrite of UPCommons which is part of #1332)

AbeJellinek:
This change is intended to address one edge case in EM author splitting: single-author items with Highwire metadata where the author has a name like (as in the test case) "García García, Antonio A." Previously, that author's name would have been split into two separate creators: "García García" and "Antonio A." Not ideal behavior.

My solution is to add a fairly restrictive set of conditions that disable comma-splitting: there is only one comma, the Highwire metadata indicates that the item is in Spanish, and the text before the comma ("García García") has exactly two space-separated parts.

I'd be interested in hearing others' opinions about whether this is a reasonable and safe heuristic.

Fixes #1333.


adam3smith:
I'm OK with this, yes. 
I
t's obviously not 100% safe -- two authors in firstname lastname format aren't _that_ rare and having them in a single tag will occur -- but among other things I think we should prioritize correctly importing well-formatted data, i.e. "reward" pages that use "family, given" by correctly importing their data even if it (rarely) breaks "given family" import. Or to put it differently, with current behavior, there's no way to format "García García, Antonio A." in EM so that it imports correctly, which I think is a problem.

PR #2585, opened 2021-07-23
GitHub: Extract repo name from URL, not `og:title`

ssauvin:
Hello,

GitHub and Zotero worked well before but something seems broken now (Zotero Connector for Firefox detects all repositories as "Embedded Metadata" instead of Software).

After a quick investigation, it seems that GitHub itself, **sometimes**, adds  prefixes (`GitHub - `) and suffixes (`: repository description`) to meta attribute `og:title`. I tried to figure out when this prefixes/suffixes are added, and it seems somewhat related to authenticated session. This attribute is not consistent either when looking at a single file, the `og:title` looks like `foo/bar.md at branch · user/repo`.

I tried my best to fix the GitHub translator and make it work, right now all tests are passing in Zotero/Scaffold (in MacOS).

Since it's my first time editing JavaScript/Zotero code, feedback is welcome! :)

Thanks!

### DIFF
--- a/GitHub.js
+++ b/GitHub.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2021-06-24 21:22:14"
+	"lastUpdated": "2021-07-23 14:14:10"
 }
 
 /**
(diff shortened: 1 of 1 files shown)


AbeJellinek:
Odd! This looks good to me. The URL is definitely more stable than `og:title`, so we probably should've been doing it that way all along.

PR #2588, opened 2021-07-27
BibTeX: Support day field, reorganize date construction

AbeJellinek:


AbeJellinek:
OK, seems that tests are still working fine. (The date change on one of the CiNii tests is because they changed their BibTeX.)

dstillman:
We should probably add a comment explaining that `day` isn't an official BibTeX field but that some sites include it. (I didn't actually research that extensively so I don't know who/what uses or specifies it, but that's my understanding.)

PR #2595, opened 2021-07-30
Add Bibliothèque nationale de France ISBN search translator

AbeJellinek:
Based on discussion [on the forums](https://forums.zotero.org/discussion/90909/book-isbn-not-found#latest). We should discuss whether we want this. I think it's a good idea and would fill some major gaps in our ISBN search support. ISBN search with BnF's SRU endpoint takes about 400 ms, whether or not an item is returned. Not incredibly fast, but better than K10plus (> 1 second per request from here).

### DIFF
--- a/BnF ISBN.js
+++ b/BnF ISBN.js
@@ -63,6 +63,16 @@ function doSearch(item) {
 			for (let marcRecord of marcRecords) {
 				marcRecord.setAttribute('xmlns:marc', 'http://www.loc.gov/MARC21/slim');
 
+				// Here we convert the XML we get from BnF from the original
+				// MarcXchange format into MARCXML by search-and-replacing
+				// the namespace. MARCXML and and MarcXchange are essentially
+				// the same format: the latter is a "generalization (mainly by
+				// weakening restrictions)" of the former. We didn't enforce
+				// those restrictions to begin with.
+
+				// MarcXchange spec:
+				// https://www.loc.gov/standards/iso25577/ISO_DIS_25577__E_.pdf
+
 				let marcxchangeText = new XMLSerializer().serializeToString(marcRecord);
 				let marcXMLText = marcxchangeText.replace(/<mxc:/g, '<marc:').replace(/<\/mxc:/g, '</marc:');
 
--- a/BnF ISBN.js
+++ b/BnF ISBN.js
@@ -1,0 +1,119 @@
+{
+	"translatorID": "f349954c-9957-4b5f-be24-1a8bb52f7fbd",
+	"label": "BnF ISBN",
+	"creator": "Abe Jellinek",
+	"target": "",
+	"minVersion": "4.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 8,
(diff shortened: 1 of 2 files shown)


adam3smith:
+1 from me. Apart from the coverage of French literature in general, the fact that France is one of Zotero's biggest "markets" would also favor this. Given the speed, might even be worth considering moving this before K10plus in the query sequence

BnF ISBN.js
```diff
@@ -0,0 +1,119 @@
+{
+	"translatorID": "f349954c-9957-4b5f-be24-1a8bb52f7fbd",
+	"label": "BnF ISBN",
+	"creator": "Abe Jellinek",
+	"target": "",
+	"minVersion": "4.0",
+	"maxVersion": "",
+	"priority": 100,
```
dstillman, line 8: Need to adjust the priority here — Open WorldCat is also 100, so the order is probably undefined.

BnF ISBN.js
```diff
@@ -0,0 +1,119 @@
...
+	mxc: 'info:lc/xmlns/marcxchange-v2'
+};
+
+function detectSearch(item) {
+	return !!item.ISBN;
+}
+
+function doSearch(item) {
+	let ISBN = ZU.cleanISBN(item.ISBN);
+	let url = `http://catalogue.bnf.fr/api/SRU?version=1.2&operation=searchRetrieve&query=bib.isbn%20all%20%22${ISBN}%22`;
```
dstillman, line 50: `https`

BnF ISBN.js
AbeJellinek: OK, I moved LoC to 97 to make room for BnF at 98.

dstillman: Do we not want WorldCat first?

dstillman: I guess @adam3smith said "between GBV/K10 and Worldcat".

How long does WorldCat take? How does WorldCat quality compare to BnF?

Did you put this before K10 because of the speed?

dstillman: (Sorry, I missed @adam3smith's comment above where he suggested this.)

adam3smith: Worldcat quality is bad. Coverage is good, so ideal as a fallback

AbeJellinek: In order to fetch metadata for a list of ISBNs from WorldCat, we have to fetch an HTML page for each one, parse it into a document, extract the ISBN, then make a GET request to an API endpoint and parse the RIS that it returns. I haven't timed anything but the last step, but even just that takes about 680 ms on my end. BnF is definitely a _lot_ faster.

BnF is definitely better quality for French publications. BnF's MARC has creators tagged correctly, while WorldCat tends to make everyone an "author." WorldCat is frequently missing series metadata and the like. Getting BnF's authoritative (UNI)MARC is almost certainly better than getting WorldCat metadata lossily exported to RIS.

AbeJellinek: (edited: extract OCLC ID, not ISBN)

dstillman: OK, sounds good.

PR #2596, opened 2021-07-30
Add translator for Dar Almandumah

AbeJellinek:
The JSTOR of the Arab world. Pretty good ENW metadata, weird MARC. (I'm a little suspicious that the MARC might be in a dialect that we don't support.)


AbeJellinek:
Will update this once MARC improvements are merged.

PR #2600, opened 2021-08-02
EBSCOhost: fix when user clicks on permalink link

cambraca:
When the user manually clicks on the "permalink" link, this markup appears on the page as part of the popup:

`<span class="icon permalink-link"></span>`

This breaks the detectWeb function, which is expecting the first instance of a `.permalink-link` element to be an anchor tag. It's safer to search for the actual link using querySelector.


AbeJellinek:
Excellent, thank you!

PR #2607, opened 2021-08-04
Add translator for Dataverse catalogs

AbeJellinek:


adam3smith:
dataverse is a repository software used by 69 data repositories, mostly with minor modifications. Could we devise a (admittedly fairly complex but not unprecedented) target that includes all of them?
https://docs.google.com/spreadsheets/d/1bfsw7gnHlHerLXuk7YprUT68liHfcaMxs1rFciA-mEo/edit#gid=0

About half explicitly use dataverse in their domain name, the others would likely have to be added individually.

AbeJellinek:
Oh, excellent. I wasn't aware that there were more. Sure, I'll work on that.

PR #2611, opened 2021-08-04
Add translator for Republik.ch (online newspaper)

deragent:
Hi,

I wrote a basic translator for the Republik.ch online newspaper.

The main functions are:
- Extract title
- Extract author list from text
- Extract introduction paragraph as abstract
- Extract date from text
- Extract attachments, both screenshot of website and PDF generated by the website CMS

Most of this information is sadly not presented in structured HTML and therefore has to be matched and extracted from the article text flow.

For now I marked this translator as "Web" only and as "Gecko (Firefox)" only, as this is the only use cases which I could test so far.
As this is my first translator, please advise on how to set these flags properly.

The translator also includes a (short) set of testcases.


Republik.ch.js
```diff
@@ -0,0 +1,388 @@
...
+	"translatorID": "b4c83513-921f-474b-adc3-7017ea690ba7",
+	"label": "Republik.ch",
+	"creator": "Johannes Wüthrich",
+	"target": "^https?://www\\.republik\\.ch/[0-9]*/[0-9]*/[0-9]*/",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "g",
```
AbeJellinek, line 11: This can be `gcsibv`! We're thankfully past the days of significant browser incompatibilities, at least when it comes to translators.

And re: being web-only, that's correct and shouldn't change. Import, search, etc., are for translators that allow you to import new metadata formats, locate items via ISBN, that kind of thing. Not relevant for a translator like this.

Republik.ch.js
```diff
@@ -0,0 +1,388 @@
...
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+function detectWeb(doc, url) {
+	// TODO match different types of articles / debates / updates / blog-posts / newsletters etc.
```
AbeJellinek, line 34: Just matching articles is fine. If possible, you should try to exclude blog posts, newsletters, that kind of thing, since few people will want to add them to Zotero and Embedded Metadata should do a fine enough job when they do.

Republik.ch.js
```diff
@@ -0,0 +1,388 @@
...
+	
+	return "newspaperArticle";
+}
+
+
+function doWeb(doc, url) {
+	// TODO adapt for multiple type of sources (see above)
+	
+	scrape(doc, url);
```
AbeJellinek, line 43: Does the site have search results that we should handle? It's a very cool design, but I couldn't really tell...

Republik.ch.js
```diff
@@ -0,0 +1,388 @@
...
+}
+
+
+function scrape(doc, url) {
+
+	var newItem = new Zotero.Item("newspaperArticle");
+
+	newItem.title = ZU.xpathText(doc, "//article//section[@class='title-block']/h1");
```
AbeJellinek, line 51: If possible, use CSS selectors instead of XPath. This could be `text(doc, 'article section.title-block h1')` (and even `text(doc, 'article h1')` might suffice). More flexible because it doesn't match on the *entire* class attribute, just the presence of one particular class, and it's easier for people who haven't used XPath to understand.

Republik.ch.js
```diff
@@ -0,0 +1,388 @@
...
+	
+	var abstract = ZU.xpathText(doc, "//article//section[@class='title-block']/h1/following::p[1]");
```
AbeJellinek, line 53: For `abstractNote` you should be able to get away with `attr(doc, 'meta[name="description"]', 'content')`. Less finicky.

Republik.ch.js
```diff
@@ -0,0 +1,388 @@
...
+		meta = abstract
+		abstract = "";
+	}
+	newItem.abstractNote = abstract;
+	
+	
+	const reg_date = /(.*),?\s+([0-3][0-9]\.[0-1][0-9]\.[1-3][0-9][0-9][0-9])$/;
+	const reg_function = /(.*)\s+\((.*)\)/
+	
+	var date = meta.match(reg_date)
```
AbeJellinek, line 70: A fun hack: could we just get the date from the URL and avoid all the trouble? Article URLs seem to take the form `/year/month/day/slug`.

We'll also want to pass it through `ZU.strToISO`, which takes a date as a string and puts it into a standard form.

Republik.ch.js
```diff
@@ -0,0 +1,388 @@
...
+		
+			authors = authors[1].split(/,\s|\sund\s/);
+			
+			// Zotero.debug(authors);
+		
+			for(var aa in authors){
+
+				var author = authors[aa].trim();
+				
+				var fction = "author";
```
AbeJellinek, line 91: Best to give this a standard name like `type` - I'm not sure what `fction` stands for.

Republik.ch.js
```diff
@@ -0,0 +1,388 @@
...
+				}
+				
+				newItem.creators.push(ZU.cleanAuthor(author, fction));
+			}
+		}
+	}
+
+	newItem.url = ZU.xpathText(doc, '//meta[@property="og:url"]/@content') || url;
+
+	newItem.attachments.push({
```
AbeJellinek, line 108: Only push a snapshot if we can't get a PDF. (So put it in an `else` after the `if (pdfURL)`.)

Republik.ch.js
```diff
@@ -0,0 +1,388 @@
...
+		url: newItem.url,
+		title: "Snapshot",
+		mimeType: "text/html"
+	});
+
+	var pdfURL = ZU.xpathText(doc, "//article//a[@title='PDF-Optionen']/@href");
+	if(pdfURL) {
+		newItem.attachments.push({
+			url: pdfURL,
+			title: "PDF",
```
AbeJellinek, line 118: Call this `Full Text PDF`.

Republik.ch.js
```diff
@@ -0,0 +1,388 @@
...
+			}
+		}
+	}
+
+	newItem.url = ZU.xpathText(doc, '//meta[@property="og:url"]/@content') || url;
+
+	newItem.attachments.push({
+		url: newItem.url,
+		title: "Snapshot",
+		mimeType: "text/html"
```
AbeJellinek, line 111:
```suggestion
		document: doc,
		title: "Snapshot"
```

So the connector doesn't need to reload the page.

Republik.ch.js
```diff
@@ -0,0 +1,388 @@
...
+				"itemType": "newspaperArticle",
+				"title": "Bruder Berg und Schwester Schwein",
+				"creators": [
+					{
+						"firstName": "Markus",
+						"lastName": "Schärli",
+						"creatorType": "author"
+					}
+				],
+				"date": "04.08.2021",
```
AbeJellinek, line 145: `strToISO` will convert this to standard form (`2021-08-04`, sort of like in the URL).

AbeJellinek (changes requested):
Thanks, this is awesome! I left some comments - let me know what you think.

In terms of general style, we'll want to lint before merging this, so that things like spacing are consistent with the style we use across the Zotero codebase. You can run `npm run lint -- Republik.ch.js --fix` if you have npm, but I'm happy to do it too.

Republik.ch.js
deragent: Done.

Republik.ch.js
deragent: Done

Republik.ch.js
deragent: Done

Republik.ch.js
deragent: Yes it does have a search function. The search (as well as the article feed) is only available to paying members.

This is because of the specific financing model of this newspaper: Articles can be shared freely without restrictions, but the virtual "front-page" and the archive of articles is only available to paying members.

Republik.ch.js
deragent: I changed this in ebf0cd6

Republik.ch.js
deragent: I changed this in 04382ae

Republik.ch.js
deragent: Changed in  46ae941

Republik.ch.js
deragent: Changed in 753c2ba

Republik.ch.js
deragent: In order to get the author list I will need to deconstruct the "meta" text line in any case.
Therefore I would leave this as is, as I do not see a lot of simplifications by taking the date from the title line.

Republik.ch.js
deragent: All content featuring pages are actually structured in a very similar way (speaking from a HTML point of view).
Therefore this translator should work on any content featuring page, be it an article, a newsletter or a discussion page.
I already filter out all non-content pages, via a restrictive target directive: ^https?://www\.republik\.ch/[0-9]*/[0-9]*/[0-9]*/

For comparison:
- This is an article: https://www.republik.ch/2021/08/05/dieser-stoff-ist-ein-einziges-minenfeld
- This a newsletter: https://www.republik.ch/2021/08/05/7-uhr-newsletter
- This a debate: https://www.republik.ch/2021/08/04/auf-welche-privilegien-wuerden-sie-zugunsten-der-natur-verzichten/diskussion

The translator extracts the available information in all of these cases.

Filtering out non articles would actually be quite tedious, as it would mean filtering based on some content. In the structure of the page there is not flag / indication if it is an article or not.

Republik.ch.js
deragent: I would prefer to leave it as is, as the current implementation also gets some pseudo-abstracts for example in the case of newsletters and similar pages: https://www.republik.ch/2021/08/05/7-uhr-newsletter.
On this page the `meta[name="description"]` element is empty, while the translator at the moment extracts the first `p` text in the title-section.

deragent:
Dear Abe,
thanks for the review and all the inputs. I addressed most of them and commented on all the other ones.
I will run the lint once we resolved all the open change requests.

PR #2613, opened 2021-08-04
Add support for EBSCO Ebooks

mjpelmear:
It seems to work fine-ish with only the change in the "target" regex.

### DIFF
--- a/EBSCOhost.js
+++ b/EBSCOhost.js
@@ -2,7 +2,7 @@
 	"translatorID": "d0b1914a-11f1-4dd7-8557-b32fe8a3dd47",
 	"label": "EBSCOhost",
 	"creator": "Simon Kornblith, Michael Berkowitz, Josh Geller",
-	"target": "^https?://[^/]+/(eds|bsi|ehost)/(results|detail|folder|pdfviewer)",
+	"target": "^https?://[^/]+/(eds|bsi|ehost)/(results|detail|folder|pdfviewer|ebookviewer)",
 	"minVersion": "3.0",
 	"maxVersion": "",
 	"priority": 100,


AbeJellinek:
Excellent, thanks! Could you find an ebook that this works for and post the permalink here? Tests for EBSCOhost are a nonstarter, but I'd like to give it a try on my end.

mjpelmear:
Here you go:
http://search.ebscohost.com.proxy.library.upenn.edu/login.aspx?direct=true&db=nlebk&AN=1463586&site=eds-live&ebv=EK&ppid=Page-__-32

AbeJellinek:
Huh, it's not returning any items for me. It seems like ebook catalog pages already worked - is this just for when you're in the reader?

mjpelmear:
Correct.
The catalog page is `/eds/detail/...`
but the reader ("EPUB Full Text") is `/eds/ebookviewer/ebook/...`

Oddly, it's not working for me now, either. I'll take a look again on Monday.

AbeJellinek:
Sounds good, thanks.

mjpelmear:
I looked at this again [and included a new commit to increment the "lastUpdated" metadata value].

When I run this without my change to the regex, I get a "webpage" itemType, where the url is the url from the browser rather than the permalink.

When I run it with my change, the itemType is "book", the url is a permalink (http://search.ebscohost.com/login.aspx?direct=true&db=nlebk&AN=1463586&site=eds-live), and the ISBN/creators/etc. are all populated.

So I think that means it actually is working?

AbeJellinek:
What a pain. I'm not seeing it working on my end, but I also can't view that exact item - I don't think my library has access to it. Could you send the ISBN and I'll give it another try?

This could be something that varies between libraries.

mjpelmear:
Sure. The catalog page lists:
Related ISBNs: 9781440841453. 9781440841460
Accession number: 1463586
("The State and Federal Courts: A Complete Guide to History, Powers, and Controversy" by Christopher P. Banks)

Alternatively, could you add info here so I can try accessing the one that isn't working for you?

PR #2614, opened 2021-08-05
HighWire 2.0: Support [med,bio]Rxiv and preprints

AbeJellinek:
No general way to detect preprints, so we just check for the Rxiv suffix for now.


dstillman:
Discussion: https://forums.zotero.org/discussion/94032/biorxiv-and-medrxiv-not-importing-properply-since-plugin-update

PR #2618, opened 2021-08-09
Add ISBN search to WHO translator

### LINKED ISSUE #2617: WHO ISBN search translator
https://forums.zotero.org/discussion/91092/whos-iris-pdf-metadata-automatic-citation-download-anyone-would-want-to-develop

I think we can just reject detection when the ISBN doesn't start with "978924". (This seems like something we should do for more things?)

I think we can just put this in the existing WHO translator and make it translatorType 12.

AbeJellinek:
Scaffold will delete search tests, so none added.

Fixes #2617.

### DIFF
--- a/WHO.js
+++ b/WHO.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 12,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2021-08-09 17:50:46"
+	"lastUpdated": "2021-08-10 16:40:28"
 }
 
 /*
@@ -152,13 +152,17 @@ function scrape(doc, url) {
 }
 
 function detectSearch(item) {
-	return item.ISBN && ZU.cleanISBN(item.ISBN).startsWith('978924');
+	return item.ISBN && (ZU.cleanISBN(item.ISBN) || '').startsWith('978924');
 }
 
 function doSearch(item) {
 	let ISBN = ZU.cleanISBN(item.ISBN);
 	let url = `https://apps.who.int/iris/discover/export?format=refman&list=discover&rpp=10&etal=0&query=${ISBN}&group_by=none&page=1`;
 	ZU.doGet(url, function (risText) {
+		if (!risText) return;
+
+		risText = risText.replace(/^SE(\s*-\s*[0-9]+).*$/m, 'SP$1');
+
 		var translator = Zotero.loadTranslator("import");
 		translator.setTranslator("32d59d2d-b65a-4da4-b0a3-bdd3cfb979e7"); // RIS
 		translator.setString(risText);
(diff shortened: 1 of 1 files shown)


WHO.js
```diff
@@ -154,6 +149,36 @@ function scrape(doc, url) {
 	});
 }
 
+function detectSearch(item) {
+	return item.ISBN && ZU.cleanISBN(item.ISBN).startsWith('978924');
```
dstillman, line 155: `cleanISBN` can return false, so this would probably throw for an invalid ISBN.

WHO.js
```diff
@@ -154,6 +149,36 @@ function scrape(doc, url) {
...
+		translator.setHandler("itemDone", function (obj, item) {
+			item.libraryCatalog = 'WHO IRIS';
+			item.archive = '';
+			
+			if (item.url) {
+				ZU.processDocuments(item.url, function (recordDoc) {
+					item.attachments.push({
+						title: 'Full Text PDF',
+						mimeType: 'application/pdf',
+						url: attr(recordDoc, 'meta[name="citation_pdf_url"]', 'content')
```
dstillman, line 174: Probably should test this before pushing in case it's missing.

WHO.js
```diff
@@ -154,6 +149,36 @@ function scrape(doc, url) {
 	});
 }
 
+function detectSearch(item) {
+	return item.ISBN && ZU.cleanISBN(item.ISBN).startsWith('978924');
+}
+
+function doSearch(item) {
+	let ISBN = ZU.cleanISBN(item.ISBN);
+	let url = `https://apps.who.int/iris/discover/export?format=refman&list=discover&rpp=10&etal=0&query=${ISBN}&group_by=none&page=1`;
```
dstillman, line 160: Is the RIS quality comparable to the EM we get normally? If it's any lossier, it seems like we should just use the standard scraping routine.

WHO.js
```diff
@@ -148,12 +145,42 @@ function scrape(doc, url) {
...
+	ZU.doGet(url, function (risText) {
```
dstillman, line 161: At least from curl, it looks like this returns an empty 200 response in the case of an unknown ISBN. Not sure what the RIS translator does on an empty string, but maybe cleaner to test for an empty `risText` here and bail.

WHO.js
AbeJellinek: It's more or less equivalent. EM is missing books' editions, the RIS has `numPages` in the wrong field (but I'll fix that). I think RIS makes more sense for a search translator.

WHO.js
AbeJellinek: The RIS translator just returns no items when it gets an empty string, so I don't think bailing will do anything but save us a little time loading the translator. But there's no reason not to, so I'll add it.

dstillman:
👍

PR #2624, opened 2021-08-10
beck-online: Allow detection of proxied URLs

### LINKED ISSUE #2598: Adjust target URL for beck-online
First of all, thank you for the great work done here (in particular @zuphilip)!
Unfortunately, the beck-online translator stopped working for me since I am accessing the site now via a webvpn of my university. The URL has changed from
> https://beck-online.beck.de/

to

> https://beck-online-beck-de.ubproxy.ub.uni-heidelberg.de

Since that, Zotero does not recognize articles on beck-online as articles anymore.
If I'm not mistaken, the removal of the dot in front of the second "beck" in the target URL could fix it without any further troubles.

Also, would it be possible to integrate the beck-online PDF download? The URL to the PDF is
> https://beck-online.beck.de/Print/CurrentMagazine?vpath=[vpath]&printdialogmode=CurrentDoc&actionname=Index&gesamtversionpath=&exportFormat=pdf&options=WithLinks

That would be great feature!

AbeJellinek:
This is a hack. EZproxy replaces dots with dashes in proxied HTTPS URLs, so the connector undoes that by replacing dashes with dots. Unfortunately, that means that `beck-online.beck.de` will be proxied to `beck-online-beck-de` and then de-proxied by the connector to `beck.online.beck.de`. Accepting a dot in addition to a dash sidesteps the issue.

Fixes #2598.

### DIFF
--- a/beck-online.js
+++ b/beck-online.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2021-08-10 17:39:39"
+	"lastUpdated": "2021-08-11 15:41:33"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


dstillman:
What URL gets saved to Zotero in this case?

AbeJellinek:
Oh, that would seem to be a separate issue - the permalink `href` was relative instead of absolute. Fixed now.

There are also some tests that aren't detecting anymore, but I'll work on that in a separate PR.

PR #2638, opened 2021-08-12
Rewrite translator for Toronto Star

qubidt:
Not sure of the last time this translator was working, but it seems to be nonfunctional for the Toronto Star website as of time of writing. I rewrote the translator. Mostly uses the JSON-LD data because it seems to be the most consistent and accurate (e.g. doesn't truncate the description).

Some issues/assumptions:

* the hardcoded `.language = "en-CA"` was kept from the original translator, as were `.publicationTitle` and `.ISSN`. 
* `.libraryCatalog` was added to match the values from the original tests (not sure where this came from originally)
* based the translator off of the Embedded Metadata translator. This seems to be common but most of the data is hand-pulled so it might be more straightforward to skip this
* The tags don't match the original tests. Partially due to the embedded metadata translator and because the tags listed seem to have been removed from the pages

It passes the linter and the existing tests have been validated

### DIFF
--- a/Toronto Star.js
+++ b/Toronto Star.js
@@ -55,10 +55,10 @@ function detectWeb(doc) {
 function getSearchResults(doc, checkOnly) {
 	var items = {};
 	var found = false;
-	var rows = doc.querySelectorAll('a[data-test-id="mediacard"]');
+	var rows = doc.querySelectorAll('a[class*="-mediacard"]');
 	for (let row of rows) {
 		let href = row.href;
-		let title = text(row, '*[data-test-id="mediacard-headline"]');
+		let title = text(row, 'h3');
 		if (!href || !title) continue;
 		if (checkOnly) return true;
 		found = true;
@@ -267,6 +267,16 @@ var testCases = [
 	{
 		"type": "web",
 		"url": "https://www.thestar.com/news/canada.html",
+		"items": "multiple"
+	},
+	{
+		"type": "web",
+		"url": "https://www.thestar.com/authors.sarrouh_maria.html",
+		"items": "multiple"
+	},
+	{
+		"type": "web",
+		"url": "https://www.thestar.com/topic.ottawa.html",
 		"items": "multiple"
 	}
 ]
--- a/Toronto Star.js
+++ b/Toronto Star.js
@@ -105,10 +105,10 @@ function scrape(doc, url) {
 				if (articleData.author) {
 					if (Array.isArray(articleData.author)) {
 						for (let author of articleData.author) {
-							item.creators.push(ZU.cleanAuthor(author.name, 'author'));
+							if (author.name) item.creators.push(ZU.cleanAuthor(author.name, 'author'));
 						}
 					}
-					else {
+					else if (articleData.author.name) {
 						item.creators.push(ZU.cleanAuthor(articleData.author.name, 'author'));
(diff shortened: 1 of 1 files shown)


Toronto Star.js
```diff
@@ -56,10 +55,10 @@ function detectWeb(doc, url) {
 function getSearchResults(doc, checkOnly) {
 	var items = {};
 	var found = false;
-	var rows = ZU.xpath(doc, '//a[span[contains(@class, "story__headline")]]');
-	for (var i = 0; i < rows.length; i++) {
-		var href = rows[i].href;
-		var title = ZU.trimInternal(rows[i].textContent);
+	var rows = doc.querySelectorAll('a[data-test-id="mediacard"]');
```
AbeJellinek, line 58: It looks like `'.c-search-results-item > a'` would be a better selector here. I'm always a little skeptical that these `data-test-id` attributes will stick around, since they're clearly meant to be internal.

Toronto Star.js
```diff
@@ -56,10 +55,10 @@ function detectWeb(doc, url) {
...
+	for (let row of rows) {
+		let href = row.href;
+		let title = text(row, '*[data-test-id="mediacard-headline"]');
```
AbeJellinek, line 61: This can just be `text(row, 'h3')`!

AbeJellinek (changes requested):
Thanks, this is really, really great work!

Just a couple suggestions about selectors below. It would also be good to add a slash onto the end of the target, and you can feel free to add your name to the metadata/copyright if you'd like.

Toronto Star.js
AbeJellinek: Hmm, I see. I didn't realize the selector you had would work on other pages. That is worth keeping, then.

Toronto Star.js
qubidt: I was about to argue that it's unlikely that the headline is exclusively `h3` nor that `h3` are exclusively headlines. But from the   pages I tested this does seem to be the case, lol.

It also made me realize that the other selector missed some edge cases

Toronto Star.js
qubidt: I realized that the selector I had still missed some articles (see `.c-feature-mediacard` on [this page](https://www.thestar.com/topic.ottawa.html)). The new selector (`a[class*="-mediacard"]`) seems to be pretty consistent and also avoids the `data-test-id` attribute

qubidt:
Issues have been fixed and added an edge case I missed

Toronto Star.js
```diff
@@ -233,28 +266,76 @@ var testCases = [
...
 					{
-						"title": "Toronto Star Snapshot",
+						"title": "Snapshot",
 						"mimeType": "text/html"
 					}
 				],
-				"tags": [],
+				"tags": [
+					{
+						"tag": "CP_URGENCY_2"
```
AbeJellinek, line 306: Sorry, last suggestion. Could we filter out tags that contain underscores? These clearly aren't meaningful in a Zotero library.

AbeJellinek (changes requested):

Toronto Star.js
qubidt: Good idea! done

AbeJellinek:
Perfect.

PR #2642, opened 2021-08-13
PKP Catalog Systems: grab abstract from header

qubidt:
Using the `DC.Description` meta tag lets us fill out `abstractNote` for some of the test cases that were missing them


AbeJellinek:
Excellent, thanks!

PR #2646, opened 2021-08-15
The Guardian: update translator

qubidt:
The translator's selectors were no longer applicable, updated to use the JSON-LD for article data (particularly the authors). 

I kept the old byline-parsing code (though refactored) because of [this test-case](https://www.theguardian.com/world/2013/mar/05/hugo-chavez-dies-cuba) where one of the authors is not officially linked. This complicates the code and it also means there are likely to be false-positives when there are more complicated bylines.

### DIFF
--- a/The Guardian.js
+++ b/The Guardian.js
@@ -2,14 +2,14 @@
 	"translatorID": "8e11559d-60f0-4a7f-bb91-99ac0c5a2d63",
 	"label": "The Guardian",
 	"creator": "Philipp Zumstein, Bao Trinh",
-	"target": "^https?://(www\\.)?(guardian\\.co\\.uk|theguardian\\.com)",
+	"target": "^https?://(www\\.)?(guardian\\.co\\.uk|theguardian\\.com)/",
 	"minVersion": "3.0",
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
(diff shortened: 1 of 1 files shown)


The Guardian.js
```diff
@@ -82,39 +92,67 @@ function scrape(doc, url) {
...
 			}
 		}
-		item.language = "en-GB";
-		// og:url does not preserve https prefixes, so use canonical link until fixed
-		var canonical = doc.querySelector('link[rel="canonical"]');
-		if (canonical) {
-			item.url = canonical.href;
+
+		let linkedData = JSON.parse(text(doc, 'script[type="application/ld+json"]'));
+		if (linkedData instanceof Array) linkedData = linkedData.find(x => x["@type"] == "NewsArticle");
```
AbeJellinek, line 121: `instanceof Array` [has some issues](https://web.mit.edu/jwalden/www/isArray.html), so it would be best to use `Array.isArray` here.

The Guardian.js
```diff
@@ -82,39 +92,67 @@ function scrape(doc, url) {
...
+		if (linkedData) {
+			if (linkedData.publisher) item.libraryCatalog = linkedData.publisher.name;
```
AbeJellinek, line 123: Will this always just be The Guardian?

AbeJellinek (changes requested):
Looks great, thanks! A couple comments.

PR #2647, opened 2021-08-16
Vice: rewrite translator

qubidt:
Update the vice.com translator to work with the current incarnation of the site.

The multiple-page tests don't pass in scaffold, I assume because the results are pulled dynamically rather than as part of the initial GET. Not sure what to do there.

### DIFF
--- a/Vice.js
+++ b/Vice.js
@@ -1,21 +1,21 @@
 {
 	"translatorID": "131310dc-854c-4629-acad-521319ab9f19",
 	"label": "Vice",
-	"creator": "czar",
-	"target": "^https?://(.+?\\.)?vice\\.com",
+	"creator": "czar, Bao Trinh",
+	"target": "^https?://(.+?\\.)?vice?\\.com/",
 	"minVersion": "3.0",
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2019-06-13 19:09:19"
+	"lastUpdated": "2021-08-15 22:11:31"
 }
 
 /*
 	***** BEGIN LICENSE BLOCK *****
 
-	Copyright © 2018 czar
+	Copyright © 2018-2021 czar, Bao Trinh
 	http://en.wikipedia.org/wiki/User_talk:Czar
 
 	This file is part of Zotero.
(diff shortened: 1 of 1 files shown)


Vice.js
```diff
@@ -148,32 +64,77 @@ function getSearchResults(doc, checkOnly) {
...
+		if (linkedData) {
+			linkedData = JSON.parse(linkedData);
+			if (linkedData && linkedData['@graph']) {
+				linkedData = linkedData['@graph'].find(x => x['@type'] == "NewsArticle");
+				if (linkedData) item.date = linkedData.datePublished;
+			}
+		}
+
+		item.creators.length = 0;
+		for (const author of doc.querySelectorAll('main.main-content .contributors .contributor .contributor__meta a')) {
```
AbeJellinek, line 88: Can we make this a bit shorter to be more resilient to site changes? Like:

```suggestion
		for (const author of doc.querySelectorAll('main .contributor__meta a')) {
```

It's less specific, but it still seems to match fine and it'll survive a reorganization of the site's outer layout elements.

Vice.js
```diff
@@ -148,32 +64,77 @@ function getSearchResults(doc, checkOnly) {
...
+			item.creators.push(ZU.cleanAuthor(author.textContent, 'author'));
+		}
+		for (const author of attr(doc, 'meta[property="article:author"]', 'content').split(",")) {
+			item.creators.push(ZU.cleanAuthor(author, 'author'));
```
AbeJellinek, line 92: What happens if an author is in the `.contributors` and `article:author`?

AbeJellinek (changes requested):
Thanks! A couple comments.

AbeJellinek:
Oh, and you can add `"defer": true` to a test's JSON object to have Scaffold wait for the page to populate dynamically before running the test.

qubidt:
> Oh, and you can add `"defer": true` to a test's JSON object to have Scaffold wait for the page to populate dynamically before running the test.

Brilliant, thank you. I'll test that out

Vice.js
qubidt: Good catch. I hadn't encountered this but it doesn't hurt to be safe. I'll throw them in a Set

Vice.js
qubidt: Good call, it seems to work well, I'll opt for that instead

AbeJellinek: Cool, sounds good.

AbeJellinek:
No problem. It doesn't show in the Testing pane, but it'll work and be preserved between populates/saves fine.

AbeJellinek:
Hey, I'd love to get this merged! Have you gotten a chance to finish it up?

owcz:
Superseded by #2761?

PR #2657, opened 2021-08-17
Fix for Tagesspiegel translator

JLueke:
- fixes abstractNote, section, tags
- updates tests
- removes print action for snapshot

### DIFF
--- a/Tagesspiegel.js
+++ b/Tagesspiegel.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2021-08-17 14:28:25"
+	"lastUpdated": "2021-08-18 12:46:30"
 }
 
 /*
@@ -32,9 +32,9 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 
 function detectWeb(doc, url) {
-	if (ZU.xpathText(doc, "//meta[@property='og:type']/@content")=="article" ){ 
+	if (ZU.xpathText(doc, "//meta[@property='og:type']/@content")=="article" ){
 		return "newspaperArticle";
-	} else if (url.indexOf('/suchergebnis/')>-1){ 
+	} else if (url.indexOf('/suchergebnis/')>-1){
 		return "multiple";
 	} else if (getSearchResults(doc, true)  ) {
 		return "multiple";
@@ -83,10 +83,10 @@ function scrape(doc, url) {
 	newItem.title = ZU.xpathText(doc, "//meta[@property='og:title']/@content");
 	newItem.date = ZU.xpathText(doc, "//time[@itemprop='datePublished']/@datetime");
 	newItem.abstractNote = ZU.xpathText(doc, '//meta[@name="description"]/@content');
-	// Note: doesn't work with subsections
+	// Note: it only grabs the top-level section
 	newItem.section = ZU.xpathText(doc, '//ul[contains(@class, "ts-main-nav-items")]/li[contains(@class, "ts-active")]/a');
 
-	// Authors 
+	// Authors
 	var author  = ZU.xpathText(doc, "//header[contains(@class, 'ts-article-header')]//a[@rel='author']");
 	//Zotero.debug(author);
 	if (author) {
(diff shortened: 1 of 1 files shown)


Tagesspiegel.js
```diff
@@ -82,8 +82,9 @@ function scrape(doc, url) {
 
 	newItem.title = ZU.xpathText(doc, "//meta[@property='og:title']/@content");
 	newItem.date = ZU.xpathText(doc, "//time[@itemprop='datePublished']/@datetime");
-	newItem.abstractNote = ZU.xpathText(doc, ".//p[@class='hcf-teaser']");
-	newItem.section = ZU.xpathText(doc, '//ul[contains(@class, "ts-main-nav-items")]/li[contains(@class, "ts-active-point")]/a');
+	newItem.abstractNote = ZU.xpathText(doc, '//meta[@name="description"]/@content');
+	// Note: doesn't work with subsections
```
AbeJellinek, line 86: So this'll always grab the top-level section? That seems fine.

Tagesspiegel.js
```diff
@@ -97,17 +98,34 @@ function scrape(doc, url) {
...
+	/* We read the tags from the initialisation of variable cmsObject.
+	 * This object and the keywords are defined in a head script tag.
+	 */
+	let tags = [];
+	let scriptItems = doc.querySelectorAll('head > script');
+	if (scriptItems) {
+		for (let i = 0; i < scriptItems.length; i++) {
+			let scriptItemText = scriptItems[i].textContent;
+			// search for script tag that declares the variable
+			if (!scriptItemText.match(/var\scmsObject\s=/)) {
```
AbeJellinek, line 117: Maybe try this:

```suggestion
			if (!scriptItemText.match(/var\s+cmsObject\s*=/)) {
```

So that we'll be OK if the script's formatting changes.

Tagesspiegel.js
```diff
@@ -97,17 +98,34 @@ function scrape(doc, url) {
...
+				continue;
+			}
+			// the pid seems to be added at the end of the keywords, so we remove it
+			let matches = scriptItemText.match(/keywords:\s"(.+?)(\,pid\d+)?"\,/);
```
AbeJellinek, line 121: And same here:

```suggestion
			let matches = scriptItemText.match(/keywords:\s*"(.+?)(\,pid\d+)?"\,/);
```

Tagesspiegel.js
```diff
@@ -154,7 +174,7 @@ var testCases = [
 				"tags": [
 					"ARD",
 					"DDR",
-					"Der Turm"
+					"Der_Turm"
```
AbeJellinek, line 177: Replace underscores with spaces in tags.

Tagesspiegel.js
```diff
@@ -185,29 +205,31 @@ var testCases = [
...
-					"Ehe für alle",
-					"Grosse-Böhmer",
+					"Ehe_fuer_alle",
+					"Grosse_Boehmer",
 					"Queerspiegel",
-					"Renate Künast",
-					"SPD"
+					"Renate_Kuenast",
+					"SPD",
+					"_btw17"
```
AbeJellinek, line 232: And remove tags starting with an underscore.

AbeJellinek (changes requested):
Looks good, thanks! A few suggestions.

Tagesspiegel.js
JLueke: Yes, I'll clarify the comment

Tagesspiegel.js
JLueke: Why not remove the underscore instead? Because it maybe is a SEO tag?
In the original test it was present as #btw17.

AbeJellinek: Yeah, I would guess that it's an SEO tag. #btw17 [is a thing](https://democracy-reporting.org/ua/btw17-social-media-in-the-german-parliamentary-elections/?la), but we already have "Bundestagswahl 2017" for that item.

AbeJellinek:
OK, great. Linted (`npm run lint -- "Tagesspiegel.js" --fix`) to avoid CI complaints; I'll go ahead and merge. Thanks so much for your work on this!

PR #2658, opened 2021-08-17
Fix YouTube translator

JLueke:
- use canonical link
- use the given metatag for date
- fix tags and channel name
- add callNumber

### DIFF
--- a/YouTube.js
+++ b/YouTube.js
@@ -1,7 +1,7 @@
 {
 	"translatorID": "d3b1d34c-f8a1-43bb-9dd6-27aa6403b217",
 	"label": "YouTube",
-	"creator": "Sean Takats, Michael Berkowitz, Matt Burton and Rintze Zelle",
+	"creator": "Sean Takats, Michael Berkowitz, Matt Burton, Rintze Zelle, Jaco Lüken",
 	"target": "^https?://([^/]+\\.)?youtube\\.com/",
 	"minVersion": "3.0",
 	"maxVersion": "",
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsbv",
-	"lastUpdated": "2021-06-03 17:51:31"
+	"lastUpdated": "2021-08-17 15:18:35"
 }
 
 /*
@@ -90,15 +90,14 @@ function doWeb(doc, url) {
 function scrape(doc, url) {
 	var item = new Zotero.Item("videoRecording");
 	item.title = text(doc, '#info-contents h1.title');
-	item.url = url;
+	item.url = ZU.xpathText(doc, '//link[@rel="canonical"]/@href') || url;
 	item.runningTime = text(doc, '#movie_player .ytp-time-duration');
 	
-	item.date = text(doc, '#info-text #date');
+	item.date = ZU.xpathText(doc, '//meta[@itemprop="datePublished"]/@content');
 	if (item.date) {
 		item.date = ZU.strToISO(item.date);
 	}
-	var author = text(doc, '#meta-contents #text-container .ytd-channel-name')
-		|| text(doc, '#text-container .ytd-channel-name');
+	var author = text(doc, '#text-container.ytd-channel-name > yt-formatted-string > a');
 	if (author) {
 		item.creators.push({
 			lastName: author,
@@ -109,6 +108,18 @@ function scrape(doc, url) {
 	var description = text(doc, '#description .content') || text(doc, '#description');
 	if (description) {
 		item.abstractNote = description;
+	}
+	let tags = ZU.xpathText(doc, '//meta[@name="keywords"]/@content');
+	if (tags) {
+		tags.split(',').forEach(function (tag) {
+			item.tags.push(tag.trim());
+		});
+	}
+	
+	// add video key as signatur
+	let signaturMatches = item.url.match(/.+v=(.+)$/);
+	if (signaturMatches) {
+		item.callNumber = signaturMatches[1];
 	}
 	
 	item.complete();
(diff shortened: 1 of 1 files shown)


YouTube.js
```diff
@@ -136,8 +147,26 @@ var testCases = [
...
+					"History",
+					"Mason",
+					"Media",
+					"Mozilia",
+					"New",
+					"Research",
+					"University",
+					"Web",
+					"Zotero",
+					"and",
```
AbeJellinek, line 165: These are a little bit junk, aren't they? My impression is that YT tags are mostly there for SEO, and we usually try to exclude SEO-focused metadata from being imported.

YouTube.js
```diff
@@ -110,6 +109,18 @@ function scrape(doc, url) {
...
 	}
+	let tags = ZU.xpathText(doc, '//meta[@name="keywords"]/@content');
+	if (tags) {
+		tags.split(',').forEach(function (tag) {
+			item.tags.push(tag.trim());
+		});
+	}
+	
+	// add video key as signatur
+	let signaturMatches = item.url.match(/.+v=(.+)$/);
```
AbeJellinek, line 120: I don't really think we need/want this. It's already in the URL and YouTube isn't actually a library catalog. I don't know of a citation style that would want something like `Zoteron. "Zotero Intro" (2007). YouTube, pq94aBrc0pY.` No one would know what that means!

YouTube.js
```diff
@@ -90,15 +90,14 @@ function doWeb(doc, url) {
 function scrape(doc, url) {
 	var item = new Zotero.Item("videoRecording");
 	item.title = text(doc, '#info-contents h1.title');
-	item.url = url;
+	item.url = ZU.xpathText(doc, '//link[@rel="canonical"]/@href') || url;
```
AbeJellinek, line 93: Prefer attr/text:

```suggestion
	item.url = attr(doc, 'link[rel="canonical"]', 'href') || url;
```

YouTube.js
```diff
@@ -90,15 +90,14 @@ function doWeb(doc, url) {
...
 	item.runningTime = text(doc, '#movie_player .ytp-time-duration');
 	
-	item.date = text(doc, '#info-text #date');
+	item.date = ZU.xpathText(doc, '//meta[@itemprop="datePublished"]/@content');
```
AbeJellinek, line 96: And here:

```suggestion
	item.date = attr(doc, 'meta[itemprop="datePublished"]', 'content');
```

CSS selectors are more widely understood than XPath and lack its odd edge cases (namespaces in XHTML/XML documents...).

YouTube.js
```diff
@@ -90,15 +90,14 @@ function doWeb(doc, url) {
...
 	if (item.date) {
 		item.date = ZU.strToISO(item.date);
 	}
-	var author = text(doc, '#meta-contents #text-container .ytd-channel-name')
-		|| text(doc, '#text-container .ytd-channel-name');
+	var author = text(doc, '#text-container.ytd-channel-name > yt-formatted-string > a');
```
AbeJellinek, line 100: If we're matching on an ID, we don't need to match on a class, too, do we? Will there ever ben a `#text-container` without the `.ytd-channel-name` class that we want to exclude?

Also, is this broken right now? It's working fine on my end, and the `#meta-contents` tweak was to prevent the wrong creator name from being scraped, so it should stay unless it's causing issues.

AbeJellinek (changes requested):
Thanks! The first two tweaks look good. I would argue against the rest - see reasoning below.

YouTube.js
JLueke: Depends on the video. I saw other entries with only a handful of tags, but only updated the already existing test case.

So should I remove the part that fetches the tags?

AbeJellinek: Yeah, I think it should be removed.

YouTube.js
JLueke: I'll revert it.
It's working, I guess confused it with another error and didn't notice that the nested element also has the "ytd-channel-name" class while testing.

JLueke:
ready for review

YouTube.js
```diff
@@ -90,10 +90,10 @@ function doWeb(doc, url) {
 function scrape(doc, url) {
 	var item = new Zotero.Item("videoRecording");
 	item.title = text(doc, '#info-contents h1.title');
-	item.url = url;
+	item.url = attr(doc, 'link[rel="canonical"]', 'href') || url;
 	item.runningTime = text(doc, '#movie_player .ytp-time-duration');
-	
-	item.date = text(doc, '#info-text #date');
+
+	item.date = attr(doc, 'meta[itemprop="datePublished"]', 'content');
```
AbeJellinek, line 96: Oh God. I remember now why we don't use the meta tags for YouTube. Try going to a video, clicking on a link to a related video (so that it opens in the same tab), then opening browser developer tools and looking at the meta/link tags. They don't change! Not even `<link rel="canonical">`. They reflect the metadata of the original video, not the new one you opened. So we just can't use them.

Could we stick with screen scraping but just update the selector? I think it should be:

```js
item.date = ZU.strToISO(text(doc, '#info-strings yt-formatted-string'));
```

AbeJellinek (changes requested):

YouTube.js
```diff
@@ -89,14 +89,18 @@ function doWeb(doc, url) {
...
+
+	/* YouTube won't update the meta tags for the user,
+	 * if they open e.g. a suggested video in the same tab.
+	 * Thus we scrape them from screen instead.
+	 */
+
 	item.title = text(doc, '#info-contents h1.title');
-	item.url = url;
+	// try to scrape only the canonical url, excluding additional query parameters
+	item.url = url.replace(/^(.+\/watch\?\bv=[0-9a-zA-Z_-]+).*/, "$1");
```
AbeJellinek, line 100: Why the `\b`? And won't this not work on valid URLs like https://www.youtube.com/watch?app=desktop&v=HfIymdQ2LWc?

AbeJellinek (changes requested):

PR #2663, opened 2021-08-18
add translator for National Gallery of Australia

benswift:
Hey all, first time contributor so if there's something I've done wrong just let me know and I'll update it.

this translator works for the NGA's searchthecollection.nga.gov.au subdomain

this translator currently only supports adding single artworks, however the
website doesn't actually have a notion of an "artwork" page; the relevant info
only shows up when there's a search query with one result

therefore, this translator

- looks for `uniqueID=` in the query params, as a proxy for the "one artwork on page" criteria

- uses CSS selectors to scrape the relevant metadata

- creates a Zotero "artwork" item

In the future, it'd be good to support creating multiple artwork items at
once (if there's more than one search result) but the current implementation of
the collection search website only displays the relevant metadata in the "one
search result" case, so we'd have to come up with some way to deal with that
issue.


National Gallery of Australia.js
```diff
@@ -0,0 +1,98 @@
...
+function parseArtwork(doc, url){
+
+	var selectorMap = {
+		artist: ".main-artist-name > span:nth-child(1) > a:nth-child(1)",
+		title: ".object-info-big-title",
+		date: ".object-info-big-date > a:nth-child(1)",
+		artworkMedium: ".object-info-left-col > div:nth-child(3) > div:nth-child(2)",
+		artworkSize: ".object-info-left-col > div:nth-child(4) > div:nth-child(2) > div:nth-child(1)",
+		callNumber: ".object-nav-vra-items > div:nth-child(1) > span:nth-child(2)", // Internal Reference Number (IRN)
+		libraryCatalog: ".object-nav-vra-items > div:nth-child(2) > span:nth-child(2)", // Accession number
```
AbeJellinek, line 55: This definitely shouldn't go into `libraryCatalog`, which is meant to hold things like "National Gallery of Australia." Which is more relevant for locating an item from a citation: the IRN or the accession number? Whichever it is, that one should go into `callNumber` and the other should go into `extra` (`National Gallery of Australia IRN: XYZ123`) or nowhere at all.

National Gallery of Australia.js
```diff
@@ -0,0 +1,98 @@
...
+		return "artwork";
+	}
+
+	return false;
+}
+
+function parseArtwork(doc, url){
+
+	var selectorMap = {
+		artist: ".main-artist-name > span:nth-child(1) > a:nth-child(1)",
```
AbeJellinek, line 49: Try to avoid `:nth-child` selectors when possible. They're almost certain to break quickly or already be broken on pages that you haven't tested on. I see that the site doesn't have classes that can easily be matched on, but it _is_ populated via JSON. Why not load the data the same way the site does? Make a GET request:

```js
let uniqueID = ...; // extract from the URL
let jsonURL = `https://searchthecollection.nga.gov.au/stcapi/service/stc/node?uniqueId=${uniqueID}`;
ZU.doGet(jsonURL, function (respText) {
	let json = JSON.parse(respText);

	let title = json.vraTitleDisplayText;

	// basic creator info:
	let creators = json.creators; // ["Alison Alder", "Redback Graphix", "NSW Women's Advisory Service", "Printcraft Pty Ltd"]

	// or, maybe better:
	let creators = JSON.parse(json.creatorsJson); // an array of objects, as follows:
	for (let creator of creators) {
		let name = creator.NamFullName; // "Alison Alder"
		let type = creator.NamPartyType; // "Person" or "Organisation"
		if (type == "Person") {
			// if we get a person, we should treat their name like
			// a personal name and attempt to parse it
			item.creators.push(ZU.cleanAuthor(name, 'artist'));
		}
		else {
			// if we get an organization, we should pass its name
			// through as-is
			item.creators.push({
				lastName: name,
				creatorType: 'contributor', // organizations probably aren't artists?
				fieldMode: 1
			});
		}
	}

	// do more processing...
});
```

This way we can avoid doing messy, fragile screen scraping altogether. I hope this helps!

National Gallery of Australia.js
```diff
@@ -0,0 +1,98 @@
...
+
+	// finally, construct a URL
+	newArtwork.url = `https://searchthecollection.nga.gov.au/object?uniqueId=${newArtwork.libraryCatalog}`;
+	newArtwork.complete();
+
+	return newArtwork;
+}
+
+function doWeb(doc, url) {
+	if (detectWeb(doc, url) == "artwork") {
```
AbeJellinek, line 95: `detectWeb` will only ever return `artwork` or false, and `doWeb` won't be called if it returns false, so we don't need this check.

National Gallery of Australia.js
```diff
@@ -0,0 +1,98 @@
...
+		parseArtwork(doc, url);
+	}
+}
```
AbeJellinek, line 98: Add tests using Scaffold: navigate to a page in the Browser tab, go to the Testing tab, and click New Web and then Save.

AbeJellinek (changes requested):
Thanks, this is really good work! We can avoid screen-scraping and dealing with nasty selectors for this one, though. I left some comments with more details. Let me know what you think!

National Gallery of Australia.js
AbeJellinek: (Examples using https://searchthecollection.nga.gov.au/object?uniqueId=4193).

Another advantage of using the JSON API is that we can handle multiples - don't even need to load the front-facing HTML for each one, just pull the `uniqueId` from each link and grab the JSON.

benswift:
Hey @AbeJellinek thanks for the detailed code review - very helpful for a new-to-zotero contributor like me.

Nice idea about the JSON thing - in hindsight I should have looked more closely about how the page worked to do something like that. Will implement the suggestions you made, and hard agree that it's nicer than the incredibly fragile screen scraping (I had a comment in the code in my feature branch expressing my distaste at the fragility of it all, but I thought it was best to remove that comment from the PR 😉 ).

Re: testing, will add some tests. I had been using scaffold for testing, but only for eyeballing the results of the `doWeb()` call---I didn't know to click save and add them as proper tests. Will do so and re-submit.

One question: I guess you want me to re-base my changes into a single commit? I'll do that (and force-push to this repo) unless you'd prefer otherwise.

I'll make the changes and re-submit the PR asap.

benswift:
hey @AbeJellinek I've made some changes along the lines you suggested - much more robust now (thanks). I'm still not 100% sure on zotero conventions (either in terms of coding conventions or the way the stored in the Zotero item) so if you'd like any more changes then just let me know.

One other question: is there an easy way to share a "local" translator; i.e. can I send the js file to colleagues and let them sideload it so they can use it on the NGA's collection straight away?

benswift:
Hey @AbeJellinek just wanted to see if you'd had a chance to see my updated version in response to your review? Happy to make any more changes if you'd like them.

AbeJellinek:
Sorry, I was off Thursday-Tuesday. Will take a look now.

AbeJellinek:
>One other question: is there an easy way to share a "local" translator; i.e. can I send the js file to colleagues and let them sideload it so they can use it on the NGA's collection straight away?

Yes! Have them copy it into `~/Zotero/translators` (or equivalent for their platform; Preferences -> Advanced -> Files and Folders -> Show Data Directory) and restart Zotero.

PR #2666, opened 2021-08-20
Fix translator for Baruch Foundation artwork

cano-endava:
The target was left empty, so it was being taken as a generic translator


dstillman:
Yikes! Thanks! Merged in 7a6b54f1. (Note that it's necessary to update the timestamp for the change to go out to Zotero users.)

AbeJellinek:
Thank you, sorry! I don't know how I overlooked that.

PR #2690, opened 2021-09-06
Update CNKI translator

l0o0:
Sorry for the late.

Updates:
 1. expand URL matching for proxied URLs
 2. Zotero Connector parameters for Chinese names and CAJ/PDF attachment downloading
 3. Get citations and CSSCI

### DIFF
--- a/CNKI.js
+++ b/CNKI.js
@@ -2,45 +2,40 @@
 	"translatorID": "5c95b67b-41c5-4f55-b71a-48d5d7183063",
 	"label": "CNKI",
 	"creator": "Aurimas Vinckevicius, Xingzhong Lin",
-	"target": "^https?://([^/]+\\.)?cnki\\.net",
+	"target": "https?://.*?/(kns8?/defaultresult/index|kns8?/AdvSearch|kcms/detail|KXReader/Detail\\?|KNavi/)",
 	"minVersion": "3.0",
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcs",
-	"lastUpdated": "2019-12-05 08:10:19"
+	"lastUpdated": "2021-03-17 10:39:20"
 }
 
 /*
 	***** BEGIN LICENSE BLOCK *****
-
 	CNKI(China National Knowledge Infrastructure) Translator
 	Copyright © 2013 Aurimas Vinckevicius
-
 	This file is part of Zotero.
-
 	Zotero is free software: you can redistribute it and/or modify
 	it under the terms of the GNU Affero General Public License as published by
 	the Free Software Foundation, either version 3 of the License, or
 	(at your option) any later version.
-
 	Zotero is distributed in the hope that it will be useful,
 	but WITHOUT ANY WARRANTY; without even the implied warranty of
 	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
 	GNU Affero General Public License for more details.
-
 	You should have received a copy of the GNU Affero General Public License
 	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
-
 	***** END LICENSE BLOCK *****
 */
 
+// Target regex for default search, advance search, detail page and journal articles pages.
 // Fetches RefWorks records for provided IDs and calls onDataAvailable with resulting text
 // ids should be in the form [{dbname: "CDFDLAST2013", filename: "1013102302.nh"}]
 function getRefWorksByID(ids, onDataAvailable) {
 	if (!ids.length) return;
-	var { dbname, filename } = ids.shift();
+	var { dbname, filename, url } = ids.shift();
 	var postData = "formfilenames=" + encodeURIComponent(dbname + "!" + filename + "!1!0,")
 		+ '&hid_kLogin_headerUrl=/KLogin/Request/GetKHeader.ashx%3Fcallback%3D%3F'
 		+ '&hid_KLogin_FooterUrl=/KLogin/Request/GetKHeader.ashx%3Fcallback%3D%3F'
@@ -50,17 +45,24 @@ function getRefWorksByID(ids, onDataAvailable) {
 			var parser = new DOMParser();
 			var html = parser.parseFromString(text, "text/html");
 			var data = ZU.xpath(html, "//table[@class='mainTable']//td")[0].innerHTML
-				.replace(/<br>/g, '\n')
+				.replace(/\n/g, '<br>')
+				.replace(/<br>\s+/g, ';')
+				.replace(/(<br>)+/g, '\n')
+				.replace(/^RT\s+Conference Proceeding/gmi, 'RT Conference Proceedings')
 				.replace(/^RT\s+Dissertation\/Thesis/gmi, 'RT Dissertation')
+				.replace(/;;/g, ';') // 保留作者中一个英文分号
+				.replace(/^ AB/g, 'AB') // 去除AB前空格
+				.replace(/vo (\d+)\n/, "VO $1\n")  // 修改vo 大小写
 				.replace(
 					/^(A[1-4]|U2)\s*([^\r\n]+)/gm,
 					function (m, tag, authors) {
 						authors = authors.split(/\s*[;，,]\s*/); // that's a special comma
 						if (!authors[authors.length - 1].trim()) authors.pop();
-						return tag + ' ' + authors.join('\n' + tag + ' ');
+						return 'A1' + ' ' + authors.join('\n' + 'A1' + ' ');  // Use A1 tag instead
 					}
 				);
-			onDataAvailable(data);
+			Z.debug(data);
+			onDataAvailable(data, url);
 			// If more results, keep going
 			if (ids.length) {
 				getRefWorksByID(ids, onDataAvailable);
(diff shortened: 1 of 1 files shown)


dstillman:
> expand URL matching for proxied URLs

No, we don't do this — the Connector handles proxy detection, not individual translators. EZproxy installations should be automatically detected, and others can be configured manually in the Zotero Connector preferences (though not all proxy schemes are supported).

You'll need revert the target change so that it's limited to a specific domain before we can accept this.

l0o0:
> 
> 
> > expand URL matching for proxied URLs
> 
> No, we don't do this — the Connector handles proxy detection, not individual translators. EZproxy installations should be automatically detected, and others can be configured manually in the Zotero Connector preferences (though not all proxy schemes are supported).
> 
> You'll need to revert the target change before we can accept this. All site-specific translators should have a target limited to a specific domain.

OK, I will revert to the origin URL patent, also I will leave a comment if someone wants to expand the URL matching.

AbeJellinek:
>I will leave a comment if someone wants to expand the URL matching

What would it be expanded to? Are there domains that the original target should've matched but didn't?

AbeJellinek:
Oh, I see - if I understand correctly, the domain change was just for proxied URLs? In that case, I don't think a comment would be appropriate. _Nobody_ should have to expand the target to match a proxy. If there's a proxy scheme that Zotero doesn't support, that's a separate issue and we should add support for it.

l0o0:
> Oh, I see - if I understand correctly, the domain change was just for proxied URLs? In that case, I don't think a comment would be appropriate. _Nobody_ should have to expand the target to match a proxy. If there's a proxy scheme that Zotero doesn't support, that's a separate issue and we should add support for it.

Hi @AbeJellinek  As I know, most colleges in China use EasyConnect for the proxy. Adding proxy in Zotero Connector can work in some cases. But some school uses dynamic proxy URL schema, it's hard to configure in Zotero Connector. Maybe expand the URL matching is the easiest way to work.

dstillman:
The problem is that the Connector's proxy support doesn't just determine whether to save — it also determines how the URL is saved. So if it doesn't detect it properly as a proxy, a proxied URL may be saved to the URL field (depending on the site), which wouldn't be appropriate for citation (or even future access) purposes.

AbeJellinek:
@l0o0: Could you give an example of a URL proxied by EasyConnect?

l0o0:
> 
> 
> @l0o0: Could you give an example of a URL proxied by EasyConnect?

These are proxied by URLs (from different colleges):

http://hfgga60aabc7d15084b00sqncnfnbuqv906uqp.fhaz.libproxy.ruc.edu.cn/kns8/defaultresult/index
https://kns-cnki-net-s.vpn2.njau.edu.cn:8118/kcms/detail/detail.aspx?dbcode=CAPJ&dbname=CAPJDAY&filename=SPFX20210129001&v=%25mmd2Ff6t9sWdgG6X7m3%25mmd2FghP9ZI7zt%25mmd2Fchv0wCAVbQV7bAYNIfz45hIrJ%25mmd2FFK3HtcKcdU3N
https://elksslcc0eb1c56d2d940cf2d0186445b0c858.casb.njucm.edu.cn/kns8/defaultresult/index
http://kns-cnki-net-443--buaa.buaa.ilibs.cn/kns8/defaultresult/index
https://kns-cnki-net-s.vpn.buaa.edu.cn:8118/kns8/defaultresult/index
http://kns-cnki-net-s.vpn.cau.edu.cn:8118/kns8/defaultresult/index
http://f3442d2d91cfa2485dde859ea753e903.3be401a9.libvpn.zuel.edu.cn/kns8/defaultresult/index
https://er.szlib.org.cn/rwt/331/https/NNYHGLUDN3WXTLUPMW4A/kns8/defaultresult/index
http://kns-cnki-net-s.vpn.ysu.edu.cn:8118/kcms/detail/detail.aspx?dbcode=CAPJ&dbname=CAPJDAY&filename=SHZN20210129001&v=UDLwfYA%25mmd2BHEE6A%25mmd2BqqSQyTOMdgNS%25mmd2FFL5DAA6GcomfOPuJdl4Qzpt7ZqbEl2SIzOVHQ
https://kns-cnki-net-s.webvpn.scut.edu.cn/kns8/defaultresult/index
https://kns-cnki-net-s.vpn3.fjnu.edu.cn/kns8/defaultresult/index
https://v.guet.edu.cn/https/77726476706e69737468656265737421fbf952d2243e635930068cb8/kns8/defaultresult/index
https://wvpn.upc.edu.cn/https/77726476706e69737468656265737421e7e056d2243e635930068cb8/?wrdrecordvisit=1612172555000
http://210.40.130.91/https/77726476706e69737468656265737421f3ff4fd228266d426d0d88e29b5b283c13d3673f/kns/defaultresult/index

Sometimes they use a dynamic prefix in the proxy schema, the prefix string of proxy URL will change when login in.

https://kns.cnki.net/kns8/defaultresult/index is CNKI search URL
https://kns.cnki.net/kcms/detail/detail.aspx? is a article page URL

l0o0:
@AbeJellinek  Hi, I've updated the URLs target and fixed the ESlint errors. But I don't know how to fix the left 4 errors.

CNKI.js
```diff
@@ -50,17 +49,24 @@ function getRefWorksByID(ids, onDataAvailable) {
...
+				.replace(/;;/g, ';') // 保留作者中一个英文分号
+				.replace(/^ AB/g, 'AB') // 去除AB前空格
+				.replace(/vo (\d+)\n/, "VO $1\n") // 修改vo 大小写
 				.replace(
 					/^(A[1-4]|U2)\s*([^\r\n]+)/gm,
 					function (m, tag, authors) {
 						authors = authors.split(/\s*[;，,]\s*/); // that's a special comma
 						if (!authors[authors.length - 1].trim()) authors.pop();
-						return tag + ' ' + authors.join('\n' + tag + ' ');
+						return 'A1' + ' ' + authors.join('\n' + 'A1' + ' '); // Use A1 tag instead
```
Lemmingh, line 65: What about

```javascript
const lines = [];

// Author names may be separated by Semicolon (U+003B), Comma (U+002C), or Full-width Comma (U+FF0C).
for (let item of authors.split(/[;，,]/)) {
    item = item.trim(); // Skip empty entries.
    if (item) {
        lines.push("A1 " + item); // Always use the `A1` tag.
    }
}

return lines.join("\n");
```

or

```javascript
return authors
    .split(/[;，,]/) // Author names may be separated by Semicolon (U+003B), Comma (U+002C), or Full-width Comma (U+FF0C).
    .map(String.prototype.trim)
    .filter(Boolean) // Remove empty entries.
    .map((s) => "A1 " + s) // Always use the `A1` tag.
    .join("\n");
```

Lemmingh (changes requested):
IMHO, there are a lot more you can improve, such as normalizing CNKI URLs, migrating from RefWorks to CNKI's E-Study, and adopting modern JavaScript. Of course, they should be done in follow-up work, not this PR.

CNKI.js
```diff
@@ -176,148 +214,173 @@ function detectWeb(doc, url) {
...
+				}
+				var moreClick = ZU.xpath(doc, "//span/a[@id='ChDivSummaryMore']");
+				if (moreClick.length) {
+					moreClick[0].click();// click to get a full abstract in a single article page
+					newItem.abstractNote = ZU.xpath(doc, "//span[@id='ChDivSummary']")[0].innerText;
+				}
+			}
+			var timestamp = new Date().toLocaleDateString().replace(/\//g, '-');
+			citeStr = cite ? `${cite} citations(CNKI)[${timestamp}]` : "";
+			newItem.extra = (citeStr + pubTypeStr).trim();
```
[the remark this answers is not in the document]
l0o0: Modifying other item fields would update modified dates, so citation following a date would make it clear. I will update codes to match the extra field format.

CNKI.js
```diff
@@ -144,6 +143,7 @@ function getItemsFromSearchResults(doc, url, itemInfo) {
 		fileXpath = "./ul/li/a";
 	}
 	else { // for search result page
+		var result = doc.querySelector('table.result-table-list');
```
AbeJellinek, line 146: Why was this added but not used?

CNKI.js
```diff
@@ -242,6 +241,7 @@ function scrape(ids, doc, itemInfo) {
 		translator.setHandler('itemDone', function (obj, newItem) {
 			// add PDF/CAJ attachments
 			var cite, citeStr, pubType, pubTypeStr = "";
```
AbeJellinek, line 243: Unused variables here should also be removed.

CNKI.js
```diff
@@ -348,35 +345,33 @@ function getPDF(doc, itemType) {
...
+		url = url.replace(/kns\.cnki\.net\/KNS8/, "oversea.cnki.net/kns");
 		attachments.push({
 			title: "Full Text PDF",
 			mimeType: "application/pdf",
-			url: pdfurl ? pdfurl : cajurl.includes("&dflag=nhdown") ? cajurl.replace('&dflag=nhdown', '&dflag=pdfdown') : cajurl + '&dflag=pdfdown'
+			url: url
 		});
-	}
-	else if (cajurl) {
+	} else if (cajurl) {
```
AbeJellinek, line 363: This introduced a code style regression and should be reverted.

AbeJellinek (changes requested):
Thanks, just a few more comments.

l0o0:
Hi @AbeJellinek , someone suggests me to add citation in extra field in format like "24 CNKI". It's easy to find citation number and to sort item by extra file(sort by value of citation number). Is it OK to store value in this format instead of key value pairs ?

skymagician:
这开发者真是撩完就跑啊，没后续了。

l0o0:
> 这开发者真是撩完就跑啊，没后续了。

I will create a new pull request in a few days. CNKI has update attachments download links,now you can download thesis PDF instead of caj.

skymagician:
> > 这开发者真是撩完就跑啊，没后续了。
> 
> I will create a new pull request in a few days. CNKI has update attachments download links,now you can download thesis PDF instead of caj.

哈哈，你误会了，我说的不是你，是这个官方维护人员。之前主动找你说希望合并到官方库，提了合并请求之后就没后续了。

PR #2709, opened 2021-09-20
Remove duplicate doWeb() function in DBLP Comp Sci

mjpelmear:
Seems like a duplicate doWeb() function was added in https://github.com/zotero/translators/pull/2616


AbeJellinek:
Oops, thanks for catching this! Added a commit to bump `lastUpdated` so that this fix goes out to clients.

PR #2720, opened 2021-10-09
Cambridge Core: Fix dates for First View

adam3smith:
reported: 
https://forums.zotero.org/discussion/comment/388528/#Comment_388528

Current example: 
https://www.cambridge.org/core/journals/journal-of-ecclesiastical-history/article/abs/simony-crisis-of-the-eleventh-century-and-the-letter-of-guido/C915507B2324FC0A8EF03B99475202EA#
Not adding tests because the underlying issue disappears as soon as the 
article is included in a published issed

### DIFF
--- a/Cambridge Core.js
+++ b/Cambridge Core.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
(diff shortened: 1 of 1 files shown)


AbeJellinek:
Excellent, thanks.

PR #2721, opened 2021-10-10
ProQuest: Fix dates for EEBO

adam3smith:
Add publisher.
https://forums.zotero.org/discussion/comment/388528/#Comment_388528


adam3smith:
Example is here: https://www.proquest.com/eebo/docview/2240875806/citation/2A8FEA5AB2944BA9PQ/120?accountid=14214 I wasn't able to create a test in Scaffold because Proquest does some weird re-directing that Scaffold doesn't handle

AbeJellinek:
Great, thanks!

PR #2742, opened 2021-11-06
ScienceDirect: Remove proxy: false from attachments

### LINKED ISSUE #2739: No pdf from ScienceDirect
Both without institution recirect with open access and with redirect never any pdf from Clinical Psychology Review.
Tried to update and reset translators without any effect.

AbeJellinek:
This was breaking relative URL resolution and doesn't seem to be necessary to pull PDFs when behind a proxy.

Fixes #2739.

### DIFF
--- a/ScienceDirect.js
+++ b/ScienceDirect.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2021-07-20 00:44:51"
+	"lastUpdated": "2021-11-06 23:29:33"
 }
 
 function detectWeb(doc, url) {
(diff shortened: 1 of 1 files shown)


dstillman:
Can you say more about what this was breaking?

We added this in 4f86533c541cc28f1323b0e210b082e3d4c357eb, and it was [necessary at the time](https://forums.zotero.org/discussion/comment/328880/#Comment_328880) for some users. The appropriateness of `proxy: false` is really less about individual proxies and more about the method of extraction — if a URL is extracted directly from the page, [it should be used](https://www.zotero.org/support/dev/translators/coding#attachments), because the URL will already be proxied or not as appropriate. We have, however, changed the ScienceDirect translator's since then, so it's possible it's not necessary for some or all of the URL extraction methods. It's possible we need to use it conditionally for certain methods.

dstillman:
Not a fix: https://github.com/zotero/translators/issues/2745#issuecomment-966547599

To continue in #2739

PR #2743, opened 2021-11-09
new translator for the PRC History Review

boan-anbo:
http://prchistory.org/the-prc-history-review/

The site uses WordPress with direct file links without a dedicated child page for individual articles in most cases, so I use less structured ways to extract the information.

It supports downloading articles on the home page of the Review (http://prchistory.org/the-prc-history-review/). However, since this home page has entries that are ever-growing, I did not include it in the test.

### DIFF
--- a/PRC History Review.js
+++ b/PRC History Review.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2021-11-09 12:23:22"
+	"lastUpdated": "2021-11-21 04:54:57"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


PRC History Review.js
```diff
@@ -0,0 +1,297 @@
...
+						authorTexts.forEach((authorText) => {
+							const authorName = authorText.split(',')[0];
+							if (authorName) {
+								newItem.creators.push(ZU.cleanAuthor(authorName, 'author', false));
+							}
+						});
+					}
+				}
+
+				if (isSingleArticle) {
```
AbeJellinek, line 233: Can these `if`s be condensed, or could this single `scrape` function be split into multiple? The logic here is difficult to follow and the calls to `scrape` with two unlabeled boolean arguments are a bit opaque.

PRC History Review.js
```diff
@@ -0,0 +1,297 @@
...
+							if (authorName) {
+								newItem.creators.push(ZU.cleanAuthor(authorName, 'author', false));
+							}
+						});
+					}
+				}
+
+				// Download pdf
+				const pdfUrl = url;
+				if (pdfUrl && pdfUrl.toLowerCase().endsWith('.pdf')) {
```
AbeJellinek, line 248: Use `checkIfLinkIsPdf`?

PRC History Review.js
```diff
@@ -0,0 +1,297 @@
...
+					});
+				}
+				newItem.complete();
+			}
+		}
+	});
+}
+
+// helper functions.
+function checkIfLinkIsPdf(url) {
```
AbeJellinek, line 261: Maybe better to rename this and the following method `isPdfUrl` and `isHomePageUrl` to follow convention.

PRC History Review.js
boan-anbo: Thanks! I tried but since every Zotero field needs to change according to these variables, it might be easier doing with one function. I did remove the unlabled variables and simplifyy the logics.

PRC History Review.js
boan-anbo: Done

PRC History Review.js
boan-anbo: Done

AbeJellinek:
Thank you! Sorry this took a while to get back to.

PR #2746, opened 2021-11-13
translator for Swiss Federal Court

KampfCaspar:
First translator pull request - please bear with me

### DIFF
--- a/Schweizerisches Bundesgericht.js
+++ b/Schweizerisches Bundesgericht.js
@@ -1,0 +1,218 @@
+{
+	"translatorID": "9e898223-938c-4c8f-8b5c-d118b6e44aae",
+	"label": "Schweizerisches Bundesgericht",
+	"creator": "Hans-Peter Oeri",
+	"target": "^https://www\\.bger\\.ch/ext/eurospider/live/(de|fr|it)/php/(clir|aza)/",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2021-11-13 20:19:13"
+}
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2020 Hans-Peter Oeri
+
+	This file is part of Zotero.
+
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+/**
+ * read query parameters from url
+ *
+ * @param   {String} url
+ * @returns {{}}
+ */
+function parseURLParameters(url) {
+	let parts = require('url').parse(url, true);
+	if (parts.pathname === "/ext/eurospider/live/de/php/aza/http/index.php") {
+		parts.query._collection = "BGer";
+	}
+	else if (parts.pathname === "/ext/eurospider/live/de/php/clir/http/index.php") {
+		parts.query._collection = "BGE";
+	}
+	else if (parts.pathname === "/ext/eurospider/live/de/php/aza/http/index_aza.php") {
+		parts.query._collection = "news";
+	}
+	return parts.query;
+}
+
+/**
+ * internal detectWeb interpreting prepared query parameters
+ *
+ * @param   {{}} parameters
+ * @returns {string|boolean}
+ * @private
+ */
+function _detectWeb(parameters) {
+	switch (parameters.type) {
+		case "simple_query":
+		case "simple_similar_documents":
+			return "multiple";
+
+		case "show_document":
+		case "highlight_simple_query":
+		case "highlight_simple_similar_documents":
+			return "case";
+	}
+	if (parameters.mode === "news") {
+		return "multiple";
+	}
+	return false;
+}
+
+/**
+ * collect items from a new rulings page
+ *
+ * Those pages are found at:
+ * /ext/eurospider/live/de/php/aza/http/index_aza.php?...mode=news
+ *
+ * @param {Document} doc
+ * @param {{}}       parameters
+ * @returns {{}}
+ */
+function getSearchResultsNews(doc, parameters) {
+	let domRows = doc.querySelectorAll("div.eit>table:nth-of-type(2)>tbody>tr");
+	let items = {};
+
+	for (let i = 1; i < domRows.length; i += 2) {
+		let date = domRows[i].children[1].innerText;
+		let url  = domRows[i].children[2].children[0];
+		let number = domRows[i].children[2].children[0].innerText;
+		let subject = domRows[i].children[4].innerText;
+		items[url] = number + " (" + date + "; " + subject + ")";
+	}
+
+	return items;
+}
+
+/**
+ * collect items from a search result page of all rulings
+ *
+ * Those pages are found at:
+ * /ext/eurospider/live/de/php/aza/http/index.php?l...type=simple_query|simple_similar_documents
+ *
+ * @param {Document} doc
+ * @param {{}}       parameters
+ * @returns {{}}
+ */
+function getSearchResultsBGer(doc, parameters) {
+	let domRows = doc.querySelectorAll("div.ranklist_content>ol>li");
+	let items = {};
+
+	for (let i = 1; i < domRows.length; i++ ) {
+		let meta = domRows[i].children[0].children[0].innerText;
+		let date = meta.substring(0, meta.indexOf(' '));
+		let number = meta.substring(date.length + 1);
+		number = number.substring(0, 2) + "_" + number.substring(3);
+		let url  = domRows[i].children[0].children[0];
+		let subject = domRows[i].children[2].children[1].innerText;
+		let subject2 = domRows[i].children[2].children[2].innerText;
+		items[url] = number + " (" + date + "; " + subject + ": " + subject2 + ")";
+	}
+	return items;
+}
+
+/**
+ * collect items from a search result page of formally published rulings
+ *
+ * Those pages are found at:
+ * /ext/eurospider/live/de/php/clir/http/index.php?...type=highlight_simple_query|highlight_simple_similar_documents
+ *
+ * @param {Document} doc
+ * @param {{}}       parameters
+ * @returns {{}}
+ */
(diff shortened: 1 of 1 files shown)


Schweizerisches Bundesgericht.js
```diff
@@ -0,0 +1,243 @@
...
+}
+
+/**
+ * read query parameters from url
+ *
+ * @param   {String} url
+ * @returns {{}}
+ */
+function parseURLParameters(url) {
+	let parts = require('url').parse(url, true);
```
AbeJellinek, line 63: We don't want to use `require` here, or anywhere in a translator - I am genuinely surprised that it's available in the sandbox at all! Try [`new URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL).

Schweizerisches Bundesgericht.js
```diff
@@ -0,0 +1,243 @@
...
+	parts.query._lang = parts.pathname.substring(21, 23);
```
AbeJellinek, line 64: Can this be a regex? Feels unstable.

Schweizerisches Bundesgericht.js
```diff
@@ -0,0 +1,243 @@
...
+ *
+ * Those pages are found at:
+ * /ext/eurospider/live/de/php/aza/http/index_aza.php?...mode=news
+ *
+ * @param {Document} doc
+ * @param {{}}       parameters
+ * @returns {{}}
+ */
+function getSearchResultsNews(doc, parameters) {
+	let domRows = doc.querySelectorAll("div.eit>table:nth-of-type(2)>tbody>tr");
```
AbeJellinek, line 115: Does the selector have to be this strict? It would be better if we could get away with `.eit table tbody tr` or similar.

Schweizerisches Bundesgericht.js
```diff
@@ -0,0 +1,243 @@
...
+ *
+ * Those pages are found at:
+ * /ext/eurospider/live/de/php/aza/http/index.php?l...type=simple_query|simple_similar_documents
+ *
+ * @param {Document} doc
+ * @param {{}}       parameters
+ * @returns {{}}
+ */
+function getSearchResultsBGer(doc, parameters) {
+	let domRows = doc.querySelectorAll("div.ranklist_content>ol>li");
```
AbeJellinek, line 140:
```suggestion
	let domRows = doc.querySelectorAll(".ranklist_content ol > li");
```

For resilience to site changes.

Schweizerisches Bundesgericht.js
```diff
@@ -0,0 +1,243 @@
...
+	let items = {};
+
+	for (let i = 1; i < domRows.length; i++ ) {
+		let meta = domRows[i].children[0].children[0].innerText;
```
AbeJellinek, line 144: Use `text` with a selector here?

Schweizerisches Bundesgericht.js
```diff
@@ -0,0 +1,243 @@
...
+		let date = meta.substring(0, meta.indexOf(' '));
+		let number = meta.substring(date.length + 1);
+		number = number.substring(0, 2) + "_" + number.substring(3);
+		let url  = domRows[i].children[0].children[0];
```
AbeJellinek, line 148: And similar for this and the following two lines.

Schweizerisches Bundesgericht.js
```diff
@@ -0,0 +1,243 @@
...
+ * @param {Document} doc
+ * @param {{}}       parameters
+ * @returns {{}}
+ */
+function getSearchResultsBGE(doc, parameters) {
+	let dom_rows = doc.querySelectorAll("div.ranklist_content>ol>li");
+	let items = {};
+
+	for (let i = 1; i < dom_rows.length; i++ ) {
+		let url = dom_rows[i].children[0].children[0];
```
AbeJellinek, line 171: As above. (Also, we can use `for .. of` instead of an indexed `for`.)

Schweizerisches Bundesgericht.js
```diff
@@ -0,0 +1,243 @@
...
+		item.attachments.push({
+			title: "Snapshot",
+			document: doc
+		});
+		if (parameters._collection === "BGE") {
+			let docket = parameters.highlight_docid;
+			docket = docket.substring(6, docket.length - 3).replace(/-/g, ' ');
+			item.title = abbrevPublished[parameters._lang] + docket;
+			item.number = item.title;
+			item.date = parseInt(docket.split(' ')[0]) + 1874;
```
AbeJellinek, line 194: ...This is unexpected! Can you give an example of what the date looks like on the page?

Schweizerisches Bundesgericht.js
```diff
@@ -0,0 +1,243 @@
...
+			item.abstractNote = ZU.trimInternal(doc.querySelector("div#regeste>div.paraatf").innerText);
```
AbeJellinek, line 195:
```suggestion
			item.abstractNote = ZU.trimInternal(text(doc, "#regeste .paraatf"));
```

Schweizerisches Bundesgericht.js
```diff
@@ -0,0 +1,243 @@
...
+
+	if (_detectWeb(parameters) === "multiple") {
+		Zotero.selectItems(getSearchResults(doc, parameters), function (items) {
+			if (items) ZU.processDocuments(Object.keys(items), scrape);
+		});
+	}
+	else {
+		_scrape(doc, url, parameters);
+	}
+}
```
AbeJellinek, line 243: Can you add tests?

AbeJellinek (changes requested):
Thanks! Some suggestions.

Schweizerisches Bundesgericht.js
KampfCaspar: Unlike `require` the global URL API is _not_ available in my Scaffold sandbox :( (v5.0.96.3)

That's why I searched `Zotero.Utilities` and found its `parseURL` that itself relies on `require('url')` but does not offer the second parameter (to parse the querystring). If `require` is prohibited and `URL` not supported, is there any other documented way to parse the query string - short of low-level parsing it oneself (which would be a shame of duplicated effort).

Schweizerisches Bundesgericht.js
KampfCaspar: The 'published' decisions are numbered in a fixed way "[year of court's existence] [chamber/collection/field of law] [first page]'. The actual date of the decision is not consistently available (especially for older decisions).

As the court (as it stands now) was instituted with the constitution of 1874 and began its work in 1875, the published decisions are numbered from that year. The decision '146 II 49' therefore was published in 2020 (1874 + 146).

NB: There is a small gap between the actual decision date and the publication date. It is therefore possible that a decision late in 2019 was published only in 2020. But as there is no reliable way to determine the actual decision date (or year) and the actual decision date of *published* decisions is in fact never used, it's better than no date at all.

Schweizerisches Bundesgericht.js
KampfCaspar: changed to path particle.

(NB the site is 20 years old (and it shows) - if they change something, everything will break)

Schweizerisches Bundesgericht.js
KampfCaspar: It actually is the second table - but I relaxed the selector as far as I'm comfortable with ;)

Schweizerisches Bundesgericht.js
KampfCaspar: I was not aware of `text` and `attr` as available functions. Could the docs be updated?

Schweizerisches Bundesgericht.js
AbeJellinek: Could you try [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)? That should be available in the sandbox. (It is in my client.)

Schweizerisches Bundesgericht.js
```diff
@@ -241,3 +241,169 @@ function doWeb(doc, url) {
...
+	},
+	{
+		"type": "web",
+		"url": "https://www.bger.ch/ext/eurospider/live/it/php/aza/http/index.php?highlight_docid=aza%3A%2F%2Faza://27-10-2021-8C_329-2021&lang=it&zoom=&type=show_document",
+		"items": [
+			{
+				"itemType": "case",
+				"caseName": "TF 8C_329/2021",
+				"creators": [],
+				"dateDecided": "27.10.2021",
```
AbeJellinek, line 367: We should try to represent the dates as ISO 8601 dates as much as possible, so `YYYY-MM-DD`. Other date formats run into issues with internationalization when they're ambiguous. (On a US-locale machine, for instance, `11.10.2021` is parsed as November 10th, 2021`.)

Schweizerisches Bundesgericht.js
AbeJellinek: Any reason why this suggestion wouldn't work?

AbeJellinek (changes requested):

Schweizerisches Bundesgericht.js
KampfCaspar: URLSearchParams works - one probably could include the 'full' URL in a future version of the sandbox ;)

Schweizerisches Bundesgericht.js
KampfCaspar: That was my oversight - corrected.

Schweizerisches Bundesgericht.js
KampfCaspar: Sorry, I seem to have missed some changes while commiting... Corrected.

Schweizerisches Bundesgericht.js
```diff
@@ -0,0 +1,411 @@
...
+		"items": [
+			{
+				"itemType": "case",
+				"caseName": "ATF 81 I 63",
+				"creators": [],
+				"dateDecided": 1955,
+				"abstractNote": "Militärpflichtersatz: Besteuerung eines Evangelisten, für dessen Unterhalt seine Glaubensfreunde sorgen.",
+				"court": "Tribunal Fédéral Suisse",
+				"docketNumber": "ATF 81 I 63",
+				"url": "https://www.bger.ch/ext/eurospider/live/fr/php/clir/http/index.php?lang=fr&type=highlight_simple_query&page=1&from_date=&to_date=&from_year=1954&to_year=2021&sort=relevance&insertion_date=&from_date_push=&top_subcollection_clir=bge&query_words=UNterhalt&part=all&de_fr=&de_it=&fr_de=&fr_it=&it_de=&it_fr=&orig=&translation=&rank=1&highlight_docid=atf%3A%2F%2F81-I-63%3Ade&number_of_ranks=714&azaclir=clir",
```
AbeJellinek, line 324: OK, final nitpick. Can we clean these URLs up a bit? Item URLs should be as close to concise permalinks as possible, with parameters like search queries and filters removed. Here, for instance, we seem to be able to get away with https://www.bger.ch/ext/eurospider/live/fr/php/clir/http/index.php?lang=fr&type=highlight_simple_query&highlight_docid=atf%3A%2F%2F81-I-63%3Ade - much shorter and more appropriate for a printed bibliography.

AbeJellinek (changes requested):

Schweizerisches Bundesgericht.js
KampfCaspar: Although federal court decisions are never referenced by URL in a bibliography (in Switzerland, at least), I agree it's common sense and much better looking. Done.

PR #2749, opened 2021-11-18
Add note text and HTML export translators

mrtcode:


AbeJellinek:
Marking this as a draft again so it's clear that we shouldn't merge until we add a repo-side block preventing this translator from going out to stable clients - they won't understand the `noteTranslator` config option and it'll show as an export translator for all items, which is incorrect.

Separately, what's the reasoning behind ditching Note Text?

dstillman:
We decided to just use Markdown for plain-text paste targets. Markdown is already intended to be readable and is widely used, so there's no point in having an alternative plain-text representation (which we'd have to reimplement ourselves, since we couldn't use Mozilla's `nsIDocumentEncoder` in a translator). Our Markdown translator will be generated in [a separate repo](https://github.com/zotero/markdown-translator-builder), so we only need to add an HTML translator here. The HTML translator will be used both as HTML for copying rich text and as plain text for copying HTML source (equivalent to the "Copy as HTML" option for the bibliographic formats).

PR #2752, opened 2021-11-23
add translators for fedlex.admin.ch and lexfind.ch (Swiss Laws)

KampfCaspar:
I'm not sure how laws should be mapped to the very US-centric field system...

NB: Searching on lexfind does not work (yet) as it is a javascript site. I'm not sure how instruct ZU.processDocuments to defer scraping after loading the HTML source code. Any hints welcome.

### DIFF
--- a/b/fedlex Swiss Laws.js
+++ b/b/fedlex Swiss Laws.js
@@ -1,0 +1,574 @@
+{
+	"translatorID": "0e53af77-2ffe-4d1e-9271-0c712acbd3e4",
+	"label": "fedlex Swiss Laws",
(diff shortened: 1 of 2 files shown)


GregiGregi:
Hi Caspar

I think you're already aware of it but there's a Zotero style created by Stephan Schlegel (UniBa) which orients on Ryser, Büschi et al:

Stlye:

https://www.zotero.org/styles/juristische-zitierweise-schweizer?source=1

Example collection:

https://www.zotero.org/groups/2044805/ryser_bschi_et_al._beispiele

PR #2757, opened 2021-12-12
fix minor error in Newsbank

reubot:
the call to itemFromRIS was sending the textarea context as the whole document, so risText was empty


AbeJellinek:
Thanks!

PR #2759, opened 2021-12-29
Crossref Unixref XML: Fix reference book with 'other' component type

dstillman:
See note on first commit


adam3smith (approved):
lgtm, thanks!

PR #2761, opened 2021-12-31
Vice rewritten to use JSON-LD

owcz:
Prior version was broken; now with fewer edge cases


owcz:
Not sure what the Linter wants by way of `'url' is defined but never used`—I don't declare it separately

AbeJellinek:
Thanks!

Issue #2760, opened 2021-12-30
ProQuest breaks on standard pages


owcz:
PQ defaults to Embedded Metadata (not the PQ translator) on basic pages such as the following:

* https://www.proquest.com/docview/557394080/74D8D0EE05304335PQ/8

Tested with proxy redirection off

Been like this for a few months, to my recollection

```
18:16:17 On Abstract page, scraping
18:16:17 WARNING: No item type specified
18:16:17 Translation using ProQuest failed: 
         Error: No title specified for item

owcz:
For posterity, logged out abstract pages are supposed to fail to EM because there is no structured metadata to scrape (just the citation listing at the top)

adam3smith:
@owcz we can close this, right?

owcz:
Yes, assuming that it's not worth the effort writing a potentially brittle translator for logged-out metadata

PR #2763, opened 2021-12-31
Repair ProQuest detection on non-Abstract pages

owcz:
Fixes
* `detectWeb` no longer matches PQ pages without an Abstract tab (#2760)
  * This saves the translator from erroring out when it cannot find the Abstract tab; prevents user confusion
* Restored auto-navigation to the Abstract tab when starting the translator on the Full text tab

cc @avram @adam3smith

### DIFF
--- a/ProQuest.js
+++ b/ProQuest.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2021-12-31 09:03:40"
+	"lastUpdated": "2021-12-31 09:11:56"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


ProQuest.js
```diff
@@ -219,23 +219,22 @@ function doWeb(doc, url, noFollow) {
...
-			|| doc.getElementById('tab-Record-null'); // Shown as Details
-		if (!(abstractTab && !abstractTab.classList.contains('active'))) {
-			Zotero.debug("On Abstract page, scraping");
+		var abstractTab = doc.getElementById('addFlashPageParameterformat_abstract');
+		if (abstractTab && abstractTab.classList.contains('active')) {
+			Zotero.debug("On Abstract tab and scraping");
 			scrape(doc, url, type);
 		}
-		else if (noFollow) {
+		else if (noFollow) { // when is this used? wouldn't it be better to throw an error instead of brute forcing?
```
owcz, line 227: when is this used? wouldn't it be better to throw an error instead of brute forcing?

AbeJellinek: Evidently to prevent infinite redirect loops: https://github.com/zotero/translators/commit/bb2a0b519ddba6043f145a1e677f3d1333f2154c

I think it's good to have in case the design changes slightly and the `abstractTab.classList.contains('active')` check no longer works. If you can remove the added comment, I think this is ready to merge.

owcz:
@AbeJellinek I removed the comment you mentioned, if this is still ready for merge

AbeJellinek:
Thanks! All looks good, test pages work in browser.

PR #2764, opened 2021-12-31
Add new translator for opentextbc.ca chapter urls

mvolz:
Add new translator for opentextbc.ca chapter urls;
uses embedded metadata translator and then
fixes itemType to bookSection
and adds bookTitle.

Doesn't capture the book URLs, as embedded
metadata translator already correctly translates
these as books.

### DIFF
--- a/OpenTextBC.js
+++ b/OpenTextBC.js
@@ -1,0 +1,89 @@
+{
+	"translatorID": "8bdd3e33-87a0-488e-aa06-05bf87bea942",
+	"label": "openTextBC",
+	"creator": "Marielle Volz",
+	"target": "^https?://opentextbc\\.ca/[a-z0-9\\-]+/chapter/[a-z0-9\\-]+",
+	"minVersion": "4.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsbv",
+	"lastUpdated": "2021-12-31 15:20:11"
+}
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2018 Sebastian Karcher
+
+	This file is part of Zotero.
+
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+function detectWeb() {
+	return "bookSection";
+}
+
+function doWeb(doc, url) {
+	var translator = Zotero.loadTranslator('web');
+	ZU.doGet(url, function () {
+		// Embedded Metadata
+		translator.setTranslator('951c027d-74ac-47d4-a107-9c3069ab7b48');
+		translator.setDocument(doc);
+		translator.setHandler("itemDone", function (obj, item) {
+			item.itemType = 'bookSection'; // Overwrite journal article
+			item.bookTitle = ZU.xpath(doc, '//*[@id="page"]/header/div[2]/nav/h1/a/text()')[0].textContent;
+			item.complete();
+		});
+		translator.translate();
+	});
+}
+
+/** BEGIN TEST CASES **/
+var testCases = [
+	{
+		"type": "web",
+		"url": "https://opentextbc.ca/clinicalskills/chapter/4-4-suture-care-and-removal/",
+		"items": [
+			{
+				"itemType":"bookSection",
+				"creators":[
+					{
+						"firstName":"Glynda Rees",
+						"lastName":"Doyle",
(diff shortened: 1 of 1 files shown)


mvolz:
Hmm, checks pass on my fork...

adam3smith:
You're good. The tests fail because the setup to run the translator tests using chromedriver fails (currently for all PRs), nothing to do with your translator

OpenTextBC.js
```diff
@@ -0,0 +1,89 @@
...
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+function detectWeb() {
+	return "bookSection";
+}
```
AbeJellinek, line 40:
```suggestion
function detectWeb(doc, _url) {
	if (attr(doc, 'meta[name="application-name"]', 'content') == 'Pressbooks'
		&& doc.querySelector('meta[name="citation_title"]')) {
		return "bookSection";
	}
	return false;
}
```

OpenTextBC.js
```diff
@@ -0,0 +1,89 @@
...
+		// Embedded Metadata
+		translator.setTranslator('951c027d-74ac-47d4-a107-9c3069ab7b48');
+		translator.setDocument(doc);
+		translator.setHandler("itemDone", function (obj, item) {
+			item.itemType = 'bookSection'; // Overwrite journal article
+			item.bookTitle = ZU.xpath(doc, '//*[@id="page"]/header/div[2]/nav/h1/a/text()')[0].textContent;
+			item.complete();
+		});
+		translator.translate();
+	});
```
AbeJellinek, line 54: The `doGet` seems completely unnecessary here.

OpenTextBC.js
```diff
@@ -0,0 +1,89 @@
...
+
+function doWeb(doc, url) {
+	var translator = Zotero.loadTranslator('web');
+	ZU.doGet(url, function () {
+		// Embedded Metadata
+		translator.setTranslator('951c027d-74ac-47d4-a107-9c3069ab7b48');
+		translator.setDocument(doc);
+		translator.setHandler("itemDone", function (obj, item) {
+			item.itemType = 'bookSection'; // Overwrite journal article
+			item.bookTitle = ZU.xpath(doc, '//*[@id="page"]/header/div[2]/nav/h1/a/text()')[0].textContent;
```
AbeJellinek, line 50: Use `text` with a less rigid selector:

```suggestion
			item.bookTitle = text(doc, '.reading-header__title a');
```

OpenTextBC.js
```diff
@@ -0,0 +1,89 @@
...
+}
+
+/** BEGIN TEST CASES **/
+var testCases = [
+	{
+		"type": "web",
+		"url": "https://opentextbc.ca/clinicalskills/chapter/4-4-suture-care-and-removal/",
+		"items": [
+			{
+				"itemType":"bookSection",
```
AbeJellinek, line 64: What's up with the lack of spacing? If you're using Scaffold, it shouldn't be producing JSON that looks like this. If you aren't, just use it to load and save the translator once to normalize the formatting here.

OpenTextBC.js
```diff
@@ -0,0 +1,89 @@
+{
+	"translatorID": "8bdd3e33-87a0-488e-aa06-05bf87bea942",
+	"label": "openTextBC",
```
AbeJellinek, line 3: This should be able to support more than just OpenTextBC! For instance, [here's a book from URegina](https://opentextbooks.uregina.ca/leadershipandinfluencingchangeinnursing/chapter/chapter-1-identifying-your-leadership-strengths-and-opportunities-for-growth/) being served by the same software. The application is called Pressbooks (some more users/"networks" listed [here](https://pressbooks.directory/))...

OpenTextBC.js
```diff
@@ -0,0 +1,89 @@
...
+	"creator": "Marielle Volz",
+	"target": "^https?://opentextbc\\.ca/[a-z0-9\\-]+/chapter/[a-z0-9\\-]+",
```
AbeJellinek, line 5: ...And we have a couple choices for how to approach matching URLs. We could do it like this and have a lot of false positives that we filter out with `detectWeb`:

```suggestion
	"target": "^https?://[^/]+/[a-z0-9\\-]+/chapter/[a-z0-9\\-]+",
```

Or we could try to come up with a regex that matches everything. A lot of Pressbooks sites are subdomains of `pressbooks.pub`, but not all; see a full list by expanding "Network" [here](https://pressbooks.directory/).

OpenTextBC.js
```diff
@@ -0,0 +1,89 @@
...
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsbv",
+	"lastUpdated": "2021-12-31 15:20:11"
+}
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2018 Sebastian Karcher
```
AbeJellinek, line 18: Update this

AbeJellinek (changes requested):
Thanks! Some comments/suggestions. We should definitely generalize this to all Pressbooks sites.

OpenTextBC.js
mvolz: Yeah, I'm not sure url matching will work for this - maybe I should just move this into the embedded metadata translator entirely instead of doing a separate translator, and check for <meta name="application-name" content="Pressbooks">?

AbeJellinek: We don't want to hardcode workarounds for specific sites/apps into EM - that's what other translators are for. Using the regex I suggested with a more specific `detectWeb` should work fine.

PR #2769, opened 2022-01-07
Update Delpher for new site, add PDF/image support

rwiermer:
Fix scraping for new website layout
Adds support for PDF and images

### DIFF
--- a/Delpher.js
+++ b/Delpher.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2022-01-20 14:21:26"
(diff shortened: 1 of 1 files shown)


AbeJellinek:
Pushed some changes to deal with lint issues/consistency with other translators. I'm having some trouble updating tests on my end with Scaffold, but they do need to be updated - if you're able to do it on your end and push, please do, otherwise I'll figure out why the page isn't loading correctly and come back to this later.

rwiermer:
Thanks! I updated the test cases. They are still running into timeouts sometimes in Scaffold (15 seconds passed). Not sure if that can (or should) be fixed.

AbeJellinek:
Thanks!

PR #2775, opened 2022-01-23
Fix Straits Times

robertsky:
major changes in the meta elements, and some changes to their byline section rendered the current version of the translator non-functional. this update retargets at the remaining available meta elements and byline section.


AbeJellinek:
Thank you!

PR #2777, opened 2022-01-24
Beobachter - rewrite

adam3smith:
Closes https://github.com/zotero/translators/issues/2486

I got rid of the issue/volume designation as this looks like it's almost 
entirely online, with no sign of published issues on the online articles

### DIFF
--- a/Beobachter.js
+++ b/Beobachter.js
@@ -36,8 +36,7 @@
 	***** END LICENSE BLOCK *****
 */
 
-// eslint-disable-next-line no-unused-vars
-function detectWeb(doc, url) {
+function detectWeb(doc, _url) {
 	if (doc.getElementsByClassName('article-header').length > 0) {
 		return "magazineArticle";
 	}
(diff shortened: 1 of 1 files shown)


AbeJellinek:
Rebased on master to get CI to work (hopefully)

Beobachter.js
```diff
@@ -183,38 +107,67 @@ var testCases = [
 	},
 	{
 		"type": "web",
-		"url": "http://www.beobachter.ch/natur/forschung-wissen/klima-wetter/artikel/blitzschlag_suche-nicht-die-buche/",
+		"url": "https://www.beobachter.ch/umwelt/blitze-suche-nicht-die-buche",
 		"items": [
 			{
 				"itemType": "magazineArticle",
+				"title": "Blitze: Suche nicht die Buche! | Beobachter",
```
AbeJellinek, line 114: We should strip `| Beobachter` from the title, but otherwise this looks good!

AbeJellinek (changes requested):

Beobachter.js
adam3smith: OK, should be all set

Issue #1745, opened 2018-09-13
Google Scholar Profiles


adam3smith:
Fail when there is no "Related" available for articles -- but I'm not seeing any other sign of the item ID. There is a numerical ID that we could see what it does, but that's quite involved.

PR #2778, opened 2022-01-30
Fixes Google Scholar profiles

adam3smith:
Closes https://github.com/zotero/translators/issues/1745

### DIFF
--- a/Google Scholar.js
+++ b/Google Scholar.js
@@ -12,37 +12,60 @@
 	"lastUpdated": "2022-01-30 19:53:00"
 }
 
-// attr()/text() v2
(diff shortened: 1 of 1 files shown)


Google Scholar.js
```diff
@@ -282,11 +306,11 @@ var scrapeCase = function (doc, url) {
 	// Citelet is identified by
 	// id="gsl_reference"
 	var refFrag = doc.evaluate('//div[@id="gsl_reference"] | //div[@id="gs_reference"]',
-					doc, null, XPathResult.ANY_TYPE, null).iterateNext();
+		doc, null, XPathResult.ANY_TYPE, null).iterateNext();
 	if (refFrag) {
 		// citelet looks kind of like this
 		// Powell v. McCormack, 395 US 486 - Supreme Court 1969
-		var item = new Zotero.Item("case");
+		// var item = new Zotero.Item("case");
```
adam3smith, line 313: This I'd want to test again -- I don't *think* we need it but let's double check (GS had me locked out by the time I tried

AbeJellinek: It doesn't look like `item` is ever used, so we can just go ahead and delete that line entirely.

adam3smith:
Addressed comment and squashed extra commit

PR #2779, opened 2022-02-01
RIS: Output ISO dates

AbeJellinek:
To fix the issue discussed [here](https://forums.zotero.org/discussion/94168/zotero-extracting-publication-month-in-my-systems-language). Instead of setting `item.date` to a date formatted in the user's locale ("1 January 2022"), it now always outputs ISO dates ("2022-01-01").

### DIFF
--- a/RIS.js
+++ b/RIS.js
@@ -17,7 +17,7 @@
 	},
 	"inRepository": true,
 	"translatorType": 3,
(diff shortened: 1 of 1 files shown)


dstillman:
This looks fine to me (but I think we can just use `String.prototype.padStart()` instead of `Zotero.Utilities.lpad()` now).

PR #2782, opened 2022-02-05
Fix Atypon for Science

adam3smith:
Prompted by 
https://forums.zotero.org/discussion/94325/safari-connector-not-recognizing-articles-on-the-science-website#latest 
though not sure it's the exact same issue

### DIFF
--- a/Atypon Journals.js
+++ b/Atypon Journals.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2021-07-07 17:32:12"
+	"lastUpdated": "2022-02-05 20:00:12"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


Atypon Journals.js
```diff
@@ -720,441 +720,6 @@ var testCases = [
 						"creatorType": "author",
 						"fieldMode": 1
 					},
-					{
```
AbeJellinek, line 723: This doesn't seem right - where'd all the authors go?

AbeJellinek (changes requested):

Atypon Journals.js
adam3smith: All those authors are part of the COVID-19 Review Consortium. They're not listed at all in the RIS etc. export and they're included in the same content attribute as COVID-19 Review Consortium in the metatags with some quite wonky HTML. I don't think we should try to fix it (even in terms of what _should_ be in the metadata it's a borderline case)

AbeJellinek: Makes sense.

PR #2784, opened 2022-02-10
Fix WestLaw case title scraping

georgegebbett:
Annoyingly, WestLaw added " | WestLaw UK" to all the page titles right as my WestLaw translator was merged. This screwed with the case scraping, and this PR gets the case names from somewhere more reliable on the page rather than the page title. 

Statute titles are not affected as their contents was already pulled from the page itself in a much more reliable way.

### DIFF
--- a/WestLaw UK.js
+++ b/WestLaw UK.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2022-02-10 22:37:23"
+	"lastUpdated": "2022-02-10 22:42:40"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


WestLaw UK.js
```diff
@@ -73,10 +73,11 @@ function scrapeCase(doc, url) {
 	translator.setHandler('itemDone', function (obj, item) {
 		const citationArray = parseCitationList(text(doc, "#co_docContentWhereReported", 0));
 
+		item.title = text(doc, "#co_docHeaderContainer");
```
adam3smith, line 76: is it worth checking whether this exists before overwriting the EM title, to make this more robust? (and otherwise just remove the addition from the existing item title)?

georgegebbett: All added

adam3smith:
Thanks! Lots of requests for this over the years, I'm glad Westlake is working now

georgegebbett:
The problem is the intersection between the people who have access to WestLaw and those who are comfortable enough with JS to have a go at writing a translator is very thin! 

I do not profess to be a JS wizz in any sense but very much hacked this one together to save me time in my degree, and my usefulness is very limited because I only have access to WestLaw UK - I am very confident that this translator would not work for any other region's WestLaw site...

PR #2786, opened 2022-02-14
nbib - Get page number from LID field when no PG field exists

### LINKED ISSUE #2783: nbib -- add article numbers
Reported here: https://forums.zotero.org/discussion/94403/nbib-article-number-missing-from-metadata#latest
We are importing article numbers into the page field from Pubmed XML and should just copy the logic

georgegebbett:
This PR fixes the scenario where an nbib file without a PG field is imported. In this case, the LID (Location Identifier) field is inspected for a PII - if one is found, this is used as the page number in the created item. 

The LID field has been chosen over the AID (Article Identifier) field as this matches the behaviour seen in the PubMed XML translator.

This PR resolves #2783

### DIFF
--- a/MEDLINEnbib.js
+++ b/MEDLINEnbib.js
@@ -8,7 +8,7 @@
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 1,
-	"lastUpdated": "2022-02-14 13:07:48"
+	"lastUpdated": "2022-02-14 13:09:41"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


adam3smith:
Thanks! Zotero currently saves unused elements in the `item` object to Extra, so this leads to: 
image

You'd need to remove `item.pagesBackup` at the end of the finalize function

georgegebbett:
Interestingly I could not reproduce that behaviour at all - at any rate I have added in the deletion :)

adam3smith:
FWIW, this only shows up in Zotero, in Scaffold it shows up as 
` -   "pagesBackup": "1042"`
Not sure if this is only in the beta, but I thought it preceded that.

georgegebbett:
Yeah I looked in Zotero and didn't see it there 🤔 I don't know the last time I updated Zotero though and I could be many many versions behind

adam3smith:
Great, thank you!

PR #2790, opened 2022-02-27
Switch arXiv to report / article

adam3smith:
Add arxiv DOI; differentiate detectWeb function


arXiv.org.js
```diff
@@ -650,9 +663,10 @@ var testCases = [
...
-				"extra": "arXiv: 1810.04805\nversion: 1",
+				"extra": "arXiv:1810.04805 [cs]\nversion: 1\ntype: article",
+				"institution": "arXiv",
 				"libraryCatalog": "arXiv.org",
-				"publicationTitle": "arXiv:1810.04805 [cs]",
+				"reportNumber": "arXiv: 1810.04805",
```
bwiernik, line 669:
```suggestion
				"reportNumber": "arXiv:1810.04805",
```

There shouldn't be a space here should there?

bwiernik (changes requested):
There shouldn't be a space between `arxiv:` and the numeric portion of the ID, should there?

https://arxiv.org/help/arxiv_identifier

dstillman:
So this is tricky — we should update this to use the new Preprint item type, which I'm planning to roll out in the next beta, but we then can't push it to older clients. (I'm actually not sure what happens if a translator tries to save to an unknown item type, but probably nothing good.)

We could just not update the timestamp and bundle it with new clients, but then we wouldn't be able to make any changes for older clients if, say, arXiv made a site change. That's probably fine — people can upgrade — but it would be a bit nicer to test for the newer version in the translator and return the right type.

@AbeJellinek, should we make the [schema version](https://github.com/zotero/zotero-schema/) available to translators? As we make additional type/field changes, it might be useful to be able to test for that and return different data conditionally.

AbeJellinek:
Sure. `Zotero.schemaVersion` within the sandbox?

We don't need to touch translation architecture to add a basic check - `ZU.fieldIsValidForType('title', 'preprint')` - but making the schema version available is better.

dstillman:
Oh, that seems fine, actually — I forgot we had that available. Unless we can think of some change that we couldn't handle with that, it seems better to rely on explicit type/field tests than special knowledge of arbitrary integers.

AbeJellinek:
Yeah, and it simplifies compatibility with Juris-M and the like.

adam3smith:
So given that it seems like we accidentally forced the 6.0 update on everyone, are we still doing the tests (which I'm happy to add, but it's a fair amount of clutter) or are we just breaking this for 5.x versions?

AbeJellinek:
I'm still in favor of adding a test and returning `report` items where `preprint` isn't supported until we cut off translator updates for <6.0 (not yet, not sure what the timeline is for that).

arXiv.org.js
```diff
@@ -251,6 +264,17 @@ function parseXML(text) {
 		translate.translate();
 	}
 	else {
+		newItem.publisher = "arXiv";
+		newItem.number = "arXiv:" + articleID;
```
adam3smith, line 268: This line relies on https://github.com/zotero/zotero/issues/2481#issuecomment-1079754076 being fixed to work for `preprint`

AbeJellinek:
Now that the beta connector supports `preprint`, are we ready to merge this?

adam3smith:
Ready from my end

AbeJellinek:
OK, although I did agree about the space before, I'm not so sure it should be that way in `extra`. Does the arXiv ID in `extra` actually do anything when interpreted by CiteProc? If it does, it should have a space after the colon, I think. If it's just for user info, I guess not having a space is fine.

bwiernik:
It doesn't currently do anything with citeproc. It could potentially be used more if/when CSL adds a flexible ID variable system

AbeJellinek:
Alright, makes sense. Then this works for now!

dstillman:
Seems like if we're saving it there should be a space. Users can enter whatever they want in Extra, of course, but we generally only save key-value pairs with spaces from translators.

Why are we saving it at all, though? Isn't it redundant with `archiveID`? If CSL did gain more flexible ID handling, Zotero would know how to parse it out of `archiveID` based on the `arXiv` prefix.

bwiernik:
Extra is the only place where the field labels are stored (eg, [stat]). Those aren't generally cited, but some ArXiV users seem to want them. I don't have a strong preference and think the Extra part could be dropped.

dstillman:
Yeah, but then it seems like if anything they should go in a separate `Classification: ` key or something without the redundant arXiv ID.

Have we saved these up to now? Really they seem like tags, and we're already saving the full strings for them as automatic tags…

fkguo:
> Extra is the only place where the field labels are stored (eg, [stat]). Those aren't generally cited, but some ArXiV users seem to want them. I don't have a strong preference and think the Extra part could be dropped.

I think it's better to have the format matches that given in the "Cite as" line on each arxiv article page. Currently, all the cross-listed categories are also listed inside the square brackets, and the primary category that the article is submitted to is not the first.

bwiernik:
In practice, the vast majority of citations to ArXiV preprints do not include the categories, and they are not canonically part of the ArXiV ID

fkguo:
> In practice, the vast majority of citations to ArXiV preprints do not include the categories, and they are not canonically part of the ArXiV ID

Here I have a different opinion. Historically the primary category was a part of the arXiv number. Up to 2007, the numbers were written as sth like primary-category/0701... . And the suggested format in the arxiv page "cited as" contains the primary category and not the cross-listed ones...

bwiernik:
Searching actual citations in Google scholar full text shows that they are very rarely included.

fkguo:
well, it tells the reader which field the authors of the paper consider it to be in, and it is in the arXiv standard... (the arxiv page for any paper has a line of "Cite as:"). Currently, the extra field includes also the cross-listed categories, which looks too clumsy.

PR #2804, opened 2022-03-23
Update NIPS for 2020-22

### LINKED ISSUE #2803: doesn't work for NeurIPS 2020 and 2021
maybe related to #2590 . For the proceedings 2020 and 2021 the zotero-connector doesn't detect any articles in the webpage when clicking 'save to zotero'. For example, this paper: https://proceedings.neurips.cc/paper/2020/hash/0060ef47b12160b9198302ebdb144dcf-Abstract.html

Is this caused by the change of webpage structure? The webpage for 2018 and 2019 has a 'metadata' link, but this is removed in the webpage for year 2020.

adam3smith:
Close #2803
Note that this won't work on some pages because there's no bibtex. Not 
detecting them on single import, not sure what to do for search/multiple

### DIFF
--- a/Neural Information Processing Systems.js
+++ b/Neural Information Processing Systems.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2022-03-23 00:51:45"
+	"lastUpdated": "2022-03-26 18:20:28"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


AbeJellinek:
I think we should fall back to an EM-based scrape when BibTeX isn't available. I haven't looked at this in detail, but the Highwire meta tags for [one BibTeX-less paper](https://proceedings.neurips.cc/paper/2021/hash/b618c3210e934362ac261db280128c22-Abstract.html) are pretty good.

AbeJellinek:
Excellent, thanks.

PR #2807, opened 2022-03-28
conference_proceeding->book

denismaier:
This adjusts the behavior for conference proceedings so they get imported as item type `book`. The current behavior (import as item type `conference paper`) doesn't seem to be a good fit IMO.


AbeJellinek:
Thanks! Do you have some `conference_proceeding` PNX we could use to add a test for this?

AbeJellinek:
(I mostly just want to see how this affects what other fields are mapped to.)

AbeJellinek:
OK, sounds good.

PR #2811, opened 2022-04-03
Add two missing Stack Exchange sites

baltpeter:
Just a very minor update. I just noticed the new Stack Exchange translator. That is incredibly helpful, thank you!

There are two more Stack Exchange sites not caught by the current target regex. This PR adds support for them (superuser.com is actually quite common, I'd say, stackapps.com… less so).

I checked this by running the following the browser console of https://stackexchange.com/sites?view=list#traffic:

```js
sites = Array.from($$('.lv-info h2 a')).map(e => e.href)
missing_sites = sites.filter(s => !s.match(/^https:\/\/([^\/]+\.)?(((stack(overflow|exchange)|serverfault|askubuntu)\.com)|mathoverflow\.net)\//))
```

### DIFF
--- a/Stack Exchange.js
+++ b/Stack Exchange.js
@@ -2,7 +2,7 @@
 	"translatorID": "26ecfee6-535d-4044-89a8-fbda31883642",
 	"label": "Stack Exchange",
 	"creator": "Abe Jellinek",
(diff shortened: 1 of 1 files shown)


AbeJellinek:
Thank you!

PR #2812, opened 2022-04-05
Update Translator Der Freitag

JLueke:
Rewrite and update Translator for "Der Freitag" with Embedded Metadata as base translator.

### DIFF
--- a/Der Freitag.js
+++ b/Der Freitag.js
@@ -9,10 +9,9 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2016-11-01 18:27:25"
+	"lastUpdated": "2022-04-05 19:08:57"
 }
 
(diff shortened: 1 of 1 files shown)


JLueke:
The lint script in package json should be updated. There was no error with the command documented here:
https://www.zotero.org/support/dev/translators

`npm ci && npm run lint -- "Your Translator.js"`

AbeJellinek:
You have to replace `"Your Translator.js"` with the translator filename. (It should return an error when passed an invalid filename - will fix.)

JLueke:
> You have to replace `"Your Translator.js"` with the translator filename. (It should return an error when passed an invalid filename - will fix.)

I know that. For me the "lint" script didn't worked. Using the _eslint_ command directly worked, _testlint_ didn't.

AbeJellinek:
Hmm, that's odd. Did it give you any output at all?

AbeJellinek:
(`eslint` works fine but doesn't support some Zotero-specific linters that use a cache of translator data.)

Issue #2815, opened 2022-04-14
wos can not be recongnized


tughv:
URL: https://www.webofscience.com/wos/alldb/full-record/DIIDW:202205717D
![image](https://user-images.githubusercontent.com/40344969/163313922-908f9305-b5a1-4135-87d1-8a0e8031c29f.png)

AbeJellinek:
@adam3smith, any chance you could take a look at this? I don't have WoS access anymore and they don't even let you see abstracts without logging in.

PR #2817, opened 2022-04-22
Add support for all DBs in WoS

adam3smith:
Closes https://github.com/zotero/translators/issues/2815

Not thoroughly tested, but should generally work


AbeJellinek:
Excellent, thank you!

PR #2821, opened 2022-05-01
Dart-Europe change domain from eu to org

mmoole:
it seems that the domain for this site got changed from eu to org - now another site is online at the eu domain.
So i just changed .eu to .org in the code, but don't know if this will fix everything....


AbeJellinek:
Be sure to bump the `lastUpdated` date, which is automatic if you use Scaffold. Otherwise this looks good. (A rebase might fix CI.)

mmoole:
thanks for the hint, hoping it now works 👍

PR #2824, opened 2022-05-13
Resolve multiple issues of Douban.js:

jacobleft:
1. fix pattern matches for info with links (a tags);
2. fix pattern matches for creators;
3. fix abstract note xpath;
4. add subtitle;
5. add original title to extra
6. remove Chinese characters in date and in page numbers

### DIFF
--- a/Douban.js
+++ b/Douban.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2020-06-04 03:46:18"
+	"lastUpdated": "2022-05-13 14:37:34"
 }
 
 /*
@@ -52,6 +52,15 @@
 // #### Local utility functions ####
 // #################################
 
+
+function getTextInaTag(text) {
+	var aTag = /<a [^>]*?>(.*?)<\/a>/;
+	if (aTag.test(text)) {
+		return aTag.exec(text)[1];
+	} else {
+		return text
+	}
+}
 function trimTags(text) {
 	return text.replace(/(<.*?>)/g, "");
 }
@@ -62,8 +71,9 @@ function trimTags(text) {
 
 function scrapeAndParse(doc, url) {
 	// Z.debug({ url })
-	Zotero.Utilities.HTTP.doGet(url, function (page) {
-		// Z.debug(page)
+
+	ZU.doGet(url, function (page) {
+		
 		var pattern;
 
 		// 类型 & URL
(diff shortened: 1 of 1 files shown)


Douban.js
```diff
@@ -72,27 +82,35 @@ function scrapeAndParse(doc, url) {
...
-			newItem.title = Zotero.Utilities.trim(trimTags(title));
-			// Zotero.debug("title: "+title);
-		}
-
-		// 又名
-		pattern = /<span [^>]*?>又名:(.*?)<\/span>/;
 		if (pattern.test(page)) {
 			var shortTitle = pattern.exec(page)[1];
-			newItem.shortTitle = Zotero.Utilities.trim(shortTitle);
+			newItem.shortTitle = ZU.trim(trimTags(shortTitle));
```
AbeJellinek, line 89: Can just use `.trim()` in modern browsers.

Douban.js
```diff
@@ -72,27 +82,35 @@ function scrapeAndParse(doc, url) {
...
-
+		
 		page = page.replace(/\n/g, "");
 		// Z.debug(page)
+		
+		// 标题 = 主标题: 副标题
+		pattern = /<span [^>]*?>副标题:<\/span>(.*?)<br/;
+		if (pattern.test(page)) {
+			var subTitle = pattern.exec(page)[1];
+			Zotero.debug("subTitle: "+subTitle);
```
AbeJellinek, line 100: No need to debug

Douban.js
```diff
@@ -72,27 +82,35 @@ function scrapeAndParse(doc, url) {
...
+			newItem.title = newItem.shortTitle + ": " + ZU.trim(subTitle);
```
AbeJellinek, line 101: Ditto with `.trim()`

Douban.js
```diff
@@ -108,69 +126,83 @@ function scrapeAndParse(doc, url) {
 						useComma = false;
 					}
 				}
-				newItem.creators.push(Zotero.Utilities.cleanAuthor(
-					Zotero.Utilities.trim(authorNames[i]),
+				newItem.creators.push(ZU.cleanAuthor(
+					ZU.trim(authorNames[i]),
```
AbeJellinek, line 130: Ditto

AbeJellinek (changes requested):
Thanks! This generally looks quite good. Do most people citing Douban want to cite with the title/names in Chinese, though? (Assuming this PR caused that change and it wasn't just on Douban's end). If so, I don't have a problem with it.

jacobleft:
> Thanks! This generally looks quite good. Do most people citing Douban want to cite with the title/names in Chinese, though? (Assuming this PR caused that change and it wasn't just on Douban's end). If so, I don't have a problem with it.

Thank you for your review. 
This PR does NOT change the language of titles/names, but adds the original title of **translated** book. This field is already on the Douban website, but was ignored by the previous version of Douban.js.

PR #2827, opened 2022-05-27
Add CSTR identifier resovle Translators

cheneyfeng3:
Add CSTR identifier resovle Metadata Web Translators and Search Translators

https://forums.zotero.org/discussion/97130/add-cstr-identifier-to-the-add-item-by-identifier-function#latest


adam3smith:
@AbeJellinek -- are you still working on translators? This has policy implication beyond translators, so I'd rather not review it.

AbeJellinek:
@adam3smith: Sorry for the inactivity here lately! Yes, I am, although less than before.

I think my main issue with this PR is that I'm not convinced of its usefulness - it seems like all documents with CSTR IDs also have DOIs, and CSTRs don't have any author info in 95%+ of cases (including the tests for this translator). Why not just use the DOI? Or at least just make this translator resolve CSTRs to DOIs and pull the metadata from the DOI?

PR #2832, opened 2022-06-13
Add Libraries Tasmania translator

wragge:
This is a translator for [Libraries Tasmania](https://libraries.tas.gov.au/). The catalogue uses SirsiDynix, but none of the existing translators worked and the system has been customised to include a range of important archival material. There's actually three catalogues rolled into one:

- the Library catalogue for all the normal library stuff (but also including things like archived websites)
- the Archives catalogue
- the Name Index catalogue -- basically transcribed entries from certain archive collections

The translator captures data from all three sections. Many of the records in the Library section have MARC views, so I was able to use the existing MARC translator for those. Some elements are generated by JS, so they took a bit of tracking down. The digital files are rendered and linked to in a variety of different ways. I tried many different solutions, but this one seems to work ok in most cases. The main limit is that if a record links to a multi-image viewer, without a reference to a specific page, no images are saved. But if it's a single image viewer, or there's a link identifying the page, the image/file will be saved -- both images and PDFs. In some cases, such as the Convict Name index, you get multiple image attachments which is pretty cool.


AbeJellinek:
Thanks, this is awesome work! That said... I think the issue might just be that we don't have a (working?) translator that targets SIRSI Dynix. [This site](https://librariestas.ent.sirsidynix.net.au/client/en_AU/names/search/detailnonmodal/ent:$002f$002fNAME_INDEXES$002f0$002fNAME_INDEXES:1384365/one) looks similar, as does [this one](https://mrcc.libsvic.ent.sirsidynix.net.au/client/en_AU/mrcc/search/results?qu=Dream+town&te=ILS) and [this one](https://wales.ent.sirsidynix.net.uk/client/en_GB/vog_en/search/detailnonmodal/ent:$002f$002fSD_ILS$002f0$002fSD_ILS:5170442/one?qu=TITLE%3D%22A+peculiar+problem%22&qu=AUTHOR%3DPauley%2C+Kimberly%2C+1973-). None of them are detected by any translator on my end. Could you test out if this translator can target other Dynix sites? If so, let's just make it more general.

wragge:
Good idea, but I think the Libraries Tasmania catalogue is too heavily customized for the translator to work with more generic SirsiDynix systems. Looking at the two examples you mentioned, there are some major differences:

- The links on items in the search results don't include any item identifiers. To get a link to an item page I think you'd have to get an identifier from the 'Place hold' button (or somewhere else) and build a url. The links in Libraries Tas go to the item page.
- The other systems don't seem to provide the MARC view that Libraries Tas includes in many of the bib records.
- It looks like most of the field/class names are different in the item records, so you'd need another set of selectors anyway.

On top of that Libraries Tas adds permalinks, digitised items, and separate custom catalogues for archival material. None of which seem to be standard.

From a quick poke around, it looks like a translator could be created to work with the more generic SirsiDynix systems you've linked to, but I don't think there'd be much overlap with the Libraries Tas translator.

wragge:
Ah looking closer it seems you can get the item identifiers from the onclick event on the titles, but it's still different to Libraries Tas which has the full item url in the href.

wragge:
Is there anything else you'd like me to do?

Libraries Tasmania.js
```diff
@@ -0,0 +1,781 @@
...
+	dvd: "videoRecording",
+	"archived website": "webpage"
+};
+
+function detectWeb(doc, url) {
+	let catType = url.match(/en_AU\/(library|tas|names)\/search/);
+	if (url.includes("results") && getSearchResults(doc, true)) {
+		return "multiple";
+	}
+	else if (catType == "tas" || url.includes("ARCHIVES_")) {
```
AbeJellinek, line 53: `catType ==` might be a little sketchy - `String#match` returns an array, or null if there was no match. Instead, how about `catType && catType[1] == "tas"` (and similar for the check below)?

Libraries Tasmania.js
```diff
@@ -0,0 +1,781 @@
...
+		return "manuscript";
+	}
+	else if (catType == "names" || url.includes("NAME_INDEXES")) {
+		return "manuscript";
+	}
+	else {
+		var formats = doc.querySelectorAll("div.displayElementText.text-p.LOCAL_FORMAT");
```
AbeJellinek, line 60: Are all these classes necessary or would just `".LOCAL_FORMAT"` work? For stability if the site changes.

Libraries Tasmania.js
```diff
@@ -0,0 +1,781 @@
...
+		Zotero.selectItems(getSearchResults(doc, false), function (items) {
+			if (!items) {
+				return true;
+			}
+			var articles = [];
+			for (var i in items) {
+				articles.push(i);
+			}
+			ZU.processDocuments(articles, scrape);
+			return false;
```
AbeJellinek, line 81: These days this can be simplified to

```suggestion
			if (!items) return;
			ZU.processDocuments(Object.keys(items), scrape);
```

Libraries Tasmania.js
```diff
@@ -0,0 +1,781 @@
...
+		if (!href || !title) continue;
+		if (checkOnly) return true;
+		found = true;
+		items[href] = title;
+	}
+	return found ? items : false;
+}
+
+function getFieldText(doc, label) {
+	let fields = doc.querySelectorAll("div.displayElementText.text-p." + label);
```
AbeJellinek, line 105: Same question about the selectors here - it seems like we could at least drop the `div` tag selector, and maybe `.displayElementText`, since that seems like a more visual and less semantic class name and is liable to change. We could avoid false positives by running `querySelectorAll()` on a smaller scope instead of the full document, like the main metadata table element if there is a consistent one.

Libraries Tasmania.js
```diff
@@ -0,0 +1,781 @@
...
+			fieldTexts.push(ZU.trimInternal(fields[i].textContent));
+		}
+		return fieldTexts.join(" | ");
+	}
+	return "";
+}
+
+function getLinkLists(doc, label, idx) {
+	var values = [];
+	let links = doc.querySelectorAll("div." + label + " a[href*='" + idx + "']");
```
AbeJellinek, line 118: Is the `div` necessary here? (Sorry for being annoying about these!)

Libraries Tasmania.js
```diff
@@ -0,0 +1,781 @@
...
+	else {
+		scrapeLibrary(doc, url);
+	}
+}
+
+function scrapeLibrary(doc, url) {
+	// Format of item
+	var format = detectWeb(doc, url);
+	// If there's a 'Library View' tab we can use MARC
+	if (doc.querySelector("div#marc-tab0")) {
```
AbeJellinek, line 303: `div` definitely not necessary with a unique ID selector.

Libraries Tasmania.js
```diff
@@ -0,0 +1,781 @@
...
+		// call MARC translator
+		var translator = Zotero.loadTranslator("import");
+		translator.setTranslator("a6ee60df-1ddc-4aae-bb25-45e0537be973");
+		translator.getTranslatorObject(function (marc) {
+			var record = new marc.record();
+			var item = new Zotero.Item();
+			var fieldTag, indicators, fieldContent;
+			for (let row of doc.querySelectorAll("div#marc-tab0 table tr")) {
+				let fields = row.querySelectorAll("td");
+				fieldTag = ZU.trim(fields[0].textContent);
```
AbeJellinek, line 313: `ZU.trim` can be replaced by `String#trim()` here and everywhere else it's used:

```suggestion
				fieldTag = fields[0].textContent.trim();
```

We eventually need to remove it - it was just a shim before browsers supported that built-in function.

Libraries Tasmania.js
```diff
@@ -0,0 +1,781 @@
...
+		let notes = getFieldText(doc, "NOTES_TALIS");
+		let combined = [summary, contents, notes].join("\n\n");
+		item.abstractNote = combined ? combined : "";
+		
+		// Physical description
+		let physDesc = getFieldText(doc, "DESC_TALIS");
+		let physParts = physDesc.split(";");
+		if (physParts.length == 2 && format == "book") {
+			item.numPages = cleanText(physParts[0]);
+			item["physical desription"] = cleanText(physParts[1]);
```
AbeJellinek, line 380: spelling

Libraries Tasmania.js
```diff
@@ -0,0 +1,781 @@
...
+		item = addPermalink(doc, item, null);
+		addDigitalFiles(doc, item);
+	}
+}
+
+function scrapeArchives(doc) {
+	var item = new Zotero.Item("manuscript");
+
+	// Types should be Agency, Series, or Item
+	let typeLabel = ZU.trim(doc.querySelector(".T245_DISPLAY_label").textContent).replace(/:$/, "");
```
AbeJellinek, line 395: Stuff like this should use `text(doc, ...)` instead of `querySelector(...).textContent` for null-safety:

```suggestion
	let typeLabel = ZU.trim(text(doc, ".T245_DISPLAY_label")).replace(/:$/, "");
```

Libraries Tasmania.js
```diff
@@ -0,0 +1,781 @@
...
+		"NI_NOMINATOR",
+		"NI_DOC_DATE",
+		"NI_FATHER",
+		"NI_P_OCCUP",
+		"NI_PRE_SCHOOL",
+		"NI_ADMISS_DATE"
+	];
+	// Add field values to Extras
+	for (let i = 0; i < fields.length; ++i) {
+		let label = doc.querySelector("div." + fields[i] + "_label");
```
AbeJellinek, line 557: Selector nitpick!

Libraries Tasmania.js
```diff
@@ -0,0 +1,781 @@
...
+		if (label) {
+			item[ZU.trim(label.textContent).replace(/:$/, "")] = getFieldText(doc, fields[i]);
+		}
+	}
+	item = addPermalink(doc, item, "NI");
+	addDigitalFiles(doc, item);
+}
+
+// Zotero.debug(item);
```
AbeJellinek, line 566: Can remove this

Libraries Tasmania.js
```diff
@@ -0,0 +1,781 @@
...
+					{
+						"lastName": "University of Tasmania (TA92)",
+						"creatorType": "contributor"
+					}
+				],
+				"date": "1890-01-19/1992-07-10",
+				"abstractNote": "Official minutes of meetings. | These records are part of the holdings of the Tasmanian Archives",
+				"callNumber": "AD940",
+				"libraryCatalog": "Libraries Tasmania",
+				"manuscriptType": "Series",
```
AbeJellinek, line 587: I don't know if these really belong in `manuscriptType`... Maybe a line in `extra`?

AbeJellinek (changes requested):
Some comments:

Libraries Tasmania.js
wragge: Where does `text()` come from?

AbeJellinek: From here: https://github.com/zotero/translate/blob/e08e318997b80685f03fbc4aae2cf6944fe46cfc/src/translation/translate.js#L1879-L1896

It's added to the translation sandbox for web and search translators. There's also `innerText`, `attr` (same idea but takes an attribute name as the last argument and returns the value of that attribute on the matched element, or the empty string), and now `request` (async HTTP).

Libraries Tasmania.js
wragge: The `.LOCAL_FORMAT` class is also used in the field label, but simplifying to `.displayElementText.LOCAL_FORMAT` seems ok.

Libraries Tasmania.js
wragge: Ok, done.

Libraries Tasmania.js
wragge: Ok, done.

Libraries Tasmania.js
wragge: Removed `div`

Libraries Tasmania.js
wragge: Corrected

Libraries Tasmania.js
wragge: Changed in this instance and in a couple of other places.

Libraries Tasmania.js
wragge: `div` removed

Libraries Tasmania.js
wragge: Done

Libraries Tasmania.js
wragge: These are the type of archival entity being described -- series, item, or agency -- so I think `manuscriptType` is the best place for it. Having it there also makes it much easier to browse/filter different types in the Zotero interface. This is also consistent with other translators including National Archives of Australia and State Records Office of WA.

adam3smith: I'm inclined to agree, but to make sure -- is this expected to be part of citations, typically? `manuscriptType` aka CSL's `genre` is used a fair amount in citation styles, so this would show up regularly. If that makes sense, I think this is definitely the right field. If it doesn't, I'd reconsider.

wragge: I've made a couple of changes. I've kept `manuscriptType` to indicate the archival entity type ('series', 'item' etc). This is consistent with other translators (those noted above plus the UK National Archives) and is most likely to be used in citations. I've also lowercased these values for consistency.

I was also using `manuscriptType` for the type of index for names index records. I've changed these so the `manuscriptType` is now 'item', and the index title is saved in 'Extras'.

AbeJellinek:
Ack. I'm really sorry I didn't get to this in time. Looks good to go. I am curious why some of the `publisher` and `label` fields aren't getting split into `place`, `publisher`/`label`, and `date`, since the regex for that looks good, but we can work on that later.

PR #2833, opened 2022-06-13
RIS - Thesis : add contributor field

stl74:
RIS format doesn't export contributor for thesis document type.
Needed to add thesis supervisor as contributor , added as "A2"


AbeJellinek:
Could we use A3 instead? That's what ProQuest Dissertations uses and it seems more appropriate.

adam3smith:
(RIS specs also say A3)

adam3smith:
Tests are just timing out -- merging this.

adam3smith:
Thanks!

Issue #2808, opened 2022-03-30
Use `preprint` in NBER, SSRN, OSF preprints, and bioRxiv


AbeJellinek:
https://forums.zotero.org/discussion/95881/change-nber-and-ssrn-translators-to-use-preprint

PR #2834, opened 2022-06-13
Update HighWire2 importer to correctly set preprint type on biorXiv/medrXiv items

meson800:
Previously, the HighWire2.0 translator was updated in #2808 in commit ba3cec3. However, due to the hard-coded `report` itemtype in `doWeb`, the updated translator code was never called for biorXiv and medrXiv, and thus was failing the biorXiv and medrXiv tests.

This PR updates the hardcoded itemtype in detectWeb for biorXiv and medrXiv to be the "preprint" type, which then passes the biorXiv and medrXiv tests and properly imports these items.

### DIFF
--- a/HighWire 2.0.js
+++ b/HighWire 2.0.js
@@ -409,7 +409,7 @@ function detectWeb(doc, url) {
 			&& !url.includes('/suppl/')
 		) {
 			if (url.includes('medrxiv.org') || url.includes('biorxiv.org')) {
-				return "preprint";
+				return preprintType;
 			}
 			else {
 				return "journalArticle";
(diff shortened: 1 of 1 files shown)


meson800:
Thanks for the suggestion! I didn't realize that the preprint type was added in Zotero 6.

adam3smith:
(not sure if CI is going to run correctly, so if this fails, probably don't worry about it)

dstillman:
(I think technically it will still use `report` for the Chrome connector, whereas the Firefox connector will use `preprint`. We'll be pushing a new version of the Chrome

PR #2846, opened 2022-07-06
Web translator for University of Chicago VuFind

bufordrat:
Here is a first stab at a web translator to work with UChicago's version of VuFind.  Our subject specialists and patrons have expressed a strong interest in being able to do accurate Zotero harvesting of catalog items that come up in our results pages, so we will be excited to have this up and running.  

This mostly follows the design of the Finna web translator, in that `doWeb` requests the MARC record for each item from VuFind, and then the bibliographic entry is built using Zotero's import translator for MARC.  The main difference is that this code also does a little "touch-up" on the output of the MARC translator, looking a few things up in the binary MARC data that are specific to University of Chicago catalog items, then adding them to the entry before finishing.

You can take a quick look at our instance of VuFind, which uses FOLIO as a back-end, here:
https://catalog.lib.uchicago.edu/vufind/

### DIFF
--- a/UChicago VuFind.js
+++ b/UChicago VuFind.js
@@ -208,6 +208,9 @@ const detectWeb = (doc, url) => {
 		else if (doc.querySelector('.format.dissertations')) {
 			return 'thesis';
 		}
+		else if (doc.querySelector('.format.archivesmanuscripts')) {
+			return 'manuscript';
+		}
 		else if (doc.querySelector('.format.audio')) {
 			return 'audioRecording';
 		}
--- a/UChicago VuFind.js
+++ b/UChicago VuFind.js
@@ -1,4 +1,4 @@
-[{
+{
 	"translatorID": "f9052879-ab4c-4056-84b4-e963dd98cb5d",
 	"label": "UChicago VuFind",
 	"creator": "Matt Teichman",
@@ -104,7 +104,7 @@ const scrapeMARC = doc => (url) => {
 	// custom UChicago tweaks to the return of the Zotero MARC translator
 	const customizeMARC = doc => item => (marc) => {
 		// put catalog URL in the entry
-		const addUrl = item => item.url = document.location.href;
+		const addUrl = item => item.url = doc.location.href;
 
 		// replace general call number with UChicago-internal call number
 		const updateCN = (item) => {
--- a/UChicago VuFind.js
+++ b/UChicago VuFind.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2022-07-26 20:06:21"
+	"lastUpdated": "2022-08-09 14:40:56"
 }
 
 /*
@@ -112,11 +112,12 @@ const scrapeMARC = (doc, url) => {
 			}
 		};
 
+		// if there's a 502 field, it should be a thesis
+		const isDissertation = marc => fieldExists('502', marc);
+
 		// correct errors in identifying dissertations, maps,
 		// manuscripts, and films
 		const fixItemType = (item) => {
-			// if there's a 502 field, it should be a thesis
-			const isDissertation = marc => fieldExists('502', marc);
 			// if the record type is 'p', it's a manuscript
 			const isManuscript = marc => marc.substring(6, 7) == 'p';
 			// if the item type is film, it's a film
@@ -143,10 +144,37 @@ const scrapeMARC = (doc, url) => {
 			}
 		};
 
+		// relocate 264|b of a dissertation to the 'archive'
+		// Zotero field
+		const archivePublisher = (item) => {
+			const publisher = lookupSubfields('264', 'b', marc);
+			if (isDissertation(marc) && publisher.length === 1) {
+				item.archive = publisher[0];
+			}
+		};
+
+		// make 710|a field into the publisher of a dissertation
+		const makeUniversityPublisher = (item) => {
+			// does 710|e say 'degree granting institution.'?
+			const degreeGranting = lookupSubfields('710', 'e', marc);
+			let isDG = false;
+			if (degreeGranting.length === 1) {
+				isDG = degreeGranting[0] == 'degree granting institution.';
+			}
+			// is there a 710|a MARC field?
+			const university = lookupSubfields('710', 'a', marc);
+			// if both conditions hold, 710|a becomes the publisher
+			if (isDissertation(marc) && isDG && university.length === 1) {
+				item.publisher = university[0];
+			}
+		};
+
 		// perform the fixes
 		addUrl(item);
 		fixItemType(item);
 		updateCN(item);
+		archivePublisher(item);
+		makeUniversityPublisher(item);
 	};
 	
 	// this part is based on the Finna translator code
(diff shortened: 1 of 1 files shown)


AbeJellinek:
> Zotero's ESlint

To be exact, run `npm run lint -- 'UChicago VuFind.js' --fix`. That will fix automatically fixable problems and output the rest that have to be fixed manually.

> I'm wondering if using MARC 502 to detect thesis isn't something we should do globally

Is that an official or at least unambiguous mapping? If so, absolutely.

UChicago VuFind.js
```diff
@@ -0,0 +1,794 @@
...
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2022-07-06 17:32:33"
+}
+
+// MARC retrieval code: run the MARC import translator, then perform a
+// few adjustments to the output by looking things up in the MARC
+// record
+
+// scrapeMARC function, overall design based on Finna translator
+const scrapeMARC = doc => url => {
```
AbeJellinek, line 20: Very Haskell :)

There's nothing technically incorrect about using currying like this, but it's a bit idiosyncratic for this codebase/JS in general. I haven't spotted any when quickly looking over the code, but are there any places here where we're taking advantage of e.g. partially applying these functions? If not, I think I'd argue that we should replace curried functions with traditional argument lists (`doc => url =>` -> `(doc, url) =>`) for clarity/ease of maintenance by others with a JavaScript background.

UChicago VuFind.js
```diff
@@ -0,0 +1,794 @@
...
+	}
+	// for any MARC field, return the length and starting position of
+	// the value
+	const lookup_in_directory = key => threes => {
+		const assocs = threes.filter(three => three[0] == key);
+		return assocs.map(x => x.slice(1, x.length));
+	}
+	// the data portion of a MARC record
+	const data_portion =
+		  table.substring(base_pos(table), table.length);
```
AbeJellinek, line 50: ditto with `substring`

UChicago VuFind.js
```diff
@@ -0,0 +1,794 @@
...
+		return 'thesis';
+	} else if (doc.querySelector('.format.audio')) {
+		return 'audioRecording';
+	} else if (doc.querySelector('.format.map')) {
+		return 'map';
+	} else {
+		return 'book';
+	}
+	// VuFind URL patterns starting with 'Search' are for search results
+	} else if (url.includes('Search')) {
```
AbeJellinek, line 199: Can we make these a little more specific? Like, if we're trying to match a URL that looks like `https://example.com/search?q=123`, `url.includes('/search?')` is a lot better and less likely to hit false positives than `url.includes('search')`.

UChicago VuFind.js
```diff
@@ -0,0 +1,794 @@
...
+				"itemType": "thesis",
+				"title": "Characterizing kinds: A semantics for generic sentences",
+				"creators": [
+					{
+						"firstName": "Matthew",
+						"lastName": "Teichman",
+						"creatorType": "author"
+					},
+					{
+						"lastName": "University of Chicago",
```
AbeJellinek, line 307: It looks like this should be in the `university` field and the current value in the `university field` (ProQuest) should be... maybe nowhere? Maybe in `archive`?

If this is a common issue, can we correct it automatically?

AbeJellinek (changes requested):
OK, some more comments! Great work overall.

bufordrat:
Thanks so much for these comments, which are totally awesome.  Sorry for the late response; I would have gotten back to you immediately but I had to stay home to take care of my sick kid.  Then she made me sick!  Anyway, most of these changes are totally straightforward, but I have a couple follow-up questions; see below.

bufordrat:
(or above?  :) GitHub is spatially confusing)

UChicago VuFind.js
bufordrat: Guilty as charged, your honor.  OCaml and Haskell are my main languages, and I also teach Haskell!  

The one place in the code where I make use of the fact that `scrapeMARC` is curried is in line 213:

https://github.com/zotero/translators/pull/2846/files#diff-0b26c1ec34be25953bbeb714f32ed65fd5d1a2ede776aa3b1eaa8a03c1910064R213

If you think it would be easier for your team to work with, I'd be happy to uncurry `scrapeMARC` and change line 213 around so that it didn't rely on partial application; just let me know what your preference is.

UChicago VuFind.js
```diff
@@ -0,0 +1,794 @@
...
+		const value_pos = parseInt(str.substring(7,12));
+		return [field, value_length, value_pos];
+		}
+		return twelves.map(process_entry);
+	}
+	// for any MARC field, return the length and starting position of
+	// the value
+	const lookup_in_directory = key => threes => {
+		const assocs = threes.filter(three => three[0] == key);
+		return assocs.map(x => x.slice(1, x.length));
```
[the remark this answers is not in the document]
bufordrat: Whoops!  Thanks.

UChicago VuFind.js
```diff
@@ -0,0 +1,794 @@
...
+	const a = Array.from(rowNodes);
+	let obj = {};
+	// extract information from each li element for output
+	const buildOutput = r => {
+	const linkElement = r.querySelector('.title.getFull');
+	const entry_url = linkElement.href;
+	const title = ZU.trimInternal(linkElement.textContent);
+	if (entry_url && title) { 
+		obj[entry_url] = title;
+	}
```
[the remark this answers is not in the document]
bufordrat: Yikes!  Great catch--thank you.

UChicago VuFind.js
bufordrat: So the rest of these changes, the linting, etc. are pretty straightforward, I can get to work on those right away.  This one is really interesting but it is also more of a rabbit hole, because it involves the MARC import translator.  I think I need to ask you a couple questions, and then you'll have a chance to look at them while I make the rest of your suggested changes.

Phenomenon No. 1: "University of Chicago" being listed as a secondary creator ("contributor") in addition to me being listed as the author.  This is definitely weird!  I believe it is the MARC import translator code that is making this happen.  One initial indication is that if you comment out line 155, the same contributor with the "last name" of University of Chicago continues to appear in the entry.  You can observe phenomenon No. 1 happening in other dissertations in our collection as well:

https://catalog.lib.uchicago.edu/vufind/Record/10168855
https://catalog.lib.uchicago.edu/vufind/Record/10773211
https://catalog.lib.uchicago.edu/vufind/Record/11715209

It looks like the MARC import translator is pulling "University of Chicago" from the 710|a MARC fields and adding it as a second author with 710|a as its "last name" and classified as a "contributor".  See here:

https://github.com/zotero/translators/blob/master/MARC.js#L409

Assuming that's right, how would you prefer to handle this?

dstillman: I'll let @AbeJellinek comment on the specifics, but just to clarify, Publisher and University are the same thing — the field is [just called University when it's a thesis](https://github.com/zotero/zotero-schema/blob/628358f04b967a7669063cc3713d6fe593750570/schema.json#L2257-L2271).

bufordrat: @dstillman thanks for the link to the schema; that explains the behavior perfectly!

AbeJellinek:
OK, thanks so much! Last couple small things:

1. Should `detectWeb` return `'manuscript'` if it finds `.format.archivesmanuscripts` on the page? That would make the last test succeed.
2. Can we have a test for a multiples page?

AbeJellinek:
(Search results)

UChicago VuFind.js
bufordrat: All the changes except a) the uncurrying of `scrapeMARC` and b) whatever we're going to do about the university as a contributor and ProQuest not being a university are now up on the branch.  Please feel free to ping me if I can help with other stuff or answer any other questions.

bufordrat:
> (Search results)

Sorry; didn't see these replies till just now (still getting used to where GitHub puts things).  Will work on these changes.

bufordrat:
>     1. Should `detectWeb` return `'manuscript'` if it finds `.format.archivesmanuscripts` on the page? That would make the last test succeed.

I was wondering why that test was failing!  Thanks for the explanation.

bufordrat:
Ok, your suggested changes (having `detectWeb` correctly identify manuscripts and adding a search results test case) should be in now.  Sorry about the delay; now that I know which part of the GitHub web interface to look at, I'll be keeping my eye on it for updates.

UChicago VuFind.js
AbeJellinek: How should these actually be cited? Is ProQuest as publisher (university) correct? If so, I think the way it works now is fine. If not, we probably need to fix the MARC translator.

adam3smith: No, ProQuest should definitely not be the University (nor CSL publisher). Here's the sample entry from bwiernik's APA library, I think following this (i.e. putting ProQuest into archive) makes sense: https://www.zotero.org/groups/2205533/items/UD3VZHS5

bufordrat: So since you brought this to our attention, our metadata analysts have been discussing the issue and think that this could be something to fix at the catalog level.  That is, for recent dissertations, ProQuest should not be listed as the publisher, as is the norm right now.  Actually doing the fix is somewhat logistically complicated, but they are looking into it and apparently there's interest.

In terms of this pull request, it's a little unclear to me what the fix would be at the level of the VuFind web translator code, but one thing I could at least do is replace that test case with a test case involving an older dissertation that lists the actual university as publisher.  For example, this one:

https://catalog.lib.uchicago.edu/vufind/Record/7143200

What do you think, would that be suitable from your point of view?

AbeJellinek: That would be fine, I think. Something like

> - check if the 710|e field is "degree granting institution"
> - if it is, change item.publisher to whatever's in the 710|a field
> - if not, change item.publisher to the empty string, and then nothing ends up populating the "university" field (or maybe you have a bibliographic field delete method somewhere in your project that could be used for that)

Also sounds fine for the time being, if that 710|e value is consistent.

AbeJellinek: (With @adam3smith's suggestion to put ProQuest in `archive`.)

bufordrat: Ok, these changes are now in.

UChicago VuFind.js
```diff
@@ -822,6 +812,95 @@ var testCases = [
...
+				"archive": "ProQuest Dissertations & Theses,",
+				"libraryCatalog": "UChicago VuFind",
+				"numPages": "1",
+				"place": "Ann Arbor",
+				"shortTitle": "Characterizing kinds",
+				"university": "University of Chicago.",
```
AbeJellinek, line 842: OK, really, really last thing. Can we strip the period off the end? To avoid messing with acronyms ("U.C.L."), how about:

```js
if (item.university) {
	item.university = item.university.replace(/([a-z])\.\s*$/, '$1');
}

AbeJellinek (changes requested):

UChicago VuFind.js
AbeJellinek: Actually, done :)

AbeJellinek:
Thank you so, so much for your excellent work on this and sorry for the somewhat protracted review!

bufordrat:
It was a pleasure working through it with you!

> Thank you so, so much for your excellent work on this and sorry for the somewhat protracted review!

PR #2847, opened 2022-07-08
Fix Proquest

### LINKED ISSUE #2839: Fix EEBO in ProQuest
There's no Abstract/Details tab, so both detect and doWeb fail

Reported:
https://forums.zotero.org/discussion/97741/zotero-not-scraping-proquest-properly#latest

adam3smith:
Closes #2839
Addresses error reported at https://forums.zotero.org/discussion/comment/411799#Comment_411799

### DIFF
--- a/ProQuest.js
+++ b/ProQuest.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2022-01-04 09:11:57"
+	"lastUpdated": "2022-07-08 14:27:35"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


dstillman:
Thanks!

owcz:
This commit broke the main changes fixed in #2763
For example, https://www.proquest.com/docview/326733957/7047FD799CEF423FPQ/11 fails and goes to Embedded Metadata (i.e., incorrect metadata) under this new commit, whereas I just manually reverted to the prior commit and the citation doesn't fail. I checked [the diff](https://github.com/zotero/translators/compare/e7f32884f9b966e6d050f3f58297f81dae4b40cd..541ccea85938e13c753e1adfa8b6532a18f4ff7a) but it looks like there is a lot going on.

adam3smith:
Sorry, thanks for catching that. Fixing momentarily (running translator tests on proxy doesn't work :( )

PR #2848, opened 2022-07-10
Fixes regression introduces in #2847

adam3smith:
FYI @owcz


adam3smith:
@owcz -- I tested this with both old & new layouts, but would appreciate a sanity check when you have a minute

PR #2851, opened 2022-07-14
RIS: Fail gracefully on 0000 in date field

adam3smith:
See https://forums.zotero.org/discussion/comment/412372#Comment_412372 for details

RIS is currently unlinted, that should probably just happen separately with >200 current errors...

### DIFF
--- a/RIS.js
+++ b/RIS.js
@@ -17,7 +17,7 @@
 	},
 	"inRepository": true,
 	"translatorType": 3,
-	"lastUpdated": "2022-07-14 08:59:07"
+	"lastUpdated": "2022-07-14 15:44:17"
 }
 
 function detectImport() {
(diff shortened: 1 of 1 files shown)


RIS.js
```diff
@@ -6968,6 +6967,82 @@ var testCases = [
...
+					{
+						"note": "<p>Export Date: 13 July 2022</p>"
+					}
+				],
+				"seeAlso": []
+			},
+			{
+				"itemType": "journalArticle",
+				"creators": [],
+				"date": "0000",
```
AbeJellinek, line 7031: Do we even want to save these 0000 dates, though? Might be better to just skip all-zero date segments.

adam3smith: I guess there's no upside to the 0000 -- I based keeping it on the MM/DD case above, but that's not actually the same since we add the year in order to keep the other info there. I'll look at removing it

adam3smith:
How is this instead?

Issue #2819, opened 2022-04-27
Project Muse automatically download all chapters of a book


boan-anbo:
I was wondering if the Project Muse translator could stop downloading all pdfs attachments for books automatically. 

https://github.com/zotero/translators/blob/7aeaea1d12e5e4e139c282e53f9550d64044c224/Project%20MUSE.js#L115

Several reasons for the change (take [this book for example](https://muse.jhu.edu/book/98273)):

1. Some books have Epub format for the entirety of books and the user might not want individual pdf chapters.
2. Zotfile cannot handle multiple pdfs (each for a chapter) for a single entry of book. This causes only the first pdf to be treated as the default attachment and the others to be handled by Zotero (and saved in a separate Zotero location for instance)
3. In my personal experience, the user might want to download all the chapters themselves and combine the pdfs before attach the combined pdf to the entry. 
4. This might cause exceeding the download rate set for Project Muse because it's not treated as a Zotero translator collection of items, where the user get to specify which to download. A single click will trigger all pdfs to be downloaded.

So I suggest remove the the functionality for book but keep it for collections for Project Muse journals.

If this is Ok, I can create a PR.

PR #2853, opened 2022-07-27
Remove lines for automatically download all PDF attachments

boan-anbo:
#2819

### DIFF
--- a/Project MUSE.js
+++ b/Project MUSE.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2021-05-18 20:04:42"
+	"lastUpdated": "2022-07-27 00:49:29"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


AbeJellinek:
Could you update the tests? Otherwise this looks good, thanks.

boan-anbo:
> Could you update the tests? Otherwise this looks good, thanks.

Oops! Just updated. Thanks Abe!

AbeJellinek:
Oh, wait, this removes _all_ attachments. How about we remove all attachments if there are multiple but keep if there's only one ("Full Text PDF")?

Project MUSE.js
```diff
@@ -111,22 +111,13 @@ function scrape(doc) {
...
-			for (let card of cards) {
-				let url = attr(card, 'a[href*="/pdf"]', 'href');
-				if (!url) continue;
-				item.attachments.push({
-					url,
-					title: text(card, '.title') || "Full Text PDF",
-					mimeType: 'application/pdf'
-				});
-			}
 			item.complete();
```
AbeJellinek, line 114: As in:

```suggestion
			// Don't download long lists of attachments (like all chapters in a book)
			if (item.attachments.length > 1) {
				item.attachments = [];
			}
			item.complete();
```

AbeJellinek (changes requested):

PR #2855, opened 2022-07-31
fixed: airiti getDocId function

cesaryuan:
fixed airiti cannot getDocId with lowercase url paramters

### DIFF
--- a/Airiti.js
+++ b/Airiti.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 12,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2022-05-11 00:55:12"
+	"lastUpdated": "2022-07-31 16:14:21"
 }
 
 function detectWeb(doc, url) {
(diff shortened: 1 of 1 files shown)


Airiti.js
```diff
@@ -20,13 +20,14 @@ function detectWeb(doc, url) {
...
 function getDocId(url) {
-	var m = url.match(/\bDocID=([^&#]+)/);
+	var m = url.match(/\bDocID=([^&#]+)/i);
 	if (!m) return;
 	return m[1];
 }
 
 async function doWeb(doc, url) {
 	var docID = getDocId(url);
+	Z.debug(`doWeb: ${docID}`);
```
AbeJellinek, line 30: This should be removed

Airiti.js
```diff
@@ -54,6 +55,7 @@ async function scrape(docIDs, itemDoneHandler) {
 	await requestDocument('https://www.airitilibrary.com/Publication/MetaExport?' + tokenParams);
 
 	var bibTeXUrl = buildQuery(docIDs);
+	Z.debug('bibTeXUrl: ' + bibTeXUrl);
```
AbeJellinek, line 58: And this

AbeJellinek (changes requested):
Looks good, just remove the debug calls!

Airiti.js
cesaryuan: Thank you. I have removed the calls.

AbeJellinek:
Thanks!

PR #2856, opened 2022-08-01
Update NLA Catalogue translator to save urls for digitised items

wragge:
The NLA catalogue includes links to digitised versions of items where available in the 856 MARC field. This PR saves this url to the Zotero item if the z subfield of 856 starts with 'National Library of Australia digitised item'. 

I also fixed up a few linting errors and added a new test.

### DIFF
--- a/National Library of Australia (new catalog).js
+++ b/National Library of Australia (new catalog).js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2022-08-01 05:49:00"
+	"lastUpdated": "2022-08-02 21:35:50"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


National Library of Australia (new catalog).js
```diff
@@ -114,7 +109,19 @@ function scrapeMarc(text) {
...
+	translator.getTranslatorObject(function (obj) {
+		var record = new obj.record();
+		// Populate the record from the MARC
+		record.importBinary(text);
+		var item = new Zotero.Item();
+		// If there is a url for the digitised item add it
+		var url = record.getFieldSubfields("856");
+		if (url && url.length > 0 && url[0].u && url[0].z && url[0].z.startsWith("National Library of Australia digitised item")) {
+			item.url = url[0].u;
+		}
```
AbeJellinek, line 121: `getFieldSubfields()` always returns an array (but it might be empty), so I think this could be simplified to:

```suggestion
		var [url] = record.getFieldSubfields("856");
		if (url && url.u && url.z && url.z.startsWith("National Library of Australia digitised item")) {
			item.url = url.u;
		}
```

AbeJellinek (changes requested):
Looks good! One suggestion.

National Library of Australia (new catalog).js
wragge: That's a neat trick to remember! Done.

PR #2858, opened 2022-08-07
index.d.ts: fix HTTPRequestParameters for Zotero.Utilities.request

aGolduck:
requestHeaders should be headers, see https://github.com/zotero/zotero-connectors/blob/master/src/common/http.js#L68


AbeJellinek:
Thank you!

PR #2861, opened 2022-08-18
Update EBSCOhost.js

troykozee:
Added src_ic to regex to make translator compatible with EBSCO's "Explora" product.

My limited testing shows that this works, but I'd be happy to run through any test suite that you have.

### DIFF
--- a/EBSCOhost.js
+++ b/EBSCOhost.js
@@ -2,7 +2,7 @@
 	"translatorID": "d0b1914a-11f1-4dd7-8557-b32fe8a3dd47",
 	"label": "EBSCOhost",
 	"creator": "Simon Kornblith, Michael Berkowitz, Josh Geller",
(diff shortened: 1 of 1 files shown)


AbeJellinek:
We unfortunately aren't able to have tests for EBSCOhost. This looks fine if you've tested locally and it works, but `dateModified` needs to be updated.

troykozee:
Updated modification date.

Further testing revealed the following

- Previous functionality on Ebsco-branded sites appears unchanged
- Results on Explora are "equivalent" when storing to Zotero from an article detail page (even from Ebsco's PDF viewer)
- Two issues when using the plugin from search results page
1. Translator will not download PDF attachments when article(s) are selected from the Explora results page. Cause unknown.
2. Some Explora search results have a "topic box." The translator can't parse this and Zotero falls back to the DOI plugin.

I don't know if these are show stopping issues for this pull request. I don't have time to work on them this moment, but can in the future.

PR #2873, opened 2022-08-24
add web translator for Index Theologicus and Relbib

socheres:


socheres:
@AbeJellinek I create a translator base on UChicago VuFind. Could you please have a look on this? Thanks in advance.

AbeJellinek:
I think you committed a duplicate file. Or are there supposed to be two?

adam3smith:
Also -- how closely do these mirror the UChicago translator? I think we'd want to try to avoid having custom translators for every VuFind flavor.

AbeJellinek:
Yeah, agreed. And we have #2852 open too, so that's three separate translators for VuFinds. I think we might end up wanting to unify them into a translator that uses MARC ("Staff View") if available, falling back to Refer/BibIX (Export -> Export to EndNote) if *that's* available, and failing to detect if neither is.

socheres:
> I think you committed a duplicate file. Or are there supposed to be two?

I changed the file name and changed fixed some errors and committed again. The last committed file is the correct one for both databases IxTheo and Relbib.

socheres:
> Yeah, agreed. And we have #2852 open too, so that's three separate translators for VuFinds. I think we might end up wanting to unify them into a translator that uses MARC ("Staff View") if available, falling back to Refer/BibIX (Export -> Export to EndNote) if _that's_ available, and failing to detect if neither is.

It would be nice, if there is a translator for different versions of vufind pages. Maybe with your suggestion (marcview > ris or bibtex export format as fallback) this could be done. Until then, it would be good if the ixtheo/relbib translator could be included in the official version. But if the general translator for vufind can be written in the foreseeable future, I can wait until then.

PR #2874, opened 2022-08-25
Add translator for VuFind

AbeJellinek:
Based on (and replaces) UChicago VuFind.

Supersedes #2873 and #2852.


socheres:
@AbeJellinek Could you please also to consider adding the below code block for volume, issue, pages for item.journals? Thanks.

```
		// add 773|g field for year, issue, volume
		const addVolumeYearNumberPages = (item) => {
			const subfieldString = lookupSubfields('773', 'g', marc);
			if (subfieldString.length > 0) {
				for (i in subfieldString) {
					if (!item.volume && subfieldString[i].match(/volume:\d+/gi) !== null) {
						item.volume = subfieldString[i].split(':')[1];
					}
					// IxTheo: 773|g value for month equal with issue volume e.g. "773 1 8 |g volume:118  |g year:2022 |g month:02 |g pages:1-2" 
					if (!item.issue && subfieldString[i].match(/(number:\d+|month:\d+)/gi) !== null) {
					item.issue = subfieldString[i].split(':')[1].replace(/^0/, '');
					}
					if (!item.pages && subfieldString[i].match(/pages:\d+/gi) !== null) {
					item.pages = subfieldString[i].split(':')[1];
					}
				}
			}
		};
```

VuFind.js
```diff
@@ -0,0 +1,1804 @@
...
+			}
+		]
+	},
+	{
+		"type": "web",
+		"url": "http://bdtd.ibict.br/vufind/Record/PUC_RIO-1_5e711c01a0c4745a4c27c1750ecfc724",
+		"items": [
+			{
+				"itemType": "book",
+				"title": "A DEPENDENCY TREE ARC FILTER",
```
adam3smith, line 972: Could we fix all caps in titles & authors? (See e.g. https://github.com/zotero/translators/pull/2876 but we could also consider building a `ZU` function that does this.)

VuFind.js
```diff
@@ -0,0 +1,1804 @@
...
+						"lastName": "Saalmann",
+						"creatorType": "author"
+					}
+				],
+				"date": "2020",
+				"ISBN": "9783956507083",
+				"callNumber": "306.6",
+				"language": "ger",
+				"libraryCatalog": "ixtheo.de",
+				"numPages": "1",
```
adam3smith, line 1016: what's going on here? This is volume 1, I don't see any reference to numPages .

VuFind.js
```diff
@@ -0,0 +1,1804 @@
...
+		]
+	},
+	{
+		"type": "web",
+		"url": "http://bdtd.ibict.br/vufind/Search/Results?lookfor=trees&type=AllFields&limit=20&sort=relevance",
+		"items": "multiple"
+	},
+	{
+		"type": "web",
+		"url": "https://ixtheo.de/Record/1698984863",
```
adam3smith, line 999: This item does have a DOI -- any way we can capture that from the metadata? (it's in the RIS and I also get it when importing the MARC directly)

VuFind.js
```diff
@@ -0,0 +1,1804 @@
...
+						"creatorType": "author"
+					}
+				],
+				"date": "2022",
+				"DOI": "10.5334/snr.151",
+				"ISSN": "2053-6712",
+				"abstractNote": "In the United States, nonreligious people face stigma, prejudice, and discrimination because they are viewed as immoral and distrustful. This is partly because of othering, by which nonreligious people are subjugated to a minority status. Othering also occurs in academic research and writing. Applying feminist principles can improve research about nonreligious populations. Grounded in results of a US-based online study, we recommend two feminist principles to facilitate the study of nonreligion: (1) rejecting othering of minority groups, and (2) intersectionality. As a result of applying these principles, the nuanced differences between nonreligious groups can be better understood and the complex identities of nonreligious people can be more accurately represented. Researchers benefit from increased accuracy and understanding of nonreligion via better informed theoretical and methodological decisions and nonreligious people benefit from their more accurate representation in academic research",
+				"callNumber": "0",
```
adam3smith, line 1135: Not sure if we can avoid this?

VuFind.js
```diff
@@ -0,0 +1,1804 @@
...
+		]
+	},
+	{
+		"type": "web",
+		"url": "https://ixtheo.de/Record/1775609642",
+		"items": [
+			{
+				"itemType": "book",
+				"title": "Vielfalt und Gemeinsinn: der Beitrag der evangelischen Kirche zu Freiheit und gesellschaftlichem Zusammenhalt: ein Grundlagentext der Kammer für Öffentliche Verantwortung der EKD",
+				"creators": [
```
adam3smith, line 1236: This might be a MARC issue, but we shouldn't import the same author twice.

adam3smith (commented):
Some quick notes -- thanks for doing this!

mathieugrimault:
Is this translator supposed to target every VuFind instance ? 

Boosting priority shouldn't shadow other specific translator ? If they exist, they are probably more accurate and if not, they need to be removed. 

I think this translator is too specific to be labeled 'VuFind' : not every VuFind instance is using MARC21 (UNIMARC user here) and while it is well handled in MARC.js, we can see here some specific code related to MARC21. And some code specific to UChicago.

As VuFind can be heavily modified, I can't think of only one translator but a low-priority standard one (this one without all specifics parts) and each catalog would implement their translator with a higher priority ?

PR #2876, opened 2022-08-29
Fix all caps titles & authors in JSTOR

adam3smith:
Related post: https://forums.zotero.org/discussion/98444/feature-request-title-case-for-author-name#latest

### DIFF
--- a/JSTOR.js
+++ b/JSTOR.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2022-07-15 14:04:13"
+	"lastUpdated": "2022-08-29 13:56:41"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


JSTOR.js
```diff
@@ -246,6 +246,19 @@ function processRIS(text, jid) {
 		item.url = item.url.replace('http:', 'https:'); // RIS still lists http addresses while JSTOR's stable URLs use https
 		if (item.url && !item.url.startsWith("http")) item.url = "https://" + item.url;
 		
+		// remove all caps from titles and authors.
+		for (i = 0; i < item.creators.length; i++) {
+			if (item.creators[i].lastName && item.creators[i].lastName == item.creators[i].lastName.toUpperCase()) {
+				item.creators[i].lastName = ZU.capitalizeTitle(item.creators[i].lastName.toLowerCase(), true);
```
AbeJellinek, line 252: Any reason for this instead of `capitalizeName()`?

AbeJellinek (changes requested):

JSTOR.js
adam3smith: I don't know `capitalizeName()` -- do you have a link to definition or code? I couldn't find either a generic js function nor something in Zotero

dstillman: Zotero.Utilities functions are all in a separate repo now:

https://github.com/zotero/utilities/blob/48d5e170545acaa9127612d6d4337c032864b564/utilities.js#L121-L141

PR #2878, opened 2022-09-06
EM: Don't save a snapshot if we save a PDF

AbeJellinek:
\+ Update tests.


adam3smith (approved):
This is great -- I forgot we have this up, just implemented this manually in a translator relying on EM

adam3smith:
I'd suggest merging once you have the conflict resolved.

PR #2882, opened 2022-09-12
AustLII translator updates

jpwarren:
A few updates to the AustLII translator to more closely align with the _Australian Guide to Legal Citation_, 4th Edition.

- Major jurisdictions are now supported and abbreviated.
- Support for Report Series abbreviations, with some unauthorised report abbreviations supplied.
- Statute names (called Acts here) more properly parsed, and the year put into dateEnacted where it should be.
- Act names are Title Cased, with support for Act Name (Parenthetical Short Name) Acts.
- Additional test cases and updated tests to match the added/changed features.

### DIFF
--- a/AustLII and NZLII.js
+++ b/AustLII and NZLII.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2022-09-12 12:09:14"
+	"lastUpdated": "2022-09-12 13:05:47"
 }
 
 /*
@@ -37,7 +37,7 @@
 
 
 // attr()/text() v2
-function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null;}function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null;}
+// function attr(docOrElem,selector,attr,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.getAttribute(attr):null;}function text(docOrElem,selector,index){var elem=index?docOrElem.querySelectorAll(selector).item(index):docOrElem.querySelector(selector);return elem?elem.textContent:null;}
 
 
 function detectWeb(doc, url) {
@@ -61,6 +61,7 @@ function detectWeb(doc, url) {
 	if (getSearchResults(doc, true)) {
 		return "multiple";
 	}
+	return false;
 }
 
 
@@ -68,7 +69,7 @@ function getSearchResults(doc, checkOnly) {
 	var items = {};
 	var found = false;
 	var rows = doc.querySelectorAll('#page-main ul>li>a');
-	for (let i=0; i<rows.length; i++) {
+	for (let i = 0; i < rows.length; i++) {
 		let href = rows[i].href;
 		let title = ZU.trimInternal(rows[i].textContent);
 		if (!href || !title) continue;
@@ -93,10 +94,13 @@ function doWeb(doc, url) {
 				articles.push(i);
 			}
 			ZU.processDocuments(articles, scrape);
-		});
+			return true;
+		}
+		);
 	} else {
 		scrape(doc, url);
 	}
+	return false;
 }
 
 /*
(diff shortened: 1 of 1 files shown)


AustLII and NZLII.js
```diff
@@ -94,37 +89,124 @@ function doWeb(doc, url) {
...
+ * The Australian Guide to Legal Citation covers this in Rule 2.2.3
+ * legal Reports in Appendix A, but I've not been able to locate
+ * a copy with the appendix intact yet.
+*/
+function abbrevCourt(fullname) {
+	var courtMap = new Map();
+	courtMap.set('Federal Court of Australia', 'FCA');
+	courtMap.set('High Court of Australia', 'HCA');
+	courtMap.set('Family Court of Australia', 'FamCA');
+	courtMap.set('Australian Information Commissioner', 'AICmr');
```
AbeJellinek, line 115: Keep this (and the `jMap` below) outside the function so we don't need to rebuild every time?

AustLII and NZLII.js
```diff
@@ -94,37 +89,124 @@ function doWeb(doc, url) {
...
+	var lastWordIndex = words.length - 1;
+	for (var i = 0; i <= lastWordIndex; i++) {
+		if (actNameDelimRegex.test(words[i])) {
+			newString += words[i];
+		}
+		else {
+			newString += ZU.capitalizeTitle(words[i].toLowerCase(), true);
+		}
+	}
+	Zotero.debug(newString);
```
AbeJellinek, line 168: No need to debug here

AbeJellinek (changes requested):
Thanks! Some comments.

jpwarren:
Sorry, put it down for a moment and then… months went by somehow.

Made the requested changes.

The github action for the CI appears to have failed partway through due to what looks like a misconfiguration of which Chrome driver it's looking for, not something to do with this PR specifically.

AustLII and NZLII.js
```diff
@@ -94,37 +89,124 @@ function doWeb(doc, url) {
...
+jMap.set('Western Australia', 'WA');
+
+function abbrevJurisdiction(fullname) {
+
+	var abbrev = jMap.get(fullname);
+	if (abbrev === undefined) {
+		abbrev = fullname;
+	}
+	return abbrev;
+}
```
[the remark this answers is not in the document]
jpwarren: Okay.

I've kept them as separate maps/dictionaries to avoid key collisions, rather than having a single jMap object.

jpwarren:
This is what I get for mis-clicking in the website.

I am unsure of how to do the rebase safely. Advice would be appreciated so I don't make things worse as I fumble about trying to figure it out.

LawData-user:
Hi 
Thanks for the work on this translator.  I hope this is the appropriate place for this feedback.  
Unfortunately I am not able to code (except by asking ChatGPT to do it), however there are a few small functional changes I would like to suggest, which would really improve this translator:
1. In the 'Extra' field it adds an entry 'Code: [jurisdiction abbreviation]'.  I don't know what this is for, but it results in citations being inserted in Word that don't comply with the citation standards for Australia (AGLC) as it adds this information in front of the Media neutral court name, eg '[2023] Cth HCA 30' which should read '[2023] HCA 30'.
2. If I also add the citation for the Authorised version of the case (for HCA it is the Commonwealth Law Reports (CLR)) then the citation also becomes messy, giving '(2023) 220 CLR HCA 29' rather than '(2023) 220 CLR 110; [2023] HCA 29'.
3. The translator only works on Austlii and not Classic AustLII, it should be an easy fix as the 'classic version' just uses classic.austlii... instead of www.austlii...  Otherwise, I think that the site is identical.  Lots of people use the classic interface as it makes better use of screen space, even if it isn't so pretty.
4. It would be great if you would have it add the pdf or rtf file as an attachment as many people read the case in Zotero and highlight (and even add inked comments), before using the tool to extract the highlighted passages (if it is a pdf) to a Note field.  If it is an rtf at least it is downloaded and can be printed from word to pdf and then added). I am unaware of an alternative workflow from a webpage that is convenient.

Thank you again for your work.
Regards

LawData-user:
Hi,  Can I reiterate what the last person said about the problems with the insertion of the jurisdiction abbreviation in the Extra field. It gets inserted by the AGLC citation style and it is just wrong. 

I have to manually delete the "Code: Cth" every time I add a case to Zotero using the translator, as well as deleting common court names and inserting the abbreviations that appear in the same place on every AustLII page.

I am writing a thesis at present and the problems with the translator and AGLC are making it really tedious 

If I can help fix this in somme way please let me know what I can do.

thanks in advance

PR #2885, opened 2022-09-20
Fix CSL JSON import for new CSL types

adam3smith:
See https://forums.zotero.org/discussion/99814/import-citations-with-html-italics-tags#latest


AbeJellinek:
Thanks!

PR #2893, opened 2022-10-01
Update Clinical Key.js

igravious:
Add a newline `\n` after the metadata block to make `Clinical Key.js` compatible with every other translator.


dstillman:
(Happy to accept, but I wouldn't rely on there being two newlines in any parsing code. It's not a requirement.)

igravious:
OK – gotcha. Thanks.

PR #2895, opened 2022-10-03
Fix typo in FreePatentsOnline attachment

nikosavola:
This PR fixes one typo in `FreePatentsOnline .js`, _Snaptshot_ $\rightarrow$ _Snapshot_


adam3smith:
Please also update the timestamp, otherwise the translator won't update for users

nikosavola:
Linting appears to fail but this is clearly not caused by the changes implemented here. Is this okay?

dstillman:
Yes, that's no problem. Thanks!

PR #2896, opened 2022-10-04
Fix EV for new records page

adam3smith:
Reported: https://forums.zotero.org/discussion/100118/compendex-engineering-village-records-import-as-web-pages#latest


dstillman:
Thanks!

PR #2899, opened 2022-10-05
Add Duke University Press Books

adam3smith:
Journals are covered by Silverchair. Requested https://forums.zotero.org/discussion/100066/clipping-biblio-info-from-university-press-websites/p1


adam3smith:
@AbeJellinek thanks for that, didn't know about `filter(Boolean)` -- this one should be ready to merge, too

PR #2907, opened 2022-10-15
Some MODS improvements for govinfo

adam3smith:
Reported here: https://forums.zotero.org/discussion/comment/419592#Comment_419592 (kind of)
There's a lot more we could do, I've focussed here on Bills and info relevant for Bluebook citations

### DIFF
--- a/MODS.js
+++ b/MODS.js
@@ -14,7 +14,7 @@
 	},
 	"inRepository": true,
 	"translatorType": 3,
-	"lastUpdated": "2021-07-14 23:47:43"
+	"lastUpdated": "2022-10-15 01:42:49"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


adam3smith:
@AbeJellinek -- looks like you did most of this in https://github.com/zotero/translators/commit/a23df534c770cdd7621162a0065989f80660eb14

Do you want to see if I'm adding anything that you haven't already and otherwise close?

adam3smith:
@AbeJellink -- ready to merge, I think. I kept the contributors out: generally agree that we can put things into Contributor (and they indeed aren't cited), but just putting the general Congress info there just seems like garbage that doesn't need to be saved.

PR #2911, opened 2022-10-25
Add abstract to various additional Atypon Sites

adam3smith:
Addresses https://forums.zotero.org/discussion/100405/science-pnas-translators-not-pulling-in-abstract-field#latest


dstillman:
Thanks!

PR #2921, opened 2022-11-09
Update Scopus for new interface

adam3smith:
Reported: https://forums.zotero.org/discussion/100981/scopus-translator-needs-an-update#latest
This will work with both now


adam3smith:
The linting issues are all from @AbeJellinek 's `requestDocument` update of the translator, wasn't sure how to address those.

PR #2929, opened 2022-11-18
Zotero RDF: Encode ? and # in filenames, always output file URIs

### LINKED ISSUE #2928: Zotero RDF: Encode special characters in file path
https://forums.zotero.org/discussion/101210/bug-zotero-wont-import-all-attachments-from-zotero-rdf-xml

`#` and other valid characters

We're properly importing, e.g., `%23`, but we export `#`, which then won't import.

AbeJellinek:
Fixes #2928

### DIFF
--- a/Zotero RDF.js
+++ b/Zotero RDF.js
@@ -519,10 +519,21 @@ function generateItem(item, zoteroType, resource) {
 }
 
 function makeValidFileURI(path) {
-	path = path.replace(/#/g, '%23').replace(/\?/g, '#3F');
+	path = path
+		.replace(/#/g, '%23')
+		.replace(/\?/g, '#3F')
+		.replace(/ /g, '%20')
+		.replace(/\\/g, '/');
+
+	// Absolute paths
 	if (path.startsWith('/')) {
 		return 'file://' + path;
 	}
+	// Absolute Windows paths with drive letter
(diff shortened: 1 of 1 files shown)


Zotero RDF.js
```diff
@@ -518,6 +518,16 @@ function generateItem(item, zoteroType, resource) {
 	if (displayTitle) Zotero.RDF.addStatement(resource, n.z+"displayTitle", displayTitle, true);
 }
 
+function makeValidFileURI(path) {
+	path = path.replace(/#/g, '%23').replace(/\?/g, '#3F');
+	if (path.startsWith('/')) {
+		return 'file://' + path;
```
dstillman, line 524: [Apparently](https://en.wikipedia.org/wiki/File_URI_scheme#Windows_2) on Windows a valid path would be `file:///c:/path/to/the%20file.txt`, with three slashes. We'd also presumably need to replace backslashes with slashes. And maybe all the other things?

Did you test that we actually import file URLs properly? And was my guess right that we handle absolute paths now as a hack? I guess I'd worry that by making these neither valid file URLs nor paths, we'll break importing altogether, because it will throw an error trying to parse it as a URL but then not be a valid path either.

We'd want to add some tests for this afterwards, at least — we have some RDF tests in importExportTest.js in zotero/zotero already.

PR #2931, opened 2022-11-21
EM: Skip adjacent repeated authors (for Optica)

AbeJellinek:
FWIW I think this is really an issue on Optica's end - repeating the same `citation_author` with different `citation_author_institution`s is kind of weird - but I think this fix is unlikely to cause problems on other sites and could potentially fix other issues we haven't caught yet. So it seemed OK to just put it in EM.

https://forums.zotero.org/discussion/comment/422712/#Comment_422712

### DIFF
--- a/Embedded Metadata.js
+++ b/Embedded Metadata.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2022-01-03 16:51:36"
+	"lastUpdated": "2022-11-21 21:41:54"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


Embedded Metadata.js
```diff
@@ -596,6 +597,11 @@ function processHighwireCreators(creatorNodes, role, doc) {
 			// skip empty authors. Try to match something other than punctuation
 			if (!creator || !creator.match(/[^\s,-.;]/)) continue;
 
+			// Skip adjacent repeated authors
+			if (lastCreator && creator == lastCreator) continue;
```
adam3smith, line 601: the only scenario in which I could see authors being repeated are different roles (i.e. editor & author being the same person, say in the intro to an edited volume). But `role` is inherited from the function call, so that doesn't seem relevant here.

adam3smith (approved):
I assume you ran the other tests & they continue to work (my Scaffold is broken, so I can't...).
Otherwise looks good -- agree that this is a bad idea data-wise, but I think it's not unlikely to come up as publishers take good data in more appropriate formats like JSON and convert it to limited metatags like this, so agree this makes sense in EM.

adam3smith:
> What's going on with it? If you have an error log, I'll see if I can debug.

https://forums.zotero.org/discussion/101275/beta-scaffold-not-showing-loaded-translators#latest

PR #2943, opened 2022-12-07
Added language Label support in Wikidata Quickstatements.js

Futur3r:


Wikidata QuickStatements.js
```diff
@@ -260,11 +259,13 @@ function zoteroItemToQuickStatements(item) {
 
 	if (item.language && (item.language.toLowerCase() in languageMapping)) {
 		let lang = item.language.toLowerCase();
+		addStatement('L' + lang, '"' + item.title + '"');
 		addStatement('P1476', lang + ':"' + item.title + '"');
 		addStatement('P407', languageMapping[lang]);
 	}
 	else {
 		// otherwise use "und" for undetermined language
+		addStatement('Len', '"' + item.title + '"');
```
adam3smith, line 268: update the comment here, please: you're adding English for undetermined labels, right? Is there any reason that undetermined cant be used (I'm fairly ignorant about QS, so this may be obvious)

adam3smith (changes requested):
One question. Also, please update the timestamp; otherwise the translator won't update for users

Wikidata QuickStatements.js
Futur3r: I've updated.
The translator was already adding the label in English without taking into account the value in the 'language field' of a Zotero item.

I changed that, but if it still doesn't recognize the language value, it will add the label in English by default. It's better than nothing.

Futur3r:
I agree, they learn the hard way. The thing is that the language value, right now, just work for properties like "title" and "language of the work", not the label of the wikidata element which is defined by "Lxx" (Len, Lfr, Lde, etc.).

Wikidata doesn't expect  ISO 639-1 codes, neither do quickstatements. QS expect a QID for "language of work", a 'value' in  ISO 639-1 at the start of the string for the "title" property (i.e fr:"xxx", en:"xxx", etc.) and a label language with Lxx as stated before. 

If it's not in  ISO 639-1, like it's already doing, it adds "und:" at the start of the string of the "title" value and nothing else.
And QS doesn't like that, it'll fail with an error.
Right now, even if you have the language value properly set in  ISO 639-1, it always adds the label of the wikidata element in English. And like I'm doing every time, you have to manually change it in a spreadsheet !
BTW, the description of the element is also always in English regardless of the language value, but changing that requires way more time, time that I don't have now.

That's why I put the label in English by default in case the language value isn't set.

PR #2952, opened 2023-01-01
Don't capture history page in Wikipedia sites

yfdyh000:


yfdyh000:
It seems that CI Lint is not suitable for [multiple files edited](https://github.com/zotero/translators/pull/2940)?

BTW, the lint tool and Zotero build steps is very uncomfortable for Windows OS, maybe this is a known issue?

yfdyh000:
> Do we actually need the `.eslintrc` change?

CI will fail without the change.
https://github.com/zotero/translators/commit/2c95b178ea1ab8196feda4371fef035c8b01b425
https://github.com/zotero/translators/actions/runs/3815251105/jobs/6490124566#step:12:12
`Error:   52:10  error  'URLSearchParams' is not defined  no-undef`

yfdyh000:
> See also [65ab6eb](https://github.com/zotero/translators/commit/65ab6eb2fb30a6b83049b7ff10a44df781d445ff) (which was supposed to address this) and #2932 (where it also occurred)

This patch does not solve the problem. https://github.com/zotero/translators/blob/2c95b178ea1ab8196feda4371fef035c8b01b425/.eslintrc have the `"es2017": true`.
#2932's CI work because of `teslint ""`, "Target file '' does not exist; skipping".

yfdyh000:
> Please explain the difficulties you're having. Do you have Node/NPM installed? Does `npm run lint -- "Some Translator.js"` not work?
> 
> (As far as I know, though, Zotero build scripts for Windows are meant for use under Cygwin. You'll likely have more luck with that. You can also try WSL 2, aka a Linux VM.)

In fact, I expect the Lint rules to work correctly in VS Code, but this seems impossible.
It is not convenient to run Lint manually every time, especially run additional Bash in Windows.
I have tried WSL but not very successful. Anyway, I succeeded build Zotero now on Windows, but not very simple.

adam3smith:
I have the linter working more or less in VSCode under Windows:
image

though it doesn't update the Problem messages for me -- I assumed that was an eslint issue, though? Given that VS Code has Terminals embedded, I don't find it particularly tedious to run the linting there, though.

PR #2958, opened 2023-01-10
HotFix for missing import from PubMed collections & clipboard

adam3smith:
https://forums.zotero.org/discussion/102235/is-it-possible-to-develop-zotero-connector-so-that-it-will-work-in-pubmed-clipboard#latest


adam3smith:
The newly added URLs are account specific and require log-in (Collections) or transitional (Clipboard) so no text, but thankfully the DOM is identical for those and search results, so there shouldn't be any issues.

PR #2961, opened 2023-01-12
MSJ

adam3smith:
https://forums.zotero.org/discussion/102280/zotero-connector-importing-a-different-open-access-pdf-even-from-an-open-access-journal#latest


adam3smith:
(sorry, branch is too messy to fix)

PR #2967, opened 2023-01-23
DOI: Pull DOIs from URL

### LINKED ISSUE #2964: DOI: Detect DOI in URL?
Testing https://onlinelibrary.wiley.com/doi/full/10.7448/IAS.15.5.18440 from https://github.com/zotero/translators/issues/2963 in translation-server, the Wiley translation failed, but then translation-server fell back to using the DOI from the URL. It occurs to me that the DOI translator could just do that itself.

If only the URL matches, it can just translate that DOI rather than returning `multiple`. If there are other DOIs on the page, it can put the one from the URL at the top.

Any reason not to do this?

AbeJellinek:
If only the URL contains a DOI (or the URL contains a DOI and the body contains only that same DOI), we now detect a single `journalArticle` item and don't show the Select Items dialog.

Also optimized the page-scraping code a little bit by replacing `Array#includes()` calls (O(n^2) overall) with `Set#has()` (O(n) overall).

Fixes #2964

### DIFF
--- a/DOI.js
+++ b/DOI.js
@@ -142,7 +142,7 @@ function detectWeb(doc, url) {
 	
 	if (!blacklistRe.test(url)) {
 		let doiOrDOIs = getDOIs(doc, url);
-		if (Array.isArray(doiOrDOIs)) {
+		if (Array.isArray(doiOrDOIs) && doiOrDOIs.length) {
 			return "multiple";
 		}
 		else {


DOI.js
```diff
@@ -35,23 +35,58 @@
...
+
+function getDOIsFromURL(url) {
+	// Split on # and ?, so that we don't allow DOIs to contain those characters
+	// but do allow finding DOIs on either side of them (e.g. a DOI in the URL hash)
+	let urlParts = url.split(/[#?]/);
+	for (let urlPart of urlParts) {
+		let match = DOIre.exec(urlPart);
+		if (match) {
+			// Only return a single DOI from the URL
+			return [match[0]];
```
dstillman, line 82: If we're only returning a single DOI from the URL anyway, why not make this `getDOIFromURL()`, return a single DOI string or `false`/`null`, and not bother with the extra `.length == 1` and `[0]` and `...` in `getDOIs()`?

dstillman: I guess because then you would have to test if `fromURL` was null and not put it in the Set at the end.… Still, a little confusing to call it `getDOIsFromURL()`, so maybe better to just deal with that in `getDOIs()`.

dstillman (approved):
Looks good otherwise!

adam3smith:
> We can add a test for single-item detection (DOI in URL and no other DOIs on the page) 

I can find you an example that has the same DOI on the page but no others, would that do?

Issue #2966, opened 2023-01-20
Support for The Markup


RayBB:
Right now dates for The Markup aren't parsed.

Example URL: https://themarkup.org/inside-the-markup/2023/01/18/five-ways-toward-a-fairer-more-transparent-hiring-process

Appreciate all the work that goes into maintaining this :)

PR #2968, opened 2023-01-23
EM: Fall back to `<time>` in `addLowQualityMetadata()`

AbeJellinek:
Addresses #2966 (although it might be worth adding a translator for that site, since EM doesn't get the publication title either). Also updated tests; the Azatliq change is because of this commit - data there is improved - but the others are unrelated.

It's possible that there could be `<time>`s that don't describe the main item on the page, but on a page that contains `<time>`s, it seems very *likely* that the first one is relevant. (And it's called low-quality metadata for a reason.)

### DIFF
--- a/Embedded Metadata.js
+++ b/Embedded Metadata.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2022-12-08 16:38:00"
+	"lastUpdated": "2023-01-23 17:12:53"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


Embedded Metadata.js
```diff
@@ -1256,7 +1260,7 @@ var testCases = [
 						"creatorType": "author"
 					}
 				],
-				"date": "2013/02/20",
+				"date": "2013/2/20",
```
dstillman, line 1263: Unrelated, but seems like we should be cleaning these…

dstillman (approved):

Embedded Metadata.js
adam3smith: I'm pretty sure the only thing we'd have is a `ZU.strToISO` in there, which would be fine (I think?) if duplicated.

PR #2974, opened 2023-02-14
Add EBSCO DIscovery Layer

adam3smith:
Seems to work pretty well. Could likely use some fine tuning, but the basics are there https://forums.zotero.org/discussion/comment/426914#Comment_426914


adam3smith:
I'll merge this since it's hard to test without credentials & it will be good to get some feedback

PR #2975, opened 2023-02-14
Change FAO publication target URL

swifterslb:
FAO has changed the URL naming scheme. Now fao.org/publications/* and fao.org/documents/* are not longer the same, and this script only applies to the former. I'll upload a script that works for the latter soon.

It's likely that FAO will use the "documents" URL as the main link for publications from now on.

### DIFF
--- a/FAO Publications.js
+++ b/FAO Publications.js
@@ -2,14 +2,14 @@
 	"translatorID": "4883f662-29df-44ad-959e-27c9d036d165",
 	"label": "FAO Publications",
 	"creator": "Bin Liu <lieubean@gmail.com>",
-	"target": "^https?://www\\.fao\\.org/(documents|publications)/",
+	"target": "^https?://www\\.fao\\.org/publications/",
 	"minVersion": "3.0",
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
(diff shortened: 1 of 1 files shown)


swifterslb:
I'm sorry I really don't understand the Lint output messages... Could anyone help me?

swifterslb:
This passed all checks a few days ago. Could you have a look? Thanks.

swifterslb:
> Wait, so if this translator only works on `/publications/` URLs now, why do we still have tests with `/documents/` URLs? And are they different enough that we need two translators?

Thanks for the comment. To clarify, now this translator works for both `/publications/` and `/documents/` URLs. Previously all the `/documents/` URLs redirect to `/publications/`, so it was set to work just with `/publications/`. And yes, the two diverge very much now. I tried my best to make the code less redundant, but there may be much room to improve.

swifterslb:
Thanks. I have changed the codes according to the suggestions and did a minor tweak (Line 232). But I don't understand why the checks failed. Could you help me?

adam3smith:
You can ignore the test failure (it's failing on building the structure to run the automated translator tests, not on actually checking your translator)

swifterslb:
> You can ignore the test failure (it's failing on building the structure to run the automated translator tests, not on actually checking your translator)

Thanks. Then I think there are no other issues. It could be merged now.

PR #2976, opened 2023-02-14
Create FAO Documents.js

swifterslb:
Related to #2975 . This is the script for the "new" FAO publications link "https://www.fao.org/documents/*". I made some adjustments on the original "FAO publications" script to accommodate the new layout. Please see if this is OK.

### DIFF
--- a/FAO Documents.js
+++ b/FAO Documents.js
@@ -326,8 +326,7 @@ function scrape(doc, url) {
 }*/
 
 function doWeb(doc, url) {
-
-/*	if (detectWeb(doc, url) == "multiple") {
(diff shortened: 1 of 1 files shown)


swifterslb:
> Couldn't we still combine this into a single translator? Even if we (at least for now) use a different logic for `documents` and `publications`, having all FAO code in one place makes sense.

Could you give me an example script on how to realize this? Some variable names and logic are largely different. Thanks.

swifterslb:
I have updated a combined single translator to #2975. This can be closed.

adam3smith:
superseded by #2975

PR #2978, opened 2023-02-22
CNKI:fix  parsing of db and file info from item pages

adam3smith:
Add some tests


adam3smith:
Merged this given the number of complaints

PR #2987, opened 2023-03-03
Update CNKI refworks output post method

l0o0:
The CNKI has deprecated the old post method for exporting citation. I've updated the post URL and post data string.


dstillman:
Thanks!

PR #2990, opened 2023-03-08
CNKI: Replace obsolete method of article-info retrieval.

zoe-translates:
- The old method based on the "zwjdown" class is removed, because the class is no longer present in the document.
- It is replaced by the method using hidden input fields in the sequence of attempts, after the URL-based one.
- As a backup, a new method based on the note-taking button is added after that.


dstillman:
Great, thanks!

PR #2991, opened 2023-03-09
New translator: Literary Hub

zoe-translates:
Literary Hub is a website hosting blog posts by writers (original literature works, translations, reviews, etc.) and syndicated content such as podcasts.

This translator currently will save blog posts by post-processing the Web translator output. It also includes some support for more complicated authorship-detection (such as translated content), but more work needs to be done in the future.

### DIFF
--- a/Literary Hub.js
+++ b/Literary Hub.js
@@ -330,7 +330,7 @@ function getByline(doc) {
 	if (!byline) {
 		// Only for certain pre-2022 posts without more
 		// semantically-clear byline info.
-		byline = text(doc, ".post_detail>a[href*='author']");
+		byline = text(doc, ".post_detail > a[href*='author']");
 	}
 	return ZU.trimInternal(byline);
 }


dstillman:
Never a need to pull in upstream commits in general for translators (though I imagine here you wanted the ESLint changes). But you'll want to rebase off `origin/master` and force-push so that the upstream commits don't appear in this PR — if you rebase properly, they won't appear here.

dstillman:
Great! We'll take a look.

Note that you still haven't properly rebased, though. Click on the Commits tab here and you can see it's showing separate versions of commits that are already on the master branch. Same in the "Files changed" tab, where it lists files from other commits that shouldn't show as changed. When you rebase properly, it will only show changes to `Literary Hub.js`.

adam3smith:
Sorry I didn't see the question above, but yes, dropping the author is the right thing to do imo.

PR #2993, opened 2023-03-15
CanLII: Breadcrumbs now contain nested spans and other changes

quink:
* And refactor.

Data Changes:
* Links now HTTPS
* Citation format changed

### DIFF
--- a/CanLII.js
+++ b/CanLII.js
@@ -9,9 +9,8 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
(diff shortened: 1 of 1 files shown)


dstillman:
Thanks, but please revert the unnecessary changes. `text()` and `attr()` are built-in helper functions and should be used whenever possible. (The shims at the top can be removed.) We don't use `const` everywhere — `var` is fine. There's certainly no need to replace clear assignments with a convoluted thing involving  `Object.entries()` and `forEach()`. We're grateful for site compatibility fixes, but not opinionated, idiosyncratic refactorings of perfectly clear code.

PR #2994, opened 2023-03-15
CanLII: Fix court retrieval (without refactoring)

quink:


CanLII.js
```diff
@@ -76,7 +71,7 @@ function scrape(doc, url) {
 		newItem.firstPage = reporterDetails[3];
 	}
 	
-	newItem.court = text('#breadcrumbs span', 2);
+	newItem.court = text('#breadcrumbs span.nowrap', 2);
```
dstillman, line 72: Probably safer to go with `#breadcrumbs span[itemprop="name"]` here — or even `#breadcrumbs *[itemprop="name"]` — rather than something presentational.

CanLII.js
```diff
@@ -76,17 +69,17 @@ function scrape(doc, url) {
 		newItem.firstPage = reporterDetails[3];
 	}
 	
-	newItem.court = text('#breadcrumbs span', 2);
+	newItem.court = text('#breadcrumbs *[itemprop="name"] a', 2);
```
dstillman, line 72: No need for the `a`, though — it doesn't add anything and just makes it slightly more brittle (e.g., if they change it to use `a` with `itemprop="name"` directly without `span`).

dstillman: (Though I guess in that case the itemprop value would be the href, so they wouldn't do that, exactly, but still, no need for the extra `a`.)

dstillman:
Thank you!

PR #3000, opened 2023-03-21
I think this will fix some failing cases, eg. https://www.sciencedire…

JavaGT:
https://www.sciencedirect.com/book/9781843344971/building-library-3-0

Some of the chapters seem to be in h5 level headers rather than h4


dstillman:
Thanks! A couple things:

1) You'll want to make translator changes through Scaffold (Tools → Developer → Translator Editor in Zotero), which will automatically update the timestamp, without which a change won't be pushed out to users.

2) We'll want to add a test case for a specific chapter that was failing before and fixed by this change. This can also be done in Scaffold.

PR #3011, opened 2023-04-06
INSPIRE: Fix PDF import

adam3smith:
add defer to item tests. https://forums.zotero.org/discussion/comment/432186#Comment_432186


dstillman:
Thanks!

PR #3014, opened 2023-04-07
fixes #3013

### LINKED ISSUE #3013: BibTeX Import removes special characters when combined with sub/superscript
Example:

```
 @article{Borissov:2855446,
               author        = "Borissov, Alexander and Solokhin, Sergei",
               collaboration = "ALICE",
               title         = "{Production of $\Sigma^{0}$ Hyperon and Search of
                                $\Sigma$$^{0}$ Hypernuclei at LHC with ALICE}",
               journal       = "Phys. At. Nucl.",
               volume        = "85",
               number        = "6",
               pages         = "970-975",
               year          = "2023",
               url           = "https://cds.cern.ch/record/2855446",
               doi           = "10.1134/S1063778823010131",
         }
 ```

(from the included URL)
title imports as 
> Production of <sup>0$</sup> Hyperon and Search of <sup>0$</sup> Hypernuclei at LHC with ALICE

i.e. makes a complete mess of things. The problem is the `mapTeXmarkup` function.

@retorquere -- you're probably (and understandably) not super interested in messing with the generic BibTeX import, but since you've thought about this about 100x as much as I have... if you would take a look that'd be wonderful.

### PR DESCRIPTION
**Pull request by retorquere**

### CODE REVIEW

**Comment by adam3smith** [comment]
Zotero has actions set to time out after 10mins, which means they'll always time out on import translators (like RIS, EM, or BibTeX) used across many web translators. I don't know if there's a solution to this -- it'd actually be nice to see if changes to import affect web output.

**Comment by adam3smith** [comment]
That fails this test still

```
@ARTICLE{Frit2,
  author = {Fritz, U. and Corti, C. and P\"{a}ckert, M.},
  title = {Test of markupconversion: Italics, bold, superscript, subscript, and small caps: Mitochondrial DNA$_{\textrm{2}}$ sequences suggest unexpected phylogenetic position
        of Corso-Sardinian grass snakes (\textit{Natrix cetti}) and \textbf{do not}
        support their \textsc{species status}, with notes on phylogeography and subspecies
        delineation of grass snakes.},
  journal = {Actes du $4^{\textrm{ème}}$ Congrès Français d'Acoustique},
  year = {2012},
  volume = {12},
  pages = {71-80},
  doi = {10.1007/s13127-011-0069-8}
}
```

(both on title and journal title, it leaves $s where it shouldn't -- though the current output is also not right)

retorquere:
New version pushed. These were long-standing errors BTW -- the regex replace would just always chuck any prefix to the sup/sub cased text.

There are always going to be edge cases for this -- math mode can't really be handled comprehensively with regexes.

dstillman:
> Zotero has actions set to time out after 10mins, which means they'll always time out on import translators

Are we sure that's the problem? It's timing out at 10 minutes, but it looks like we also truncate the list to 10 translators, and I wouldn't expect 10 translators to take 10 minutes:

> Over 10 translators need to be tested, but this will take too long and timeout the CI environment. Truncating to 10.

There's no output after it says it's truncating translators, so it's hard to tell what's going on.

dstillman:
Maybe with 10 translators it's just likely that for one of them the site is slow or no longer accessible and timing out at 30 seconds per request or something?

Do we not normally get translator output in the test output?

adam3smith:
> Do we not normally get translator output in the test output?

Alas, no. From https://github.com/zotero/translators/pull/3015/checks

>  Will run tests for translators ["DOI"]
All translator tests passed

PR #3015, opened 2023-04-07
DOI web overhaul

zoe-translates:
Here's a fairly large changeset to the DOI web translator. The overall goal is to make DOI detection more accurate, while making the code easier to understand.

To summarize,

1.  When parsing for DOIs from the `href` attribute values, we decode the URL components before attempting to match with regular expression.
2.  DOIs are case-normalized for deduplication, because DOIs are not case-sensitive.
3.  In "free text", we use a different regular expression to exclude common characters that tend to belong to surrounding context rather than the DOI itself. This is less of a concern for attribute values.
4.  DOI prefixes are not restricted to `\b10\.[0-9]{4,}/`, according to the DOI spec and in particular the examples in it.
5.  In the matched string, any unbalanced or badly-nested parentheses, brackets, braces, .. etc. are sanitized with a more robust approach. Although no rule in the DOI spec says they must balance, IRL the unbalanced puncts (esp. in free text) likely indicate that we matched too eagerly and "overflew" into the surrounding text. So we try to salvage a more likely candidate (the maximal slice from the beginning that is balanced).
6.  The DOM-scraping function is re-written for greater clarity. Now the code nests less deeply, no longer performs unnecessary checks, and repeats itself less often.
7.  Some tests were meant to test for the handling of one single item (journal article), but they no longer work, because the page at the test URL now implements listings (also-reads or bibliography) with DOIs from other sources. This turns the detected type into "multiple". We update the tests to restore the intended behaviour (using a hack: point the test URL to an "export citation" page, still containing the DOI but without other articles).
8.  There are lots of new comments, hopefully able to help others read and maintain the code.

### DIFF
--- a/DOI.js
+++ b/DOI.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2023-04-14 07:11:22"
+	"lastUpdated": "2023-04-17 03:26:22"
 }
 
 /*
@@ -142,12 +142,12 @@ const hrefValueDOIRe = /\b10\.[^/]+\/.+/g;
  * @return {string | string[]} A single string if the URL contains a DOI
  * 		and the document contains no others, or an array of DOIs otherwise
  */
-function getDOIs(doc, url) {
+function getDOIs(doc, url, checkOnly = false) {
 	let fromURL = getRawDOIFromURL(url); // this doesn't sanitize or normalize
 	if (fromURL) {
 		fromURL = sanitizePairedPunct(fromURL).toUpperCase();
 	}
-	let fromDocument = getDOIsFromDocument(doc);
+	let fromDocument = getDOIsFromDocument(doc, checkOnly);
 	if (fromURL) { // We got a DOI from the URL, and...
 		if (fromDocument.size === 0 // None from the document,
 			// Or one from the document, but the same one that was in the URL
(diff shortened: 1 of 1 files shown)


adam3smith:
Haven't looked at the code yet, but the summary all sounds great.

DOI.js
```diff
@@ -71,67 +160,175 @@ function getDOIs(doc, url) {
...
-	// Split on # and ?, so that we don't allow DOIs to contain those characters
-	// but do allow finding DOIs on either side of them (e.g. a DOI in the URL hash)
-	let urlParts = url.split(/[#?]/);
-	for (let urlPart of urlParts) {
-		let match = DOIre.exec(urlPart);
-		if (match) {
-			// Only return a single DOI from the URL
-			return match[0];
+// NOTE: We should case-normalize according to the DOI spec, which says the
+// registry uses all uppercase internally. We follow this convention.
```
dstillman, line 164: I don't think we want to capitalize all DOIs. Regardless of how DOIs are stored in the registry, different publishers have different case conventions, and I think we can assume that they're rendered in the preferred manner, which tends to be much more commonly lowercase.

The place where we want to normalize is when doing comparisons (e.g., a recent bug in the app regarding retraction checks).

dstillman: Oh, right. But then is there any point in normalizing in this translator?

dstillman: Ah, got it — that makes sense.

PR #3018, opened 2023-04-14
Cambridge Core: remove asterisk or "1" from end of title if exists

brendan-oconnell:
Issue reported on Forums:
https://forums.zotero.org/discussion/103018/translator-for-cambridge-core

### DIFF
--- a/Cambridge Core.js
+++ b/Cambridge Core.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2023-04-14 14:03:21"
+	"lastUpdated": "2023-04-14 14:06:56"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


Cambridge Core.js
```diff
@@ -156,6 +156,10 @@ function scrape(doc, url) {
 			if (item.date.includes("undefined")) {
 				item.date = attr('meta[name="citation_online_date"]', "content");
 			}
+			// remove asterisk or 1 at end of title, e.g. https://www.cambridge.org/core/journals/american-political-science-review/article/abs/violence-in-premodern-societies-rural-colombia/A14B0BB4130A2BA6BE79E2853597526E
```
adam3smith, line 159: wouldn't it be better to actually test for the string at the end of the title? I see two potential issues here:
1. If Cambridge ever fixes their metadata and remove the asterisk from `citation_title`, we'll randomly chop of the last character
2. This won't work properly if there are multiple hyperlinked characters at the end of a title (which I don't have an example for, but is generally not implausible in scholarly publishing.

Alternatively, we could consider getting/replacing the title straight from h1 and just remove the hyperlinked content.

PR #3021, opened 2023-04-19
Update Zenodo for Dataset item type

adam3smith:


Zenodo.js
```diff
@@ -120,36 +127,35 @@ function doWeb(doc, url) {
...
+			if (!ZU.fieldIsValidForType('DOI', item.itemType) && doi) {
 				item.extra = "DOI: " + doi;
 			}
-			//workaround while we don't have proper item type for data
-			//if (schemaType && schemaType.includes("Dataset")) {
-			if (ZU.xpathText(doc, '//span[@class="pull-right"]/span[contains(@class, "label-default") and contains(., "Dataset")]')) {
+
+			// workaround for pre 6.0.24 versions that don't have proper item type for data
+			// if (schemaType && schemaType.includes("Dataset")) {
+			if (!datasetType == "dataset" && ZU.xpathText(doc, '//span[@class="pull-right"]/span[contains(@class, "label-default") and contains(., "Dataset")]')) {
```
dstillman, line 158: > !datasetType == "dataset"

This looks like a mistake.

adam3smith: I don't think it is? `datasetType` is [defined here](https://github.com/zotero/translators/pull/3021/files#diff-23ef15323eb9526928ac756f18c84013122c75bdb657077836608330105c9a77R38) and this code should only run where dataset isn't a valid item type, which this tests for.

dstillman: But you're testing whether `!dataset` is equal to a string, which will always be false.

Zenodo.js
```diff
@@ -155,7 +155,7 @@ function scrape(doc, url) {
 
 			// workaround for pre 6.0.24 versions that don't have proper item type for data
 			// if (schemaType && schemaType.includes("Dataset")) {
-			if (!datasetType == "dataset" && ZU.xpathText(doc, '//span[@class="pull-right"]/span[contains(@class, "label-default") and contains(., "Dataset")]')) {
+			if (!(datasetType == "dataset") && ZU.xpathText(doc, '//span[@class="pull-right"]/span[contains(@class, "label-default") and contains(., "Dataset")]')) {
```
dstillman, line 158: You're just going for `datasetType != "dataset"`, right? There's no need for parentheses.

adam3smith: Thanks, I'm clearly a bit out of it post workshop ;)

dstillman:
Now it looks like you pasted in another `scrape` function?

adam3smith:
This should now actually be ready. No idea why that was so hard...

PR #3022, opened 2023-04-20
Add Dataset to RIS

adam3smith:
Also -- lint...

### DIFF
--- a/RIS.js
+++ b/RIS.js
@@ -17,7 +17,7 @@
 	},
 	"inRepository": true,
 	"translatorType": 3,
-	"lastUpdated": "2023-04-20 16:20:25"
+	"lastUpdated": "2023-04-20 19:03:44"
 }
 
 /*
@@ -164,6 +164,14 @@ var importTypeMap = {
 var ty;
 for (ty in exportTypeMap) {
 	importTypeMap[exportTypeMap[ty]] = ty;
+}
+
+// for Zotero clients <6.0.24: set dataset back to document
+if (!ZU.fieldIsValidForType('title', 'dataset')) {
+	delete importTypeMap.DATA;
+	delete importTypeMap.DBASE;
+	importTypeMap.DBASE = 'document';
+	importTypeMap.DATA = 'document';
 }
 
 //merge degenerate export type map into main list


adam3smith:
Question while I'm working on this @dstillman @AbeJellinek  -- There are ~20 `Z.debug` statements in here that seem to slow import down noticeably (going by how quickly the tests in Scaffold run with and without). I don't actually think they help us catch any issues on import, so would it be OK (and, I would argue, desirable, since people use RIS for large import a lot) to comment them out? Or are they only doing anything when debug output is enabled (in which case speed would matter much less).

RIS.js
```diff
@@ -57,83 +81,83 @@ var DEFAULT_EXPORT_TYPE = 'GEN';
...
+	blogPost: "BLOG",
+	book: "BOOK",
+	bookSection: "CHAP",
+	case: "CASE",
+	computerProgram: "COMP",
+	conferencePaper: "CONF",
+	dictionaryEntry: "DICT",
+	encyclopediaArticle: "ENCYC",
+	email: "ICOMM",
+	dataset: "DATA",
```
adam3smith, line 96: Note -- this doesn't check if dataset is available in the importing Zotero version: I didn't know how to do that here in the object and without that it'd be a mess to hack into the translator

adam3smith: No, that's not a mistake. `ExportTypeMap` holds only bi-directional, unique mappings, and [is merged with](https://github.com/zotero/translators/blob/246071dd68bc3d7bf068cb8afff63e19f227408e/RIS.js#L165) `ImportTypeMap` (see also the comments above each of them).
So this does impact import -- could we do the item type switch all the way at the end in post-processing? That seems cleanest, I think.

adam3smith: Cool, actually put the code all the way at the top so the item type would be treated correctly during import throughout. Works nicely best I could test.

dstillman:
Slowdown is due to [this line](https://github.com/zotero/zotero/blob/7473515d53489d0acd2216ab91851be45aa80696/chrome/content/scaffold/scaffold.js#L1147).

adam3smith:
@AbeJellinek lmk when/if this is good to merge.

PR #3024, opened 2023-04-20
Update DV to work properly for datasets

adam3smith:
Note that a bunch of the tests are currently timing out, so I updated them manually -- this appear to affect all current Dataverse installations (the ones that still run use older versions) so we want to keep the tests in there and hope that Z7 fixes this.

### DIFF
--- a/Dataverse.js
+++ b/Dataverse.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2023-04-20 18:10:33"
+	"lastUpdated": "2023-04-24 16:02:02"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


adam3smith:
@AbeJellinek you [created the original regex](https://github.com/zotero/translators/pull/2607). There have since been ~30 additional installations. 
Going by the tests and some clicking around, couldn't we get away with the much simpler ` /\/dataverse\/|\/file.xhtml\?|\/dataset.xhtml\?/`, i.e. simply the part after the URL. That shouldn't be much more costly then the regex we're running now (I've also asked on their mailing list, but maybe you remember why you didn't go by that in the original version).

adam3smith:
Nervermind -- the root collection/dataverse is the problem.
Here's an R script to check for unsupported collections

```
library(jsonlite)

dvInstants <- fromJSON("https://iqss.github.io/dataverse-installations/data/data.json")

dvInstants <- dvInstants$installations

dvRegex <- "^https?:\\/\\/(www\\.)?((open|research-?|hei|planetary-|osna|in|bonn|borealis|lida\\.|archaeology\\.|entrepot\\.recherche\\.|archive\\.|redape\\.)?(data|e?da[td]os)|dvn|sodha\\.be|repositorio(\\.|dedados|pesquisas)|darus|abacus\\.library\\.ubc\\.ca|dorel\\.univ-lorraine\\.fr|dunas\\.ua\\.pt|edmond\\.mpdl\\.mpg\\.de|keen\\.zih\\.tu-dresden\\.de|rdr\\.kuleuven\\.be|portal\\.odissei\\.nl)"


unsupportedInstance = list()
for (url in dvInstants$hostname) {
  url <- paste0("https://", url)
  if (!grepl(dvRegex, url)) {
    unsupportedInstance <- append(unsupportedInstance, url)
  }
} 

print(unsupportedInstance)
```
the resulting list should be empty (or lists all unsupported hosts)

PR #3030, opened 2023-04-27
Don't detect J-Stage PDFs

adam3smith:
https://forums.zotero.org/discussion/104661/cannot-download-papers-from-jstage#latest

@dstillman @AbeJellinek -- there's a larger question here about how much effort we should make to actually detect & translate on PDF pages. Here, it's easy to a) detect this is JSTAGE and b) construct the article landing page from the PDF page, so doing a proper translation would require construction that and a processDocument call -- is that worth it or not (FWIW, recognize PDF doesn't work super reliably for JSTAGE PDFs because they have wonky encoding issues)?

### DIFF
--- a/J-Stage.js
+++ b/J-Stage.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2019-06-15 19:04:11"
+	"lastUpdated": "2023-04-27 13:09:03"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


dstillman:
I've been thinking about this lately, and I think a general goal for the project needs to be to reduce the number of places where people can do the "wrong" thing. No matter how many times we say that you should save this or that way, people are going to do what they do, so we should aim to make it not matter as much as possible. (There are various other contexts where this applies.)

So, if a translator would detect and break regular PDF saving, like here, we should obviously fix that.

If metadata retrieval doesn't work well and we can do the save, we should do it, so that we don't have to tell people they're doing something wrong when they click the save button on the PDF.

If metadata retrieval works well, though, I think it's fine to rely on that. (Metadata updating should also be able to help fill in abstracts or other things that might be missed.)

PR #3032, opened 2023-04-29
Add CFF export

adam3smith:
https://forums.zotero.org/discussion/101421/from-zotero-to-citation-cff#latest

### DIFF
--- a/CFF.js
+++ b/CFF.js
@@ -69,8 +69,8 @@ function writeAuthors(itemCreators) {
 
 function doExport() {
 	var item;
-	// eslint-disable-next-line no-cond-assign
 	Zotero.write('# This CITATION.cff file was generated with Zotero.\n');
+	// eslint-disable-next-line no-cond-assign
 	while (item = Zotero.nextItem()) {
 		// Only use for dataset and software
 		if (item.itemType != "dataset" && item.itemType != "computerProgram") {
--- a/CFF.js
+++ b/CFF.js
@@ -8,30 +8,30 @@
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 2,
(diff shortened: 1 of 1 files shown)


adam3smith:
A couple of open questions:
- Should we do something with other item types? CFF-type is explicitly only valid for software and dataset, so I'd be inclined to say no.
- Should we add some sort of message at the top? E.g. this is what cff files generated with cffinit display. I find the second line tacky, but clarifying origin might be worthwhile?:

> \# This CITATION.cff file was generated with cffinit.
> \# Visit https://bit.ly/cffinit to generate yours today!

- The are various places for the URL. It could be identifier: - URL or repository-code: I currently don't handle it at all, but it should probably be somewhere?

dstillman:
> Should we do something with other item types? CFF-type is explicitly only valid for software and dataset, so I'd be inclined to say no.

Yeah, I think that's fine.

> Should we add some sort of message at the top?

Sure, `This CITATION.cff file was generated with Zotero.` seems appropriate.

> The are various places for the URL.

Just `url:` seems fine.

adam3smith:
> Just url: seems fine.

oh, I had overlooked that, that makes sense.

I think this is read then. One last question: there is not a way I can set the filename of the exported file, is there? That should be `CITATION.cff`, not `Exported Items.cff`, ideally.

dstillman:
> there is not a way I can set the filename of the exported file, is there?

Hmm, no, though we could just hard-code that for this translator in the app. We could add proper support for it, but I'm not sure we've ever had another need for it.

adam3smith:
> We could add proper support for it, but I'm not sure we've ever had another need for it.

I agree -- I think it'd be nice to hard-code this for CFF, but otherwise it has never come up.

PR #3035, opened 2023-05-12
NYU Press

18kimn:
# Summary

This PR attempts to add a translator for NYU Press, a major university publishing house that is currently recognized as having embedded metadata only, and that often fails or produces incorrect translations. It only adds single books, though I can implement the `multiple` type if that is considered desirable? 

Looking for comments on idiomatic code, since I'm new to the norms of this codebase -- 

## Glimpse of provided test cases:

- [Trickster Tales](https://nyupress.org/9780889779563/trickster-tales-and-the-stories-of-chief-thunderchild/), having one editor, one contributor, and one author
- [The Health of Newcomers](https://nyupress.org/9780814789216/the-health-of-newcomers/), having two authors 
- [Keywords for Gender and Sexuality Studies](https://nyupress.org/9781479808151/keywords-for-gender-and-sexuality-studies/), having a non-person author, being part of a series, and having no subtitle

### DIFF
--- a/NYU Press.js
+++ b/NYU Press.js
@@ -1,0 +1,276 @@
+{
+	"translatorID": "2ca2c054-7084-458f-b493-86e388284720",
+	"label": "NYU Press",
+	"creator": "Nathan Kim",
+	"target": "https?://nyupress\\.org/([0-9]){13}",
+	"minVersion": "5.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2023-05-12 00:33:10"
+}
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2022 Nathan Kim
+
+	This file is part of Zotero.
+
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+function detectWeb() {
+	return "book";
+}
+
+function doWeb(doc, url) {
+	var item = new Zotero.Item("book");
+	item.date = doc.querySelector(".sp__published").textContent.replace("Published: ", "");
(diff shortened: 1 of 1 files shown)


NYU Press.js
```diff
@@ -0,0 +1,276 @@
...
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+function detectWeb() {
```
adam3smith, line 38: We'd want support for multiples (e.g. browse & search results) here as well (that also means amending the target regex)

NYU Press.js
```diff
@@ -0,0 +1,276 @@
...
+	return "book";
+}
+
+function doWeb(doc, url) {
+	var item = new Zotero.Item("book");
+	item.date = doc.querySelector(".sp__published").textContent.replace("Published: ", "");
+	item.ISBN = doc.querySelector(".sp__isbn13").textContent;
```
adam3smith, line 45: use `text(doc, 'sp__isbn13')` instead (it's more robust & simpler)

NYU Press.js
```diff
@@ -0,0 +1,276 @@
...
+
+	for (let roleNode of doc.querySelectorAll('.sp__the-author')) {
+		let role;
+		if (roleNode.textContent.match("Edited")) {
```
adam3smith, line 49: use `.includes()` for literal matches

NYU Press.js
```diff
@@ -0,0 +1,276 @@
...
+		}
+		else {
+			role = "author";
+		}
+		for (let authorNode of roleNode.querySelectorAll(".sp__author-link")) {
+			item.creators.push(ZU.cleanAuthor(authorNode.textContent, role));
+		}
+	}
+
+	item.numPages = doc.querySelector(".sp__the-pages").textContent.replace(" Pages", "");
```
adam3smith, line 63: for this and other fields that may possibly be missing, first test for the existence, then use replace, so something like
`if(text(doc, '.sp__the-pages')) { item.numPages = text(doc, '.sp__the-pages').replace(" Pages", "");

NYU Press.js
```diff
@@ -0,0 +1,276 @@
...
+	item.abstractNote = doc.querySelector(".tabs__panel--description p").textContent;
+	let series = doc.querySelector(".book-wrapper__top-section--details h3");
+	item.series = series ? series.textContent : '';
+	item.publisher = doc.querySelector(".sp__the-publisher").textContent.replace("Published by: ", "");
+
+	item.url = url;
```
adam3smith, line 69: remove -- the URL field should only be used for full-text sources (which I assume these almost never are) -- it'll be used in citations. You'll save the link via the Snapshot.

adam3smith (changes requested):
Thanks -- a couple of change requests to make this cleaner & more robust, plus we'll definitely want multiples

18kimn:
apologies, yes, thanks for the review @adam3smith. I've been a bit busy but will update this PR soon!

PR #3036, opened 2023-05-20
National Library of Poland ISBN search

Eccenux:
Hi. I've added BN search for Zotero, to be used hopefully on Wikipedia (Citoid) in future.

This is a simple marcXml search and BN API is rather stable. I've created the test in the Scaffold IDE and it seems to parse data without problems.

### DIFF
--- a/National Library of Poland ISBN.js
+++ b/National Library of Poland ISBN.js
@@ -5,12 +5,12 @@
 	"target": "",
 	"minVersion": "4.0",
 	"maxVersion": "",
-	"priority": 97,
+	"priority": 98,
 	"inRepository": true,
 	"translatorType": 8,
-	"browserSupport": "gcsibv",
+	"browserSupport": "gcsbv",
 	"lastUpdated": "2023-05-20 15:41:08"
-};
+}
 
 /*
 	***** BEGIN LICENSE BLOCK *****
--- a/National Library of Poland ISBN.js
+++ b/National Library of Poland ISBN.js
@@ -1,0 +1,71 @@
+{
+	"translatorID": "7dda2b16-6a04-4d22-afd1-21d59f8d0a4c",
+	"label": "National Library of Poland ISBN search",
+	"creator": "Maciej Nux Jaros",
+	"target": "",
+	"minVersion": "4.0",
+	"maxVersion": "",
+	"priority": 97,
+	"inRepository": true,
+	"translatorType": 8,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2023-05-20 21:23:00"
+}
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2023 Maciej Nux Jaros
+
+	This file is part of Zotero.
+
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+function detectSearch(item) {
+	// Other supported parameters (BN docs):
+	// https://data.bn.org.pl/docs/bibs
+	
+	// for now only using ISBN
+	return !!item.ISBN;
+}
+
+function doSearch(item) {
+	const isbn = ZU.cleanISBN(item.ISBN);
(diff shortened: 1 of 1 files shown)


Eccenux:
I just noticed I've copy&pasted wrong `browserSupport`. Not sure if that matters. Using ES6, so no IE support.

National Library of Poland ISBN.js
```diff
@@ -0,0 +1,118 @@
+{
+	"translatorID": "aa7f310e-10d3-4209-91dc-88301e7070c6",
+	"label": "National Library of Poland ISBN",
+	"creator": "Maciej Nux Jaros",
+	"target": "",
+	"minVersion": "4.0",
+	"maxVersion": "",
+	"priority": 97,
```
adam3smith, line 8: We'd want the priority lower -- I think this should probably run after the K10+ catalog or after the BnF

adam3smith:
Thanks -- looks good in general. 
There is a larger question about how many different ISBN searches we're going to add. I think in general this one makes sense -- we don't have good coverage of Poland/Eastern Europe in the current ISBN searches -- but this is run on every single ISBN that people input and makes things a bit slower, so there is a real downside to keep adding these.
@dstillman what's your view on this? One option would be to just run this on ISBNs with Polish [region code](https://en.wikipedia.org/wiki/List_of_ISBN_registration_groups) -- that's surely going to miss some stuff (e.g. other Eastern countries well represented in the NLP) but it would avoid the issue above.

National Library of Poland ISBN.js
```diff
@@ -0,0 +1,118 @@
...
+	"inRepository": true,
+	"translatorType": 8,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2023-05-20 15:41:08"
+};
```
adam3smith, line 13: that's the ; the linter is complaining about

adam3smith (changes requested):

National Library of Poland ISBN.js
Eccenux: Lower priority is higher number? Changed to the same as bnf (98)

National Library of Poland ISBN.js
Eccenux: Should be fine now. That came from editing in VSC.

Eccenux:
> ...One option would be to just run this on ISBNs with Polish [region code](https://en.wikipedia.org/wiki/List_of_ISBN_registration_groups) -- that's surely going to miss some stuff (e.g. other Eastern countries well represented in the NLP) but it would avoid the issue above.

I happen to have access to full dump of BN. I'll check for patterns in ISBN.

Eccenux:
Filtering by `/^(97883|83)/` might work. Stats from the database:
* pl83: 1 085 412
* total with ~valid isbn: 1 433 909

Stats SQL:
```
select 'pl83' as type, sum(isbn_count) from temp_isbn_short
--select * from temp_isbn_short
where isbn_prefix ~ '^97883' or isbn_prefix ~ '^83'
union all
select 'total with ~valid isbn', sum(isbn_count) from temp_isbn_short
;
```

Prefix data of BN:
[data-isbn-short.csv](https://github.com/zotero/translators/files/11523245/data-isbn-short.csv)

adam3smith:
Oh yes, I think that's a strong case for prefiltering ISBNs. (which would then also allow us to add many more ISBN translators). @dstillman would it make more sense to do this in the identifier code or the translators individually (which I think we can already do?)

dstillman:
Much better to do it in the translator detect code — then it'll just work everywhere.

Eccenux:
Done. Added a simple filter which should be fast.

mvolz:
> Thanks -- looks good in general. There is a larger question about how many different ISBN searches we're going to add. I think in general this one makes sense -- we don't have good coverage of Poland/Eastern Europe in the current ISBN searches -- but this is run on every single ISBN that people input and makes things a bit slower, so there is a real downside to keep adding these. @dstillman what's your view on this? One option would be to just run this on ISBNs with Polish [region code](https://en.wikipedia.org/wiki/List_of_ISBN_registration_groups) -- that's surely going to miss some stuff (e.g. other Eastern countries well represented in the NLP) but it would avoid the issue above.

Sorry for my ignorant question, but there is any way to do this less syncronously to avoid this issue? Or is that too hard with the current infrastructure. 

There is a long standing request for us to support the national library in Japan which I was hoping to look into next. https://phabricator.wikimedia.org/T192528

Eccenux:
@adam3smith Hi. I think I resolved everything here, but please let me know if there is more work for me to do? Otherwise, could you re-run tests and merge?

Eccenux:
Hi, I've updated the base for my pull request so merging should be easy.

Maybe this linter will work too... Although I don't know where the version number '65161a4967b2...' came from in the script. From what I see it's some change from `2022-10-28`:
OUP: Lint, add authors from DOM if necessary (#2898)

adam3smith:
(I believe CI fails if there have been any CI-relevant changes on the `master` branch while the PR is under review)

adam3smith:
Sorry for the delay -- I wanted to spend some time testing this to make sure this works as intended, since we haven't used the idea of restricting to ISBN regions before -- it's looking great and I'm very excited that this means we'll be able to add other national (or quasi-national) libraries to ISBN search without real costs.

PR #3038, opened 2023-05-26
Lexis+: Oops

bfahrenfort:
I discovered a bug with word capture today, turns out `/act/i` will match Interactive, leading cases to be flagged as statutes and fail. Easy fix.

Edit: Lexis+ dynamically changes titles so it probably wouldn't ever happen, but my regex should probably match if it's at the end of a string.

### DIFF
--- a/Lexis+.js
+++ b/Lexis+.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2023-05-26 03:42:30"
+	"lastUpdated": "2023-05-26 04:09:26"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


Lexis+.js
```diff
@@ -40,7 +40,7 @@ function detectWeb(doc, _url) {
 		return "multiple";
 	}
 	else if (/[a-zA-Z. ]+\s§\s\d+/.test(doc.title)
-	|| /\W*(act)\W|\W*(act)$/i.test(doc.title)
+	|| /\W(act)\W|\W(act)$|\W(acts)\W|\W(acts)$/i.test(doc.title)
```
dstillman, line 43: How about just `/\W(acts?)(\W|$)/` ?

Also good to add a comment with some real-life examples.

bfahrenfort: Perfect, thanks!

PR #3040, opened 2023-05-31
Updated Electronic Colloquium on Computational Complexity translator

morganfshirley:
I've updated the translator for the Electronic Colloquium on Computational Complexity. The changes:

- Previously, all of the metadata was scraped from the text on the webpage. Some of the data is available in meta tags, and the translator now gets what it can from there.
- Similarly to the Arxiv.org translator, we now use the preprint type if available and fall back to the report type otherwise.
- Abstract scraping was broken. This is now fixed.
- The tests are updated, including a new test to ensure that abstracts with multiple paragraphs work with the new scraping code.
- Made some style fixes for compliance with current guidelines.

This is my first PR here; please let me know if I've done something incorrectly!

### DIFF
--- a/Electronic Colloquium on Computational Complexity.js
+++ b/Electronic Colloquium on Computational Complexity.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2023-05-31 16:38:35"
+	"lastUpdated": "2023-05-31 20:54:48"
 }
 
 /*
@@ -74,6 +74,7 @@ function scrape(doc) {
 	
 	if (hasPreprint) {
 		newItem.archiveID = ZU.xpathText(doc, metaXPath("technical_report_number"));
+		newItem.publisher = "Electronic Colloquium on Computational Complexity";
 	}
 	else {
 		newItem.reportNumber = ZU.xpathText(doc, metaXPath("technical_report_number"));
@@ -162,6 +163,7 @@ var testCases = [
 				"abstractNote": "We study the impact of combinatorial structure in congestion games on the complexity of computing pure Nash equilibria and the convergence time of best response sequences.  In particular, we investigate which properties of the strategy spaces of individual players ensure a polynomial convergence time. We show, if the strategy space of each player consists of the bases of a matroid over the set of resources, then the lengths of all best response sequences are polynomially bounded in the number of players and resources. We can also prove that this result is tight, that is, the matroid property is a necessary and sufficient condition on the players' strategy spaces for guaranteeing polynomial time convergence to a Nash equilibrium. In addition, we present an approach that enables us to devise hardness proofs for various kinds of combinatorial games, including first results about the hardness of market sharing games and congestion games for overlay network design. Our approach also yields a short proof for the PLS-completeness of network congestion games.",
 				"archiveID": "TR06-067",
 				"libraryCatalog": "Electronic Colloquium on Computational Complexity",
+				"repository": "Electronic Colloquium on Computational Complexity",
 				"url": "https://eccc.weizmann.ac.il/report/2006/067/",
 				"attachments": [
 					{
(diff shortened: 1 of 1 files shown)


Electronic Colloquium on Computational Complexity.js
```diff
@@ -1,94 +1,122 @@
...
 function detectWeb(doc, url) {
-	var singleRe   = /^https?:\/\/eccc\.weizmann\.ac\.il\/report\/\d{4}\/\d{3}/;
 	var multipleRe = /^https?:\/\/eccc\.weizmann\.ac\.il\/(title|year|keyword)\//;
-	if (singleRe.test(url)) {
-		return "report";
-	} else if (multipleRe.test(url)) {
+	if (multipleRe.test(url)) {
 		return "multiple";
 	}
+	else if (ZU.fieldIsValidForType('title', 'preprint')) {
```
adam3smith, line 43: Could you use the `const` approach as e.g. here:
https://github.com/zotero/translators/blob/master/OSF%20Preprints.js#L39

a) it doesn't require you to duplicate code quite as much and b) it's what we've done pretty consistently in other translators, so we can remove the report code more quickly once it's no longer needed

Electronic Colloquium on Computational Complexity.js
```diff
@@ -1,94 +1,122 @@
...
+		return "preprint";
+	}
+	else {
+		return "report";
+	}
+}
+
+// Get a meta tag from the header
+function metaXPath(name) {
+	return `/html/head/meta[contains(@name,'citation_${name}')]/@content`;
```
adam3smith, line 53: I understand what you're trying to do, but I don't think this is a great approach here.
a) Using a CSS selector with `attr` is barely longer than calling your custom function and more readable and
b) if the translator uses a lot of `citation_` "highwire" metatags, I'd recommend calling the embedded metadata translator and letting that do most of the work. There's a template for this availabel in the translator editor/Scaffold

adam3smith (changes requested):
Two larger things to fix -- it'd be good to general modernize this, but I'm OK just taking the general improvement with the two changes requested

Electronic Colloquium on Computational Complexity.js
morganfshirley: Done. The original code was based on the arXiv.org translator; that one should probably be changed as well.

Electronic Colloquium on Computational Complexity.js
morganfshirley: Switched to using the Embedded Metadata translator. The template code in Scaffold does not work. Specifically, the getTranslatorObject function appears to fail silently. I fixed this by using the getTranslatorObject code from the OSF Preprint translator mentioned above.

Electronic Colloquium on Computational Complexity.js
```diff
@@ -207,17 +189,14 @@ var testCases = [
...
 				"archiveID": "TR23-078",
-				"libraryCatalog": "Electronic Colloquium on Computational Complexity",
+				"language": "en",
+				"libraryCatalog": "eccc.weizmann.ac.il",
 				"repository": "Electronic Colloquium on Computational Complexity",
 				"shortTitle": "Toward Better Depth Lower Bounds",
 				"url": "https://eccc.weizmann.ac.il/report/2023/078/",
 				"attachments": [
 					{
-						"title": "ECCC Snapshot",
```
morganfshirley: Using the Embedded Metadata translator means that we don't get a snapshot anymore. Should it be added back?

adam3smith: I don't think the Snapshot adds much here -- people will principally want the PDF and already get the abstract in the metadata, so I'd leave it out.

adam3smith:
Could you try rebasing this against upstream master to see if that runs CI properly?

morganfshirley:
My apologies. I'm afraid I don't know what I'm supposed to do here.

adam3smith:
Second try worked, thanks

morganfshirley:
@zoe-translates Sorry, I don't know where to find the suggestions/diffs you've made. I'd be happy to replace `doWeb()` with a template if you point me towards documentation or an example translator where that change has already been implemented.

Electronic Colloquium on Computational Complexity.js
```diff
@@ -1,116 +1,120 @@
...
+			item.abstractNote += abstractParagraphs[i].innerText + "\n";
+		}
+		item.complete();
+	});
 
-	newItem.complete();
+	await translator.getTranslatorObject(function (trans) {
+		trans.itemType = preprintType;
+		trans.doWeb(doc, url);
+	});
```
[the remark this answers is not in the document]
morganfshirley: Strange, I also can't reproduce. Replacing with the template-generated code.

Electronic Colloquium on Computational Complexity.js
```diff
@@ -128,40 +132,99 @@ var testCases = [
...
 				],
-				"seeAlso": [],
+				"notes": [],
+				"seeAlso": []
+			}
+		]
+	},
+	{
+		"type": "web",
+		"url": "https://eccc.weizmann.ac.il/keyword/13486/",
```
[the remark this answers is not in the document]
morganfshirley: Applied this suggestion, thanks!

PR #3042, opened 2023-06-04
Add Optimization Online translator

quachpas:
New translator for https://optimization-online.org/. 

From their website (https://optimization-online.org/about/):
> Optimization Online is a repository of Eprints about optimization and related topics. It facilitates quick dissemination of new research of interest to the optimization community. This service is free.
> Optimization Online submissions are electronically handled by a team of volunteer coordinators.

Unfortunately, I have not managed to parse the abstract consistently as the structure is not fixed.

Thanks in advance!

### DIFF
--- a/Optimization Online.js
+++ b/Optimization Online.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2023-06-04 09:53:10"
+	"lastUpdated": "2023-06-18 08:16:06"
 }
 
 /*
@@ -82,15 +82,13 @@ async function scrape(doc, url) {
 	translator.setHandler('itemDone', (_obj, item) => {
 		// Item
 		item.itemType = 'preprint';
-		let title = doc.querySelector('h1.entry-title');
+		let title = text(doc, 'h1.entry-title');
 		if (title) {
-			item.title = ZU.capitalizeTitle(title.innerText);
+			item.title = ZU.capitalizeTitle(title);
 		}
 		let tags = doc.querySelectorAll('div.entry-meta > span.tags-links > a[rel="tag"]');
-		if (tags.length) {
-			for (var i = 0; i < tags.length; i++) {
-				item.tags.push(ZU.cleanTags(tags[i].text));
-			}
+		for (let tag of tags) {
+			item.tags.push(ZU.cleanTags(tag.text));
 		}
 		let abstractNote = doc.querySelector('div.entry-content > div > p');
 		if (abstractNote) {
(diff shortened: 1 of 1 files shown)


Optimization Online.js
```diff
@@ -0,0 +1,364 @@
...
+				"archiveID": "23021",
+				"language": "en-US",
+				"libraryCatalog": "optimization-online.org",
+				"repository": "Optimization Online",
+				"url": "https://optimization-online.org/2023/05/political-districting-to-optimize-the-polsby-popper-compactness-score/",
```
adam3smith, line 160: Let's save the short URL, which will be much more convenient for citing

Optimization Online.js
```diff
@@ -0,0 +1,364 @@
...
+async function scrape(doc, url) {
+	let translator = Zotero.loadTranslator('web');
+	// Embedded Metadata
+	translator.setTranslator('951c027d-74ac-47d4-a107-9c3069ab7b48');
+	translator.setDocument(doc);
+
+	translator.setHandler('itemDone', (_obj, item) => {
+		// Item
+		item.itemType = 'preprint';
+		let title = doc.querySelector('h1.entry-title');
```
adam3smith, line 85: just use Zotero's built-in  `text(doc, selector)` instead of `querySelector` and `innerText`

Optimization Online.js
```diff
@@ -0,0 +1,364 @@
...
+		if (title) {
+			item.title = ZU.capitalizeTitle(title.innerText);
+		}
+		let tags = doc.querySelectorAll('div.entry-meta > span.tags-links > a[rel="tag"]');
+		if (tags.length) {
+			for (var i = 0; i < tags.length; i++) {
+				item.tags.push(ZU.cleanTags(tags[i].text));
+			}
+		}
+		let abstractNote = doc.querySelector('div.entry-content > div > p');
```
adam3smith, line 95: ditto on using `text()`

adam3smith (changes requested):
Looks good in general. A couple of minor changes requested, I'll run the auto-linter once you've made thus

Optimization Online.js
quachpas: When I use `text()` instead, it duplicates mathjax in the abstract.
I left it as it is for now, would you know how to avoid that?
See https://optimization-online.org/?p=23021 for an example

quachpas:
Thank you for the review, I have integrated most changes but need some further discussion on the abstract parsing. 
On another note, do I add you to the credits now or will you do it after the review?

Optimization Online.js
quachpas: Thank you for the snippet @zoe-translates! I ended up using `innerText()` instead, which was simpler, thank you @AbeJellinek 
I also modified it for the title and added a test case.

quachpas:
Thank you for the help! I believe all brought-up issues were resolved in one way or another.
Would it be ok to run the auto-linter @adam3smith?

Optimization Online.js
```diff
@@ -0,0 +1,418 @@
...
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+function detectWeb(doc, url) { // eslint-disable-line no-unused-vars
+	if (doc.body.classList.contains('single-post')) {
+		return 'preprint';
```
adam3smith, line 40: @AbeJellinek @dstillman  -- at one point do we decide that it's OK if translators don't work for versions anymore? Preprint was introduced in 6.0, almost 18 month ago, I feel like it's fine to just return this as a type now (as is done here) rather than working with the alternates for pre 6.0 versions, thoughts? 
Since this will come up repeatedly with new item types, it'd be good to have an approximate policy/guideline for this.

adam3smith:
One general question for core devs (See e.g. [OSF Preprints](https://github.com/zotero/translators/blob/master/OSF%20Preprints.js#L39) for what we did right after introducing the `preprint` item type), once we have agreement on this, this lgtm -- thanks!

Optimization Online.js
dstillman: Yes, totally fine to break 5.0 compatibility at this point.

adam3smith:
Great, thanks!

PR #3043, opened 2023-06-04
Google Scholar: Add delays between consecutive network requests.

### LINKED ISSUE #3037: Google Scholar: Add delay between metadata requests
https://forums.zotero.org/discussion/comment/435456/#Comment_435456

zoe-translates:
When there are multiple items to be translated, the previous implementation fires network requests to Google Scholar server in rapid succession. This increases the likelihood of being rate-limited or denied by the server.

To fix this, an artificial delay of at least 500ms is added between successive network requests.

In addition, calls to the callback-based `doGet()` are replaced with promise-based asynchronous calls.

A few stylistic fixes (such as `===` vs. `==`) are also included.

Resolves #3037.

### DIFF
--- a/Google Scholar.js
+++ b/Google Scholar.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2023-06-04 11:32:20"
+	"lastUpdated": "2023-06-05 08:15:25"
 }
 
 /*
@@ -155,11 +155,12 @@ async function scrape(doc, idsOrUrl, type) {
 }
 
 
-async function processCitePage(citeURL, context) {
+async function processCitePage(citeURL, context, referrer) {
 	// Note that the page at citeURL has no doctype and is not a complete HTML
 	// document. The browser can parse it in quirks mode but ZU.requestDocument
 	// has trouble with it.
-	const citePage = await ZU.requestText(citeURL);
+	let reqOptions = { headers: { Referer: referrer || "" } };
+	const citePage = await ZU.requestText(citeURL, reqOptions);
 
 	let m = citePage.match(/href="((https?:\/\/[a-z.]*)?\/scholar.bib\?[^"]+)/);
 	if (!m) {
@@ -179,8 +180,13 @@ async function processCitePage(citeURL, context) {
 	}
 	const bibTeXURL = ZU.unescapeHTML(m[1]);
 
+	// Pause between obtaining the citation info page and sending the request
+	// for the BibTeX document.
 	await delay(500);
-	const bibTeXBody = await ZU.requestText(bibTeXURL);
+	// NOTE: To emulate the web app, the referrer for the BibTeX text is always
+	// set to the origin (i.e. https://scholar.google.com/)
+	reqOptions.headers.Referer = "https://scholar.google.com/";
+	const bibTeXBody = await ZU.requestText(bibTeXURL, reqOptions);
 
 	const translator = Z.loadTranslator("import");
 	translator.setTranslator("9cb70025-a888-4a29-a210-93ec52da40d4"); // BibTeX
@@ -292,17 +298,18 @@ async function scrapeIds(doc, ids) {
 			context = doc;
 		}
 
-		let citeUrl = '/scholar?q=info:' + ids[i] + ':scholar.google.com/&output=cite&scirp=1';
+		let citeUrl = '/scholar?q=info:' + ids[i] + ':scholar.google.com/&output=cite&scirp=0&hl=en';
 		// For 'My Library' we check the search field at the top
 		// and then in these cases change the citeUrl accordingly.
 		let scilib = attr(doc, '#gs_hdr_frm input[name="scilib"]', 'value');
 		if (scilib && scilib === 1) {
-			citeUrl = '/scholar?scila=' + ids[i] + '&output=cite&scirp=1';
+			citeUrl = '/scholar?scila=' + ids[i] + '&output=cite&scirp=0&hl=en';
 		}
 
-		await processCitePage(citeUrl, context);
+		await processCitePage(citeUrl, context, doc.location.href);
+		// Pause for 1s before processing next citeUrl.
 		if (i !== ids.length - 1) {
-			await delay(500);
+			await delay(1000);
 		}
 	}
 }
--- a/Google Scholar.js
+++ b/Google Scholar.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2023-07-03 04:50:33"
+	"lastUpdated": "2023-07-11 05:43:37"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


dstillman:
Oh, Google almost certainly doesn't include a Retry-After. That would be offered for a documented public API, not a site like Google trying to stop bots.

Approximately how many items were you able to save before you got a 429 with the 500ms delay?

Something like the token idea might help, but I think we'd have to implement it in the Connector itself and enforce it based on HTTP request domain, since we'd want it to apply across pages. And we wouldn't do it with `setInterval` — it would just have to be mathematical based on an in-memory record of recent translator HTTP requests for that domain or URL pattern. So we can do that separately from this commit.

Google Scholar.js
```diff
@@ -101,48 +102,40 @@ function getProfileResults(doc, checkOnly) {
...
-				}
-				// here we need open these pages before calling scrape
-				ZU.processDocuments(articles, scrape);
-			});
+			const items = await Z.selectItems(getProfileResults(doc, false));
+			if (!items) {
+				return;
+			}
+			// here we need open these pages before calling scrape
+			for (const item of Object.keys(items)) {
```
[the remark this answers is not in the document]
dstillman: `item` → `url` (key, not value), but probably need to rework the `for` loop anyway — see below

Google Scholar.js
```diff
@@ -101,48 +102,40 @@ function getProfileResults(doc, checkOnly) {
...
+				await scrape(await requestDocument(item));
```
dstillman, line 123: We need a `delay()` here if not the last key.

dstillman (changes requested):

dstillman:
Switch on Developer mode and view the extension's background page to see the translator requests.

dstillman:
> I have great difficulty getting any consistent behaviour. Now I'm getting 403 Forbidden instead of 429, even with 1s delay.
>
> Moreover there are too many moving parts (IP address, browser/version, prior interactions with the servers, etc.) for me to deduce anything for now.

Yeah, we've always had the view that it's essentially a black box and not something we can really solve in a reliable way. But anything in the direction of looking more human-like is likely to help to some extent. That may mean a 1s minimum rather than 500ms, at least between items.

adam3smith:
Thanks for working on this. Did you still want my input on something -- I'm not quite sure I understand on what? If it's exact strategies on construction the URLs, I have no recollection. I (or whoever last changed this) almost certainly did exactly what you're doing now, replicating the behavior of the website, which may well have changed.

Google Scholar.js
```diff
@@ -661,6 +570,630 @@ ItemFactory.prototype.saveItemCommonVars = function () {
...
+	}
+
+	// Attach linked page as snapshot if available
+	if (row.directLink && row.directLink !== row.attachmentLink) {
+		item.attachments.push({
+			url: row.directLink,
+			title: "Snapshot",
+			mimeType: "text/html"
+		});
+	}
```
dstillman, line 1194: I'd remove snapshot saving. We certainly don't want them when PDFs are available, but there's also no reason to think that every link on GS is going to produce useful full-text content. In general, we save too many snapshots, and we should reserve them for things like newspaper articles where we know we're capturing the full text (and also the page might go behind a paywall).

Google Scholar.js
```diff
@@ -661,6 +570,630 @@ ItemFactory.prototype.saveItemCommonVars = function () {
...
+ * @property {TransPipeline?} next - Next pipeline in the chain.
+ * @property {RowObj[]} bin - Row objects that failed all translations.
+ * @property {Promise} current - The current task-queue tail to which further
+ * tasks will be appended.
+ * @property {Promise[]} trace - Array holding references to each task promise
+ * that passes through the pipeline.
+ * @property {boolean} restLeads - Whether the "rest" shall apply to the
+ * leading edge of a work task, in addition to the trailing edge.
+ */
+function TransPipeline(identify, work, rest, context, label = "[unnamed]") {
```
dstillman, line 718: Can just call this `TranslationPipeline`

Google Scholar.js
```diff
@@ -661,6 +570,630 @@ ItemFactory.prototype.saveItemCommonVars = function () {
...
+// Returns a promise that resolves (to undefined) after the miminum time delay
+// specified in milliseconds.
+function delay(ms) {
+	return new Promise(resolve => setTimeout(resolve, ms));
+}
+
+// Reusable external-search utility for both DOI and ArXiv.
+
+/**
+ * Search using external ID.
```
dstillman, line 589: Here and throughout: No periods in comments for single sentences/fragements

Google Scholar.js
```diff
@@ -661,6 +570,630 @@ ItemFactory.prototype.saveItemCommonVars = function () {
...
 };
 
+
+/*
+ * #########################
+ * ### Utility Functions ###
+ * #########################
+ */
+
+// Returns a promise that resolves (to undefined) after the miminum time delay
```
dstillman, line 580: "minimum"

Google Scholar.js
```diff
@@ -661,6 +570,630 @@ ItemFactory.prototype.saveItemCommonVars = function () {
...
+ *
+ * @async
+ * @param {string[]} pfURLS - URLs selected by the user, as returned by
+ * getProfileResults().
+ * @param {Document} pfDoc - The profile-page document being scraped.
+ * @yields {Promise<RowObj?>} Promise that resolves to the row object, obtained
+ * by requesting the "Article View" document at the selected URL, or undefined
+ * in the case of failure.
+ */
+function* rowsFromProfile(pfURLs, pfDoc) {
```
dstillman, line 915: `pfURLs` → `profileURLs`, `pfDoc` → `profileDocs`

Google Scholar.js
```diff
@@ -101,198 +112,96 @@ function getProfileResults(doc, checkOnly) {
...
+
+		if (getSearchResults(doc, true/* checkOnly */)) {
+			let items = await Z.selectItems(getSearchResults(doc, false));
+			if (!items) {
+				return;
+			}
+
+			multiScraper = makeGSScraper(Object.keys(items),
+				rowsFromSearchResult,
+				new URL(doc.location));
```
dstillman, line 135: Clearer indentation, here and below:

```
multiScraper = makeGSScraper(
    Object.keys(items),
    rowsFromSearchResult,
    new URL(doc.location)
);

File: Google Scholar.js [hunk 9]
```diff
@@ -101,198 +112,96 @@ function getProfileResults(doc, checkOnly) {
...
 	else {
-		var related = ZU.xpathText(doc, '//a[contains(@href, "q=related:")]/@href');
-		if (!related) {
-			throw new Error("Could not locate related URL");
-		}
-		var itemID = related.match(/=related:([^:]+):/);
-		if (itemID) {
-			scrapeIds(doc, [itemID[1]]);
+		// Stand-alone "Article View" page
+		const pfName = text(doc, "#gsc_sb_ui > div > a");
```
dstillman, line 165: Clearer to not abbreviate. Letters are cheap.

Google Scholar.js
```diff
@@ -101,198 +112,96 @@ function getProfileResults(doc, checkOnly) {
...
+		let emuRowObj = parseArticleView(doc);
```
dstillman, line 166: Same. `Obj` is fine, but should spell out whatever `emu` is, unless this is actually about [emus](https://www.google.com/search?q=emus&tbm=isch).

Google Scholar.js
```diff
@@ -101,198 +112,96 @@ function getProfileResults(doc, checkOnly) {
...
-				});
-				translator.translate();
-			});
-		});
-	}
+/**
+ * Returns an emulated search URL for a GS search with the profile name as the
+ * search term.
+ * @param {string} pfName - Name of the profile's owner (e.g.,
+ * "Sebastian Karcher")
```
dstillman, line 188: - → `profileName`
- Remove name — don't need to use a real name here

dstillman (commented):
Haven't tried this or studied the logic, but some quick style/clarity/policy notes

Google Scholar.js
```diff
@@ -661,6 +570,630 @@ ItemFactory.prototype.saveItemCommonVars = function () {
...
+		item.complete();
+	});
+	return translator.translate();
+}
+
+function addAttachment(item, row) {
+	// attach linked document as attachment if available
+	if (row.attachmentLink) {
+		let	attachment = {
+			title: "Full Text (via Google Scholar)",
```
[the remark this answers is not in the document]
dstillman: Hmm, yeah. This is tough.

I think "Linked Document" is too vague. Linked from where?

For files from Unpaywall, we use "Submitted Version", "Accepted Version", and "Published Version", but we don't know what those are here. Could be "Google Scholar Linked Version", but that's not really appropriate for the same reason "Full Text (via Google Scholar)" isn't really appropriate — the fact that it came from Google Scholar isn't really relevant to what it is. "Available Version"? "Publicly Available Version"?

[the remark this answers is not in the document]
adam3smith: FWIW, I actually liked the (via Google Scholar) option because it allows quick searching/filtering by the attachment title and also makes clear to users why they get a 'wrong' (e.g., old working paper) PDF. Otoh, with "Available Version" it's unclear where it's available from and, where it's wrong, it looks like Zotero mishandled it.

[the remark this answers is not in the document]
dstillman: That's fair. I guess the practical aspect of this outweighs any conceptual objections to putting that in the title. OK, @zoe-translates, so we can revert to that.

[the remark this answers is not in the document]
dstillman: OK, sorry, let's go with `Available Version (Google Scholar)`. `Available Version via Google Scholar` is vaguely ungrammatical, but `Available Version` is nice and consistent, and `(Google Scholar)` makes sense as a sort of tag. Put them together and we have a winner.

Google Scholar.js
```diff
@@ -661,6 +579,646 @@ ItemFactory.prototype.saveItemCommonVars = function () {
...
+ * @constructor
+ * @param {string} searchKey - The property to be used for the ID in the
+ * item-stub object used as the basis of search (e.g. "DOI", "arXiv", etc.)
+ * @param {string} translatorUUID - The id of the search translator
+ * @param {boolean} [doAttach=false] - Whether we should use a workaround to
+ * save attachments
+ * @property {Z.Translate<Z.SearchTranslator>} translator
+ * @property {string} key
+ */
+function ExternalSearch(searchKey, translatorUUID, doAttach = false) {
```
dstillman, line 609: So echoing https://github.com/zotero/translators/pull/3062#issuecomment-1621544494, as a general rule, let's just say no constructors in translators. There are likely a few bad examples of this in existing translators (RIS comes to mind), but most translators just use functions and it is good and fine.

There's just too much engineering here. Translators need to be simple and easy to understand. There can't be constructors, methods, pipelines, factory functions, generators — it's just not appropriate for a translator. I understand that you're trying to do some things here that other translators don't have to deal with, but if the logic needs to be slightly less optimized to keep the translator simpler, so be it.

dstillman:
Yes, that's all correct.

Optimizing for concurrency is a worthy goal, but not at the expense of readability and maintainability. I suspect there are some simpler ways we can achieve some level of concurrency without so much extra machinery.

PR #3052, opened 2023-06-15
zbMATH: Updates and multiple fixes.

### LINKED ISSUE #3039: zbMATH does not record the url nor the Zbl identifier
The translator for zbMATH pages needs some serious update.
You can quickly check with scaffold that its "scrape" function does not find the Zbl identifier
(id variable is empty), even though it uses BibTex translator which knows it !
As a result, no url nor identifier is recorded in Zotero. I'm not conversant enough with java to
even suggest corrections, but it should be simple for someone who is.
Thanks in advance.
PS: btw, I added a modified translator in my Zotero data/translators directory, 
restarted, updated translators multiple times, to no avail...

zoe-translates:
- Updated the selectors/XPaths to match the current state of the site.
- Prefer selector to XPath to simplify code.
- Made the scrape()/doWeb() functions async.
- Changes to keyword/tag handling: the returned tags now contain MSC numbers, their readable labels, and the "Keywords" content.
- Strip the duplicated characters in MSC labels and abstracts that had been caused by inline MathML rendered by MathJax. In abstracts, the math content is replaced by their LaTeX annotation, surrounded by the dollar signs ($ $), to mark the places where math text appeared.
- Prefer the cleaner permalinks in URL fields.
- Updated test cases.

Resolves #3039

### DIFF
--- a/zbMATH.js
+++ b/zbMATH.js
@@ -177,6 +177,10 @@ function cleanText(element) {
  * the duplicate-causing elements removed and the LaTeX math text converted to
  * text nodes (surrounded with $ $ if laTeXify = true).
  *
+ * NOTE: If the node containes unrendered MathJax source, the surrounding
+ * delimiters \( and \) will be removed if laTeXify is false, or replaced with
+ * $ $ if laTeXify is true.
+ *
  * @param {HTMLElement} element - The DOM element node to operate on
  * @param {boolean} [laTeXify=true] - Whether to replace the MathML nodes with
  * LaTeX source surrounded by $ $. If set to false, the replacement will be the
@@ -194,10 +198,22 @@ function cleanupMath(element, laTeXify = true) {
 		node.parentNode.replaceChild(textNode, node);
 	};
 
-	// Top-level contentful MathML elements. If this doesn't exist, there's no
-	// math.
+	// Top-level contentful MathML elements.
 	let baseMathList = dup.querySelectorAll("span[id^='MathJax-Element']");
 	if (!baseMathList.length) {
+		// When we're saving one of the multiples from getSearchResults, even
+		// if there's math, MathJax code will not be run. In this case, we do a
+		// text replacement
+		dup.normalize();
+		let walker = doc.createTreeWalker(dup, 4/* SHOW_TEXT */);
+		let currentNode;
+		while ((currentNode = walker.nextNode())/* assignment */) {
+			currentNode.textContent
+				= currentNode.textContent.replace(
+					/\\\(|\\\)/g, // \( or \); MathJax source text
+					laTeXify ? "$" : "",
+				);
+		}
 		return dup;
 	}
 


zbMATH.js
```diff
@@ -158,16 +216,16 @@ var testCases = [
 				"date": "2012",
 				"DOI": "10.1002/rsa.20472",
 				"ISSN": "1042-9832",
-				"abstractNote": "We prove that a given tree TT on n vertices with bounded maximum degree is contained asymptotically almost surely in the binomial random graph G(n,(1+ε)lognn)G\\left(n,\\frac {(1+\\varepsilon)\\log n}{n}\\right) provided that TT belongs to one of the following two classes: \n\n(1)TT has linearly many leaves; (2)TT has a path of linear length all of whose vertices have degree two in TT.",
-				"extra": "MSC2010: 05C05 = Trees\nMSC2010: 05C80 = Random graphs (graph-theoretic aspects)\nZbl: 1255.05045",
+				"abstractNote": "We prove that a given tree $T$ on n vertices with bounded maximum degree is contained asymptotically almost surely in the binomial random graph $G\\left(n,\\frac {(1+\\varepsilon)\\log n}{n}\\right)$ provided that $T$ belongs to one of the following two classes: \n\n(1)$T$ has linearly many leaves; (2)$T$ has a path of linear length all of whose vertices have degree two in $T$.",
```
adam3smith, line 219: I don't quite know what to do about this -- we don't actually support TeX in Zotero fields (other then the new notes), so this is a bit messy, but I'm also not sure what else we could do.

zbMATH.js
```diff
@@ -35,105 +35,163 @@
...
+		item.complete();
 	});
+
+	trans.translate();
 }
 
+// Clean up the MathJaX-rendered text in elements. Returns a clone of the node
+// with the duplicate-causing elements removed and the LaTeX math text
+// converted to text nodes (surrounded with $ $ if laTeXify = true).
+function cleanupMath(element, laTeXify = true) {
```
adam3smith, line 165: I'd want to hear from @dstillman or @AbeJellinek what Zotero's view is on handling LaTeX/MathJaX in fields. It's currently not supported, so adding things like $$ doesn't do any good, but given the nature of the translator it might still make sense?

zbMATH.js
```diff
@@ -35,105 +35,163 @@
...
+	let rows = doc.querySelectorAll(".content-result .list > article");
+	for (let row of rows) {
+		let href = attr(row, ".title a", "href");
+		let title = ZU.trimInternal(text(row, ".title"));
+		if (!href || !title) continue;
+		if (checkOnly) return true;
+		found = true;
+		items[href] = title;
+	}
+	return found && items;
```
adam3smith, line 61: Why are you returning both here? The standard would be `	return found ? items : false;`

zbMATH.js
```diff
@@ -35,105 +35,163 @@
...
-					item.place = item.publisher.substr(0, publisherSeparation);
-					item.publisher = item.publisher.substr(publisherSeparation + 1);
-				}
+async function scrape(doc, url) {
+	let bibURL = attr(doc, ".bib", "href");
+	if (!bibURL) {
+		Z.debug(`Error: document at ${url} does not contain bibTeX link`);
+		return;
+	}
+	let bibTeXDoc = await ZU.requestText(bibURL);
```
adam3smith, line 83:
```suggestion
	let bibTeXDoc = await requestText(bibURL);
```
No need for `ZU.` for the newer translator utility functions

adam3smith (changes requested):
A couple of small things -- I'd want dstillman or AbeJellinek to chime in on the handling of LaTeX in fields

zbMATH.js
dstillman: Yeah, I think this is a reasonable approach. And, honestly, we could probably support math in abstract fields pretty easily (just showing as `$…$` in edit mode).

dstillman:
> This will go into the full-text PDF filename. So we'd prefer less special characters such as $ or \

That doesn't matter. Zotero will automatically remove any characters that aren't valid in filenames.

dstillman:
Ah, got it.

adam3smith:
I don't love identifiers in callNumber fields in the first place. We are typically putting identifiers (as opposed to actual call numbers) into Extra. To make sense in Extra, they do need the prefix
arxiv: is part of the Archive ID because arXiv includes arXiv as part of the identifer. Generally, where identifiers aren't otherwise self-identifying (like, say, DOIs), it makes sense to store them with a namespace-y prefix.

PR #3053, opened 2023-06-15
Web of Science Tagged: More robust against trailing space characters.

zoe-translates:
If there are trailing whitespace characters after the EF "tag", the script fails because it interprets "EF" as an unknown tag rather than end of file.

This is fixed by interpreting EF as intended, see the table of export field tags at Clarivate:
https://webofscience.help.clarivate.com/en-us/Content/export-records.htm#mc-dropdown-body6dc16a03-4cb8-43e1-b72a-ff5a3197f8f0

Bug reported by gxy1932582, see
https://forums.zotero.org/discussion/comment/436794/#Comment_436794

### DIFF
--- a/Web of Science Tagged.js
+++ b/Web of Science Tagged.js
@@ -8,7 +8,7 @@
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 1,
-	"lastUpdated": "2023-06-15 10:22:48"
+	"lastUpdated": "2023-06-15 10:54:59"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


Web of Science Tagged.js
```diff
@@ -229,6 +229,11 @@ function doImport(text) {
 	while ((rawLine = Zotero.read()) !== false) {    // until EOF
 		//Z.debug("line: " + rawLine);
 		let split = rawLine.match(/^([A-Z0-9]{2})\s(?:([^\n]*))?/);
+		// EF is equivalent to the end of file.
+		if (/^EF\s*/.test(rawLine)) {
```
dstillman, line 233: The `\s*` isn't doing anything here.

Web of Science Tagged.js
```diff
@@ -229,6 +229,11 @@ function doImport(text) {
...
+			Z.debug("Encountered EF (equivalent to End of File)");
```
dstillman, line 234: Comment out

Web of Science Tagged.js
dstillman: At least assuming every line is guaranteed to start with a tag… How are multi-line fields (e.g., multi-line notes) handled?

dstillman:
trimEnd() isn't available in Zotero 6, so we can't use it yet

https://github.com/zotero/translators/commit/cc7c3c75a068aab2c2aa087f3ecbc72cf8eb72d4

PR #3054, opened 2023-06-15
Import html rich text as TEI elements

glorieux-f:
Tested with malformed html in Zotero fields.


TEI.js
```diff
@@ -59,17 +59,27 @@
...
+const ns = {
 	tei: "http://www.tei-c.org/ns/1.0",
 	xml: "http://www.w3.org/XML/1998/namespace"
 };
 
 
-var exportedXMLIds = {};
-var generatedItems = {};
-var allItems = {};
+const exportedXMLIds = {};
```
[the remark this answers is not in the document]
glorieux-f: Hi Zoë,

Will you stay the person reviewing my propositions ? In fact I have completely rewritten TEI.js now
[new TEI.js](https://github.com/glorieux-f/translators/blob/tei_export/TEI.js)
There are some other features (more TEI conformant, more zotero fields supported, indentation…). I tried to interest other people on that, [not a success](https://forums.zotero.org/discussion/105716/xml-tei-translator-what-are-your-most-complex-records#latest) :o)

I will carefully follow all your advice for the future. I’m not a git guru for pull requests.

I’m coming from Java. I use const to protect a name. Even if JS is poor typing, exportedXMLIds will at least, stay an object (used as an HashMap). Bad practice ?

TEI.js
```diff
@@ -59,17 +59,27 @@
...
+const generatedItems = {};
+const allItems = {};
+
+
+// build one time
+const xmlparser = new DOMParser();
+const xmlser = new XMLSerializer();
```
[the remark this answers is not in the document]
glorieux-f: Is RDF.js a good example to follow ?
“ser” is (was ?) a common shortcut for serializers in XML dev, but my naming style has probably too much decades now. OK.

TEI.js
```diff
@@ -638,11 +689,14 @@ function doExport() {
 		outputElement = listBibls[0];
 	}
 	else {
-		outputElement = teiDoc.createElement("empty");
+		outputElement = teiDoc.createElement("empty");
```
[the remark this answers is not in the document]
glorieux-f: Sorry, I’m not well configured for JS, I will search for.

TEI.js
```diff
@@ -94,6 +104,50 @@ function replaceFormatting(title) {
...
 
+/**
+ * Append possible rich text html as TEI element (<hi rend="…">)
+ * @param {*} node 
+ * @param {*} html 
+ */
+function appendXML (node, html) {
+	if (!html) return;
+	// import html as dom and export is as xml to avoid xml malformations
+	let dom = xmlparser.parseFromString(html, "text/html");
```
dstillman, line 115: This won't work. These fields aren't HTML and can't be parsed as HTML. They just use a subset of HTML for specific rich-text formatting. I'm not totally clear what this PR is doing, but if the goal is to treat Zotero fields as unrestricted HTML, it's not valid and should be closed.

`replaceFormatting()` was technically already wrong here. Zotero's citation processor ignores unmatched tags, so a title `<b> is not <strong>` is treated literally without interpreting as rich text. In other words, single regexes should be matching ungreedily on each opening and closing tag and capturing the value in between, instead of there being separate regexes for the opening and closing tags. It would be good to fix this, though it probably doesn't come up much.

But what you have is much worse, because it's treating the string as full HTML. So `<b> is not <strong>` would get converted to:

```xml
<title level="a">
	<hi rend="bold"> is not <strong/>
	</hi>
</title>
```

which we obviously don't want.

Everything that's not the specific supported matching tags should be encoded as plain text. If that's what the translator was doing before, it was already correct (other than replacing non-matching tags).

TEI.js
```diff
@@ -59,17 +59,27 @@
...
+const xmlser = new XMLSerializer();
 
-// replace formatting with TEI tags
+/**
+ * Replace formatting with TEI tags.
+ * [2023-06 FG] Legacy, @see appendXML(), to have formatting as element node
+ * @param {*} title 
+ * @returns 
+ */
 function replaceFormatting(title) {
```
dstillman, line 83: If you're replacing something, you should replace it — not just leave behind an unused copy with duplicate logic.

TEI.js
glorieux-f: Hi Dan,

«This won't work.» It works on some hundreds of real life records, for example on [this group](https://www.zotero.org/groups/5018778/rougemont/library), but not on your case, you are right.

Users with brackets in their titles will have some hand work, like for all TEI edition (ex: italic in a title is usually a title, Zotero can’t infer it, a TEI editor should), time is saved for most common cases. That said… if TEI.js is reworked, let’s try to do the best.

Have you some clue to find the code doing this nice job in Zotero desktop ?
image

The RDF solution is not an option for TEI. This is not semantic tagging. The original author of TEI.js, (C) 2010 Stefan Majewski, was right from a TEI point of view, even if his solution is not yet perfect.
`Some people put &lt;b&gt;bold&lt;/b&gt; in &lt;i&gt;titles&lt;/i&gt; ?`

TEI.js
glorieux-f: I thought it was easier for reviewers to compare with the old code, it is of course deleted on my branch, quite far from that now. This little PR is a first contact with you, to learn your good practices.

TEI.js
```diff
@@ -94,6 +104,50 @@ function replaceFormatting(title) {
...
+	// import html as dom and export is as xml to avoid xml malformations
+	let dom = xmlparser.parseFromString(html, "text/html");
+	html = xmlser.serializeToString(dom.getElementsByTagName("body").item(0));
+	// transform html to tei, xslt not available
+	// https://forums.zotero.org/discussion/comment/344940/
+	// tags supported by Zotero
+	// https://www.zotero.org/support/kb/rich_text_bibliography
+	let xml = html
+		// changing namespace of root element
+		.replace(/http:\/\/www.w3.org\/1999\/xhtml/g, 'http://www.tei-c.org/ns/1.0')
```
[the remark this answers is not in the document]
glorieux-f: You think about something like that in Endnote XML.js ?
https://github.com/zotero/translators/blob/master/Endnote%20XML.js#L896

@dstillman has seen a more critic point to resolve before.
https://github.com/zotero/translators/pull/3054#discussion_r1252591654

If a good direction can be given on that topic, I will be able to propose a new PR, just on that topic for now.

TEI.js
glorieux-f: This will be done for next submission, if I can solve your point upper.

TEI.js
dstillman: > «This won't work.» It works on some hundreds of real life records, for example on [this group](https://www.zotero.org/groups/5018778/rougemont/library), but not on your case, you are right.

I don't mean "won't work" as in it won't produce the correct result in many cases. I mean it doesn't work conceptually. It's not an HTML field and therefore can't be treated like it is.

> Have you some clue to find the code doing this nice job in Zotero desktop ?

https://github.com/zotero/zotero/blob/04496fadcd73ad7db85d7a9c662fcbffac80d72e/chrome/content/zotero/itemTree.jsx#L2570-L2658

glorieux-f: NICE ! Thanks. Because it is documented as “Private methods”, should I guess the good approach is to acclimate such code in TEI.js ? It seems that this PR needs a full rewrite. What should I do ? Submit another PR when work is done ?

dstillman: Nothing to do with "private" — that code just isn't remotely related to the translator environment. (Most zotero/zotero code has no connection to translators.)

Sure, probably easiest to close this and open a new one when want us to look at something new. Thanks!

glorieux-f: Hi @dstillman,
I had to continue other works on TEI.js (other features features like indentation, or more fields supported especially some extra used by APA7…). I have now acclimate the piece of code you point to me. No more regex are going through html.
https://github.com/glorieux-f/translators/blob/tei_export/TEI.js#L140
May I submit this new version ? It’s quite a full rewrite of previous version (~10 years). Feel free to ask for more upgrades to conform the code with present Zotero standards.

dstillman: Sure!

glorieux-f: There is maybe a place where I’m not perfect, parsing the Extra field. You probably know where is the piece of code doing this good job before sending to Style. 
https://github.com/glorieux-f/translators/blob/tei_export/TEI.js#L307
Other thing, I don’t know how to work well with Pull Request. Isn’t it your responsibility to close it ?

dstillman: You can certainly always close your own PRs, but I'll close this one.

We can talk about Extra on the new PR.

PR #3059, opened 2023-06-16
Update datacite names & dataset handling

adam3smith:
https://forums.zotero.org/discussion/105639/bad-name-handling-when-adding-zenodo-items-by-doi#latest


adam3smith:
@dstillman @AbeJellinek it does seem like CI hangs up & cancels after 10mins for all PRs with an import or search translator included. It'd be great to get this to work again.

PR #3062, opened 2023-06-20
WoS Tagged: Reimplement the core algorithms.

zoe-translates:
The previous iteration of the import translator for Web of Science Tagged Format is becoming a bit difficult to maintain because of the tight coupling of logic, data, and actions.

To improve the separation of concerns, in the updated version several measures have been taken:

- Use separate methods for distinct functions such as line parsing/validation (purely formal text-processing), intermediate transformations (data normalization based on semantics, etc.), and data-to-item mapping.
- Use a lookup-table approach to help with mapping tagged data to object properties, which is easier to debug/maintain. It can also better support "polymorphism", the unfortunate fact that the same tag can mean different things depending on the item type.
- Implement better text-processing (e.g. collapsing spaces when necessary, cleaning up line noise, more robust handling of author names, etc.)

Overall the goal is to make the translator more robust and easier to reason with.

TODO: To ensure best compatibility of behaviour, none of the test cases has been updated for now. After this commit, I'll introduce temporary devices to ensure the new code, with its new underlying structure, produces the same output as the old one (to the point of bug-compatible except for the most egregious). When that is achieved, the temporary compatibility devices will be removed, and the test cases will be updated and manually verified. After that, the new test cases will serve as the basis for incremental improvements of the new code.

### DIFF
--- a/Web of Science Tagged.js
+++ b/Web of Science Tagged.js
@@ -120,7 +120,6 @@ var TAG_HANDLERS = {
 
 // Tags whose values can be assumed to be short enough (i.e. fit on a line) and
 // "atomic" (denoting one entity rather than an array thereof), such that a simple assignment will suffice
-// TODO: Further normalization
 var SIMPLE_FIELDS = {
 	AR: "pages", // article number
 	AW: "url",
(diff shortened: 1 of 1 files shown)


dstillman:
I'm afraid this is just way too complicated. This translator went from being trivial to understand to being something that someone would have to study deeply to figure out what it's doing.

Translators really aren't supposed to be complicated programs in their own right where someone stepping into the code needs to figure out some custom software architecture each time — they're supposed to be simple, easy-to-understand, fairly procedural bits of code that someone uninitiated and/or inexperienced can make tweaks to in a few minutes. If the question is ever "should I do something clever/elegant/object-oriented/formal/functional-programming-based or should I not do that?", the answer should almost always be "not do that".

That certainly applies to creating an `ItemMap` object with tons of methods, and it possibly even applies to the `getArrayJoiner()`/etc. stuff. Maybe the latter somehow makes things much easier — I haven't studied this (which is sort of the point) — but if I'm someone fixing this translator a year from now, I feel like I'd much rather see this:

```
else if (field == "PU") {
		item.publisher = content;
```

than this:

```
PU: getArrayJoiner("publisher", true, true), // mitigate all-caps
```

and then have to figure what in the world `getArrayJoiner()` is.

Bug fixes are great, but the previous translator was written exactly how almost all our other translators are written, and unless there's an extremely compelling reason to change it, it should stay that way.

dstillman:
Much better!

PR #3066, opened 2023-06-25
ePrint: fix lint errors, warnings and a broken test

alex-ter:
I have some ideas for refactoring/enhancing/modernizing this translator in mind, but as this is my first contribution (and hopefully not the last!), let me start with this simple one to get the hang of the review process.

Notable items:
1. One of the lint complaints was absence of the license block, so I've added one. Two things here:
   1. I put the currently listed author in the copyright message + mentioned "Zotero IACR ePrint translator contributors" to cover everyone else, not sure if this is okay or not. I can extract the data from past commits to make the list more specific if needed - please let me know.
   1. The year range is based on the git history, let me know if e.g., only the current one is preferred
1. There was one broken test (the entry title and abstract have changed), I've updated it and bundled it into the same commit, which seems to be ok based on what I saw in other commits. Please let me know if you'd like me to split it out into a separate commit for better logical separation.

While at it, for my future refactoring I wanted to clarify the project's stance on `var` vs `let`/`const` - is replacing one with the other discouraged, encouraged, tolerated, something else?

### DIFF
--- a/ePrint IACR.js
+++ b/ePrint IACR.js
@@ -9,34 +9,61 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2022-10-28 16:15:17"
+	"lastUpdated": "2023-06-25 18:34:47"
 }
+
+/*
+    ***** BEGIN LICENSE BLOCK *****
+
+    Copyright © 2010-2023 Jonas Schrieb and Zotero IACR ePrint translator contributors
+
+    This file is part of Zotero.
+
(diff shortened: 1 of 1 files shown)


adam3smith:
Yeah, so for the process -- we generally discourage updates just to fix linting, there's little reason for that.
If you make relatively minor changes, we suggest putting the linting in one commit and then the changes in another one, for faster review.
In this case, though, you're basically going to re-write most of the translator, so I'd just go ahead and do that. Take advantage of the templates built into Scaffold/the translator editor.

PR #3075, opened 2023-07-09
Update call number field in Primo Normalized XML

malkaPre:
Before fix:
Call-number field was taken from '//p:display/p:availlibrary' root in the pnx or '//p:browse/p:callnumber' root in the pnx
and that matched primo pnx format.
In primoVE the call-number field exists in a differnt root '//p:delivery/p:bestlocation/p:callNumber' in the pnx.

After fix:
if call-number is empty it will fetch from '//p:delivery/p:bestlocation/p:callNumber'.

### DIFF
--- a/Primo Normalized XML.js
+++ b/Primo Normalized XML.js
@@ -314,14 +314,6 @@ function doImport() {
 			}
 		}
 	}
-	if (!callArray.length) {
-		callNumber = ZU.xpath(doc, '//p:delivery/p:bestlocation/p:callNumber', ns);
-		for (let i = 0; i < callNumber.length; i++) {
-			if (callNumber[i].textContent.search(/\$\$2.+\$/) != -1) {
-				callArray.push(callNumber[i].textContent.match(/\$\$2\(?(.+?)(?:\s*\))?\$/)[1]);
-			}
-		}
-	}
 	if (callArray.length) {
 		// remove duplicate call numbers
 		callArray = dedupeArray(callArray);
--- a/Primo Normalized XML.js
+++ b/Primo Normalized XML.js
@@ -314,6 +314,14 @@ function doImport() {
 			}
 		}
 	}
+	if (!callArray.length) {
+		callNumber = ZU.xpath(doc, '//p:delivery/p:bestlocation/p:callNumber', ns);
+		for (let i = 0; i < callNumber.length; i++) {
+			if (callNumber[i].textContent.search(/\$\$2.+\$/) != -1) {
+				callArray.push(callNumber[i].textContent.match(/\$\$2\(?(.+?)(?:\s*\))?\$/)[1]);
+			}
+		}
+	}
 	if (callArray.length) {
 		// remove duplicate call numbers
 		callArray = dedupeArray(callArray);


malkaPre:
In this URL when you save to Zotero and show empty fields, call number field is empty:
https://quicksearch.lib.iastate.edu/discovery/fulldisplay?docid=alma9920251487302756&context=L&vid=01IASU_INST:01IASU&lang=en&search_scope=MyInst_and_CI&adaptor=Local%20Search%20Engine&tab=Everything&query=any,contains,fred&offset=0

malkaPre:
Is there any updates on this?

malkaPre:
@AbeJellinek  I checked all the comments - and I did whatever was requested:
1. Updated the translator files using Scaffold.
2.  Gave an example where it used to fail without the change.
3. You agreed that the following syntax is okay.
https://github.com/zotero/translators/pull/3075#discussion_r1262942934
image

Am I missing out on something?
If yes, I would like you to explain to me what.

malkaPre:
The customer is requesting for updates. Can you please review and approve the change being requested

nilinatan:
Hi @AbeJellinek @zoe-translates 

We would like your help in closing this issue, since we have Primo customers that are waiting for this fix.
@malkaPre is encountering with issues while testing using the Scaffold.

We have already implemented  your remarks on the regex.
please help to optimize the process as we do want to commit the change and provide the fix.

please suggest if possible options to help and speed this up.

All the best
Nili Natan
Primo, Product Manager 
Nili.Natan@Clarivate.com

smachefert:
> > @malkaPre is encountering with issues while testing using the Scaffold.
> 
> What are those issues?
> 

I can't speak regarding the other issues / discussions but I believe there is in fact an issue with Scaffold regarding the way it handles the detailed records pages from Primo. I have been trying to have a look at this issue with call numbers and am encountering the same issue when trying to create a new test case (with or without the changes, it does not work even with the unchanged current translator). 

This is for example a page that currently does not have a call number in Zotero but should and that would be used as a new test :

https://bcujas-catalogue.univ-paris1.fr/discovery/fulldisplay?context=L&vid=33CUJAS_INST:33CUJAS_INST&search_scope=MyInstitution&tab=LibraryCatalog&docid=alma990004764520107621

If I load this url under the browser tab of scaffold then load "Primo 2018" translator. I can "Run detect*" → book, then "Run do*" → full item returned.

If from the browser tab, I click on "Create Web Test", a new test is added to my tests. If I go to the tests tab then click on the Run button for this test, I get a message "16:54:22 TranslatorTester: Primo 2018 Test 1: failed (Detection failed)"

I am eager to provide fix to the others comments but I have not been able so to create new tests cases to prove this fix is working fine. Any idea @AbeJellinek why this test case is now working properly when manually testing is working fine?

timtomch:
Just a note that I've also tried resolving this issue and adding automated tests but am running into the same problem that @smachefert is highlighting (Detection failed error even though manual tests are fine).

adam3smith:
Have you tried `defer: true` on the test?

timtomch:
I have but I'm still getting that issue. Possibly I'm doing something else wrong.

Screen Shot 2024-09-16 at 09 15 15

adam3smith:
(you should run Zotero 7 , which I think you aren't?) 
OK, I can reproduce that on the iastate Primo, including when using the permalink for items instead of the full display (e.g. https://quicksearch.lib.iastate.edu/permalink/01IASU_INST/1s27pim/alma9921367771002756 ) Abe would need to look at that. 

That said -- tests aren't always possible anyway (some sites don't have any permalinks) so you can absolutely proceed with implementing changes and just provide a sample URL in a code comment.

timtomch:
Indeed, that screenshot was from an older version that I tried to see if the error would go away. As you note, it's still there in Zotero 7.

I believe @smachefert has some fixes ready to submit.

smachefert:
@adam3smith the newly opened PR #3360 seems to work for me. Per your suggestion I have removed the test that were not working properly. And added the XML directly in the import translators, allowing us to test that everything works fine, even if the test does not work from Primo 2018.js. 

Let me know if you see any improvement that should be done before merge is possible.

symac:
Thanks for the explanation regarding when Primo creates the elements. I think we can live without an emulation for this event, especially as a new version of Primo is on the way and the behavior might change.

PR #3077, opened 2023-07-12
[HotFix] HAL: Hotfix for identification failures

### LINKED ISSUE #3041: Importing Reports from HAL
When importing a report from HAL (example https://hal.science/hal-01512585v1 ), they are inappropriately imported as a thesis.

### LINKED ISSUE #3045: City not imported when importing an oral communication from HAL
Example: https://hal.science/hal-04120705

When importing such reference, the city of the conference is not imported

zoe-translates:
The HAL translator did not work at all with the new HAL site at a different domain.

- Update target regex for the new domain `hal.science`
- Update the element selector for the content-type label
- Do some cleaning with raw content-type label text; update some labels
- Fix a large number of ESLint errors

Resolves #3041 (Edit: also resolves #3045)

### DIFF
--- a/HAL Archives Ouvertes.js
+++ b/HAL Archives Ouvertes.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2023-07-12 07:30:59"
+	"lastUpdated": "2023-07-12 08:47:33"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


HAL Archives Ouvertes.js
```diff
@@ -90,34 +98,33 @@ function doWeb(doc, url) {
 			for (var i in items) {
 				articles.push(i);
 			}
-			Zotero.Utilities.processDocuments(articles, scrape)
+			Zotero.Utilities.processDocuments(articles, scrape);
+			return true;
```
dstillman, line 102: We don't actually check the return value of the `selectItems` callback (anymore?), so the `consistent-return` fix here is just to change the first `return true;` to `return;`.

Issue #3067, opened 2023-06-30
EBSCOhost: Asyncify


dstillman:
https://forums.zotero.org/discussion/105903/page-not-working-with-zotero-connector

Looks like we're calling processDocuments in a loop, resulting in a 429

dstillman:
If you want to try anyway, I can test it and make final tweaks.

dstillman:
That's fine to start. It's the burst that's the worst.

PR #3078, opened 2023-07-14
EBSCOhost: Asyncify network requests and use native API for encoding

zoe-translates:
- Convert the callback-based processDocuments and doGet into promise-based replacements.
- Remove the custom implementation of btoa() and related functions; replace them with built-in equivalents.

This addresses #3067.

@dstillman, I did a blind transformation of the code into the async style. Could you please test?


adam3smith:
I'm currently not getting PDFs on EBSCO (with a 404), though I think this may also affect the old version. Might be worth trying to fix (though I realize Zoe can't do that without access)

PR #3079, opened 2023-07-16
ePrint: refactor to modernize, fix issues, and add functionality

alex-ter:
This is a major refactor of the translator. Main changes:

* Fix lints and bugs, asyncify and restructure according to current translator best practices
* Add processing of all search types (except "Compact view", due to performance reasons)
* Properly populate Archive, Archive ID, Library Catalog, Note fields and add useful information from "Publication info" item into Extra field
* Update existing and add new tests for added functionality

Notes and opens (the latter are also marked with TODO in the code, will remove those when clarified). While clarifying why I went the way I did, I'm open to alternative approaches.
* This is a follow-up to my initial PR #3066 and our discussion there. Given the number of changes, the diff may not be as helpful as usual, just as a fair warning. The changes were intertwined enough so I couldn't come up with any meaningful split into commits and went with a single one.
* I went with straight `https://` as (a) `http://` doesn't exist anyway (redirects to `https://`) and (b) in this day and age, it's highly unlikely ePrint will ever get back to plain HTTP.
* I watched several discussions in other translator-related PRs and tried to keep the code simple and well-commented for better maintainability. As suggested by dstillman in one of the recent reviews, simplicity had priority, so e.g., I kept `getTextSearchResults()` and `getTimeSearchResults()` separate due to different looping logic and selectors. It could be solved, but alternatives looked more convoluted, and the code duplication is not huge.
* Right now Archive and Library Catalog field values vary slightly (the latter has ePrint FQDN appended). Looking through the translators, I see every possible combination implemented for these two fields, so I went with my understanding based on what I saw in other translators, reading the `preprint` item type PR discussion and the item schema description.
* Can we get rid of the `preprint` vs `report` item type check [here](https://github.com/zotero/translators/compare/master...alex-ter:eprint-refactor?expand=1#diff-546a2c246c178f91aaf2791151cf24aad81c586ffa17f3e9a7c27ecadb364d76R39)? The former was added approx. a year ago, so maybe a translator minimum supported version bump would be enough?
* The trick with processing direct links to PDFs [here](https://github.com/zotero/translators/compare/master...alex-ter:eprint-refactor?expand=1#diff-546a2c246c178f91aaf2791151cf24aad81c586ffa17f3e9a7c27ecadb364d76R221) does not really work (see the TODO comment there). In essence, a built-in PDF saving logic seems to take precedence even though [the code](https://github.com/zotero/zotero-connectors/blob/master/src/browserExt/background.js#L603-L611) seems to be properly putting translators first. Is this some known issue? Am I missing something here? As implemented now, that logic doesn't work, so I'm thinking of either fixing or removing it altogether.
   * ADDED LATER: this may be some sort of Firefox-specific bug or behavior, looks like Dan mentioned something very similar in a [recent forum post](https://forums.zotero.org/discussion/comment/438395/#Comment_438395). I used Firefox so far, will check it out in Chrome - but the Scaffold-related piece still stays anyway.

Side note - [translator test results page](https://zotero-translator-tests.s3.amazonaws.com/index.html) referenced in [the docs](https://www.zotero.org/support/dev/translators/testing#automated_testing) seems to be broken (lists no tests). Is that kind of testing no longer used?

### DIFF
--- a/ePrint IACR.js
+++ b/ePrint IACR.js
@@ -208,7 +208,8 @@ async function doWeb(doc, url) {
 			await scrape(await requestDocument(url));
 		}
 	}
-	// TODO: This is never called except if URL value is modified in the function directly to append ".pdf". Instead, the PDF link gets directly opened by the system PDF reader when in Scaffold (and e.g., the test times out) and in browser the standard "Save to Zotero (PDF)" icon appears instead.
+	// Firefox restrictions will prevent this convenience clause from running (Chrome works as expected)
+	// Standard PDF saving (and metadata retrieval) logic will be used instead
 	else if (/\.pdf$/.test(url)) {
 		// Go to the landing page to scrape
 		url = url.replace(/\.pdf$/, "");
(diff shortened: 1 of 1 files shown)


dstillman:
Firefox restrictions prevent us from running regular site translation on a PDF URL the way we can in other browsers, so only PDF saving without a popup is offered on PDF URLs. If you're using Firefox, you'll need to save from the main webpage or rely on PDF metadata retrieval.

PR #3081, opened 2023-07-19
NBER: add support for Series Number field

chris-vecchio:
This pull request adds support for populating the Series Number field.

### DIFF
--- a/National Bureau of Economic Research.js
+++ b/National Bureau of Economic Research.js
@@ -151,6 +151,7 @@ function scrapeWithBib(doc, url, bibURL) {
 			if (item.itemType == 'report' && detected == 'preprint') {
 				item.itemType = 'preprint';
 				item.archiveID = item.reportNumber;
+				item.seriesNumber = item.reportNumber.replace(/^w/, '');
 				delete item.reportNumber;
 			}
 


adam3smith:
What's the goal of duplicating the ID information? (If we were to accept, this would need a bumped timestamp and updated tests, but let's first clarify conceptually)

chris-vecchio:
Thanks for the quick reply! Data librarians at the Cleveland Fed have scripts that run on multiple sources and use the Series Number field to get the series number. They would prefer not to modify their scripts for an individual case. Thanks!

adam3smith:
Hmm -- that's a bit of an idiosyncratic reason. I like data librarians as much as anyone, but "a small group of people doesn't like it" is not a good reason to change this in Zotero. The question is what makes most sense in the context of the Zotero metadata model. My view is that repo & id is for preprints & working papers published on a given platform: arXiv + arXivID, SSRN + SSRN ID, NBER + NBER ID, etc. That's why this appears as above. Series and series number are for series that are separate from the platform on which they're published, such as "[Bielefeld Working Papers in Economics and Management No. 24-2013"  for https://papers.ssrn.com/sol3/papers.cfm?abstract_id=2362788

PR #3090, opened 2023-07-28
Add doctoral thesis to type map

mvolz:
Map doctoral thesis (https://www.wikidata.org/wiki/Q56462344) to type thesis in type map


dstillman:
Thanks!

PR #3097, opened 2023-08-04
Open WorldCat: Fall back to archived build ID (and asyncify)

AbeJellinek:
We try Google's cached version of the homepage, then Archive.org's, which should both be relatively recent. Old build IDs should remain valid for a little while (that's why Next.js uses them AFAIK - it lets old cached versions of the static frontend continue fetching server-side data for a while after a backend update). This does run the risk of breaking translation if the build ID expires before the cache is updated, but that's better than translation always failing for people who get 403'd by Cloudflare.


dstillman:
Merged in f3d77f2327149d76fbfe69c27e09dcc881db4832 and 27f711219b2762488fb7cf0ebd9b6384f4e35e7e

PR #3098, opened 2023-08-04
XML/TEI for titles and notes, better TEI conformance, more fields supported, indentation, refactoring

glorieux-f:


glorieux-f:
Hi Zotero team,
2 new things from last submission

- eslint OK
- html tags in fields handled like in Zotero
- robust handling of extra field, inspired from Zotero source

Even if it’s not perfect, it is surely a lot better than the actual official version. Be sure I’m a TEI guru for now 20 years.

glorieux-f:
Try to put it upper on the stack.

PR #3130, opened 2023-09-05
DOI web translator: Offer the current page as a choice among the multiple

### LINKED ISSUE #3126: DOI: Offer current webpage as first option if not present in DOI results
https://github.com/zotero/zotero-connectors/issues/392#issuecomment-1598564174

zoe-translates:
If the DOI translator turns out to be the one that matches the current page, and if the DOI (this) translator detects the item type as "multiple", offer the current page as one of the choices (indeed, the first choice).

In the item selection dialog, this special choice will show up as "Current Webpage", followed by the page title in parenthesis if the title is present.

The specific situation that triggers this behavior is the following

1. None other than the DOI web translator (lowest priority) matches the current page
2. The current page itself is not the main resource identified by DOI present on the current page (e.g. a page with its own DOI in URL)

Resolves #3126.

### DIFF
--- a/DOI.js
+++ b/DOI.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2023-10-18 11:11:59"
+	"lastUpdated": "2023-10-18 13:29:55"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


dstillman:
> Webpage -> Web Page, I think. We use that spelling in the Connector ("Web Page with Snapshot") and in the name of the item type.

I've been meaning to change that for about a decade, so "Webpage" is fine, but it should really be a localized string from the connector. Can we just return the current URL and leave it to the connector to sub in a different string there if it matches?

dstillman:
> Until the connector can do the substitution, it will fall back to displaying the URL, which may look a bit weird but could be a bigger problem with screen readers.

It should just be the current page title, no?

dstillman:
Unless I'm misunderstanding something, it doesn't need to do anything other than put the current page first.

dstillman:
I serialized the DOI Content Negotiation requests. Can you rebase on the latest DOI.js?

PR #3151, opened 2023-09-26
IEEE Xplore: Don't use export format; save full text PDF

zoe-translates:
The "endpoint" for export formats (BibTeX, RIS, etc.) may be unreliable unless the referrer is sent, but this is not yet feasible for Chromium-family browsers in Connector.

(See: https://forums.zotero.org/discussion/108011/ieee-xplore-falied; cf. #3150)

This is changed to a scraper that looks at just the page content, therefore avoiding the endpoint call.

The pages are rendered on the client side, so we read the data source as JSON directly, without scraping the DOM.

In addition, the following changes are made --

- The full-text PDF file is now saved as an attachment (if the user has access to it).
- The translator now also works on IEEE Xplore's own full-text PDF viewer web app. This is done via an additional network request to the abstract page where the JSON metadata are hosted. A test case for this is added.
- Tags (keywords) are deduplicated.
- Only the electronic-media ISBN and ISSN are kept.
- Other minor bugfixes.

### DIFF
--- a/IEEE Xplore.js
+++ b/IEEE Xplore.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2023-10-09 10:41:18"
+	"lastUpdated": "2023-10-10 02:08:12"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


dstillman:
We're fixing the Referer thing (for now) in the Connector. Do we still want to do this if Referer works? It sounds like there are improvements, but could those have been made to the API-based version? Presumably this is more fragile?

dstillman:
> I could've sworn we had code in the client to extract PDFs from iframes, but I can't find it now.

We have code in the Connector that (if I recall) shows the PDF icon if there's a PDF in a frame and the parent page doesn't detect (and maybe that puts PDF as an option in the Save to Zotero menu even if there is a translator?).

PR #3157, opened 2023-10-10
Fix J-Stage translator to handle Japanese names properly.

aldente-hu:
J-Stage translator handles Japanese names improperly, treating their first name as their last name and vise versa.
This problem is caused by the BibTex records of J-Stage. So I've fixed it to look at the RIS records instead.


adam3smith:
I think the goal should be to provide optimal output with Zotero's native format (i.e. CSL via citeproc-js). 
I'd have to recheck (there are also some threads on the forum), but IIRC, citeproc-js handles Japanese names (as identified by Kanji) correctly when the family name is entered as last name. That's then the name that gets printed in form="short" but (and that's the crucial citeproc-js part), it prints the family name (i.e., Zotero last name) first when rendering the full name. 
I don't know ottomh how exactly that's determined, e.g. ,if it also applies to Latin script names with jp in the language fields (I think?) and other languages (like Hungarian). I'm pretty sure it doesn't handle the complexity of names in India at all.

The standard recommendation, [e.g. by W3C](https://www.w3.org/International/questions/qa-personal-names#splitting), is to use family and given/other when splitting names is necessary (as it is for citation generation), so Zotero's field naming isn't idea here. I'm not sure if renaming these in the UI is up for grabs or if there are concerns about confusing users.

aldente-hu:
@zoe-translates 
Talking about the particular case, I want Zotero to treat our(=Japanese people's) family name as `lastName`.

This is because:

It seems many(all?) parts of Zotero regard `lastName` field as their family name. For instance, Zotero displays 'Creators' name with their `lastName`, and uses their `lastName` field when citing their articles with Harvard referencing style. If our first/given names were stored in `lastName` field, Zotero would be very troublesome (and might be useless) for Japanese people.

Practically, storing our family name into `lastName` field seems the best solution, even if it might not be ideal.

aldente-hu:
Thanks @zoe-translates . Your suggestion looks so good! I've committed it.

aldente-hu:
Thanks as well! I'll touch another one (such as CiNii translator).

aldente-hu:
What makes this PR remain unmerged?

jajagara:
> Thanks as well! I'll touch another one (such as CiNii translator).

Hi! I have been struggling with a similar problem (family name and given name getting swapped) when importing items from CiNii. Is it possible for you to fix this problem with CiNii as well? 
(I tried mimicking your script in CiNii.js without any success. . .)

aldente-hu:
@jajagara 
I suppose the current CiNii.js is not adjusted to the brandnew "CiNii Research". ( CiNii translator was not applyed, wasn't it?)
Now I'm fixing it!

jajagara:
Oh! That's what it is! Thank you SOOO much!! とても助かります!!

aldente-hu:
It's my treasure!
Now I've done. #3363

Issue #3172, opened 2023-10-27
Generic VuFind web translator


zoe-translates:
Summary: We should support a generic VuFind web translator, because of the ubiquity of VuFind and the needs of its userbase. Cf. https://github.com/zotero/translators/pull/2969#issuecomment-1777846439

> This is a pretty common catalog in Germany, were lots of libraries used to support (just) the Citavi software, which has become much less attractive because of both pricing and declining customer support (I'm being told), with lots of libraries looking at Zotero instead -- given the customizabilitiy of VuFind, this would likely be a bit tricky, but have high impact.

With one, Zotero will work reasonably well out of the box for VuFind sites while making it easy to write one with customizations for a specific site.

---

How to do this (general plan):

- [x] Prepare a generic VuFind translator based on #2969 (EDIT 👉🏼: The draft PR is #3173)
- [ ] Update existing translators to use this generic VuFind translator (see the comment linked above)
- [ ] In the PR backlog, identify the new translators that could benefit from this

---

Technical details for step 1:

- [x] Housekeeping based on #2969 (rebase, update metadata, transform code style into our usual one, etc.)
- [x] Turn it into a generic translator with configurable options that can be set from parent translator. Mostly, this is about choosing the best input format (MARC, RIS, EndNote, etc..) for the parent translator's particular target. For different institutions, different subsets of export formats are supported, and among them some format provides better metadata than others (details and examples below)
- [ ] Investigate the feasibility of using the "bulk export" facility for handling `multiple`. However, this doesn't appear to be a common feature (supported by IxTheo and Finland National Library, but not by Wellesley College)

An example (for ixtheo) is a GET request to `https://ixtheo.de/Cart/doExport?f=RIS&i[]=Solr|1640914242&i[]=Solr|1165461676&i[]=Solr|1643208144` which exports the three items as one RIS file. If this can be done in a generic way, it will be more efficient than handling each selected item separately. UI-wise, in the Zotero item selector window, we can support pre-selecting those already selected on the web app search page. For what the search page looks like, see https://ixtheo.de/Search/Results?lookfor=christus+victor&type=AllFields&botprotect=

---

The logic of setting the input format

Depending on the institution, some formats may be better than others. For example:
* Wellesley College: https://libcat.wellesley.edu/Record/ebs14973473e Correct type (`book`) from RIS, incorrect (`journalArticle`) from EndNote
[...]

PR #2969, opened 2023-01-30
Add Library Catalog (VuFind).js

mathieugrimault:
Attempt to provide a base translator for a generic VuFind catalog. The translator relies on scraping the HTML and using exports functionalities, MARC export first, then EndNote and lastly RIS. If none are found, Zotero will probably use Z3988 COinS. 

VuFind is heavily customisable (templates and items formats) so results are less than perfect. The translator provide placeholders for customizations for a specific catalog.

Code is based on https://github.com/zotero/translators/blob/master/Finna.js and work by Abe Jellinek https://github.com/zotero/translators/pull/2874

PR #3173, opened 2023-10-27
New translator: VuFind (generic web translator)

zoe-translates:
This addresses #3172; see there for further context and information. The code is forked off #2969.


adam3smith:
@AbeJellinek there's still a lot of interest in this -- could you take a look and see how much is left to do here?

PR #3178, opened 2023-11-05
Hyphenated "access-date" in Wikipedia Citation Templates.js

thewub:
Wikipedia citation templates have preferred hyphenated "access-date" parameter rather than "accessdate" for many years. See https://en.wikipedia.org/wiki/Help:Citation_Style_1#Access_date


dstillman:
Thanks!

PR #3203, opened 2023-12-07
update DOI handling in EndnoteXML

denismaier:
Currently, the Endnote XML import does't seem to import DOIs for certain document types (e.g. books, book chapters, generic documents). As discussed in the forums it would make sense to add them to `Extra` instead,. which is what this PR does. Maybe there's a better way to support this. If so, I'll happily make the necessary adjustments.

### DIFF
--- a/Endnote XML.js
+++ b/Endnote XML.js
@@ -16,7 +16,7 @@
 	},
 	"inRepository": true,
 	"translatorType": 3,
-	"lastUpdated": "2021-07-14 20:41:42"
+	"lastUpdated": "2023-12-07 12:38:47"
 }
 
 function detectImport() {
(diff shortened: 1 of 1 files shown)


denismaier:
Why are the tests failing? The error message is not particularly helpful...

adam3smith:
tests on import translators are broken

denismaier:
Bump. Any chance this could be merged soonish? (If not, I know probably just overwrite the local translator at the scholar's installation who needs this feature, but I thought I'd ask anyway...)

Is there something I should improve?

And what about the remaining issue (`pub-location`)?

denismaier:
> Just setting `item.DOI` no matter whether the item type claims to support it should work:
> 
> ```js
> if (zfield == 'DOI' || ZU.fieldIsValidForType(zfield, newItem.itemType)) {
> ```

You mean on line 559, right?

denismaier:
> `pub-location` is CSL `publisher-place` ("Geographic location of the publisher")? If so, setting `item.place` or `Place:` in `extra` should work.

Indeed. It should correspond to `publisher-place`. (My question wasn't so much about how to add it, but rather if it should be added. @adam3smith raised some concerns about this on the forums, at least the question is if it should be added to all types (like the DOI), or if if should be added only for some types (like conferences papers).

denismaier:
Thanks for the diff.

I've now had a chance to test your suggestion from above:

> Just setting `item.DOI` no matter whether the item type claims to support it should work:
> 
> ```js
> if (zfield == 'DOI' || ZU.fieldIsValidForType(zfield, newItem.itemType)) {
> ```

Indeed, that works and this is obviously much easier than the workaround in this PR.

I guess I should use this approach then. Should I also convert the `label`/`citation key` handling to this method? But will we deal with the failing tests then?

PR #3205, opened 2023-12-11
Translator for the UK National Archives case law archive.

timcowlishaw:
Hi there folks,

I work on the [Find Case Law](https://caselaw.nationalarchives.gov.uk/) service at the UK National Archives, which provides public access to court judgments and tribunal decisions in England and Wales and thought it might be useful for people using zotero for legal research work to have a translator which imports these into Zotero!

Here's a first attempt - it follows the example of the existing [BAILII](https://www.bailii.org/) translator for formatting UK case law as Zotero entries (judges as authors, neutral citation number in a note), and relies on the LegalDocXML representation of each case which is made available on the national archives site as a canonical source of metadata.

It will import individual judgments, as well as allowing 'multiple' import on search result and collection pages.

Any feedback very welcome - i did this much as a learning exercise as anything, so happy to make any changes required if you think it'd be a useful addition!

thanks,

Tim


timcowlishaw:
Lol, first issue is that i clearly don't have the linter set up properly. I'll have a look at that now!

PR #3209, opened 2023-12-20
Added Wetten.overheid.nl

CaspervanWetten:
Added translator for Wetten.overheid.nl, the website where all Dutch laws are published

### DIFF
--- a/Wetten.overheid.nl.js
+++ b/Wetten.overheid.nl.js
@@ -1,0 +1,277 @@
+{
+	"translatorID": "6f5cd8d6-0de9-4d61-9e62-50c078c6dfd3",
+	"label": "wetten.overheid.nl",
+	"creator": "Casper van Wetten",
+	"target": "^https?://wetten\\.overheid\\.nl/",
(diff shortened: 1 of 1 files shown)


adam3smith:
That's a Chrome button that's failing in the Selenium tests; nothing to do with your code (see here for the last time this was fixed: https://github.com/zotero/translators/pull/3003 ).

I haven't done a thorough review, but we'd want a) multiples included and b) new translators to use async calls (i.e. `requestText` or `requestDocument` instead of `doGet`

CaspervanWetten:
What do you mean with "multiples included"?
Do you have any examples of new translators that use async calls I can look at?

aborel:
The translator should support importing multiple references, typically from search results on the platform, not just one reference from its own web page.
The Cairn.info translator https://github.com/zotero/translators/blob/master/Cairn.info.js looks fairly simple, it contains a requestText() call, and it has been recently updated by an experienced developer. I guess it is a good example to follow.

PR #3210, opened 2023-12-28
Update selenium test to support `system-app`

saschanaz:
For now I did this, but can we use something like this: https://developer.chrome.com/docs/extensions/reference/manifest/key#keep-consistent-id ?

### DIFF
--- a/.ci/pull-request-check/selenium-test.js
+++ b/.ci/pull-request-check/selenium-test.js
@@ -113,15 +113,11 @@ var allPassed = false;
 
 		// No API to retrieve extension ID. Hacks, sigh.
 		await driver.get("chrome://system/");
-		await driver.wait(until.elementLocated({id: 'btn-extensions-value'}), 60*1000);
-		// Chrome 89+ has the extension list expanded by default
-		try {
-			let extBtn = await driver.findElement({css: '#btn-extensions-value'});
-			await extBtn.click();
-		} catch (e) {}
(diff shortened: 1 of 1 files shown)


dstillman:
Can you explain what this is for?

saschanaz:
It's to fix the CI failure: https://github.com/zotero/translators/pull/3209#issuecomment-1864770355

saschanaz:
It seems that didn't work: https://github.com/zotero/translators/actions/runs/7469963343/job/20327889161

Are you sure the extension ID is guaranteed to be constant?

saschanaz:
Hmm, actually that failure happens earlier than getting the URL, hmm.

aborel:
In https://github.com/zotero/translators/actions/runs/7469963343/job/20327889161 it seems that ```translatorServer.filenameToTranslator[translatorFilename]``` is undefined (for some value of ```translatorFilename```.  The lint test in the same log is run on ```TinREAD.js```, which is apparently part of another PR https://github.com/zotero/translators/pull/3223#commits-pushed-a85b22f . Based on the comments there, maybe the file should not even be there at all?

saschanaz:
Per https://github.com/zotero/translators/actions/runs/7540520982/job/20525278512 no, they all look proper files with metadata.

saschanaz:
Submitted #3231.

saschanaz:
Alright, #3213 is green now, I'll close this one. Thanks!

PR #3212, opened 2023-12-28
Run `npm run lint` only when there's any translator change

saschanaz:
Fixes #3211

PR #3213, opened 2023-12-28
Replace newline to space with xargs

saschanaz:
So that the resulting command would be `eslint file1 file2` rather than `eslint file1\nfile2`.


saschanaz:
(The failure here should be fixed by #3212)

saschanaz:
Actually, files can have spaces in their names and in that case xargs is not enough.

aborel:
The GNU xargs option to use only the new line as a file separator should do the trick: `xargs -d '\n'` instead of xargs

saschanaz:
I'm not sure that's enough? `foo bar.js\nbaz.js` will still become `foo bar.js baz.js` which is not a proper input.

aborel:
You are right, I had misunderstood the problem. Here's an idea: maybe you can use xargs in `.ci/lint.sh` to run lint on just one file at a time, while the final pipe in `.ci/helper.sh' would protect the spaces of filenames instead of processing the file list. Something like this:
```
TRANSLATORS_TO_CHECK=$(git diff HEAD^2 $branch_point --name-only | { grep -e "^[^/]*.js$" || true; } | sed 's/ /\\ /g)
```
then
```
if [ -n "$TRANSLATORS_TO_CHECK" ]; then
	echo  "$TRANSLATORS_TO_CHECK" | xargs npm run lint --
fi
```

aborel:
Thanks @AbeJellinek , I think we can trust you on this one.

saschanaz:
Hmm, in that case I don't think we need `-d` or `tr` at all, we can just move `xargs` to lint.sh and that should work.

saschanaz:
Oh hmm, actually not. Let me try your suggestion...

saschanaz:
(Still needs #3210 but otherwise sounds like it should work)

saschanaz:
> What's the status on this? Is this PR still needed after the other CI fixes?

Yes, otherwise changing multiple files would trigger the linter failure.

PR #3216, opened 2024-01-05
Use https when access export.arxiv.org

zeyugao:
This avoid some failure when save arxiv due to using http.


dstillman:
Thanks, but lastUpdated needs to be updated for this to go out to users. That happens automatically when you edit the translator with the Translator Editor in Zotero.

zeyugao:
I see. I have updated the `lastUpdated` field.

dstillman:
Thank you!

PR #3222, opened 2024-01-09
Fixed some errors in translator for "ABC News Australia"

franklindyer:
Contributed the following to `ABC News Australia.js`:

1. Made `contentType` comparison for detection case-insensitive. Case-sensitivity was causing one of the test cases to fail because the `contentType` was `video` rather than `Video`.
2. Added code to detect authors by grabbing their names from elements with a `data-uri` attribute starting with `coremedia://person/`, when such elements exist. Previous method of author name detection is now used as a fallback. The previous method fails when author titles such as "weather correspondent" or "global affairs editor" are present.
3. Added a third test case to reflect the above quirk.

Before these modifications, the first test case failed and the second failed detection. Now the first test case passes and the second passes detection, but still fails due to an incorrect `abstractNote`. But this appears to be an issue with a different translator that is being used here (whichever one has GUID equal to `951c027d-74ac-47d4-a107-9c3069ab7b48`).

P.S. This is my first contribution, just learning how Zotero translators work! :-)

### DIFF
--- a/ABC News Australia.js
+++ b/ABC News Australia.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2024-01-09 21:09:32"
+	"lastUpdated": "2024-01-12 21:22:12"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


franklindyer:
Hi! For some reason, my pull request is always failing at the `Test pull request` stage, even though all of my test cases succeed on my end. But I'm noticing that several of the most recently merged pull requests also fail at this stage.

The error message is kind of opaque. Is it a problem that my PR is failing here?

aborel:
The tests have been broken for a while, work to fix them is in progress in PR #3210 .

Issue #3225, opened 2024-01-13
Add support for Harvard case.law


adam3smith:
https://cite.case.law/

Requested: https://forums.zotero.org/discussion/110768/zotero-connector-and-case-law-courtlistener/p1

Looks like you can easily get to API-based JSON results that should work well

franklindyer:
I was tinkering with this and noticed some really strange query selector behavior, maybe someone can point out what I'm doing wrong. I'm using [this entry](https://cite.case.law/so-2d/57/40/9903854/) as my test page.

First step was to extract the `api.case.law` URL corresponding to the given `cite.case.law` page. At first glance it might look like we can snatch the case ID e.g. `9903854` directly from the URL, but this isn't reliably possible, see for example [this page](https://cite.case.law/am-samoa/2/3/) in which the `cite` page URL doesn't contain the ID used in the `api` page URL. So it will be necessary to actually extract the `api` page URL from an HTML element's `href` attribute.

When I use the following query selector in my `scrape` function, it yields no results:

```
attr(doc, "a[href*='api.case.law/v1/cases/']", 'href');
```

But when I open the page in another browser, the corresponding query selector *does* yield results:

```
document.querySelectorAll("a[href*='api.case.law/v1/cases/']")
```

Also, using a regex-based text search of the entire document's `innerHTML` (which is awful, but I needed a sanity check) *also does* yield a match:

```
doc.body.innerHTML.match(/api\.case\.law\/v1\/cases\/([0-9]+)/)[0];
```

Any ideas what's going wrong here? Can anyone reproduce this? Maybe there is a silly mistake in my query selector. But I can't see why the same query selector that fails in `scrape` would succeed in another browser (I also have `"defer": true`, so I don't think it's a timing issue) even when the desired element is actually present in the raw text of the page's HTML.

adam3smith:
How exactly are you testing this? I just loaded https://cite.case.law/am-samoa/2/3/ into the Scaffold browser, created a new translator with the Web Translator template, changed detect so it always detects as a case and then put
`Z.debug(attr(doc, "a[href*='api.case.law/v1/cases/']", 'href'))` into the scrape function and ran detect. That returned  https://api.case.law/v1/cases/206939/ as expected. 
[...]

PR #3230, opened 2024-01-15
Fixes zotero/translators#3225 by adding Harvard Caselaw Access Project translator

franklindyer:
Closes [#3225](https://github.com/zotero/translators/issues/3225).

### DIFF
--- a/Harvard Caselaw Access Project.js
+++ b/Harvard Caselaw Access Project.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2024-01-16 05:45:02"
+	"lastUpdated": "2024-01-16 20:41:04"
 }
 
 /*
@@ -103,7 +103,7 @@ async function scrape(doc, url = doc.location.href) {
 	caseItem.shortTitle = abbrvCaseName;
 	caseItem.court = caseJson.court.name;
 	caseItem.dateDecided = caseJson.decision_date;
-	caseItem.docketNumber = caseJson.docket_number;
+	caseItem.docketNumber = caseJson.docket_number.replace(/[Nn]o\.?\s*/g, "");
 	caseItem.reporter = caseJson.reporter.full_name;
 	caseItem.reporterVolume = caseJson.volume.volume_number;
 
@@ -122,7 +122,7 @@ var testCases = [
 				"creators": [],
 				"dateDecided": "1947-02-07",
 				"court": "High Court of American Samoa",
-				"docketNumber": "No. 2-1944",
+				"docketNumber": "2-1944",
 				"language": "en-US",
 				"reporter": "American Samoa Reports",
 				"reporterVolume": "2",
@@ -145,7 +145,7 @@ var testCases = [
 				"creators": [],
 				"dateDecided": "1971-02-19",
 				"court": "Adams County Court of Common Pleas",
-				"docketNumber": "no. 47",
+				"docketNumber": "47",
 				"language": "en-US",
 				"reporter": "Pennsylvania District and County Reports",
 				"reporterVolume": "51",
@@ -193,7 +193,7 @@ var testCases = [
 				"creators": [],
 				"dateDecided": "1952-01-14",
 				"court": "Louisiana Supreme Court",
-				"docketNumber": "No. 40041",
+				"docketNumber": "40041",
 				"language": "en-US",
 				"reporter": "Southern Reporter, Second Series",
 				"reporterVolume": "57",
--- a/Harvard Caselaw Access Project.js
+++ b/Harvard Caselaw Access Project.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2024-01-17 00:57:03"
+	"lastUpdated": "2024-01-17 01:06:01"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


Harvard Caselaw Access Project.js
```diff
@@ -0,0 +1,175 @@
+{
+	"translatorID": "2a1cafb9-6f61-48d3-b621-c3265fde9eba",
+	"label": "Harvard Caselaw Access Project",
+	"creator": "Franklin Pezzuti Dyer",
+	"target": "^https://(.*\\.)*case\\.law",
```
adam3smith, line 5: is this every anything other then cite.case.law? If not, let's just use `(cite\\.)?` (also, don't use `* `where `?` suffices)

Harvard Caselaw Access Project.js
```diff
@@ -0,0 +1,175 @@
...
+	return false;
+}
+
+function getSearchResults(doc, checkOnly) {
+	var items = {};
+	var found = false;
+	var rows = doc.querySelectorAll('div.result-title > div > a');
+	for (let row of rows) {
+		let href = row.href;
+		Z.debug(href);
```
adam3smith, line 55: comment out all but the most essential debug statements. Most translators won't have any.

Harvard Caselaw Access Project.js
```diff
@@ -0,0 +1,175 @@
...
+}
+
+async function scrape(doc, url = doc.location.href) {
+	let apiUrl = attr(doc, "a[href*='api.case.law/v1/cases/']", 'href');
+	let caseText = await requestText(apiUrl);
+	let caseJson = JSON.parse(caseText);
+
+	let caseItem = new Zotero.Item("case");
+	caseItem.language = "en-US";
+	caseItem.url = url;
```
adam3smith, line 85: I left the URL out of the CourtListener translator, but could go either way on this -- do you have any opinion on this? We definitely will want to save an attachment, though -- either a snapshot by using `document: doc` or a PDF. Again, open for either, but we'll definitely want to capture some full text.

Harvard Caselaw Access Project.js
```diff
@@ -0,0 +1,175 @@
...
+
+	caseItem.caseName = caseJson.name;
+	caseItem.shortTitle = caseJson.name_abbreviation;
+	caseItem.court = caseJson.court.name;
+	caseItem.dateDecided = caseJson.decision_date;
+	caseItem.docketNumber = caseJson.docket_number;
+	caseItem.reporter = caseJson.reporter.full_name;
+	caseItem.reporterVolume = caseJson.volume.volume_number;
+
```
adam3smith, line 94: I'm capturing the parallel citation in `history` in the Court Listener translator and I think that makes sense. They're often required in citations and while we can't 100% do this with Zotero, we can at least approximate it. You could iterate over citations.cite and concat with a comma, then write to history.

Harvard Caselaw Access Project.js
```diff
@@ -0,0 +1,175 @@
...
+		"type": "web",
+		"url": "https://cite.case.law/am-samoa/2/3/",
+		"items": [
+			{
+				"itemType": "case",
+				"caseName": "PASA of FAGATOGO, Plaintiff v. FAIISIOTA of FAGANEANEA, Defendant",
+				"creators": [],
+				"dateDecided": "1947-02-07",
+				"court": "High Court of American Samoa",
+				"docketNumber": "No. 2-1944",
```
adam3smith, line 110: I'd be inclined to strip `[Nn]o\\.?\s*` from `docketNumber`, but it is typically included in citations, so I could to either way.

Harvard Caselaw Access Project.js
```diff
@@ -0,0 +1,175 @@
...
+
+/** BEGIN TEST CASES **/
+var testCases = [
+	{
+		"type": "web",
+		"url": "https://cite.case.law/am-samoa/2/3/",
+		"items": [
+			{
+				"itemType": "case",
+				"caseName": "PASA of FAGATOGO, Plaintiff v. FAIISIOTA of FAGANEANEA, Defendant",
```
adam3smith, line 106: we'll probably have to leave this as is, but I'm not loving the all caps here, if we can get rid of them? Problem is that I don't see a way in which we wouldn't also get rid of (properly) all capsed acronyms that way.

Harvard Caselaw Access Project.js
franklindyer: It can also be just `case.law`, namely on the search pages. I can make this more specific by narrowing it down to `(cite\\.)?case\\.law`.

Harvard Caselaw Access Project.js
franklindyer: One brute-force solution occurs to me:
1. Split the title into words and detect which ones are all-caps.
2. For each of those words, search the case text body for the first (case-insensitive) match.
3. If an all-caps match is found, chances are that it's an acronym and should be left capitalized.
4. Otherwise, capitalize the word like a name. (I'm betting on the likelihood that if it's an acronym, it will appear again within the case text.)
This should deal with acronyms properly, but it involves searching an arbitrary amount of the case's text. Matches can probably be found early when they're present at all, so it may suffice to just look at the first couple paragraphs. Overkill?

adam3smith: That's clever -- and this is on triggered actions (i.e. within `doWeb`) so we don't need to go crazy chasing efficiency (and it doesn't make additional requests to the site. Let's try it if you don't mind?

Harvard Caselaw Access Project.js
franklindyer: I have no preference, but I'll strip them for the sake of consistency with your new CourtListener translator and with [this one](https://github.com/zotero/translators/blob/master/CanLII.js).

Harvard Caselaw Access Project.js
franklindyer: Went ahead and added the PDF attachment. Do you see any harm in keeping both the case URL and the full text PDF URL? I like redundancy :-)

adam3smith: Yeah, I think I'll add it to CourtListener.

For context: Nothing wrong with redundancy -- a URL in the URL field should generally be citeable, though (otherwise it goes into an attached link) -- for academic works, we use it only when it goes to the full text of the work, e.g. For legal materials I think the criteria might be a little different, but I'm not sure -- most people will cite cases without URL, but I'm inclined to see where they are adding the URL, it should be the one they actually retrieved the material from.

adam3smith:
Great, thanks!

PR #3233, opened 2024-01-21
Add translator for Public Record Office Victoria

wragge:
New translator for checking. I think it's pretty straightforward as the metadata is all coming from their public API. Thanks!

### DIFF
--- a/Public Record Office Victoria.js
+++ b/Public Record Office Victoria.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2024-01-21 01:09:39"
+	"lastUpdated": "2024-01-21 03:13:11"
 }
 
 /*
@@ -191,7 +191,7 @@ var testCases = [
 				"url": "https://prov.vic.gov.au/archive/0C7B792B-F7F4-11E9-AE98-39C0B3AF8E48?image=1",
 				"attachments": [
 					{
-						"title": "Download PDF",
+						"title": "Download file as PDF",
 						"mimeType": "application/pdf",
 						"snapshot": false
 					},
@@ -199,6 +199,11 @@ var testCases = [
 						"title": "IIIF manifest",
 						"mimeType": "application/json",
 						"snapshot": false
+					},
+					{
+						"title": "Page 1 image",
+						"mimeType": "image/jpeg",
+						"snapshot": true
 					}
 				],
 				"tags": [],
--- a/Public Record Office Victoria.js
+++ b/Public Record Office Victoria.js
@@ -1,0 +1,211 @@
+{
+	"translatorID": "99592877-f698-4e14-b541-b6181f6c577f",
+	"label": "Public Record Office Victoria",
+	"creator": "Tim Sherratt (tim@timsherratt.au)",
+	"target": "^https?://prov\\.vic\\.gov\\.au/(archive|search_journey)/*",
+	"minVersion": "5.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2024-01-21 01:09:39"
+}
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2024 Tim Sherratt
+
+	This file is part of Zotero.
+
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+async function detectWeb(doc, url) {
+	if (getSearchResults(doc, true)) {
+		return "multiple";
+	// match items but not series, agencies, or functions (VPRS, VA or VF)
+	}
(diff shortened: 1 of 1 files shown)


Public Record Office Victoria.js
```diff
@@ -0,0 +1,216 @@
...
+
+	***** END LICENSE BLOCK *****
+*/
+
+async function detectWeb(doc, url) {
+	if (getSearchResults(doc, true)) {
+		return "multiple";
+	// match items but not series, agencies, or functions (VPRS, VA or VF)
+	}
+	else if (url.match(/archive\/(?!VPRS|VA|VF)[A-Z0-9-]+/)) {
```
adam3smith, line 43: use `test` for true/false tests on regex for efficiency.

Public Record Office Victoria.js
```diff
@@ -0,0 +1,216 @@
...
+	return false;
+}
+
+function getSearchResults(doc, checkOnly) {
+	let items = {};
+	let found = false;
+	let rows = doc.querySelectorAll(".sub_heading_search_result");
+	for (let row of rows) {
+		let href = row.href;
+		href = href.match(/archive\/(?!VPRS|VA|VF)[A-Z0-9-]+/) ? href : null;
```
adam3smith, line 55: ditto here -- don't use `match` if you don't actually need the matched string.

Public Record Office Victoria.js
```diff
@@ -0,0 +1,216 @@
...
+	let apiJSON = await requestJSON(apiURL);
+	let record = apiJSON.response.docs[0];
+	
+	// Create the Zotero item
+	let item = new Zotero.Item("manuscript");
+	item.title = record.title;
+	item.type = record.category;
+	item.archive = "Public Record Office Victoria";
+	item.archiveLocation = record.citation;
+	item.libraryCatalog = "PROV API";
```
adam3smith, line 92: I think something more readable -- like "Public Record Office Victoria" is perferable here.

Public Record Office Victoria.js
```diff
@@ -0,0 +1,216 @@
...
+	let endDate = record.end_dt.replace("T00:00:00Z", "");
+	// Discard default
+	endDate = endDate != "3000-12-31" ? endDate : "";
+
+	// If there's a date range use 'issued', otherwise use 'date'
+	if (startDate == endDate) {
+		item.date = startDate;
+	}
+	else {
+		item.issued = startDate + "/" + endDate;
```
adam3smith, line 111: @AbeJellinek -- could you weigh in on this. This will work (i.e. write the string to Extra), but I'm not sure how comfortable we are using this behavior in translators.

Public Record Office Victoria.js
```diff
@@ -0,0 +1,216 @@
...
+	// Create the Zotero item
+	let item = new Zotero.Item("manuscript");
+	item.title = record.title;
+	item.type = record.category;
+	item.archive = "Public Record Office Victoria";
+	item.archiveLocation = record.citation;
+	item.libraryCatalog = "PROV API";
+	item.url = url;
+	item.rights = record.rights_status.join(", ");
+	item.place = record.location.join(", ");
```
adam3smith, line 95: Note that this should be the place the manuscript was written/published. The location of the archive is `archive-place` in CSL (usable via `Extra`). It seems like that's what `location` is referring to?

adam3smith (changes requested):
A couple of nits, one question for you, one for @AbeJellinek ; otherwise this does look straightforward, yes.

Public Record Office Victoria.js
wragge: I was going on what has been recommended in the forums for date ranges. Eg: https://forums.zotero.org/discussion/63073/date-year-range and https://forums.zotero.org/discussion/105949/date-range

Public Record Office Victoria.js
wragge: Thanks for the quick response! This is fixed.

Public Record Office Victoria.js
wragge: Also fixed.

Public Record Office Victoria.js
wragge: Yep, changed to 'Public Record Office Victoria API'.

Public Record Office Victoria.js
wragge: Yes you're right. Changed to use `archive-place`.

Public Record Office Victoria.js
adam3smith: Right, I was referring to using item.issued to putting this into Extra -- the approach to date ranges using `Extra` is definitely right.

I kind of like the less verbose version using `item.issued`  -- is there actually a reason not to do this? will also be much faster to convert if we do get a better date field.

wragge: I've changed all the `extra` additions to use the more verbose form. I also changed `series` to `Archive Collection` to be in line with CSL.

wragge:
I think this is ready to go now?

wragge:
Sorry, getting back to this after multiple deadlines/distractions...

> I'm not sure we want to set the Library Catalog to something containing "API" - that field isn't really used in most citation styles, but it's still a little confusing and not really relevant. The default (the translator label) is fine.

Removed

> "Download file as PDF" - "Full Text PDF"? For consistency.

The attachment is a link not the actual PDF because the behaviour of the url varies -- sometimes it downloads the PDF, but if there are multiple pages it will often redirect you a a page that tells you the PDF is being generated and can take up to an hour (!). Given that it's a link, I thought "Download file as PDF" would make more sense, but perhaps "Link to PDF" would be better. In any case, as in many archival collections, they're not full text but images of documents.

> Is the manifest useful? We don't normally attach machine-readable-only files in translators. We could put it behind a hidden pref, but it seems to me like the sort of thing that you'd attach manually if you wanted it.

Yes, it's very useful. Indeed, the reason I started working on this translator was that I wanted to get the manifest urls to import digitised files into Tropy. The manifest urls aren't directly exposed through the web interface, so the translator is the only way users can get at them without using the API themselves. Also, I'm only attaching the link, not the JSON file.

> If we're attaching the PDF, why attach the image? Are the PDFs often very large? (I'm not confident that snapshot: false will prevent a non-text/html attachment from being attached...)

As noted above, it's only attaching the link to the PDF. In many cases, such as digitised photographs, the image is what you want.

PR #3237, opened 2024-01-27
Unbreak NYPL translator when annotatedMarc = null

owcz:
NYPL Catalog translator does not work in multiple instances when there bib.annotatedMarc = null in the JSON

E.g., https://www.nypl.org/research/research-catalog/bib/cb2043942

By adding a check, the translator won't fail and it'll fall back to the second (scraper) option


adam3smith:
Thanks -- bump the timestamp, please?

adam3smith:
Thanks!

PR #3244, opened 2024-01-30
Support Substack with a custom domain

EruditeLying:
Remove the substack.com domain from the target regex to support Substacks with a custom domain

In detectWeb, manually check for the old target regex first, then for one of the Substack footer buttons in the page DOM. If neither matches, return false

### DIFF
--- a/Substack.js
+++ b/Substack.js
@@ -2,14 +2,14 @@
 	"translatorID": "ac3b958f-0581-4117-bebc-44af3b876545",
 	"label": "Substack",
 	"creator": "Abe Jellinek",
-	"target": "^https://[^.]+\\.substack\\.com/(p/|archive)",
+	"target": "/p/|/archive",
 	"minVersion": "3.0",
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2022-10-05 15:16:38"
+	"lastUpdated": "2024-01-29 20:12:55"
 }
 
 /*
@@ -37,6 +37,8 @@
 
 
 function detectWeb(doc, url) {
+	if (!url.match(/^https:\/\/[^.]+\.substack\.com\/(p\/|archive)/) && !text(doc, "a.footer-substack-cta"))
+		return false;
 	if (url.includes('/p/')) {
 		return "blogPost";
 	}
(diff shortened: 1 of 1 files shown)


johannrichard:
@AbeJellinek I've been struggling a long time to get Substack metadata right and I would like to respectfully chime in this PR if I may. Maybe you can reconsider it for integration (once mergeable)? 

Here's some evidence which, I hope, can support that:

There are in my opinion, two reasons why the approach proposed with matching `/p/` and `/archive`, and maybe even matching  "`open.substack.com/pub/(.*)/p/`" would be favourable over the EM approach:

-  **Wrong date**: While it is true that EM already gets a date, more often than not, this date is plain wrong (e.g. as in wrong year, month, date). I have not investigated in detail, but EM (or whatever translator matches substack posts on custom domains) grabs a date which is off from the real post date. Easy to miss, and since there is no way to easily identify posts in the Library, even harder to fix. This happens particularly when sharing to Zotero on iPadOS and iOS (which, as far as I understand, also uses the translators under the hood). Case in point: [`https://contraptions.venkateshrao.com/p/knowledge-under-siege`](https://contraptions.venkateshrao.com/p/knowledge-under-siege), if opened on iOS and shared to the App will get a wrong date, and no date if shared via `zotero.org/save`, whereas [`https://klementoninvesting.substack.com/p/if-i-ran-a-gym-id-ask-for-tariffs`](https://klementoninvesting.substack.com/p/if-i-ran-a-gym-id-ask-for-tariffs) will be parsed flawlessly both with the App and via `zotero.org/save`).
- **Sharing from the Substack app**: When sharing from the Substack app, Substack will use an `https://open.substack.com` redirect URL in the share sheet (for example, [`https://open.substack.com/pub/contraptions/p/knowledge-under-siege`](https://open.substack.com/pub/contraptions/p/knowledge-under-siege)). For some reason, Zotero does not match these as “Substack”, even though the app itself or an underlying iOS mechanism will eventually redirect to the target. I can only assume that it matches the original URL first, and only follows the redirect, missing the Substack type, and putting a wrong date. 

Both misses were so frustrating that I have 
- wrote a script that will iterate through all (unprocessed) items with an URL in my Zotero Library, grab them, parse for `application/json-ld` and when it's a Substack post, will then update these items from the metadata (also fixing the "Blog Title" which is also missing from EM). 
- implemented a Safari Shortcut that will open the shared URL in a Safari Overlay from which I then share to Zotero to at least get all those posts on`substack.com` right

While the script has *also* allowed me to remove URL tracking parameters from these and other URLs, it would be much nicer to have a translator that matches custom domains and `open.substack.com` redirects, too.

PR #3261, opened 2024-02-20
Add responsive html as arXiv attachment

dest1n1s:
This PR aims to add articles from [arXiv.org](https://arxiv.org/) as responsive HTML5 web pages from [ar5iv](https://ar5iv.labs.arxiv.org/). This format of article is more suitable for human-reading and annotating, especially now that the latest Zotero 7 Beta provides [support for reading html snapshots](https://forums.zotero.org/discussion/106716/available-for-beta-testing-updated-reader-with-epub-snapshot-support-and-new-annotation-types/p1).

### DIFF
--- a/arXiv.org.js
+++ b/arXiv.org.js
@@ -230,6 +230,14 @@ function parseXML(text) {
 		url: pdfUrl,
 		mimeType: "application/pdf"
 	});
+	
+	var htmlUrl = "https://ar5iv.labs.arxiv.org/html/" + articleID + (version ? "v" + version : "");
+	newItem.attachments.push({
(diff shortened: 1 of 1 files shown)


dstillman:
We likely don't want to do this until we add the ability to set format priority in Zotero:

https://forums.zotero.org/discussion/comment/451101/#Comment_451101

dest1n1s:
I see. Hope the format priority feature to be available soon!

dstillman:
No need to close this. We just wouldn't merge it yet, since it would start creating redundant attachments for everyone.

PR #3275, opened 2024-03-18
Fix longstanding bug for L with stroke

adam3smith:
https://forums.zotero.org/discussion/70322/ive-notice-that-zotero-has-a-problem-with-polish-letter-l-k#latest


adam3smith:
(won't fix the BibTeX linting for this, though it might be worthwhile to do separately)

Issue #1676, opened 2018-06-13
APA PsycNET "Buy" page


bwiernik:
Detection isn't working for me in Firefox on APA PsychNET purchasing pages:
http://psycnet.apa.org/buy/2008-12294-004

Works fine (in the normal buggy way since the translator was updated) on the DOI drop page:
http://psycnet.apa.org/record/2008-12294-004

adam3smith:
Fulltext issue reported here: https://forums.zotero.org/discussion/comment/314319/#Comment_314319
for http://psycnet.apa.org/fulltext/2016-59847-001.html

bwiernik:
Is this just a URL detection issue?

adam3smith:
It may be -- current detect function doesn't include fulltext -- but I have no way of testing. You have access?

adam3smith:
I don't follow. http://psycnet.apa.org/fulltext/2016-59847-001.html redirects to http://psycnet.apa.org/buy/2016-59847-001 for me, which fails as you describe.

PR #3276, opened 2024-03-18
Fix APA PsycNET `fulltext` match issue

const7:
Fix the APA PsycNET save issue when URL containing `fulltext` (#1676, the `buy` issue has been fixed by others)
Example URL: https://psycnet.apa.org/fulltext/2022-40433-002.html

### DIFF
--- a/APA PsycNET.js
+++ b/APA PsycNET.js
@@ -207,10 +207,10 @@ async function getIds(doc, url) {
 	}
 
 	// try to extract uid from the url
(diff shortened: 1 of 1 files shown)


const7:
I'm not sure what caused the test failure here. The test cases work fine in the local browser (but not in the `Scaffold`).

const7:
thanks for your help and the test case for `/fulltext/` has been added.

PR #3286, opened 2024-03-26
Update TimesMachine

adam3smith:
https://forums.zotero.org/discussion/113113/nyt-includes-timesmachine-foxnews-no-snapshots-for-ages-and#latest

### DIFF
--- a/TimesMachine.js
+++ b/TimesMachine.js
@@ -1,7 +1,7 @@
 {
 	"translatorID": "86c3832a-ccc6-40ab-b5e9-83892423df11",
 	"label": "TimesMachine",
-	"creator": "Abe Jellinek",
+	"creator": "Abe Jellinek, Sebastian Karcher",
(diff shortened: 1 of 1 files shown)


TimesMachine.js
```diff
@@ -64,12 +64,9 @@ function scrape(doc, url) {
 		item.pages = ZU.trimInternal(text(doc, '#page_num_content'))
 			.replace(/ ,/g, ',');
 		
-		if (item.abstractNote) {
-			let place = item.abstractNote.match(/^([A-Z]+)\b/);
```
adam3smith, line 68: I don't think this is right -- 'place' is the place of publication across item types, not the place the reporter is reporting from (which this returns)

adam3smith:
Not sure how to update tests: at least for me, authentication doesn't work in the testing framework so it doesn't capture most of the data

PR #3293, opened 2024-04-02
YouTube: Fix missing fields when run from translation-server

### LINKED ISSUE #3290: YouTube titles not being read, causing YouTube translator to be skipped
When parsing a YouTube URL, the server attempts to use the YouTube translator, but ultimately skips it because it fails to read the title with the below error:
```
(1)(+0000000): Translation using YouTube failed

(1)(+0000000): Error: No title specified for item

    Error: No title specified for item
        at Object._itemDone (C:\zotero\modules\translate\src\translation\translate.js:615:32)
        at Object._itemDone (C:\zotero\src\translation\sandboxManager.js:96:17)
        at Object.complete (C:\zotero\modules\translate\src\translation\translate.js:1965:26)
        at scrape (eval at <anonymous> (C:\zotero\src\translation\sandboxManager.js:70:4), <anonymous>:148:7)
        at doWeb (eval at <anonymous> (C:\zotero\src\translation\sandboxManager.js:70:4), <anonymous>:75:3)
        at Zotero.Translate.Web.rest (C:\zotero\modules\translate\src\translation\translate.js:1346:49)
        at C:\zotero\modules\translate\src\translation\translate.js:1331:39
        at processTicksAndRejections (node:internal/process/task_queues:96:5)
  ```
These lines in the translator are responsible for parsing the title:
https://github.com/zotero/translators/blob/7837c4267095812c345c4ea86a3691b10ba07063/YouTube.js#L106C2-L108C59

The titles are read in properly in both in Scaffold and the Zotero desktop client, so this seems to be an issue with the translation server.

geofferb:
Fixes #3290 by adding an extra fallback for Title, runningTime, Date, Duration, and Description fields based on a regex search for the microformat object, which can be found in the HTML regardless if the page's JS is loaded.

### DIFF
--- a/YouTube.js
+++ b/YouTube.js
@@ -1,7 +1,7 @@
 {
 	"translatorID": "d3b1d34c-f8a1-43bb-9dd6-27aa6403b217",
 	"label": "YouTube",
-	"creator": "Sean Takats, Michael Berkowitz, Matt Burton and Rintze Zelle",
+	"creator": "Sean Takats, Michael Berkowitz, Matt Burton, Rintze Zelle,  and Geoff Banh",
 	"target": "^https?://([^/]+\\.)?youtube\\.com/",
 	"minVersion": "3.0",
 	"maxVersion": "",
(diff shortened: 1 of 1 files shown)


geofferb:
> 1. I think we can get all this data from page `<meta>` tags. Title is in `<meta name="title">`, description is in `<meta name="description">`, duration is in `<meta itemprop="duration">`, etc.
> 2. We should not use any of these fallbacks if we aren't on the server (`Zotero.isServer`), because they won't change as you browse around the site and navigate to other videos in the mobile/desktop SPA.

@AbeJellinek 
I started with the Meta tags but switched to the microformat when I realized that the description meta tag only contained the beginning of the description. If that's okay I'll use the Meta tags for everything.

PR #3294, opened 2024-04-03
Web of Science Nextgen: Support .cn top-level domain

jiaojiaodubai:
I have made the `target` more relaxed to address the issue of some Chinese users being unable to use the WOS converter for a long time. 
BTW, I'm not sure if the test case I added is redundant.

### DIFF
--- a/Web of Science Nextgen.js
+++ b/Web of Science Nextgen.js
@@ -2,14 +2,14 @@
 	"translatorID": "4a3820a3-a7bd-44a1-8711-acf7b57d2c37",
 	"label": "Web of Science Nextgen",
 	"creator": "Abe Jellinek",
-	"target": "^https://www\\.webofscience\\.com/",
+	"target": "^https://(www\\.webofscience\\.com|webofscience\\.clarivate\\.cn)/",
 	"minVersion": "3.0",
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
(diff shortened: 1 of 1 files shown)


Web of Science Nextgen.js
```diff
@@ -2,14 +2,14 @@
 	"translatorID": "4a3820a3-a7bd-44a1-8711-acf7b57d2c37",
 	"label": "Web of Science Nextgen",
 	"creator": "Abe Jellinek",
-	"target": "^https://www\\.webofscience\\.com/",
+	"target": "^https://(www\\.)?webofscience\\..*/wos/",
```
dstillman, line 5: While false positives would be unlikely here, if there are known domains we should just use those (e.g., `^https://(www\\.webofscience\\.com/|webofscience\\.clarivate\\.cn)/`).

Web of Science Nextgen.js
```diff
@@ -2,14 +2,14 @@
 	"translatorID": "4a3820a3-a7bd-44a1-8711-acf7b57d2c37",
 	"label": "Web of Science Nextgen",
 	"creator": "Abe Jellinek",
-	"target": "^https://www\\.webofscience\\.com/",
+	"target": "^https://(www\\.webofscience\\.com/|webofscience\\.clarivate\\.cn)/",
```
dstillman, line 5: Looks like there's an extra `/` in the non-China one that might break this for everyone outside of China.

jiaojiaodubai: I was careless, you were right

dstillman:
Thanks!

PR #3298, opened 2024-04-10
Note HTML & Note Markdown: Add annotation key to all reader links

AbeJellinek:
Not sure why I didn't do this in #3253 - no good reason not to add the key to EPUB/snapshot links.

@mrtcode: Let me know if this looks OK to apply to Note Markdown as well.


mrtcode:
Yeah, it makes sense!

PR #3305, opened 2024-05-10
Update DHQ translator to include PDF

joeytakeda:
This pull request updates the translator for [Digital Humanities Quarterly (DHQ)](https://www.digitalhumanities.org/dhq/) to include the full text PDF provided for each article.


joeytakeda:
Hi there! I’ve realized that the DHQ translator must have been broken for some time; using the Chrome plugin, it says that an error occurs and tries to use COINS instead and I couldn’t get this to compile in the Zotero IDE. 

Given that, I’m going to close this PR and open a new one (in the next day or two) that has the full rewrite

PR #3318, opened 2024-06-13
EM: Always look for meta tags in head and body

### LINKED ISSUE #3313: Embedded Metadata: Look for `meta` tags anywhere in document
See https://github.com/zotero/zotero-connectors/commit/55a1f75a5c48fb69285c0f72f1f7552fec2ae06a

This can also revert 2682e4c58adc850a818f28c04fc5434d56d43f6b.

AbeJellinek:
Fixes #3313. Updated PubFactory Journals; Brill needs more fixes and I'll do that separately.

### DIFF
--- a/Embedded Metadata.js
+++ b/Embedded Metadata.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2024-03-27 05:49:28"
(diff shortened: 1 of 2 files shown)


dstillman:
I changed it to just search for `meta`. Probably no difference, but if anything just searching for the tag name might be faster than specifying parent elements.

dstillman:
Merged

PR #3327, opened 2024-07-01
Improve citation of archived urls from Wayback

mvolz:
People citing Wayback urls presumably want the metadata from the website or news article itself.

Using the embedded metadata translator improves these results i.e. including the author, and the date the original article was published instead of the archive date.

Ideally we would actually want to use the translators specific to the particular url itself but this is a good first pass that improves things.

This also sets the access date to the date the page was archived since this is the date the metadata is being extracted from.


mvolz:
Another thing to note is there's no obvious field to put the archive.org url in. In this change I've put the original url in (because that's what we want upstream!) but ideally we'd leave it in... maybe for this change you want to leave the archive.org link as the main url instead of what I've done. 

Ideally we'd want a separate field to include both?

dstillman:
Yeah, I'm pretty sure we want the archive.org URL as the main URL. If someone is saving from the Wayback Machine, there's a decent chance the original URL is no longer available or has changed, and so the archive.org URL is where you go online (e.g., double-clicking in Zotero) to see the content in question, and the point of including a URL in a citation is to enable others to see the original content. There's also the epistemological argument that, while we can believe that IA is trustworthy and not covertly changing the contents of archived pages, it's just factually inaccurate to claim that the original URL represents the content being referenced.

Why do you want the original URL on Wikipedia?

PR #3335, opened 2024-07-13
Fix JAMA (Silverchair)

adam3smith:
https://forums.zotero.org/discussion/115948/cannot-import-pdf-file-from-jama#latest


dstillman:
Thanks!

PR #3354, opened 2024-08-18
Adding translator for Library of Congress Digital Collections

atombravo:
Adding translator for Library of Congress Digital Collections, newspaper only at the moment.  This will pull and attach the PDF of the page being viewed. Note: the digital collections are hosted at www.loc.gov, and the API is separate from the LC catalog (catalog.loc.gov). 

This code pulls the JSON of the currently displayed page and parses that instead of scraping the page being viewed.

### DIFF
--- a/Library of Congress Digital Collections.js
+++ b/Library of Congress Digital Collections.js
@@ -1,0 +1,131 @@
+{
+	"translatorID": "11614156-f421-4e89-1111-a5e69ce3ebed",
+	"label": "Library of Congress Newspaper Page",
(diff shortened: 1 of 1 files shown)


atombravo:
I updated the authorship - given that you've significantly expanded on what I started, I'd rather your name be first.  Manually testing the resource types, this all looks to work perfectly.

I just remembered that the translaterID guid at the top in my code is bogus--I had intended to track down how to generate that correctly, but I didn't get to it.  That probably needs to be replaced.

PR #3356, opened 2024-08-26
minor fix for Queensland State Archives

wragge:
The data structure of the API has changed slightly – there's a list of objects where there used to be a list of strings. I've changed the translator to extract the strings from the objects.


wragge:
Not sure why the tests have started failing in CI. Everything works fine in Scaffold and browser.

wragge:
Found some examples with multiple PDFs, so I've updated the translator to attach them all.

I think they might have started blocking some IP addresses which is what's causing the test failures in CI. As I said it works fine for me.

PR #3358, opened 2024-09-10
Fix OSF attachment download issue & Code refraction

const7:
## Fix

1. Fix the attachment download issue from OSF

## Update

1. Refracting previous code for a more concise structure
2. Fetch metadata from OSF API for both single/multiple modes.
3. Replace deprecated functions
4. Update meta info for test cases


OSF Preprints.js
```diff
@@ -64,84 +64,58 @@ function getSearchResults(doc, checkOnly) {
...
-function doWeb(doc, url) {
+async function doWeb(doc, url) {
 	if (detectWeb(doc, url) == "multiple") {
-		Zotero.selectItems(getSearchResults(doc, false), function (items) {
-			if (items) ZU.doGet(constructAPIURL(Object.keys(items)), osfAPIImport);
+		// only work for preprint type like previous code
+		Zotero.selectItems(getSearchResults(doc, false), async function (items) {
+			if (items) {
+				for (let url of Object.keys(items)) {
+					await scrape(doc, url);
```
const7, line 73: I think the multiple results issue comes from `doc` parameter here. It should be the document of the corresponding URL based on current scrap logic. But I failed to fix it

PR #3359, opened 2024-09-12
Update Crossref REST translator

mrtcode:
I noticed there are some differences in the available data compared to the XML API. For example, fields like ISBN, pages, editors, contributors, and language, etc. can sometimes be missing, but not always. However, this translator performs better with various book-related items.

It was tested with all item types provided by the Crossref API, and each item type is also included in the tests.

All in all, I think it's pretty good now.

**The hyphen has been removed from the translator file name to match the actual translator name (there was an issue with Scaffold). I wonder if this could have any negative consequences?**


mrtcode:
To run test with Scaffold `detectSearch` has to be changed to return `true`.

dstillman:
We should set Library Catalog to just "Crossref", not "Crossref REST".

Crossref REST.js
```diff
@@ -0,0 +1,1733 @@
...
+			item.date = parseDate(result.issued);
+		}
+
+		if (result.DOI) {
+			if (ZU.fieldIsValidForType('DOI', item.itemType)) {
+				item.DOI = result.DOI;
+			}
+			// Add DOI to extra for unsupported items
+			else {
+				item.extra = (item.extra ? item.extra + '\n' : '') + 'DOI: ' + result.DOI;
```
dstillman, line 307: We shouldn't need to do this manually. It should get set automatically if not valid for the field.

dstillman: (But that might not show up in tests.)

mrtcode: Yes, it was exactly like that.

mrtcode:
> We should set Library Catalog to just "Crossref", not "Crossref REST".

Fixed. However, I imagine it will be [overridden](https://github.com/zotero/translators/blob/c140b805886f37e6021b8b62925e41f23f994007/DOI%20Content%20Negotiation.js#L82), as is the case with the XML translator.

PR #3368, opened 2024-10-03
ScienceDirect: Don't add `isDTMRedir` URL param

### LINKED ISSUE #3367: ScienceDirect: isDTMRedir=true is causing failure to download PDF attachments via Zotero captcha viewer
I'm not sure if that URL parameter is useful in any context anymore, but it may be being used by CloudFlare bot protection to filter out and restrict requests from Zotero.

Sample page `https://www.sciencedirect.com/science/article/pii/S0378429019313863?via%3Dihub`
Failing URL to download via Zotero `https://www.sciencedirect.com/science/article/pii/S0378429019313863/pdfft?isDTMRedir=true&download=true`
URL that works via Zotero `https://www.sciencedirect.com/science/article/pii/S0378429019313863/pdfft?download=true`

Moreover, the `isDTMRedir=true` URL after being blocked on Zotero no longer opens on any browser, but the URL without that parameter works fine.

AbeJellinek:
I kept the *check* for `isDTMRedir` in `getPDFLink()` - not sure if that branch ever works anymore, but it seems safer to leave it be. That method should, separately, be overhauled in general, since the `click()`-based branch no longer works with MV3.

Fixes #3367

### DIFF
--- a/ScienceDirect.js
+++ b/ScienceDirect.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2024-06-13 19:43:00"
+	"lastUpdated": "2024-10-03 14:17:12"
 }
 
 function detectWeb(doc, url) {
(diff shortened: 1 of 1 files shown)


adomasven:
If we need `click()` to function in MV3 we can restore it. Just let me know.

PR #3370, opened 2024-10-10
Extend Citavi6 translator to import BibTex-Keys

JAC28:
- Workaround to work with syntax of Better BibTex plugin
- Writes Keys into Extra-Field
- Provided Sample-Data contains Demo-Project with auto-generated Keys. First entry is without Key for testing.


adam3smith:
Something weird happening with the commit -- looks like you modified every line (EOL character or indenting, I assume). I'd guess linting locally would prevent that.

JAC28:
I'll create a new PR. I think the scaffold tool has added the testcases to the translator file.

PR #3371, opened 2024-10-10
Extend Citavi6 translator to also import BibTex-Keys

JAC28:
- Workaround to work with syntax of Better BibTex plugin
- Writes Keys into Extra-Field
- Provided Sample-Data contains Demo-Project with auto-generated Keys. First entry is without Key for testing.

Note: First try failed due to included test cases.

### DIFF
--- a/Citavi 5 XML.js
+++ b/Citavi 5 XML.js
@@ -1,7 +1,7 @@
 {
 	"translatorID": "e7243cef-a709-4a46-ba46-1b1318051bec",
 	"label": "Citavi 5 XML",
-	"creator": "Philipp Zumstein, Tomasz Najdek",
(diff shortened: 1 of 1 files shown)


JAC28:
**Reason for the extension:**

When transferring from Citavi to Zotero, the existing BibTex keys are not transferred, if I see it correctly. Using the *Better BibTex* plug-in ensures that the keys are re-issued, which means that compatibility with previous work based on the Citavi BibTex keys is lost. 
If the extra fields contain a *Citation Key: BibTeyKey* pair, Better BibTex will adopt it. This ensures backwards compatibility of the keys even after transfer to Zotero. 

**Note** This is the same change as initially proposed in #3370 but failed due to linting problems

JAC28:
I think this can be closed, as the change is included in the still open PR #3183

PR #3384, opened 2024-11-02
Do not set arXiv DOI

thebluepotato:
After the recent changes to the arXiv translator, items imported via identifier or from arXiv search pages no longer include the arXiv issued DOI, as intended. However, when imported from the webpage, the DOI would still be set.

I believe the consensus is that the DOI should reflect the published version. We have easy access to the arXiv page anyway.


thebluepotato:
Oops sorry, my branches got mixed up... I'll probably end up deleting and re-doing #3161

adam3smith:
I don't understand why we wouldn't want the DOI for arXiv? That's why they started issuing them, just like all.other preprint servers.

PR #3385, opened 2024-11-08
feat: Update IMDb.js to add URL to item

johev09:
## Changes
`IMDb.js`:
  - Adding URL to item in `scrape(...)` method

## Notes
- This changes populates the URL field when saving a IMDb page to Zotero using Zotero Connector.
- Populating the URL makes the Zotero item clickable
- Opens IMDb page on double-clicking the Zotoer item.

## Testing
- Tested on my local machine.
- Verified URL is getting populating on saving an IMDb link.

### DIFF
--- a/IMDb.js
+++ b/IMDb.js
@@ -45,7 +45,7 @@ function detectWeb(doc, url) {
 			return "film";
 		}
 	}
-	else if (url.includes('/find?') && getSearchResults(doc, true)) {
+	else if (url.includes('/find') && getSearchResults(doc, true)) {
 		return "multiple";
 	}
 	return false;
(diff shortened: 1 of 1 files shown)


johev09:
Hi, Thanks for checking my PR.

Oh, I didn't know that about the URL field. I just checked the [documentation](https://www.zotero.org/support/kb/item_types_and_fields#:~:text=URL%20(web%2Daddress)%20where%20the%20full%20item%20was%20accessed.%20Do%20not%20use%20this%20field%20for%20links%20to%20catalog%20records%20(e.g.%2C%20libraries%20or%20Google%20Books)%20or%20abstracts%E2%80%94these%20should%20be%20instead%20added%20as%20links.) and it does mention the same. Thanks for clariyfing that.

Regarding the xpath changes, the search page xpaths have changed and was making the test cases fail so tried to update them but couldn't make it work.

I will go ahead and close this PR as the primary change is not needed.

PR #3387, opened 2024-11-15
Add TIND Repository translator

thms-rmb:
TIND provides a SaaS solution for libraries and digital repositories to host various materials. There may be many institutions running TIND. This PR introduces a simple translator that should work for most institutions.

There is already an existing translator for the ILS flavor of TIND:

https://github.com/zotero/translators/blob/bc846072a55a679866608b0d16009d2e4cfb8f0b/Library%20Catalog%20(TIND%20ILS).js

That translator uses the MARCXML content directly. The downside to that is that it makes it harder to customize the mapping to the Zotero data model. TIND provides a [Schema.org](https://schema.org/) representation of records, which is much more customizable by the institution, and which I think is a more suitable target for translating into the Zotero data model.

Disclaimer: I work for TIND.

### DIFF
--- a/Schema.org.js
+++ b/Schema.org.js
@@ -1,0 +1,324 @@
+{
+	"translatorID": "5e81b2f5-fd88-427b-9e0c-53809de98582",
+	"label": "Schema.org",
+	"creator": "Thomas Rambø",
+	"target": "json",
+	"minVersion": "3.0",
+	"maxVersion": "",
+	"priority": 750,
+	"inRepository": true,
+	"translatorType": 1,
+	"lastUpdated": "2024-12-04 09:32:40"
+}
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	This file is part of Zotero.
+
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+const SCHEMA_ORG_ZOTERO_TYPE_MAPPING = {
+	"CreativeWork": "document",
+	"Article": "journalArticle",
+	"ScholarlyArticle": "journalArticle",
+	"Book": "book",
+	"Chapter": "bookSection"
+};
+
+/**
+ * @returns {object | undefined}
+ */
+function getSchemaOrg() {
+	const schemaOrgText = [];
+
+	while (true) {
+		const line = Zotero.read();
+
+		if (!line) {
+			break;
+		}
+
+		schemaOrgText.push(line);
+	}
+
+	let schemaOrg;
+
+	try {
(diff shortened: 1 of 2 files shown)


thms-rmb:
> Thanks! A little hard to evaluate this without test cases. Could you use Scaffold (Tools -> Developer -> Translator Editor) and add some?

@AbeJellinek done!

thms-rmb:
@AbeJellinek I have revised this with the following changes:

+ Use tabs for indentation
+ TIND Repository is purely a web translator, not an import translator anymore
+ Added a separate Schema.org import translator with actual translation logic and tests
+ Schema.org import translator looks specifically for Schema.org `@type` property and specific values in it, so should not break other JSON imports
+ TIND Repository is now a small wrapper that - apart from detectWeb - simply uses Schema.org import translator

thms-rmb:
I see, that makes sense.

> Aren't https://cicero.tind.io/record/71 and https://cicero.tind.io/record/128 (from your import tests) URLs we could use for web tests?

Yes, these currently work but there is no guarantee of their persistence - they might change content or stop working.

I will investigate if we can set up some persistent static pages that could be used for tests, and then consolidate the import translator and the web translator into a single web translator as you request!

thms-rmb:
@AbeJellinek I can see why this is confusing, sorry about that!

This translator will be interesting for TIND sites with Schema.org data on the detailed record page, which is the case for the ILS, IR and RDM flavors of the TIND product.

At the moment, the Schema.org metadata is embedded in a script element that looks like this:

```html
<script type="application/ld+json">
    {
        "@context": "https://schema.org",
        "@id": "https://digitallibrary.un.org/record/4068194"
    }
</script>
```

The script element does not have any id attribute as you point out, but we are in the process of rolling out a new release which adds the id attribute that this translator is using. If you go e.g. here: https://digitallibrary.un.org/record/406819, open the browser dev console, and search for an element that matches the css selector `[type='application/ld+json']` you should be able to see the Schema.org data even now.

Here is another example record that has a more elaborate Schema.org embedded metadata: https://knowledge.uchicago.edu/record/14212

> Is there no public-facing production instance anywhere on the internet that we could use as a test case

I can certainly come up with some production instance pages that could be used - I just assumed that it might be problematic from a stability point of view. Although they should be fairly stable, and so this concern is more hypothetical: it could happen that there is downtime, the customer site deletes a record or makes it restricted, the customer changes the metadata in such a way that the test no longer passes, etc.

As an alternative, if it works for you, we could also set up some records on an instance that is not in use by a customer, and so we would have more control over the persistence of the metadata.

thms-rmb:
@AbeJellinek what you're saying makes a lot of sense. I will try to describe where we are coming from:

There are a number of cases where the current translator for TIND sites ends up with an unexpected item type.

+ For example, https://socialmediaarchive.org/record/60 should be a dataset but ends up being a book
+ https://arodes.hes-so.ch/record/14994 should be a conference paper but ends up being a journal article
+ https://knowledge.uchicago.edu/record/13629 should be a presentation but ends up being a book

Most of the time, the reason is probably that the MARC data does not fully correspond to what the MARC translator expects. This is a fact we have to accept to some degree, while in other cases we might be able to adjust the MARC data model. But the idea in this PR was to use the Schema.org metadata we have, which can more flexibly map from various MARC fields on an instance-by-instance basis.

I am on board with this approach not being the right one, however I am still interested in improving the accuracy of item types if possible. You mention:

> depending on each repository's most native representation [...] use bits of Schema.org metadata to do things like detect the item type better.

Would it be of interest to see an addition to the current Library Catalog (TIND ILS) so that it can e.g. use the embedded Schema.org metadata to capture the item type? Essentially, to use everything as-is but have an additional step that looks for Schema.org to discover the item type.

thms-rmb:
Hi @AbeJellinek, did you have time to consider my previous comment?

thms-rmb:
I will close this and try another approach

PR #3392, opened 2024-11-22
Embedded Metadata: Clean up long ISO datetimes, update tests

AbeJellinek:
Requested here:

https://forums.zotero.org/discussion/comment/479862#Comment_479862

`strToISO()` would chop off the time part, but we don't want to run all EM dates through `strToISO()`, so let's just clean up the ones that are clearly identifiable as ISO datetimes. Pretty low-risk improvement, I think.

ScholarWorks requires `defer: true` and meta tags are pretty lacking now; we should add a separate translator for it. Hindawi merged with Wiley, but the test is still OK. Other test changes are minor.


AbeJellinek:
Definitely wouldn't mind! The main hold-up, though, is making sure that this doesn't break anything, including other translators that depend on EM (which is a lot of translators).

PR #3395, opened 2024-11-25
RIS: Disambiguate journal abbreviation and pub title based on length

### LINKED ISSUE #3394: RIS: Move journalAbbreviation to publicationTitle when empty, instead of copying?
https://github.com/zotero/translators/blob/4cadb3e1455e6438b0b7c530d509cddca5f3e1c6/RIS.js#L1767-L1770

Not sure if there's a reason we leave the "abbreviation" in both fields, but moving it might be more correct.

https://forums.zotero.org/discussion/119961/journalabbreviation-contains-full-journal-name-with-silverchair-translator

AbeJellinek:
Fixes #3394

### DIFF
--- a/RIS.js
+++ b/RIS.js
@@ -17,7 +17,7 @@
 	},
 	"inRepository": true,
 	"translatorType": 3,
-	"lastUpdated": "2024-11-25 17:28:41"
+	"lastUpdated": "2024-11-25 18:57:30"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


RIS.js
```diff
@@ -1764,8 +1770,18 @@ function completeItem(item) {
 		if (cleanDOI) item.DOI = cleanDOI;
 	}
 
-	// hack for sites like Nature, which only use JA, journal abbreviation
-	if (item.journalAbbreviation && !item.publicationTitle) {
+	// Use the shorter journal abbreviation, and move the longer one to
```
adam3smith, line 1773: You're not worried that the logic here is getting a bit convoluted? It obviously works for the issue at hand, but it's hard to keep track of what's going on in the code, which will make future troubleshooting harder.
Otherwise this looks fine.

RIS.js
```diff
@@ -1770,20 +1770,18 @@ function completeItem(item) {
...
-				item.publicationTitle = item.journalAbbreviation;
-			}
+	// If we have a journalAbbreviation, use it to fill publicationTitle if missing
+	if (item.journalAbbreviation) {
+		if (!item.publicationTitle) {
+			item.publicationTitle = item.journalAbbreviation;
+		}
+		// If we got multiple potential journalAbbreviations, keep the longest one
+		if (item.backupJournalAbbreviation
+				&& item.backupJournalAbbreviation.length < item.journalAbbreviation.length) {
```
alex-ter, line 1780: This `<` seems to contradict the logic described in the comment, as it's the shorter of the two that will be assigned to `item.journalAbbreviation` inside the `if` at line 1780/1781. Should be changed to `>`, probably?

Additionally, just to make sure - the logic in this whole block is not identical to the previous one (before simplification), as it's not the longest abbreviation that will be used for the title. Is that intentional?

AbeJellinek: Yeah, you're right, and this logic is way too convoluted, as @adam3smith said. I'm just going to add a fix for Silverchair specifically. We haven't seen this anywhere else.

AbeJellinek:
Superseded by e0398e97d2c9b4009d5a686abcb2ec623fc39ea4

PR #3396, opened 2024-11-26
New Translator: Australian Parliament House (Bills & Legislation)

drzax:
This translator adds support for importing documents related to bills and enacted legislation from the Australian
Parliament House website (aph.gov.au).


drzax:
@AbeJellinek — thanks for all your comments on this, it's cleared a few things up for me. I'll work on fixing/committing your suggestions when I can.

As a first time translator author I found it very difficult to find good documentation about the structure and functions available to translators and the field mappings for different citation types. Can you recommend any good documentation I might have missed?

PR #3399, opened 2024-12-03
Tf hotfix

### LINKED ISSUE #3398: T&F Import broken because of broken bibTeX
Reported in various threads on the forums, including https://forums.zotero.org/discussion/120189/paper-saved-as#latest

This is caused by broken bibTeX provided by T&F

```
@article{Nattermann01011988},
author = {T. Nattermann and J. Villain},
title = {Random‐field ising systems: A survey of current theoretical views},
journal = {Phase Transitions},
volume = {11},
number = {1-4},
pages = {5--51},
year = {1988},
publisher = {Taylor \& Francis},
doi = {10.1080/01411598808245480},
URL = { 
        https://doi.org/10.1080/01411598808245480
},
eprint = { 
    
        https://doi.org/10.1080/01411598808245480
}
,
    abstract = { Ising or Ising-like models in random fields are good representations of a large number of impure materials. The main attempts of theoretical treatments of these models--as far as they are relevant from an experimental point of view--are reviewed. A domain argument invented by Imry and Ma shows that the long-range order is not destroyed by weak random-fields in more than D = 2 dimensions. This result is supported by considerations of the roughening of an isolated domain wall in such systems: domain walls turn out to be well defined objects for D \&gt; 2, but arbitrarily convoluted for D \&lt; 2. Different approaches for calculating the roughness exponent ζ yield ζ= (5 - D)/3 in random-field systems. The application of ζ in incommensurate-commensurate critical behaviour is discussed. Non-classical critical behaviour occurs in random-field systems below D = 6 dimensions which is determined in general by three independent exponents. The new exponent yJ
                = θ= D/2 - σ corresponds to random-field renormalization or, to say it differently, to the irrelevance of the temperature at the zero-temperature fixed point, which is believed to rule the critical behaviour. The inequalities satisfied by these exponents are investigated, as well as the relations between the eigenvalue and the critical exponents and their numerical values found in the literature. Domain wail roughening due to random fields produces also metastability and hysteresis. It is shown that when cooling a 3D system into the low-temperature phase in an applied random field, the system runs into a metastable domain state, in agreement with the experimental observation. The further approach of the system to the ordered equilibrium state is hindered by pinning which leads to domain size increasing only logarithmically in time. Domain wall roughness and pinning in random bond systems is also considered. }
}
```

Note the closing `}` after the citekey, which ends the entry. I'll report to T&F, but we should look at working around this (probably just detecting and deleting the curly brace

adam3smith:
closes #3398

### DIFF
--- a/Taylor and Francis+NEJM.js
+++ b/Taylor and Francis+NEJM.js
@@ -145,16 +145,11 @@ async function scrape(doc, url = doc.location.href) {
 	var unescapeFields = ['title', 'publicationTitle', 'abstractNote'];
 	for (var i = 0; i < unescapeFields.length; i++) {
 		if (item[unescapeFields[i]]) {
-			Z.debug("checking item fields");
-			Z.debug(unescapeFields[i]);
-			Z.debug(item[unescapeFields[i]]);
 			item[unescapeFields[i]] = ZU.unescapeHTML(item[unescapeFields[i]]);
-			Z.debug(item[unescapeFields[i]]);
 		}
 	}
 	
 	item.bookTitle = item.publicationTitle;
-	Z.debug(item);
 	if (!item.title) item.title = "<no title>";	// RIS title can be even worse, it actually says "null"
 	if (risItem.date) item.date = risItem.date; // More complete
 	if (item.date && /^\d{4}$/.test(item.date)) {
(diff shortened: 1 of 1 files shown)


adam3smith:
@AbeJellinek 3 reports on the forum within a day of this, would be great to merge asap

adam3smith:
Thanks I took out the m flag. I don't think we need/want the linebreak -- the m flag was just a leftover.
FWIW, even with m flag, this would never match  valid regex because there would be a comma before the first closing bracket.

PR #3400, opened 2024-12-06
Bluesky translator

urschrei:
This adds the ability to save individual Bluesky posts. Details:

- Saves snapshot
- If a post is being quoted, the quoted post is added as a note identified as such
- Replies to the post are added as individual notes. I believe the Bluesky API currently limits API responses to 25 items, so there won't be more than that. I'm happy to add a check and manually truncate if required.

The translator isn't working on Safari / Safari Technology Preview (I tested successfully with Scaffold, Firefox, and Chrome). I can't see any reason for this besides an issue on Safari's side.

PR #3401, opened 2024-12-09
Create Bluesky.js

rmzelle:
I see somebody else already created #3400 for Bluesky, which I noticed right before creating this pull request. My version just does screen scraping without querying the API.


rmzelle:
And per https://forums.zotero.org/discussion/comment/481076/#Comment_481076, I think it would be great if the Twitter/X, Mastodon, and Bluesky translator would all process their item titles in the same way (remove newlines, trim to 140 chars), and write the full post to the abstract field.

PR #3438, opened 2025-04-01
Update Primo Normalized XML.js

ezellohar:
Trying to fix a problem with the title in Italy: the title field is catalogued as follows "`Title words / Author's name and surname`", so I added a regex to keep only what's before the bar (and the space before the bar).
This should be the one fix that should address the issue?


ezellohar:
> Actually, that new test case is slightly broken too: it ends up as "The letters of J.R.R. Tolki".

Ok, maybe this regex should work better: `(.+?)(?=\s\/\s)` so the code could be changed with `.replace(/(.+?)(?=\s\/\s)/, '$1');`

Oh, nevermind, after writing all this I noticed you already fixed the code. Thank you very much

PR #3443, opened 2025-05-06
Update The Straits Times

robertsky:
simplify byline detection for straits times. update author list


AbeJellinek (changes requested):

The Straits Times.js
```diff
@@ -133,7 +117,7 @@ function getMultipleItems(doc, url) {
 			return (!!item.match(/^https:\/\/www\.straitstimes.com/));
 		});
 		if (items.length) {
-			ZU.processDocuments(items, scrape);
+			ZU.requestDocument(items, scrape);
```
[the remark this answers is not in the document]
robertsky: acknowledged. will be updating soon.

The Straits Times.js
```diff
@@ -105,14 +93,20 @@ function scrape(doc, url) {
...
+	return new Promise(resolve => setTimeout(resolve, ms));
+}
+
+async function getMultipleItems(doc, url) {
 	var items = [];
 	var rows;
-	if (url.includes('/search?') && url.includes('searchKey')) {
+	Zotero.debug(url);
+	if (url.includes('/search?') && url.includes('searchkey')) {
+		await sleep(1000);
```
robertsky, line 105: @AbeJellinek apparently the third party tool that the site uses for search loads much slower, hence the sleep here. but i am not sure if this is the right way to go about it... Also when I tried to run and update the test, it went into a perpetual 'updating' mode?

AbeJellinek:
(`detectSearch()` and `doSearch()` are for *search translators* that handle standard identifiers like DOIs, ISBNs, and PMIDs, not web translators that support search pages.)

The Straits Times.js
robertsky: I am unable to get the monitorDOMChanges change to work.

PR #3445, opened 2025-05-14
Add Chicago Journal of Theoretical Computer Science translator

morganfshirley:
This is a translator for CJTCS. The website is fairly bare-bones, and does not include metadata, but does provide a bibtex file. The bibtex file is occasionally malformed, and this translator will attempt to fix that.

I also added rudimentary support for adding multiple articles; there is no search function, but one can view an entire volume or special issue and the translator will correctly list all of the articles.

My git is a bit rusty so please let me know if the pull request was done incorrectly.


AbeJellinek:
Thanks, this is great.

PR #3450, opened 2025-05-20
WIP: Fix CI

AbeJellinek:


.ci/pull-request-check/selenium-test.js
```diff
@@ -2,15 +2,17 @@
...
 const chalk = require('chalk');
 
 const translatorServer = require('./translator-server');
 
-const chromeExtensionDir = path.join(__dirname, 'connectors', 'build', 'chrome');
+const chromeExtensionDir = path.join(__dirname, 'connectors', 'build', 'manifestv3');
 const KEEP_BROWSER_OPEN = 'KEEP_BROWSER_OPEN' in process.env;
-const ZOTERO_CONNECTOR_EXTENSION_ID = 'ekhagklcjbdpajgpjgmbionohlpdbjgc';
+const CI = 'CI' in process.env;
+const ZOTERO_CONNECTOR_EXTENSION_ID = 'noadpmpcjagenlgkmdgffkabnlbhcbgd';
```
dstillman, line 15: What is this id?

AbeJellinek: That's the one I get when I load the source build in Chrome, and it seems to be the same on CI.

dstillman: Hmm. I would think that would just be randomly generated, but for me it's `eefmkbdibfgpjbomibjbnfobkoiebkli` on two different computers.

AbeJellinek: That's bizarre.

Looking back at the discussion in this PR: https://github.com/zotero/zotero-connectors/pull/456
It looks like setting `CHROME_EXTENSION_KEY` should give us a fixed ID, and [we are doing that](https://github.com/zotero/translators/blob/b4a6852cd16621e0ba0c4e5af03a5b3903fdfe96/.ci/pull-request-check/check-pull-request.sh#L32). So I have no idea why it's varying across devices. Did you build the Connector by running `check-pull-request.sh`? Maybe try cleaning the build directory there first?

dstillman: Oh, sorry, is that only the id when running that script? `eefmkbdibfgpjbomibjbnfobkoiebkli` is what I get when I just load it as an unpacked extension.

AbeJellinek: OK, figured this out. Chrome derives the ID from the extension key, so we want to build our Connector with the same extension key that the Web Store uses so we get `ekhagklcjbdpajgpjgmbionohlpdbjgc`. Due to [some issues with always fixing the key](https://github.com/zotero/zotero-connectors/pull/456#issuecomment-1946383394), `check-pull-request.js` sets `CHROME_EXTENSION_KEY` when building the Connector, and the Connector's gulpfile code reads that and injects it into the manifest.

But at some point after the MV3 transition, that gulpfile code broke. When there's no extension key, Chrome just hashes the "normalized" path to the extension. That was consistent between me and the CI runner: `~/translators/.ci/pull-request-check/connectors/build/manifestv3`. I'm guessing you were building from a different directory, but the same one on both machines.

Opened a PR to fix the gulpfile code: https://github.com/zotero/zotero-connectors/pull/549

Once that's merged, we should get the old ID again when building with `check-pull-request.js`, so I'll revert the extension ID change here.

PR #3452, opened 2025-05-26
Create Jade.io translator for legal cases in Australia

LawData-user:
I’ve written (well, used ChatGPT to write) a custom translator for JADE.io to extract Australian case citations including parallel reports. It supports paragraph citations (e.g. FLC ¶93–380), CLR/ALJR/ALR-style citations, and logs remaining citations to extra.

Sample URL is https://jade.io/article/81345

It extracts case name, media neutral citations (as the AustLII translator does) and also the reported citations (the first one in the appropriate fields and the others in the extra field.

I have tested it locally and it seems to work

Tried to follow the instructions for GIT, but ended up followin Chat-PGT instructions - happy to be corrected on how to do this.

### DIFF
--- a/Jade_case
+++ b/Jade_case
@@ -1,0 +1,117 @@
+{
+	"translatorID": "c7e07d17-cc0b-4f6c-b12a-6f3a9de0410b", // generate with `uuidgen` if needed
+	"label": "JADE Case Extractor",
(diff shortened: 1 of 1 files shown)


AbeJellinek:
Yup, thanks.

PR #3461, opened 2025-06-17
IMDb: Improve CSL specification adherence

adunning:
Do not import text such as 'drama' to `genre`, which is a type/class/subtype rather than a style or category. Import the place to `event-place` rather than the nonexistent `event-location` variable.


AbeJellinek:
Thanks!

PR #3462, opened 2025-06-17
Internet Archive: ignore contributors

adunning:
It is unhelpful to import names such as digitizing libraries as contributors, since this should refer to contributors to a work, meaning that in some styles these names erroneously appear in citations.

### DIFF
--- a/Internet Archive.js
+++ b/Internet Archive.js
@@ -153,19 +153,6 @@ function scrape(doc, url) {
 				}
 			}
 		}
-		var contributors = obj.contributor;
-		if (contributors) {
-			for (i = 0; i < contributors.length; i++) {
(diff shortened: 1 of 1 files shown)


Internet Archive.js
```diff
@@ -489,13 +461,6 @@ var testCases = [
 			{
 				"itemType": "tvBroadcast",
 				"title": "The Rachel Maddow Show : MSNBCW : January 13, 2017 6:00pm-7:01pm PST",
-				"creators": [
-					{
-						"lastName": "MSNBCW",
-						"creatorType": "contributor",
-						"fieldMode": 1
-					}
-				],
```
alex-ter, line 498: As the lint says, this should be an explicit empty array instead, see e.g., line 434 in the original file.

PR #3468, opened 2025-06-19
Update Wikidata QuickStatements.js

pmartinolli:
They modify the Description of :
  - encyclopedia entries (which encyclopedia),
  - book sections (which book),
  - thesis (which university),
  - conference papers (which conference).
  
They add a "default for all languages" Label.

They manage the permanent identifiers from :
  - OpenAlex: P10283
  - Semantic Scholar CorpusID: P8299
  - Web of Science WOS: P8372
  - Microsoft Academic Graph MAG: P6366

### DIFF
--- a/Wikidata QuickStatements.js
+++ b/Wikidata QuickStatements.js
@@ -1,9 +1,9 @@
 {
 	"translatorID": "51e5355d-9974-484f-80b9-f84d2b55782e",
 	"label": "Wikidata QuickStatements",
-	"creator": "Philipp Zumstein (with minor edits by Pascal Martinolli)",
+	"creator": "Philipp Zumstein and contributors",
 	"target": "txt",
-	"minVersion": "3.1",
+	"minVersion": "3.0",
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
@@ -15,7 +15,7 @@
 /*
 	***** BEGIN LICENSE BLOCK *****
 
-	Copyright © 2017 Philipp Zumstein
+	Copyright © 2017-2025 Philipp Zumstein and contributors
 
 	This file is part of Zotero.
 
@@ -34,29 +34,6 @@
 
 	***** END LICENSE BLOCK *****
 */
-
-/*
-	***** BEGIN REVISION BLOCK *****
-
-	In 2025, Pascal Martinolli made some revisions to the original code.
-	
-	They modify the Description of :
-	   - encyclopedia entries (which encyclopedia),
-	   - book sections (which book),
-	   - thesis (which university),
-	   - conference papers (which conference).
-	   
-	They add a "default for all languages" Label.
-	
-	They manage the permanent identifiers from :
-	   - OpenAlex: P10283
-	   - Semantic Scholar CorpusID: P8299
-	   - Web of Science WOS: P8372
-	   - Microsoft Academic Graph MAG: P6366
-
-	***** END REVISION BLOCK *****
-*/
-
 
 var typeMapping = {
 	// Zotero types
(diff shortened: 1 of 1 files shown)


pmartinolli:
Can somebody help with this issue ? The code is clean and working.

Wikidata QuickStatements.js
```diff
@@ -1,14 +1,14 @@
 {
 	"translatorID": "51e5355d-9974-484f-80b9-f84d2b55782e",
 	"label": "Wikidata QuickStatements",
-	"creator": "Philipp Zumstein",
+	"creator": "Philipp Zumstein (with minor edits by Pascal Martinolli)",
 	"target": "txt",
-	"minVersion": "3.0",
+	"minVersion": "3.1",
```
alex-ter: This is the minimum supported Zotero version, not the translator's version. I'd suggest leaving this untouched.

Wikidata QuickStatements.js
```diff
@@ -1,14 +1,14 @@
 {
 	"translatorID": "51e5355d-9974-484f-80b9-f84d2b55782e",
 	"label": "Wikidata QuickStatements",
-	"creator": "Philipp Zumstein",
+	"creator": "Philipp Zumstein (with minor edits by Pascal Martinolli)",
```
alex-ter: While there's no specific policy about it (that I know of), most recent edits were simply adding "and contributors", to avoid making the lines longer and longer. You might want to consider this. Your name will still be preserved in git.

You'd also generally want to add the same and "-2025" to the copyright statement at line 18.

Wikidata QuickStatements.js
```diff
@@ -35,6 +35,28 @@
...
+	
+	They manage the permanent identifiers from :
+	   - OpenAlex: P10283
+	   - Semantic Scholar CorpusID: P8299
+	   - Web of Science WOS: P8372
+	   - Microsoft Academic Graph MAG: P6366
+
+	***** END REVISION BLOCK *****
+*/
+
```
alex-ter: This should be better put into the commit message instead.

Wikidata QuickStatements.js
```diff
@@ -148,7 +170,11 @@ var identifierMapping = {
 	"Open Library ID": "P648",
 	OCLC: "P243",
 	"IMDb ID": "P345",
-	"Google-Books-ID": "P675"
+	"Google-Books-ID": "P675",
+	OpenAlex: "P10283",
+	CorpusID: "P8299",
+	WOS: "P8372",
+	MAG: "P6366"
```
alex-ter, line 155: Given the way this table is used, it doesn't seem like anything else than `OpenAlex` is going to work, because none of the translators adds those fields into `extra`. Can you maybe provide examples of records that would use those, for me to test?

Wikidata QuickStatements.js
```diff
@@ -260,6 +298,7 @@ function zoteroItemToQuickStatements(item) {
 	if (item.language && (item.language.toLowerCase() in languageMapping)) {
 		let lang = item.language.toLowerCase();
 		addStatement('L' + lang, '"' + item.title + '"');
+		addStatement('Lmul', '"' + item.title + '"');
```
alex-ter: Would it be meaningful to add this one unconditionally (as it's a default anyway), and not only in this branch of the `if`?

pmartinolli:
Thank you @alex-ter for following up, yes I will try to solve the points. I might ask you where to send the revised version once I review it because I am not familiar with Git. Thank you!

Wikidata QuickStatements.js
pmartinolli: CorpusID is retrieved by the plugin Cita with this DOI for example : 10.29304/jqcm.2023.15.1.1142
WOS is retrieved when searching this paper (https://www.nature.com/articles/nature14539) in Web of Science. Then, MAG is retrieved for the same paper when the plugin Cita is used on the reference (Get identifier > CorpusID).

pmartinolli (commented):
Following the all the revisions and advices of @alex-ter, I resubmit a modified version of the code.

pmartinolli (commented):
I reverted the change asked "add this one unconditionally (as it's a default anyway)" because I am not able to make it work.

pmartinolli:
Thank you @alex-ter , Yes I mean that. It makes sense but I cannot code it.

Wikidata QuickStatements.js
alex-ter: Thanks for elaborating. I've analyzed the Cita plugin logic and the WOS plugin, indeed they'd add those pieces into Extra when doing their thing. This one LGTM then.

AbeJellinek:
Thank you @pmartinolli and @alex-ter!

PR #3470, opened 2025-06-24
Add translator for NASA ADS key export

rmglas:
This translator exports the NASA ADS keys for the given bibliography items. We look for a NASA ADS key in the URL and extra fields of each item. If no key is found, we look for an arXiv ID (in the URL, DOI, archiveID, and extra fields) and convert it to a NASA ADS key.

This translator enables an efficient work flow based on latex + biblatex (with NASA ADS keys as citation keys) and Zotero as reference manager.

### DIFF
--- a/NASA ADS Key.js
+++ b/NASA ADS Key.js
@@ -1,0 +1,163 @@
+{
+	"translatorID": "0620079f-8ebb-4bb3-88bd-6f6903075201",
+	"label": "NASA ADS Key",
+	"creator": "Robert Glas",
+	"target": "",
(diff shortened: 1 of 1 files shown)


AbeJellinek:
Hmm, I don't really know. I appreciate your work on this, but it seems like a better fit for a plugin, or a script for an existing plugin like [Actions & Tags](https://github.com/windingwind/zotero-actions-tags). Export translators are really for structured metadata formats, not raw text.

I'm going to close for now. Sorry about that.

PR #3474, opened 2025-07-16
Update EUR-Lex.js

dorjor:
This commit improves the import of files as statutes, i.e. EU regulations. The "bill" file type in Zotero already has a field named "Legislative Body". The file type 'statute' does not. When importing a file of this type, the translator writes the file type in the 'Extra' field. You cannot have unofficial variables in the extra field if you want your style to be registered in the official Zotero style repository. The solution is to use the recognised file variables. The correct CSL field mapping for 'Legislative Body' is 'authority', so 'authority' must be imported into the 'Extra' field for statutes.

### DIFF
--- a/EUR-Lex.js
+++ b/EUR-Lex.js
@@ -205,7 +205,12 @@ function scrape(doc, url) {
 		for (let i = 0; i < passedBy.length; i++) {
 			passedByArray.push(passedBy[i].getAttribute("resource").split("/").pop());
 		}
-		item.legislativeBody = passedByArray.join(", ");
+		
+		if (type == "statute") {
+			item.authority = passedByArray.join(", ");
+		} else {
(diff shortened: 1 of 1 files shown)


dorjor:
@alex-ter, I have made the required changes. Thanks for pointing them out to me. Regarding my proposed change please think of the output the unmodified translator produces right now: visit https://eur-lex.europa.eu/legal-content/EN/TXT/?uri=CELEX:32015R2283 and import the regulation into zotero. Watch the extra field: "Legislative Body: EP, CONSIL". It is to my understanding that "Legislative Body:" as is, can't be used by CSL. So it is there but can't be used without modifying it manually into "Authority:". If my style uses authority for "statute, bill .etc" my proposed modification just seems to work. It imports into the extra field "Authority: EP; CONSIL" and any citation style that cites authority just picks it up."

PR #3477, opened 2025-07-23
Add Usenix.org translator

timlst:
Conference hosted on usenix.org include BibTex style syntax for capitalization in their page metadata.
E.g. [this paper](https://www.usenix.org/conference/usenixsecurity18/presentation/bock) includes 5 brackets within the title

- PR adds a simple translator which removes brace pairs from title field
- tries to account for escaped braces intended in the title, but could not find an example

### DIFF
--- a/usenix.org.js
+++ b/usenix.org.js
@@ -1,0 +1,158 @@
+{
+	"translatorID": "b97462fa-f20b-4a1e-8a73-3a434a81518b",
+	"label": "usenix.org",
+	"creator": "Tim Leonhard Storm",
+	"target": "^https://www\\.usenix\\.org/conference/.*/presentation",
+	"minVersion": "5.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2025-07-23 15:57:59"
+}
+
+/*
+	***** BEGIN LICENSE BLOCK *****
+
+	Copyright © 2025 Tim Leonhard Storm
+
+	This file is part of Zotero.
+
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	Zotero is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


usenix.org.js
```diff
@@ -0,0 +1,158 @@
+{
+	"translatorID": "b97462fa-f20b-4a1e-8a73-3a434a81518b",
+	"label": "usenix.org",
```
AbeJellinek: The convention is to name translators after the site's human name, not the domain, so this should be "USENIX" (and the file should be USENIX.js).

usenix.org.js
```diff
@@ -0,0 +1,158 @@
...
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+/* Remove pairs of unescaped braces (note that braces in title are quite unlikely anyway) */
```
AbeJellinek: Add a newline before this

usenix.org.js
```diff
@@ -0,0 +1,158 @@
...
+
+function detectWeb(doc, url) {
+	if (url.includes('/presentation/')) {
+		return 'conferencePaper';
+	}
+	return false;
+}
+
+async function doWeb(doc, url) {
+	if (detectWeb(doc, url)) {
```
AbeJellinek: No need for this check - `doWeb()` won't be called if we didn't detect.

usenix.org.js
```diff
@@ -0,0 +1,158 @@
...
+}
+
+/** BEGIN TEST CASES **/
+var testCases = [
+	{
+		"type": "web",
+		"url": "https://google.com/",
+		"detectedItemType": false,
+		"items": []
+	},
```
AbeJellinek: Remove

usenix.org.js
timlst: woops, copied from some example

usenix.org.js
timlst: makes sense^

PR #3479, opened 2025-07-25
Update NASA ADS translator to support new app scixplorer.org

shinyichen:
Update NASA ADS translator to support new app scixplorer.org

### DIFF
--- a/NASA ADS.js
+++ b/NASA ADS.js
@@ -1,15 +1,15 @@
 {
 	"translatorID": "7987b420-e8cb-4bea-8ef7-61c2377cd686",
 	"label": "NASA ADS",
-	"creator": "Tim Hostetler, Abe Jellinek, and Zoë C. Ma",
-	"target": "^https://ui\\.adsabs\\.harvard\\.edu/(search|abs)/",
+	"creator": "Tim Hostetler, Abe Jellinek, Zoë C. Ma, Jennifer Chen",
+	"target": "^(https://ui\\.adsabs\\.harvard\\.edu/(search|abs)/)|(^https://scixplorer\\.org/(search|abs/))",
 	"minVersion": "6.0",
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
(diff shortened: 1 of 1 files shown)


NASA ADS.js
```diff
@@ -1,15 +1,15 @@
 {
 	"translatorID": "7987b420-e8cb-4bea-8ef7-61c2377cd686",
 	"label": "NASA ADS",
-	"creator": "Tim Hostetler, Abe Jellinek, and Zoë C. Ma",
-	"target": "^https://ui\\.adsabs\\.harvard\\.edu/(search|abs)/",
+	"creator": "Tim Hostetler, Abe Jellinek, Zoë C. Ma, Jennifer Chen",
+	"target": "^(https://ui\\.adsabs\\.harvard\\.edu/(search|abs)/)|(^https://scixplorer\\.org/(search|abs/))",
```
AbeJellinek, line 5: Can simplify if these have the same path structure:

```suggestion
	"target": "^https://(ui\\.adsabs\\.harvard\\.edu|scixplorer\\.org)/(search|abs)/",
```

NASA ADS.js
```diff
@@ -35,16 +35,36 @@
 	***** END LICENSE BLOCK *****
 */
 
-function getSearchResults(doc, checkOnly = false) {
+function getVariant(url) {
+	let hostname = new URL(url).hostname;
+	if (hostname === "ui.adsabs.harvard.edu") {
+		return "ads";
+	} else {
```
AbeJellinek: House style:


```suggestion
	}
	else {
```

AbeJellinek (changes requested):
Thanks! A few comments. Non-lint-related test failures are false negatives - something is wrong with multiples... I'll figure that out.

AbeJellinek:
Actually not really false negatives. The translator detects `journalArticle` at first, and then switches to `multiple`. There's not really a good way to test that.

NASA ADS.js
shinyichen: Not quite the same, for search in ADS:
```https://ui.adsabs.harvard.edu/search/...```
for SciX:
```https://scixplorer.org/search?.....```

PR #3482, opened 2025-08-05
(relisted) Add Israel's national newspaper library (only)

SomeoneFighter:
Bypassed test for Israel National Library due to cloudflare that makes it impossible to test it. To test it yourself, see for example https://www.nli.org.il/en/newspapers/ajnm/1945/08/10/01/article/26/?e=-------en-20--1--img-txIN%7CtxTI--------------1

### DIFF
--- a/Israel National Newspaper Library.js
+++ b/Israel National Newspaper Library.js
@@ -55,7 +55,7 @@ function detectLanguageFromText(text) {
 async function scrape(doc, url) {
 	const item = new Zotero.Item("newspaperArticle");
 
-	if (url.includes("/article/")) {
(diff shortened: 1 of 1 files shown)


SomeoneFighter:
@AbeJellinek I do not know why the tests are not working, but everything is 100% ready for deployment

SomeoneFighter:
@adam3smith Could you take a look? some of the translators have been waiting to be added for months

AbeJellinek:
adam3smith doesn't work for Zotero. We will merge this, but there isn't anyone on the team whose main focus is translator maintenance right now so it can take a while to work through the backlog. Sorry about that.

PR #3483, opened 2025-08-06
RDF: Move issue to number (don't copy) if issue isn't valid

AbeJellinek:
Fixes the bug reported here: https://forums.zotero.org/discussion/comment/496947/#Comment_496947

It makes sense to put issue info in `number` for item types that don't support `issue`, since `number` is a more general field with more mappings, but it doesn't seem right to put the data in both fields, or to apply the transformation in cases where `issue` is perfectly valid.

Instead, we now move `issue` to `number`, only on item types that don't support `issue`.

### DIFF
--- a/RDF.js
+++ b/RDF.js
@@ -12,7 +12,7 @@
 	},
 	"inRepository": true,
 	"translatorType": 1,
-	"lastUpdated": "2023-04-24 15:51:01"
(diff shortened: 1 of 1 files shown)


adam3smith (approved):
I didn't even realize that's why we kept getting number in Extra. This looks good, thanks!

PR #3485, opened 2025-08-23
Attempt at new LoC translator

adam3smith:
As noted in various form threads, they've redesigned their site and the current translator doesn't work. Unfortunately, the site uses some sort of internal unique ID with an internal API that I'm not able to make work, so currently mutliples don't work.
@AbeJellinek -- could you take a look? I tried both just requesting the document as in the current version and getting the API JSON (commented out) which contains the link to the MarcXML.


AbeJellinek:
Sorry, missed your comment, but I just pushed some additional changes to generalize this to handle all EBSCO Locate catalogs. Sorry if that causes conflicts with your detection work! But I'm done pushing changes now. Let me know if everything still looks OK.

adam3smith (commented):
OK -- I think this is good to merge now, thanks for doing this, that'd have taken me forever to figure out.

PR #3488, opened 2025-09-07
translator for gyorplusz.hu

homope:
date and abstract field from unconventionale source

### DIFF
--- a/gyorplusz.js
+++ b/gyorplusz.js
@@ -1,7 +1,7 @@
 {
 	"translatorID": "67b93a38-5182-45e5-ab35-3b95764f6979",
 	"label": "Győr Plusz",
-	"creator": "Gemini/homope",
+	"creator": "Gemini",
 	"target": "^https?://(www\\.)?gyorplusz\\.hu",
 	"minVersion": "3.0",
 	"maxVersion": "",
@@ -47,18 +47,35 @@ function scrape(doc, url) {
 	item.title = ZU.cleanString(doc.querySelector('div.content-header h1').textContent);
 	item.publicationTitle = "Győr Plusz";
 
-	const authorMeta = doc.querySelector('meta[name="author"]');
-	if (authorMeta) {
-		const authorName = authorMeta.getAttribute('content');
+	// Extract author from the new HTML structure, and fall back to the old one
+	const newAuthorElement = doc.querySelector('.author .name');
+	if (newAuthorElement) {
+		const authorName = ZU.cleanString(newAuthorElement.textContent);
 		item.creators.push(ZU.cleanAuthor(authorName, 'author'));
+	} else {
+		const authorMeta = doc.querySelector('meta[name="author"]');
+		if (authorMeta) {
+			const authorName = authorMeta.getAttribute('content');
+			item.creators.push(ZU.cleanAuthor(authorName, 'author'));
+		}
 	}
-
-	const timeElement = doc.querySelector('.article-info .time');
-	if (timeElement) {
-		const dateString = ZU.cleanString(timeElement.textContent);
+	
+	// Extract the date from the new HTML structure, and fall back to the old one
+	const newTimeElement = doc.querySelector('.author .date');
+	if (newTimeElement) {
+		const dateString = ZU.cleanString(newTimeElement.textContent);
+		// Example date string: "2025.09.05. 12:42"
 		const formattedDate = dateString.replace(/(\d{4})\.\s*(\d{2})\.\s*(\d{2})\.\s*(\d{2}:\d{2})/, '$1-$2-$3T$4:00');
 		item.date = formattedDate;
+	} else {
+		const timeElement = doc.querySelector('.article-info .time');
+		if (timeElement) {
+			const dateString = ZU.cleanString(timeElement.textContent);
+			const formattedDate = dateString.replace(/(\d{4})\.\s*(\d{2})\.\s*(\d{2})\.\s*(\d{2}:\d{2})/, '$1-$2-$3T$4:00');
+			item.date = formattedDate;
+		}
 	}
+
 
 	const leadParagraph = doc.querySelector('p.lead');
 	const abstractMeta = doc.querySelector('meta[property="og:description"]');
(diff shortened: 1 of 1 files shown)


homope:
Fix: Corrected line endings and updated selectors in gyorplusz.js

gyorplusz.js
```diff
@@ -0,0 +1,150 @@
+{
```
alex-ter: Looks like the whole file was generated with wrong EOL characters - you'd want to fix all those lints.

gyorplusz.js
```diff
@@ -0,0 +1,150 @@
+{
+	"translatorID": "67b93a38-5182-45e5-ab35-3b95764f6979",
+	"label": "Győr Plusz",
+	"creator": "Gemini",
```
alex-ter: Here and in the copyright it should be a person, the model/app should not be listed.

gyorplusz.js
```diff
@@ -0,0 +1,150 @@
+{
+	"translatorID": "67b93a38-5182-45e5-ab35-3b95764f6979",
+	"label": "Győr Plusz",
+	"creator": "Gemini",
+	"target": "^https?://(www\\.)?gyorplusz\\.hu",
+	"minVersion": "3.0",
```
alex-ter: Should be at least 5.0 or really 6.0 these days.

gyorplusz.js
```diff
@@ -0,0 +1,150 @@
...
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero. If not, see <http://www.gnu.org/licenses/>.
+
+	***** END LICENSE BLOCK *****
+*/
+
+function detectWeb(doc, url) {
+	if (doc.querySelector('div.content-header h1')) {
+		return "magazineArticle";
```
alex-ter: This selector doesn't look right - it will return `magazineArticle` if it matches, but it's generic enough, so there's no logical connection with the type being returned. If there any better point to be used for spotting articles? Moreover, looking at the page used in the test, this doesn't match at all.

gyorplusz.js
alex-ter, line 1: 1. I'm not sure if that's codified (hopefully Abe can chime in), but typically all translators are named the same/close to the actual website, so in your case a more customary name for this file would be `Gyor Plusz.js`.
2. Why do you think a new translator is needed at all? The default Embedded Metadata one does a good job saving all the information, with the only questionable piece being type == `Blog post`, but then you could simply call the EM translator and correct that afterwards.

gyorplusz.js
```diff
@@ -0,0 +1,150 @@
...
+	return false;
+}
+
+function scrape(doc, url) {
+	let item = new Zotero.Item('magazineArticle');
+	item.title = ZU.cleanString(doc.querySelector('div.content-header h1').textContent);
+	item.publicationTitle = "Győr Plusz";
+
+	// Extract author from the new HTML structure, and fall back to the old one
+	const newAuthorElement = doc.querySelector('.author .name');
```
alex-ter: 1. No need for `const`
2. This element does exist on the test page, but doesn't actually contain the name. The `meta` element used as the fallback does, so you'll want to revisit this selection.

gyorplusz.js
```diff
@@ -0,0 +1,150 @@
...
+	if (newAuthorElement) {
+		const authorName = ZU.cleanString(newAuthorElement.textContent);
```
alex-ter: I don't think there's `cleanString()`, so this looks like model's hallucination. You could use the existing `text()` function.

gyorplusz.js
```diff
@@ -0,0 +1,150 @@
...
+			item.creators.push(ZU.cleanAuthor(authorName, 'author'));
+		}
+	}
+	
+	// Extract the date from the new HTML structure, and fall back to the old one
+	const newTimeElement = doc.querySelector('.author .date');
+	if (newTimeElement) {
+		const dateString = ZU.cleanString(newTimeElement.textContent);
+		// Example date string: "2025.09.05. 12:42"
+		const formattedDate = dateString.replace(/(\d{4})\.\s*(\d{2})\.\s*(\d{2})\.\s*(\d{2}:\d{2})/, '$1-$2-$3T$4:00');
```
alex-ter: `ZU.strToISO()` is a better choice.

gyorplusz.js
```diff
@@ -0,0 +1,150 @@
...
+}
+
+function doWeb(doc, url) {
+	scrape(doc, url);
+}
+
+/** BEGIN TEST CASES **/
+var testCases = [
+	{
+		"type": "web",
```
alex-ter: This test doesn't pass because even the `detectWeb()` doesn't work.

gyorplusz.js
homope: Thank you for the feedback! You are absolutely right — the lack of handling for the lead paragraph (abstract) is precisely why a custom translator is justified.

Gyor Plusz.js
```diff
@@ -0,0 +1,175 @@
...
+	"label": "Győr Plusz",
+	"creator": "homope",
+	"target": "^https?://(www\\.)?gyorplusz\\.hu",
+	"minVersion": "6.0",
+	"maxVersion": "",
+	"priority": 100,
+	"inRepository": true,
+	"translatorType": 4,
+	"browserSupport": "gcsibv",
+	"lastUpdated": "2025-12-10 12:15:00"
```
AbeJellinek: A translator from the future? (I really highly recommend using Scaffold, Tools -> Translator Editor, rather than editing this stuff by hand.)

Gyor Plusz.js
```diff
@@ -0,0 +1,175 @@
...
+
+	// 3. Abstract (Prefer lead paragraph, fall back to meta description)
+	let leadParagraph = doc.querySelector('p.lead');
+	let abstractMeta = doc.querySelector('meta[property="og:description"]');
+	
+	// Use cleanString and textContent for paragraph
+	if (leadParagraph && ZU.cleanString(leadParagraph.textContent).length > 0) {
+		item.abstractNote = ZU.cleanString(leadParagraph.textContent);
+	// Use cleanString and getAttribute('content') for meta tag
+	} else if (abstractMeta) {
```
AbeJellinek: Lint

PR #3493, opened 2025-09-25
Embedded Metadata: do not add duplicates bylines when getting authors

quachpas:
This fixes an issue when duplicate bylines (strict true/false) are added to the list that leads to empty creators, cf. new webtest.

I ran the other webtests and they are working properly for creators. 2 links changed, but in other fields.

### DIFF
--- a/Embedded Metadata.js
+++ b/Embedded Metadata.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2025-06-10 14:42:15"
+	"lastUpdated": "2025-09-25 08:28:33"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


Embedded Metadata.js
```diff
@@ -761,7 +761,9 @@ function getAuthorFromByline(doc, newItem) {
 			Z.debug(`Found ${byline.length} elements with '${bylineClass}' class (strict: ${isStrict})`);
 			for (let bylineElement of byline) {
 				if (!bylineElement.innerText?.trim()) continue;
-				bylines.push(bylineElement);
+				if (!bylines.includes(bylineElement)) {
+					bylines.push(bylineElement);
+				}
```
AbeJellinek: Formatting nit - let's do this:

```suggestion
				if (bylines.includes(bylineElement)) continue;
				bylines.push(bylineElement);
```

PR #3494, opened 2025-09-30
adding plural versions of some resource types for Primo XML

cbaksik:
plural versions of resource types are new defaults in Primo VE. Also adding new resource type from CDI.


cbaksik:
Hi - I just tried adding another test. You should see a citation for a video titled "Cat and mouse." Did I add it correctly? ty!

PR #3497, opened 2025-10-07
Papers Past: Enhanced metadata extraction for newspapers

almostsentient:
Enhanced newspaper metadata extraction using JSON-LD and structured data.

### DIFF
--- a/Papers Past.js
+++ b/Papers Past.js
@@ -1,21 +1,21 @@
 {
 	"translatorID": "1b052690-16dd-431d-9828-9dc675eb55f6",
 	"label": "Papers Past",
-	"creator": "Philipp Zumstein and Abe Jellinek",
+	"creator": "Philipp Zumstein, Abe Jellinek, and Jason Murphy",
 	"target": "^https?://(www\\.)?paperspast\\.natlib\\.govt\\.nz/",
-	"minVersion": "3.0",
+	"minVersion": "5.0",
 	"maxVersion": "",
-	"priority": 100,
+	"priority": 200,
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2021-07-12 17:17:15"
+	"lastUpdated": "2025-09-30 06:00:00"
 }
 
 /*
 	***** BEGIN LICENSE BLOCK *****
 
-	Copyright © 2017-2021 Philipp Zumstein and Abe Jellinek
+	Copyright © 2025 Philipp Zumstein, Abe Jellinek, and Jason Murphy
 
 	This file is part of Zotero.
 
@@ -35,15 +35,31 @@
 	***** END LICENSE BLOCK *****
 */
 
+/*
+	Papers Past Translator
+	
+	Enhanced metadata extraction for newspaper articles using structured data sources
+	(JSON-LD, Highwire Press, Dublin Core) with fallbacks to screen scraping.
+	Legacy scraping maintained for other collection types (periodicals, manuscripts, parliamentary papers). 
+*/
+
+// ============================================================================
+// Detection and Routing
+// ============================================================================
 
 function detectWeb(doc, url) {
+	// Check for newspaper article
+	if (isNewspaperArticle(doc, url)) {
+		return "newspaperArticle";
+	}
+
+	// Handle search results
 	if (/[?&]query=/.test(url) && getSearchResults(doc, true)) {
 		return "multiple";
 	}
+	
+	// Handle other collection types
 	else if (ZU.xpathText(doc, '//h3[@itemprop="headline"]')) {
-		if (url.includes('/newspapers/')) {
-			return "newspaperArticle";
-		}
 		if (url.includes('/periodicals/')) {
 			return "journalArticle";
 		}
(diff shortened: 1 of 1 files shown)


Papers Past.js
```diff
@@ -1,21 +1,21 @@
...
 	"translatorID": "1b052690-16dd-431d-9828-9dc675eb55f6",
 	"label": "Papers Past",
-	"creator": "Philipp Zumstein and Abe Jellinek",
+	"creator": "Philipp Zumstein, Abe Jellinek, and Jason Murphy",
 	"target": "^https?://(www\\.)?paperspast\\.natlib\\.govt\\.nz/",
-	"minVersion": "3.0",
+	"minVersion": "5.0",
 	"maxVersion": "",
-	"priority": 100,
+	"priority": 200,
```
AbeJellinek: Website-specific translators should almost always have priority 100. No need to change this.

Papers Past.js
```diff
@@ -84,20 +88,126 @@ function doWeb(doc, url) {
...
+		if (checkOnly) return true;
+		found = true;
+		items[href] = title;
+	}
+	return found ? items : false;
+}
+
+function scrapeLegacy(doc, url) {
 	var type = detectWeb(doc, url);
+	if (!type) return false;
```
AbeJellinek: No need for this. `doWeb()` won't run if detection fails.

Papers Past.js
```diff
@@ -84,20 +88,126 @@ function doWeb(doc, url) {
...
+	
 	var item = new Zotero.Item(type);
 	var title = ZU.xpathText(doc, '//h3[@itemprop="headline"]/text()[1]');
-	item.title = ZU.capitalizeTitle(title.toLowerCase(), true);
+	if (title) {
+		item.title = ZU.capitalizeTitle(title.toLowerCase(), true);
```
AbeJellinek: I know we were already doing this, but it isn't great and should be changed. The only situation in which force-title casing makes sense is when the title is in all caps (`title.toUpperCase()  === title`), so we could put it behind that check.

Papers Past.js
```diff
@@ -136,30 +248,31 @@ function scrape(doc, url) {
...
 	}
 	
-	item.abstractNote = text(doc, '#tab-english');
-
+	item.abstractNote = ZU.xpathText(doc, '//div[@id="tab-english"]');
 	item.url = ZU.xpathText(doc, '//div[@id="researcher-tools-tab"]/input/@value');
-	if (!item.url) item.url = text('#researcher-tools-tab p');
-	if (!item.url || !item.url.startsWith('http')) item.url = url;
+	if (!item.url) {
+		item.url = ZU.xpathText(doc, '//div[@id="researcher-tools-tab"]//p');
```
AbeJellinek: Try not to introduce new XPaths when a CSS selector will do:


```suggestion
		item.url = text(doc, '#researcher-tools-tab p');
```

Papers Past.js
```diff
@@ -136,30 +248,31 @@ function scrape(doc, url) {
...
+		item.url = url;
+	}
 	
 	item.attachments.push({
 		title: "Snapshot",
 		document: doc
 	});
 	
-	let imagePageURL = attr(doc, '.imagecontainer a', 'href');
+	var imagePageURL = ZU.xpathText(doc, '//div[@class="imagecontainer"]//a/@href');
```
AbeJellinek: Revert this - there was nothing wrong with the existing code!

Papers Past.js
```diff
@@ -136,30 +248,31 @@ function scrape(doc, url) {
...
 	if (imagePageURL) {
 		ZU.processDocuments(imagePageURL, function (imageDoc) {
 			item.attachments.push({
 				title: 'Image',
 				mimeType: 'image/jpeg',
-				url: attr(imageDoc, '.imagecontainer img', 'src')
+				url: ZU.xpathText(imageDoc, '//div[@class="imagecontainer"]//img/@src')
```
AbeJellinek: Ditto

Papers Past.js
```diff
@@ -169,6 +282,147 @@ function scrape(doc, url) {
...
+			dc[name.replace(/^dc\./, "DC.")] = content;
+		}
+	}
+	return { hw, dc };
+}
+
+// Parse bibliographic details from the on-page citation text
+function parseBibliographicDetails(doc) {
+	const cite = doc.querySelector('#researcher-tools-tab .citation, .tabs-panel .citation, p.citation');
+	const text = cite ? cite.textContent : "";
```
AbeJellinek: Avoid `const` for things that aren't actually static constants; use `text()` utility:

```suggestion
	let text = text(doc, '#researcher-tools-tab .citation, .tabs-panel .citation, p.citation');
```

Papers Past.js
```diff
@@ -169,6 +282,147 @@ function scrape(doc, url) {
...
+	const out = { publicationTitle: "", volume: "", issue: "", date: "", pages: "" };
```
AbeJellinek:
```suggestion
	let out = { publicationTitle: "", volume: "", issue: "", date: "", pages: "" };
```

and below

Papers Past.js
```diff
@@ -169,6 +282,147 @@ function scrape(doc, url) {
...
+
+	const pageMatch = text.match(/Page\s+([0-9A-Za-z\-]+)/i);
+	if (pageMatch) out.pages = clean(pageMatch[1]);
+
+	return out;
+}
+
+// Trim and normalize whitespace
+function clean(s) {
+	return s ? s.replace(/\s+/g, " ").trim() : "";
```
AbeJellinek: This does the same thing as `ZU.trimInternal()`.

Papers Past.js
```diff
@@ -169,6 +282,147 @@ function scrape(doc, url) {
...
+}
+
+// Return first non-duplicate value from array
+function dedupeFirst(arr) {
```
AbeJellinek, line 319: What issue is this preventing? The logic is kind of hard to follow, but since it always exits after adding to `seen`, isn't it just returning the first non-empty value in the array? `seen` will never be read after being written to.

Papers Past.js
```diff
@@ -169,6 +282,147 @@ function scrape(doc, url) {
...
+		if (!seen.has(k)) {
+			seen.add(k);
+			return v;
+		}
+	}
+	return arr.find(Boolean) || "";
+}
+
+// Fix all-caps titles to title case, avoiding issues with apostrophes
+function fixTitleCase(str) {
```
AbeJellinek, line 323: OK, this is good, and we should use it for the existing code where we force-title case the title.

Papers Past.js
```diff
@@ -169,6 +282,147 @@ function scrape(doc, url) {
...
+	if (!str) return str;
+	const letters = str.replace(/[^A-Za-z]/g, "");
+	if (!letters) return str;
+	const uppers = (letters.match(/[A-Z]/g) || []).length;
+	const upperRatio = uppers / letters.length;
+	
+	// If more than 60% uppercase, convert to title case
+	if (upperRatio > 0.6) {
+		// Only capitalize after spaces or at start, not after apostrophes
+		return str.toLowerCase().replace(/(^|\s)\w/g, c => c.toUpperCase());
```
AbeJellinek: Does `ZU.capitalizeTitle()` not already handle this correctly?

Papers Past.js
```diff
@@ -169,6 +282,147 @@ function scrape(doc, url) {
...
+}
+
+// Remove query parameters and hash from URL
+function stripQueryAndHash(u) {
+	try {
+		const x = new URL(u);
+		x.search = "";
+		x.hash = "";
+		return x.toString();
+	} catch (e) {
```
AbeJellinek: Lint: `npm run lint -- --fix "Papers Past.js"`

Papers Past.js
```diff
@@ -178,42 +432,18 @@ var testCases = [
 	},
 	{
 		"type": "web",
-		"url": "https://paperspast.natlib.govt.nz/newspapers/EP19440218.2.61",
+		"url": "https://paperspast.natlib.govt.nz/newspapers/MT19390701.2.6.3",
```
AbeJellinek, line 390: We should keep the old tests in addition to the new ones, rather than replacing them.

PR #3498, opened 2025-10-08
New Translator: FAOLEX Database

swifterslb:
This is a translator for FAOLEX Database (https://www.fao.org/faolex). Please note that the function of multiple items doesn't work because the site use dynamic elements.


swifterslb:
`"defer": true` was used but still get mixed results. Not all tests passed.

PR #3499, opened 2025-10-08
Adding new translator Anarchist Library.js

pretty-weed:
This translator handles metadata for theanarchistlibrary.org. Development done in https://github.com/anti-toil-at-seminary/zotero-translator-anarchist-library. 

This translator is still a work in progress, it is overzealous in downloading attachments, and I want to make that behavior more opt-in, and there is useful work to be done in improving source backtracking to add to notes and extra, but this WIP is currently functional

### DIFF
--- a/Anarchist Library.js
+++ b/Anarchist Library.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2025-10-07 11:18:38"
+	"lastUpdated": "2025-10-25 01:31:43"
 }
 
 /*
@@ -163,9 +163,7 @@ async function doWeb(doc, url) {
 	}
 }
 
-
 /** BEGIN TEST CASES **/
-
 var testCases = [
 	{
 		"type": "web",
@@ -219,7 +217,6 @@ var testCases = [
 			}
 		]
 	},
-
 	{
 		"type": "web",
 		"url": "https://theanarchistlibrary.org/library/errico-malatesta-the-general-strike-and-the-insurrection-in-italy",
@@ -290,12 +287,20 @@ var testCases = [
 					}
 				],
 				"notes": [
-					{ "note": "Translated by Alexia Grosjean." },
-					{ "note": "Source: Retrieved on 11th March 2025 from www.skbl.se" }
+					{
+						"note": "Translated by Alexia Grosjean."
+					},
+					{
+						"note": "Source: Retrieved on 11th March 2025 from www.skbl.se"
+					}
 				],
 				"tags": [
-					{ "tag": "Sweden" },
-					{ "tag": "biography" }
+					{
+						"tag": "Sweden"
+					},
+					{
+						"tag": "biography"
+					}
 				],
 				"date": "2018-03-08",
 				"seeAlso": [],
@@ -331,13 +336,23 @@ var testCases = [
 					}
 				],
 				"notes": [
-					{ "note": "Translated by Reddebrek." },
-					{ "note": "Source: Provided by the translator." }
+					{
+						"note": "Translated by Reddebrek."
+					},
+					{
+						"note": "Source: Provided by the translator."
+					}
 				],
 				"tags": [
-					{ "tag": "history" },
-					{ "tag": "individualism" },
-					{ "tag": "proto-anarchism" }
+					{
+						"tag": "history"
+					},
+					{
+						"tag": "individualism"
+					},
+					{
+						"tag": "proto-anarchism"
+					}
 				],
 				"date": "1933",
 				"seeAlso": [],


Anarchist Library.js
```diff
@@ -0,0 +1,455 @@
+{
+	"translatorID": "1a31e4c5-22ed-4b5b-a75f-55476db29a44",
+	"label": "Anarchist Library",
+	"creator": "Sister Baæ'l",
+	"target": "https://theanarchistlibrary.org/(latest|library|stats/popular|category/topic|category/author|special/index|search)",
```
AbeJellinek: Escape regex special chars:

```suggestion
	"target": "https://theanarchistlibrary\\.org/(latest|library|stats/popular|category/topic|category/author|special/index|search)",
```

Anarchist Library.js
```diff
@@ -0,0 +1,455 @@
...
+If you do any work on this translator, please add yourself here <3.
+*/
+
+var urlBase = "https://theanarchistlibrary.org/";
+
+// ToDo: Localization
+var languageNames = new Intl.DisplayNames(['en'], { type: 'language' });
+
+var allAttachmentTypes = {
+	"Plain PDF": ".pdf",
```
AbeJellinek: Let's just grab the PDF for now. From the couple random examples I've looked at, it seems to be the prettiest and most lossless - the HTML and EPUB lose some of the nice layout, and the TeX isn't really useful in Zotero. We can revisit this later, though.

Anarchist Library.js
```diff
@@ -0,0 +1,455 @@
...
+	// todo copied from below
+	let items = {};
+	let results = doc.querySelectorAll('a.list-group-item');
+	for (let i = 0; i < results.length; i++) {
+		items[results[i].href] = ZU.trimInternal(text(results[i], "strong"));
+	}
+	return items;
+}
+
+function getSearchItems(doc) {
```
AbeJellinek: If possible, we should use the standard `detectWeb()`/`doWeb()`/`getSearchResults()` boilerplate from the web translator template (Scaffold: Plus button -> Web Translator). It seems like it'll fit this translator fine, with less indirection that might confuse future maintainers.

Anarchist Library.js
```diff
@@ -0,0 +1,455 @@
...
+		items[results[i].href] = ZU.trimInternal(text(results[i], "strong"));
+	}
+	return items;
+}
+
+async function doLibraryItem(doc, url = doc.location.href) {
+	// ToDo: get fancier here, allow other types
+	let item = new Zotero.Item('webpage');
+
+	let language = languageNames.of(attr(doc, "html", "lang"));
```
AbeJellinek: `language` should actually contain an ISO code, so there's no need to resolve it to a human-readable name.

Anarchist Library.js
```diff
@@ -0,0 +1,455 @@
...
+	item.url = url;
+	item.language = language;
+
+	let itemType = attr(doc, '[property~="og:type"]', 'content');
+	const tagNodeList = doc.querySelectorAll(`[property~="og:${itemType}:tag"]`);
+	let description = attr(doc, '[property~="og:description"]', 'content');
+	let author = attr(doc, `[property~="og:${itemType}:author"]`, 'content');
+	let authorFirstName = author.substring(0, author.indexOf(' '));
+	let authorLastName = author.substring(author.indexOf(' ') + 1);
+	item.creators.push({ creatorType: "author", firstName: authorFirstName, lastName: authorLastName });
```
AbeJellinek:
```suggestion
	item.creators.push(ZU.cleanAuthor(author, "author"));
```

should do the trick.

Anarchist Library.js
```diff
@@ -0,0 +1,455 @@
...
+		LaTeX: allAttachmentTypes.LaTeX,
+		Snapshot: allAttachmentTypes.Snapshot
+	};
+
+	item.accessed = new Date().toString();
+	item.url = url;
+	item.language = language;
+
+	let itemType = attr(doc, '[property~="og:type"]', 'content');
+	const tagNodeList = doc.querySelectorAll(`[property~="og:${itemType}:tag"]`);
```
AbeJellinek: We generally avoid `const` for things that aren't static constants, so:

```suggestion
	let tagNodeList = doc.querySelectorAll(`[property~="og:${itemType}:tag"]`);
```

Anarchist Library.js
```diff
@@ -0,0 +1,455 @@
...
+		let translatedMatch = description.match(re);
+		if (translatedMatch) {
+			let translator = { creatorType: "translator" };
+			if (translatedMatch[0].match(/ /)) {
+				translator.firstName = translatedMatch[0].substring(0, translatedMatch.indexOf(' '));
+				translator.lastName = translatedMatch[0].substring(translatedMatch.indexOf(' ') + 1);
+			}
+			else {
+				translator.lastName = translatedMatch[0];
+			}
```
AbeJellinek: Can simplify this using `ZU.cleanAuthor()` as above.

Anarchist Library.js
```diff
@@ -0,0 +1,455 @@
...
+
+	let date = getPreambleVal(doc, "textdate");
+	let notes = getPreambleVal(doc, "preamblenotes");
+	// misses link here: https://theanarchistlibrary.org/library/margaret-killjoy-it-s-time-to-build-resilient-communities
+	let source = getPreambleVal(doc, "preamblesrc");
+
+	let tags = [];
+	for (let i = 0; i < tagNodeList.length; i++) {
+		tags = tags.concat(tagNodeList[i].content);
+	}
```
AbeJellinek: Simplify:

```suggestion
	for (let tagNode of tagNodeList) {
		item.tags.push({ tag: tagNode.content });
	}
```

and remove the assignment below.

Anarchist Library.js
```diff
@@ -0,0 +1,455 @@
...
+	item.title = title;
+	item.tags = tags;
+	item.date = date;
+	if (notes) {
+		item.notes.push({ note: ZU.trimInternal(notes) });
+	}
+	if (source) {
+		item.notes.push({ note: `Source: ${ZU.trimInternal(source)}` });
+	}
+	item.attachments = [];
```
AbeJellinek: This is the default value, so no need.

Anarchist Library.js
```diff
@@ -0,0 +1,455 @@
...
+
+	for (const [typeName, attachmentExt] of Object.entries(attachmentTypes)) {
```
AbeJellinek: As I mentioned above, we can pare this down to just the PDF for now (no need to add a snapshot when there's a PDF with the same content). The object added to `item.attachments` needs to have a `mimeType` member, which should be `"application/pdf"`.

Anarchist Library.js
```diff
@@ -0,0 +1,455 @@
...
+
+
+var matchers = {
+	list: { matcher: listRe, type: "multiple", do: null, get: getListItems },
+	search: { matcher: searchRe, type: "multiple", do: null, get: getSearchItems },
+	document: { matcher: libraryRe, type: "webpage", do: doLibraryItem, get: null }
+};
+
+
+function detectWeb(doc, url) {
```
AbeJellinek, line 138: As I mentioned above, let's use the standard boilerplate for this.

Anarchist Library.js
pretty-weed: Thanks so much for the feedback! I'm a relative noob to js, appreciate getting some guidance.

Anarchist Library.js
pretty-weed: This is the stuff I was always such a hawk about when I was working 🤦🏻‍♀️. Appreciate that you're looking out for this, the feedback is really excellent in general

Anarchist Library.js
pretty-weed: So, on one of my test cases, where the Author's last name is "O' Malley", this causes test failures. I get if I should still go this way because the library is generally reliable and maybe make a bug report somewhere? or maybe not?

pretty-weed: Same with the author `Voltairine de Cleyre`, the `de` gets moved into the first name 
https://theanarchistlibrary.org/library/voltairine-de-cleyre-report-of-the-work-of-the-chicago-mexican-liberal-defense-league

Anarchist Library.js
pretty-weed: I left the variable here, and just snag what I want from it in `scrape()`, that okay with you or should I remove this map?

Anarchist Library.js
pretty-weed: Having issues with last names such as "O'Malley" or "De Cleyre", see other comment

Anarchist Library.js
pretty-weed: resolved in 6ca9aa65aca0b7f5d840aeb00a8f6d845bc0e99e

Anarchist Library.js
pretty-weed: resolved in 6ca9aa65aca0b7f5d840aeb00a8f6d845bc0e99e

Anarchist Library.js
pretty-weed: done in 6ca9aa65aca0b7f5d840aeb00a8f6d845bc0e99e

Anarchist Library.js
pretty-weed: resolved in 6ca9aa65aca0b7f5d840aeb00a8f6d845bc0e99e

Anarchist Library.js
pretty-weed: I think I resolved this in 6ca9aa65aca0b7f5d840aeb00a8f6d845bc0e99e, can you verify that my updates matches what you'd want to see?

Anarchist Library.js
pretty-weed: fixed here: https://github.com/zotero/translators/pull/3499/commits/6ca9aa65aca0b7f5d840aeb00a8f6d845bc0e99e#diff-919fd3a0ec282c9abba906673b09529fbfb5d45f0671665c960286c73039693eR79

Anarchist Library.js
pretty-weed: removed here: [6ca9aa65](https://github.com/zotero/translators/pull/3499/commits/6ca9aa65aca0b7f5d840aeb00a8f6d845bc0e99e#diff-919fd3a0ec282c9abba906673b09529fbfb5d45f0671665c960286c73039693eL144)

Anarchist Library.js
pretty-weed: though my route would have issues with an author such as "James Thomas Frey" or, my name, "Dandelion Tyler Mark Good", I'm happy either way, just wanted to see if I should raise an issue somewhere else, really.

pretty-weed: Used `cleanAuthor` since either has problems :)

Anarchist Library.js
```diff
@@ -0,0 +1,452 @@
...
+This translator was developed by Dandelion Good.
+
+If you do any work on this translator, please add yourself here <3.
+*/
+
+var allAttachmentTypes = {
+	"Plain PDF": { ext: ".pdf", mimeType: "application/pdf" },
+	"A4 PDF": { ext: ".a4.pdf", mimeType: "application/pdf" },
+	"Letter PDF": { ext: ".lt.pdf", mimeType: "application/pdf" },
+	EPub: { ext: ".eupb", mimeType: "application/epub+zip" },
```
AbeJellinek: Didn't comment on this before because I thought we might just remove it, but if we're keeping it around:


```suggestion
	EPub: { ext: ".epub", mimeType: "application/epub+zip" },
```

Anarchist Library.js
```diff
@@ -0,0 +1,452 @@
...
+async function scrape(doc, url = doc.location.href) {
+	// ToDo: get fancier here, allow other types
+	let item = new Zotero.Item('webpage');
+
+	// These may be expanded on in the future
+	let attachmentTypes = {
+		PDF: allAttachmentTypes["Plain PDF"],
+	};
+
+	item.accessed = new Date().toString();
```
AbeJellinek: This is filled automatically - can remove.

Anarchist Library.js
```diff
@@ -0,0 +1,452 @@
...
+
+	for (const [typeName, typeInfo] of Object.entries(attachmentTypes)) {
+		let attachment = {
+			title: typeName,
+			url: `${doc.location.href}${typeInfo.ext}`,
+			mimeType: typeInfo.mimeType
+		};
+
+		if (typeInfo.ext == "snapshot") {
+			attachment.snapshot = true;
```
AbeJellinek: This is the default for HTML attachments - you just override it to false if you _don't_ want a snapshot.

Anarchist Library.js
```diff
@@ -0,0 +1,452 @@
...
+		}
+
+		item.attachments.push(attachment);
+	}
+
+	return item.complete();
```
AbeJellinek: `complete()` only returns a promise in async import translation (it's complicated), and `scrape` doesn't need to return anything, so:

```suggestion
	item.complete();
```

Anarchist Library.js
```diff
@@ -0,0 +1,452 @@
...
+}
+
+var libraryRe = new RegExp(String.raw`library/`);
+
+
+function detectWeb(doc, url) {
+	if (url.match(libraryRe)) {
```
AbeJellinek:
```suggestion
	if (libraryRe.test(url)) {
```

Anarchist Library.js
```diff
@@ -0,0 +1,452 @@
...
+			attachment.snapshot = true;
+		}
+
+		item.attachments.push(attachment);
+	}
+
+	return item.complete();
+}
+
+var libraryRe = new RegExp(String.raw`library/`);
```
AbeJellinek:
```suggestion
var libraryRe = /library\//;
```

Anarchist Library.js
```diff
@@ -0,0 +1,452 @@
...
+		}
+	}
+	else {
+		await scrape(doc, url);
+	}
+}
+
+/*
+var unusedTestCases = [
+	// fails tests because Voltairine de Cleyre parses to Cleyre, Voltairine de
```
AbeJellinek: We can remove these; re-add if we get a better name splitting routine in the future. It's always best if sites expose author names in `Last, First` format so it's unambiguous. If they don't... it is what it is. `cleanAuthor()` is fairly naive right now, but the assumption that only the last particle is the family name holds for like 95% of western names, so until we add some better heuristics for name splitting, it's the best we have. If we assumed that only the first particle is the first name, we'd be correct less often.

Anarchist Library.js
```diff
@@ -0,0 +1,452 @@
...
+			return true;
+		}
+		items[results[i].href] = ZU.trimInternal(text(results[i], "strong"));
+	}
+	return items;
+}
+
+async function scrape(doc, url = doc.location.href) {
+	// ToDo: get fancier here, allow other types
+	let item = new Zotero.Item('webpage');
```
AbeJellinek: Forgot to comment this before, but I don't think `webpage` is really right here. These seem more like `manuscript`s to me, even though they're published on the web.

AbeJellinek (changes requested):

Anarchist Library.js
```diff
@@ -0,0 +1,452 @@
...
+function getSearchResults(doc, checkOnly) {
+	let items = {};
+	let results = doc.querySelectorAll('a.list-group-item');
+	for (let i = 0; i < results.length; i++) {
+		if (checkOnly) {
+			return true;
+		}
+		items[results[i].href] = ZU.trimInternal(text(results[i], "strong"));
+	}
+	return items;
```
AbeJellinek: Let's just align this fully with the standard format:

```suggestion
	var items = {};
	var found = false;
	var rows = doc.querySelectorAll("a.list-group-item");
	for (let row of rows) {
		let href = row.href;
		let title = ZU.trimInternal(text(row, "strong"));
		if (!href || !title) continue;
		if (checkOnly) return true;
		found = true;
		items[href] = title;
	}
	return found ? items : false;
```

Anarchist Library.js
pretty-weed: Much better! wasn't sure about the type but had forgotten to ask

Anarchist Library.js
```diff
@@ -0,0 +1,455 @@
...
+		if (checkOnly) return true;
+		found = true;
+		items[href] = title;
+	}
+	return found ? items : false;
+}
+
+async function scrape(doc, url = doc.location.href) {
+	// ToDo: get fancier here, allow other types
+	let item = new Zotero.Item('webpage');
```
pretty-weed:
```suggestion
	let item = new Zotero.Item('manuscript');
```

Anarchist Library.js
```diff
@@ -0,0 +1,455 @@
...
+
+	// These may be expanded on in the future
+	let attachmentTypes = {
+		PDF: allAttachmentTypes["Plain PDF"],
+	};
+
+	item.accessed = new Date().toString();
```
pretty-weed:
```suggestion
```

Anarchist Library.js
pretty-weed: removed! thanks!

Anarchist Library.js
pretty-weed: ah, had meant to have this _and_ the attachment.document = doc, removed the snapshot, added what I intended back in

pretty-weed:
Thank you _so much_ for all this feedback @AbeJellinek ! (absolutely no sarcasm!) you're so helpful!

pretty-weed:
@AbeJellinek have I missed changes that you wanted me to make? (not meaning to rush, just seeing if I missed something)

AbeJellinek:
Could you just open this with Scaffold and re-save? Seems like some test case stuff was reformatted externally. I don't seem to have commit access on this PR - otherwise I'd do it myself.

Other than that, looks good to me.

AbeJellinek:
Thank you!

PR #3502, opened 2025-10-14
Update Prime 9ja Online.js

vanderwaalforces:
Fixes issues like authors that are not humans (Agency reports), they should be excluded from being authors. Adjust to cover pidgin.prime9ja.com.ng and not just www.prime9ja.com.ng. And several other adjustments

### DIFF
--- a/Prime 9ja Online.js
+++ b/Prime 9ja Online.js
@@ -311,7 +311,7 @@ var testCases = [
 						"creatorType": "author"
 					}
 				],
-				"date": "2025-05-24T18:10:00+01:00",
+				"date": "2025-05-24",
 				"ISSN": "3092-8907",
 				"abstractNote": "AKURE —  The Ondo State Governorship Election Petitions Tribunal will deliver its verdict on June 4 in the series of suits challenging the e...",
 				"libraryCatalog": "Prime 9ja Online",
@@ -344,7 +344,7 @@ var testCases = [
 						"creatorType": "author"
 					}
 				],
-				"date": "2025-05-27T01:11:00+01:00",
+				"date": "2025-05-27",
 				"ISSN": "3092-8907",
 				"abstractNote": "On “CFMF”  — the fourth track from Davido’s 2025 album 5ive  —   the artist trades club-ready bravado for inward reflection. Featuri...",
 				"libraryCatalog": "Prime 9ja Online",
@@ -378,7 +378,7 @@ var testCases = [
 						"creatorType": "author"
 					}
 				],
-				"date": "2025-05-23T22:38:00+01:00",
+				"date": "2025-05-23",
 				"ISSN": "3092-8907",
 				"abstractNote": "ABUJA —  A major network of cybercriminals allegedly responsible for infiltrating the Computer-Based Testing (CBT) infrastructure of Nigeria...",
 				"libraryCatalog": "Prime 9ja Online",


Prime 9ja Online.js
```diff
@@ -2,13 +2,13 @@
 	"translatorID": "eb0bbbf8-7f57-40fa-aec2-45480d396e93",
 	"label": "Prime 9ja Online",
 	"creator": "VWF",
-	"target": "^https?://(www\\.)?prime9ja\\.com\\.ng/\\d{4}/\\d{2}/[^/]+\\.html",
+	"target": "^https?://(www\\.|pidgin\\.)?prime9ja\\.com\\.ng(/.*)?$",
```
AbeJellinek:
```suggestion
	"target": "^https?://(www\\.|pidgin\\.)?prime9ja\\.com\\.ng/",
```

Prime 9ja Online.js
```diff
@@ -34,141 +34,317 @@
...
+		return 'newspaperArticle';
+	}
+
+	// 2) explicit index/list URL
+	if (isIndexURL(url)) {
+		return 'multiple';
+	}
+
+	// 3) conservative listing detection
+	let listingCount = countListingAnchors(doc);
```
AbeJellinek: Is there a reason why this has to change to something so complicated and nonstandard? I'd really rather keep the standard `getSearchResults()` format we had before.

Prime 9ja Online.js
```diff
@@ -34,141 +34,317 @@
...
+		item.language = data.inLanguage || meta(doc, 'og:locale') || 'en';
+
+		item.publicationTitle = (data.publisher && (data.publisher.name || data.publisher.title)) || 'Prime 9ja Online';
+
+		// --- handle ISSN safely (lint-friendly formatting) ---
+		if (data.publisher) {
+			let pub = data.publisher;
+			let issn
+				= pub.issn
+				|| (pub.identifier && /\d{4}-\d{3}[\dxX]/.test(pub.identifier)
```
AbeJellinek: What are we handling with this? Is there not just one ISSN for this publication?

Prime 9ja Online.js
```diff
@@ -34,141 +34,317 @@
...
-					if (author.name) {
-						item.creators.push(ZU.cleanAuthor(author.name, 'author'));
+			let authors = Array.isArray(data.author) ? data.author : [data.author];
+			for (let a of authors) {
+				let name = (a && (a.name || a['@name'] || a)) || '';
+				if (name) {
+					let lower = name.toString().toLowerCase();
+					if (/news agency|agency|news desk|publish desk|prime 9ja|prime9ja|online media|media|staff|bureau/i.test(lower)) {
+						// skip org-like bylines
+					} else {
```
AbeJellinek: `npm run lint -- --fix "Prime 9ja Online.js"`

Prime 9ja Online.js
```diff
@@ -34,141 +34,317 @@
...
+	if (!item.abstractNote || !item.abstractNote.trim()) {
+		item.abstractNote = ZU.unescapeHTML(
+			meta(doc, 'og:description')
+			|| meta(doc, 'description')
+			|| ''
+		);
+	}
+
+	// If date still empty, try article:published_time meta (often ISO)
+	if (!item.date || !item.date.trim()) {
```
AbeJellinek, line 249: This is so complicated - what are we doing here that `ZU.strToISO()` doesn't do?

Prime 9ja Online.js
```diff
@@ -34,141 +34,317 @@
...
-						"firstName": "Chima Joseph",
-						"lastName": "Ugo",
-						"creatorType": "author"
-					}
-				],
-				"date": "2025-05-24T18:10:00+01:00",
-				"ISSN": "3092-8907",
-				"abstractNote": "AKURE —  The Ondo State Governorship Election Petitions Tribunal will deliver its verdict on June 4 in the series of suits challenging the election of Governor Lucky Aiyedatiwa, who emerged victorious in the last gubernatorial poll. Justice Benson Ogbu, wh…",
-				"language": "en",
+				"title": "Authors and Contributors",
```
AbeJellinek: - Keep the existing test
- This doesn't look like a valid test case - just looks like an author listing page to me.

Prime 9ja Online.js
vanderwaalforces: @AbeJellinek Thank you so much for your comments. For some strange reason, the date param is always missing, so this is more or less a fallback. I have also modified the method to `ZU.strToISO()`. I hope it looks good

vanderwaalforces:
I have also addressed other comments, sir!

Prime 9ja Online.js
```diff
@@ -74,74 +102,206 @@ function getSearchResults(doc, checkOnly) {
...
+			|| ''
+		);
+	}
+
+	// If date still empty, try article:published_time meta (often ISO)
+	if (!item.date || !item.date.trim()) {
+		let metaDate = meta(doc, 'article:published_time') || '';
+		if (metaDate) {
+			// meta often already ISO with timezone; keep it, else use ZU.strToISO
+			if (metaDate.includes('T') !== -1) {
```
AbeJellinek: Just run it through `strToISO()` no matter what IMO. Zoned datetimes are hard to read and don't really provide much more useful information for a citation.

Prime 9ja Online.js
```diff
@@ -74,74 +102,206 @@ function getSearchResults(doc, checkOnly) {
...
+				}
 			}
 		}
 	}
 
+	if (!item.url || !item.url.trim()) {
+		item.url = meta(doc, 'og:url') || url;
+	}
+
+	if (!item.publicationTitle) {
```
AbeJellinek, line 266: This is already set above in all cases.

Prime 9ja Online.js
```diff
@@ -74,74 +102,206 @@ function getSearchResults(doc, checkOnly) {
...
+		item.publicationTitle = 'Prime 9ja Online';
+	}
+
+	if (!item.ISSN) {
```
AbeJellinek, line 270: Also already set above

Prime 9ja Online.js
```diff
@@ -74,74 +102,206 @@ function getSearchResults(doc, checkOnly) {
...
+	}
+	
+	// If no creators yet, try common DOM byline selectors (skip org-like)
+	if (item.creators.length === 0) {
+		let cand = meta(doc, 'article:author')
+			|| text(doc, '.meta-author-author')
+			|| text(doc, '.meta-author')
+			|| text(doc, '.author-name')
+			|| text(doc, '.byline a')
+			|| text(doc, '.meta-el.meta-author a');
```
AbeJellinek, line 281: Do these actually all occur on the site?

AbeJellinek (changes requested):

Prime 9ja Online.js
vanderwaalforces: @AbeJellinek I just did, you can see it fails without those? What do you think?

Prime 9ja Online.js
vanderwaalforces: Yes, they all do.

Prime 9ja Online.js
vanderwaalforces: Also, this was added as a fallback, which looks like what works for us.

AbeJellinek: Right, but it's already explicitly set to that value above on line 199.

vanderwaalforces: I resolved this, and the above one too.

PR #3507, opened 2025-10-24
Add translator for ChatGPT

AbeJellinek:
All chats get basic info (title and website title). Shared chats (`/share/` links) get model info and a URL; private chats don't.

I'm not confident about the text I'm putting in `websiteType` - "Private AI conversation" or "Conversation with [model name]". @adam3smith, any thoughts on what's appropriate there?

### DIFF
--- a/ChatGPT.js
+++ b/ChatGPT.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2025-10-24 16:06:16"
+	"lastUpdated": "2025-10-24 21:25:43"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


AbeJellinek:
The model name isn't quite a version of ChatGPT itself, but rather the version of the model it's using. So you could be using ChatGPT with GPT-5, but I don't think it would be correct to say "ChatGPT 5." The model name could also be something that doesn't include "GPT," like "o3" or "o4-mini". Could we just put the model name in parentheses, like "ChatGPT (GPT-5)"?

ChatGPT.js
```diff
@@ -0,0 +1,158 @@
...
+		headers.Authorization = `Bearer ${accessToken}`;
+	}
+	
+	let json = await requestJSON(apiURL, { headers });
+
+	item.title = json.title;
+	item.date = ZU.strToISO(new Date((json.update_time || json.create_time) * 1000).toISOString());
+	if (json.model) {
+		item.creators[0].lastName += ` ${json.model.title}`;
+	}
```
adam3smith, line 122: I liked your idea to add the model as a parenthetical to the websiteTitle better than as an author
```suggestion
	if (json.model) {
		item.websiteTitle += ` (${json.model.title})`;
	}
```

AbeJellinek:
(We can omit the parentheses when the name of the chat software is the same as the name of the model, like in the Claude and Gemini examples from APA. But that's not a very important detail here.)

PR #3510, opened 2025-10-25
Create The Nation (Nigeria) and Leadership Newspaper translators

vanderwaalforces:
Add translators for The Nation (Nigeria) and Leadership Newspaper


vanderwaalforces:
Just like #3513, the same thing happens with this; tests usually time out for some reason.

PR #3537, opened 2025-11-28
BibTeX: round-trip citation keys.

kousu:
Imported .bib keys were being replaced by citeKeyFormat = "%a_%t_%y"; Preserve them so users can sync pre-existing LaTeX/Typst projects with Zotero.

### DIFF
--- a/BibTeX.js
+++ b/BibTeX.js
@@ -18,7 +18,7 @@
 	},
 	"inRepository": true,
 	"translatorType": 3,
-	"lastUpdated": "2024-03-25 14:51:02"
+	"lastUpdated": "2025-11-28 19:44:26"
 }
 
 /*
@@ -976,7 +976,11 @@ function beginRecord(type, closeChar) {
 					}
 					delete item.backupLocation;
 				}
-				
+				if (item.itemID) {
+					item._extraFields.push({ field: "citation key", value: item.itemID });
+					//item.citationKey = item.itemID;
+				}
+
 				if (!item.date) {
 					item.date = dateFieldsToDate(item.year, item.month, item.day);
 				}
@@ -3133,6 +3137,7 @@ var testCases = [
 					}
 				],
 				"date": "2001",
+				"extra": "citation key: Adams2001",
 				"itemID": "Adams2001",
 				"pages": "493–518",
 				"publicationTitle": "Bulletin of Volcanology",
@@ -3175,6 +3180,7 @@ var testCases = [
 				],
 				"date": "1964",
 				"edition": "ninth Dover printing, tenth GPO printing",
+				"extra": "citation key: abramowitz+stegun",
 				"itemID": "abramowitz+stegun",
 				"place": "New York",
 				"publisher": "Dover",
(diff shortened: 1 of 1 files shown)


kousu:
If anyone wants some context: I'm motivated by trying to migrate my existing project into Typst. I've [set up syncing](https://typst.app/docs/web-app/reference-sync/) and replaced my bibliography.bib with the synced zotero.bib, but all my references are broken! [The forums](https://forum.typst.app/t/how-to-prevent-duplicate-citekeys-when-syncing-bibfile-from-zotero/3141/2) turned me onto the `extra: citation key` feature; I don't want to fix everything manually, I wish the seam was invisible.

By the way, Scaffold auto-formatted everything and it looks like the formatter hasn't been run on this file in ages; other than my patch here:

```
$ git diff --stat
 BibTeX.js | 3422 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++---------------------------------------------------------------------------------------
 1 file changed, 1715 insertions(+), 1707 deletions(-)
```

I didn't commit them because I thought that would have been toooo much.  But maybe that would be something to do in a follow-up? Except it [made the linter fail](https://github.com/zotero/translators/actions/runs/19772874166/job/56660321676?pr=3537#step:12:14) :/ ?

Except for that [Scaffold](https://www.zotero.org/support/dev/translators/testing) was SUPER easy to use. I spent a while grepping to get oriented but once I knew to look for the word "translator" I got this done in a snap. I thought it would take me weeks of picking at it and I would never get it done, but there's a full IDE built in for working on these. I'm really impressed.

adam3smith:
The point is that you can just install the BBT add-on in Zotero and it will import with citekey because it's intended exactly for your use case (the plugin is mentioned in the thread you link to -- I don't quite understand the limitation they mention. Zotero's API BibTeX _will_ pick up citekeys stored in `Extra`. 
The built-in BibTeX translator has multiple use cases and having that to default to the citekey in the BibTeX could be quite confusing.

kousu:
I didn't realize BBT will generate "extra: Citation Key:" entries.  That's not mentioned in the thread or the BBT site. It's a twisty series of steps for something pretty basic.

I tried to run the tests locally to avoid the CI timeout; I got as far as having selenium pop open but I saw a bunch of content-security errors so I wasn't actually able to test. And I tried `act` just in case it was an env problem and that ran through to completion but in that case I suspect it didn't actually run the test cases.  Tips welcome, I'm sort of stumped. I can run tests in Scaffold fine but I can't get the integration tests working.

Anyway, I have a good hope that this is not really disruptive and only improves things going forward. I'm looking forward to seeing Abe's take.

---

Another approach [exposes `itemID` as the citekey](https://github.com/kousu/zotero-translators/commit/8a33b3fe56f123f8bbe816308faa3dc9b945dcf2)? It already seems to be storing the original citekey there. But this would change citekeys from pre-existing databases, so maybe not.

BibTeX.js
```diff
@@ -976,7 +976,11 @@ function beginRecord(type, closeChar) {
 					}
 					delete item.backupLocation;
 				}
-				
+				if (item.itemID) {
+					item._extraFields.push({ field: "citation key", value: item.itemID });
+					//item.citationKey = item.itemID;
```
kousu, line 981: I don't know if this is a bug that maybe someone wants to look into but `.citationKey` doesn't seem to get saved to the database, or even exported to the output in the tests, which is odd because `doExport()` references it.

PR #3571, opened 2026-01-07
ACM: update PDF download logic for the new design

### LINKED ISSUE #3570: ACM Digital Library: Always grab OA PDFs
https://forums.zotero.org/discussion/128850/acm-digital-library-fail-to-grab-pdf-without-subscription

alex-ter:
Fixes #3570.


alex-ter:
The test failure seems to have something to do with general infra, not the change at hand. BTW, they're now using Cloudflare, so e.g., running tests from the Scaffold doesn't work, so I tested it manually.

PR #3588, opened 2026-02-22
Annual Reviews: Update target URL patterns and improve BibTeX handling

mjthoraval:
https://forums.zotero.org/discussion/115080/cannot-import-pdf-file-from-annual-reviews-journal

### DIFF
--- a/Annual Reviews.js
+++ b/Annual Reviews.js
@@ -2,18 +2,19 @@
 	"translatorID": "5f22bd25-5b70-11e1-bb1d-c4f24aa18c1e",
 	"label": "Annual Reviews",
 	"creator": "Aurimas Vinckevicius and Abe Jellinek",
-	"target": "^https?://[^/]*annualreviews\\.org(:[\\d]+)?(?=/)[^?]*(/(toc|journal|doi)/|showMost(Read|Cited)Articles|doSearch)",
+	"target": "^https?://[^/]*annualreviews\\.org/(toc|journal|doi|content/journals|search|showMost(Read|Cited)Articles|doSearch)",
 	"minVersion": "3.0",
 	"maxVersion": "",
 	"priority": 150,
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2022-08-02 12:48:22"
+	"lastUpdated": "2026-02-22 10:54:38"
 }
 
 /**
 	Copyright (c) 2012-2021 Aurimas Vinckevicius and Abe Jellinek
+	Modified 2026
 
 	This program is free software: you can redistribute it and/or
 	modify it under the terms of the GNU Affero General Public License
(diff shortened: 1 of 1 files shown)


Annual Reviews.js
```diff
@@ -30,121 +31,229 @@
...
+	for (let link of links) {
+		var href = link.getAttribute('href');
+		var doiMatch = href.match(/10\.\d+\/(?:annurev[.-])[^/]+/);
+		if (!doiMatch) continue;
+		
+		var doi = doiMatch[0];
+		var title = ZU.trimInternal(link.textContent);
+		
+		if (!title || title.length < 3) continue;
+		if (title.toLowerCase().includes('more') || title.toLowerCase().includes('less')) continue;
```
AbeJellinek: I personally think we shouldn't write off good scientific research just because it has the word "more" or "less" in its title :)

Is this intended to handle a specific case?

mjthoraval: Thanks for the review.
I have tested again on the test URLs, and I cannot find any problem without this line.
So it should probably be removed indeed.
It is probably some leftovers from the iterations I did where I still had some wrong items.

mjthoraval:
Thanks for the review! The flagged line is removed — you were right, it was a debugging leftover and everything works without it.

I've also pushed a substantial update after systematically testing every page type against the live site:

- **Topic collection pages** (`/page/...`, e.g. [Top 100 in 2025](https://www.annualreviews.org/page/top-2025)) are now supported. Some articles on these pages are only linked through the publisher's `arevie.ws` URL shortener with no DOI in the DOM, so `doWeb` (now async) resolves those links and reads the DOI/title from the target page's metadata. On one collection this recovers 8 articles that were otherwise invisible.
- **Legacy URL forms removed**: `/toc/`, `/journal/`, and `/doi/` now all 301-redirect to `/content/journals/` equivalents, and the `showMost*Articles`/`doSearch` action pages are gone (redirect/404), so the target is down to `content/journals|search|page`. Test cases updated to the canonical URLs, with fasttrack and topic-collection cases added (13 total).

**About the failing CI**: I reproduced the harness locally — every failure is Cloudflare blocking the automated browser (`403` on `/cite/bibtex`, challenge pages on listings), not a translator issue. From a real logged-in browser session, all test cases save correctly with full metadata and PDFs (verified field-by-field against the expected items). So I don't think this PR can go green in CI as-is; happy to adapt if there's a preferred way to handle Cloudflare-blocked sites in the test suite.

mjthoraval:
https://forums.zotero.org/discussion/132648/zotero-cant-save-annual-review-of-physical-chemistry

mjthoraval:
Pushed one change since the last review: short-link resolution now uses `requestDocument()` instead of the deprecated `ZU.processDocuments()` (4f0f93e). Behaviour is unchanged — the resolved page's metadata still supplies the DOI and title, an unresolvable link still leaves the entry without a DOI so it gets dropped, and such a failure still doesn't abort the rest of the translation.

The check is still red, and I can now pin down why: **annualreviews.org serves a challenge to headless browsers**, so this check can't go green for this site regardless of the translator's correctness.

Running the repo's own harness on the current commit and changing nothing but the browser mode:

| run | result |
|---|---|
| `node .bin/run-tests.mjs "Annual Reviews.js" --no-dependents --headed` | **13/13 succeeded** |
| `node .bin/run-tests.mjs "Annual Reviews.js" --no-dependents` (headless) | 13/13 failed |

The headless run reproduces [run 31026808181](https://github.com/zotero/translators/actions/runs/31026808181/job/92377259745) exactly — 9 × `Detection failed`, 4 × `Error`, including the same 403s:

```
HTTP GET https://www.annualreviews.org/content/journals/10.1146/annurev-linguistics-081720-111352/cite/bibtex failed with status code 403
```

Same machine, same IP, same test cases — only `--headed` differs. The `Detection failed` results are that same block one step earlier: the article page itself returns a challenge, so `detectWeb()` never sees an article.

Verification on the current commit:

- `npm run lint -- "Annual Reviews.js"` → clean (CI's lint step passes too)
- `node .bin/run-tests.mjs "Annual Reviews.js" --headed` → 13/13
- all 13 cases also run through Zotero's own translation stack → 13/13
- for the API change specifically, I compared before/after on a collection page that uses short links: same 20 picker entries, same 20 resolved DOIs, same order
- every URL in the test block matches `target`, so detection isn't failing on the pattern

Notes for merging:

- master's `Annual Reviews.js` is unchanged since 2022-08-04, so this still applies cleanly. The shipped `target` is still the old `/(toc|journal|doi)/` form, which no longer matches the current site layout.
- 6d4490de (BibTeX: don't save `_eprint`/`_eprinttype` to Extra) doesn't interact with this PR: Annual Reviews' BibTeX payloads carry no `eprint`/`eprinttype` fields, and no test case here asserts on `extra`.

Is there a preferred convention for test cases on sites that block headless browsers? Happy to adjust anything you'd like.

PR #3592, opened 2026-03-09
Fix: correctly source expandedTitle (the full title of the journal) in ACM Digital Library.

Bluesbreaker45:
# Problem

When saving a paper from ACM Digital Library, the snippet to get the full title of the journal wrongly grabs `span.epub-section__title`, which points to the artifact of the paper:

```html
<span class="epub-section__title">{The title of the paper} (Artifact)</span>
```

which results in wrong journal field `{The title of the paper} (Artifact)`.

To reproduce the problem, one can try to save the example in [this link](https://dl.acm.org/doi/10.1145/3763081).

# Fix

The bug is caused by the wrong specification of the class. The correct class is `span.serial-title`:

```html
<span class="serial-title">Proceedings of the ACM on Programming Languages</span>
```

### DIFF
--- a/ACM Digital Library.js
+++ b/ACM Digital Library.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2026-01-07 17:48:30"
+	"lastUpdated": "2026-03-09 12:52:09"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


Bluesbreaker45:
Hi @AbeJellinek, just a gentle ping to see if you’ve had a chance to look at this PR. I’m happy to make any necessary adjustments based on your feedback. Thanks!

AbeJellinek:
`.serial-title` means an element with that class name (`serial-title`). No such element exists on the page as far as I can tell.

`attr(doc, 'meta[name="citation_journal_title"]', 'content')` would work.

Bluesbreaker45:
Thanks for the clarification --- I misunderstood the `.serial-title` selector notation in your previous comment.

I’ve updated the PR to use `attr(doc, 'meta[name="citation_journal_title"]', 'content')` as suggested. Please let me know if anything else needs to be changed. Thanks!

PR #3595, opened 2026-03-15
Refactor regex patterns to use native RegExp

tobiasdiez:
Insead of relying on the somewhat (out)dated XRegex package, native regex are now used. The unicode properties feature that is used in those regex is supported since about 5  years in all browsers.


AbeJellinek:
Need to bump `lastUpdated`

tobiasdiez:
> Need to bump `lastUpdated`

Thanks, done now!

tobiasdiez:
@AbeJellinek does this look good to you now?

AbeJellinek:
Thanks!

tobiasdiez:
@AbeJellinek Thanks for the review!

PR #3604, opened 2026-04-13
New translator for State Library of Victoria

wragge:
This new translator works in both the SLV's main catalogue and its digital item viewer. In the main catalogue it's now getting some metadata that wasn't being captured by the generic 'Primo 2018' translator. In the digital file viewer, the translator captures both metadata and digital files (PDFs and images). In both cases the translator is getting data from an internal JSON API.

### DIFF
--- a/State Library of Victoria.js
+++ b/State Library of Victoria.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2026-04-17 02:46:07"
+	"lastUpdated": "2026-04-28 10:15:53"
 }
 
 /*
@@ -95,33 +95,41 @@ async function doWeb(doc, url) {
 const FORMATS = {
 	article: "journalArticle",
 	audio: "audioRecording",
-	Books: "book",
-	Diaries: "manuscript",
+	book: "book",
+	diaries: "manuscript",
 	dissertation: "thesis",
 	ead: "manuscript",
-	Ephemera: "manuscript",
-	images: "artwork",
+	ephemera: "manuscript",
+	image: "artwork",
 	manuscript: "manuscript",
 	map: "map",
-	Maps: "map",
-	Papers: "manuscript",
-	Photographs: "artwork",
-	Periodicals: "book",
+	paper: "manuscript",
+	photograph: "artwork",
+	periodical: "book",
 	realia: "artwork",
 	video: "videoRecording"
 };
 
+// Check if the specified label is in the format keys
+function checkFormat(label) {
+	if (Object.keys(FORMATS).includes(label)) {
+		return true;
+	}
+	return false;
+}
+
 // Look for values in the specified format field that have mappings in the list above
 // If there's no matches, return the default format
 function getFormat(metadata, formatField, defaultFormat) {
-	let format;
-	if (Object.keys(FORMATS).includes(metadata[formatField][0])) {
-		format = FORMATS[metadata[formatField][0]];
+	let label = metadata[formatField][0].toLowerCase();
+	if (checkFormat(label)) {
+		return FORMATS[label];
 	}
-	else {
-		format = defaultFormat;
+	// try with trailing 's' removed
+	else if (checkFormat(label.replace(/s$/, ""))) {
+		return FORMATS[label.replace(/s$/, "")];
 	}
-	return format;
+	return defaultFormat;
 }
 
 // MAIN CATALOGUE
--- a/State Library of Victoria.js
+++ b/State Library of Victoria.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2026-04-13 04:09:15"
+	"lastUpdated": "2026-04-17 02:46:07"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


State Library of Victoria.js
```diff
@@ -0,0 +1,738 @@
...
+	Periodicals: "book",
+	realia: "artwork",
+	video: "videoRecording"
+};
+
+// Look for values in the specified format field that have mappings in the list above
+// If there's no matches, return the default format
+function getFormat(metadata, formatField, defaultFormat) {
+	let format;
+	if (Object.keys(FORMATS).includes(metadata[formatField][0])) {
```
AbeJellinek: Can we `toLowerCase()` this and use lowercase keys in the map? Maybe also strip the plural suffix? The varying capitalization/random mix of taxonomies doesn't make me too confident that this data is clean. (If this is coming from some set list of formats somewhere, then it's fine!)

State Library of Victoria.js
```diff
@@ -0,0 +1,738 @@
...
+
+/** BEGIN TEST CASES **/
+var testCases = [
+	{
+		"type": "web",
+		"url": "https://viewer.slv.vic.gov.au/?entity=IE1542628&mode=browse",
+		"items": [
+			{
+				"itemType": "artwork",
+				"title": "Penguin [picture].",
```
AbeJellinek, line 448: Probably want to strip this format info from titles (and the call number info in the next item?)

State Library of Victoria.js
```diff
@@ -0,0 +1,738 @@
...
+				],
+				"notes": [],
+				"seeAlso": []
+			}
+		]
+	},
+	{
+		"type": "web",
+		"url": "https://viewer.slv.vic.gov.au/?entity=IE13142663&file=FL19626988&mode=browse",
+		"detectedItemType": "artwork",
```
AbeJellinek, line 528: Any way we can detect correctly here? Not a big deal if not.

State Library of Victoria.js
```diff
@@ -0,0 +1,738 @@
...
+		item.callNumber = jsonData.delivery.bestlocation.callNumber || null;
+	}
+	else {
+		item.callNumber = identifier;
+	}
+	// Add tags
+	item.tags = getMetadataValues(metadata, "subject");
+	// Construct persistent(ish) url
+	let almaId = jsonUrl.match(/alma\d+/)[0];
+	item.url = `https://find.slv.vic.gov.au/discovery/fulldisplay?vid=61SLV_INST:SLV&docid=${almaId}`;
```
AbeJellinek, line 280: Ideally we only want to add `url` if the item is actually available online at that location.

AbeJellinek (changes requested):
Thank you! Some suggestions

State Library of Victoria.js
wragge: Good idea! I've cleaned up the format list and now check for lowercase and depluralised forms. The reason why there appears to be some inconsistency is that the main catalogue and the digitised item viewer use different sets of values.

State Library of Victoria.js
wragge: I'm reluctant to do this the square bracket stuff often includes important additional info for original and pictorial collections, and it's not always going to match the format anyway. In this case the format value is "Photographs" not "Picture".

In the second case, the call number value is actually added to the title by the translator:

```javascript
let [attachment, label] = prepareAttachment(digitalData, url);
// Some images are grouped under a single title,
// so add the image label to the title for clarity where necessary
if (label) {
	item.title = `${item.title} (${label})`;
}
```
As the notes indicate, this is to try and distinguish between items grouped under the one title. The `label` value is actually drawn from the `label` field in the METS metadata that distinguishes between the different items. In most cases I've seen this is a more human friendly label like 'Image 1' or 'Page 1' rather than an accession or call number. And this label value is not usually elsewhere in the metadata. So basically this gives the user an important way of telling grouped items apart.

State Library of Victoria.js
wragge: No, even if you waited for the page content to settle (which can take a while) there's still no format values displayed, even though they're in the metadata from the JSON API.

State Library of Victoria.js
wragge: These urls are reliable. The reason I don't use the urls from the browser is that the system urls are stupidly long and full of parameters that don't really do anything. There is a system generated 'permalink', but that requires an additional API call to retrieve, and while it's shorter, it's still not very 'perma', and the only thing it does is to encode info about the current view, which you don't need if all you want is the default view. I experimented with all the different parameters until I found a version that was short, reliable, and could be easily constructed from the system identifier. I even wrote [a blog post](https://updates.timsherratt.org/2025/09/23/exploring-slv-urls.html) about this if you want to read it! :-)

AbeJellinek: What I mean is that we don't put URLs in citations (and hence in the item's `url` field) unless there's full-text content at that URL. So a URL for an image that's available online, yes, but a URL for a book in the catalog that needs to be checked out physically, no.

State Library of Victoria.js
AbeJellinek: There is always some tension between creating good items for non-citation use cases and creating good citations, but putting a call number in the title is just too much IMO. Fine to add the `label` value, just not the call number. If you have a lot of items with the same title and no labels, you can enable the Call Number column in Zotero.

wragge: Ok, I've added a check so the label is only added if it doesn't include the call number.

State Library of Victoria.js
wragge: Really? I've always thought it was important for translators to maintain the link between the metadata and its source. In any case, this catalogue includes a lot of undigitised, original material (manuscripts, photos etc), with additional metadata such as finding aids and hierarchical relationships. It's really quite crucial to have the url to the original record.

AbeJellinek: Can add a web link attachment (`{ url, mimeType: 'text/html', snapshot: false }`) to maintain a link even if it's not appropriate to put in a citation.

AbeJellinek: But no, we don't want to fill in the `url` field for undigitized resources, because that would produce misleading citations.

PR #3605, opened 2026-04-14
Rewrite The Boston Globe for modern bostonglobe.com

notUserX:
The existing translator only targeted archive.boston.com and was intentionally disabled for the modern site (77683d27). This rewrite supports bostonglobe.com (Arc Publishing platform) by delegating to the Embedded Metadata translator and post-processing with LD+JSON structured data (schema.org/NewsArticle).

Changes:
- Updated target regex to match bostonglobe.com
- Rewrote scrape() to extract author, date, section, and URL from LD+JSON
- Handles editorial board as single-field creator
- Strips 'Opinion |' prefix from headline (captured in section field instead)
- Falls back to meta[name=author] with cleanup for comment-count contamination
- Preserves translatorID for seamless update
- Three test cases (opinion, metro, feature)

Previously: translator was silently failing on bostonglobe.com, causing fallback to Embedded Metadata which produced garbled author fields (e.g. 'Comments67, The Editorial BoardView').

### DIFF
--- a/The Boston Globe.js
+++ b/The Boston Globe.js
@@ -1,7 +1,7 @@
 {
 	"translatorID": "1f245496-4c1b-406a-8641-d286b3888231",
 	"label": "The Boston Globe",
-	"creator": "Adam Crymble, Frank Bennett, Sebastian Karcher, Matthew Weymar",
+	"creator": "Matthew Weymar",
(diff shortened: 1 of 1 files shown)


AbeJellinek:
Thank you!

PR #3606, opened 2026-04-15
[PICA] Handling electronic theses on sudoc.fr

symac:
Hello,
this request adds a test that will handle electronic theses as thesis rather than fallback on the book type on a record like : https://www.sudoc.fr/282427384

This should be improved to handle records when the language is set to German or English (and Thèse is changed to _Dissertation_) but right now, my scaffold version is not working with my test page https://www.sudoc.fr/282427384 (Linux, Zotero 9 but I have tried to reinstall 8.0.4 and it does not fix the issue). I assume it might be related to some redirect on sudoc's website which think scaffold is some sort of robot or something like that. It was working some weeks ago when I developed this before forgetting to create the PR.

### DIFF
--- a/Library Catalog (PICA).js
+++ b/Library Catalog (PICA).js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2024-11-26 15:26:20"
+	"lastUpdated": "2026-03-27 21:31:39"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


Library Catalog (PICA).js
```diff
@@ -94,6 +94,10 @@ function detectWeb(doc, _url) {
 				else if (type.includes('map.')) {
 					return "map";
 				}
+				else if (type.includes('binary.') && (ZU.xpath(doc, '//tr/td[@class="rec_lable" and .//span[starts-with(text(), "Thèse")]]').length)) {
```
AbeJellinek, line 97: Let's avoid adding more complicated XPaths, especially since XPath `class` matching breaks if the element has multiple class names:

```suggestion
				else if (type.includes('binary.') && text(doc, 'td > td.rec_lable').includes('Thèse')) {
```

should do it, I think.

AbeJellinek (changes requested):

symac:
Hello @AbeJellinek ,
thanks for your suggestion, the issue is that (with a tr at the begining for selector rather than the td in your comment which I assume was a typo) :
```text(doc, 'tr > td.rec_lable')```

Returns the first label of the record ( Identifiant pérenne de la notice :  ) when the label "Thèse : " that is of interest in this case is way down in the record. 

I have not found any option with text() to match property this line, do you think of a solution? 

Thanks,
Sylvain

PR #3610, opened 2026-05-17
fix(arXiv): omit empty category brackets from Extra field

wali-reheman:
When the arXiv API doesn't return a primary_category (articleField is empty), the Extra field was being set to 'arXiv:XXXX []' — the empty brackets were still included. The fix adds a conditional so the space and category brackets are only added when articleField has an actual value.

### DIFF
--- a/arXiv.org.js
+++ b/arXiv.org.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 12,
 	"browserSupport": "gcsibv",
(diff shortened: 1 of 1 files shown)


AbeJellinek:
Wait, but I don't think the situation this was supposed to fix (empty brackets appended) was actually possible. If `articleField` was empty, the `if (articleField) articleField = "[" + articleField + "]"` condition above wouldn't pass, so `articleField` would remain empty and it would just append a space (stripped on `complete()`).

Where did you see the issue?

PR #3612, opened 2026-05-19
InvenioRDM: Add CDS to target

sakshamarora1:
- the new CERN Document Server is InvenioRDM based (repository.cern)
- the Zenodo Sandbox environment should not be used to cite articles on wikipedia (if there is another use case for still including it, happy to add it back) as it is used only for testing purposes


AbeJellinek:
Thanks!

PR #3619, opened 2026-06-05
Remove format=json from Anne Frank API requests

VDK:
I figured out what was going wrong with the API requests: for individual subjects, locations, events, and people, the data can be retrieved directly from the API by using the identifier in the URL. Passing that identifier as a search query triggers a redirect, which appears to be blocked in some Zotero configurations, notably Citoid. I’ve submitted this follow-up patch to request the direct API URLs instead.

### DIFF
--- a/Anne Frank Knowledge Base.js
+++ b/Anne Frank Knowledge Base.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2026-06-01 15:40:31"
+	"lastUpdated": "2026-06-05 12:00:00"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


dstillman:
Avoiding an unnecessary redirect is fine and good, but if Citoid can't handle redirects, that's a bug and you should report it to them. Translators shouldn't need to be rewritten to work around basic HTTP behavior being broken.

AbeJellinek:
I immediately get blocked locally, so I can't test this. CI appears to be blocked as well.

If you're involved with/have contacts at the organization, could you see about easing the blocking threshold a bit, or using a more standard bot detection system? It seems to apply a domain-wide IP ban after a few API requests, which sort of defeats the point of having an API. The ban is at the network level, so there isn't even a user-friendly error message - it just drops the connection. And it affects annefrank.org as a whole! If I tried to save some items while on mobile data, a bunch of people around me could be blocked from reserving museum tickets.

And I agree that Citoid needs to fix the underlying issue on their end! As long as tests were passing for you, though, this seems fine for now.

VDK:
I've let them know

VDK:
[Citoid devs](https://phabricator.wikimedia.org/T428915) tell me they don't have a schedule for importing translators. They have done so because of my bug report, an have since imported my first version. There's no change in behavior on the Wikipedia side of things, so it's still down to the Anne Frank House to improve their infrastructure.

PR #3620, opened 2026-06-05
YouTube: Extract collaborators as authors

sebnow:
Saving a YouTube video with multiple channel collaborators to Zotero currently captures zero creators. The translator picks the channel name from one of three CSS selectors, and on a collaborator video YouTube renders the owner area as a single composite anchor reading e.g. `CreatorA and ChannelB` with `href="javascript:void(0);"` and no per-channel link. None of the selectors match that layout, so `author` ends up empty and nothing gets pushed to the item.

To recover the individual names, the creator block now reads `ytInitialData`, the JSON blob YouTube embeds in the initial HTML to bootstrap its own client. Its `videoOwnerRenderer` node lists collaborators explicitly when present, so we no longer depend on the rendered owner area for this case. Single-owner videos still fall through to the existing selector chain, since those keep working and the chain handles layouts I haven't surveyed.

YouTube preserves the original page's inline `ytInitialData` script across SPA navigations, so on a same-tab transition the JSON describes whichever video was loaded first. The workaround detects the mismatch by checking that the current `videoId` appears in the parsed data, and refetches the canonical watch URL via `requestDocument()` when it doesn't.

A new test case covers `73Do0OScoOU` (two collaborators); the existing `pq94aBrc0pY` case continues to pass with `Zoteron` as the sole author, so the single-channel path is unaffected.

### DIFF
--- a/YouTube.js
+++ b/YouTube.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2026-06-04 21:14:01"
+	"lastUpdated": "2026-06-05 16:06:50"
 }
 
 /*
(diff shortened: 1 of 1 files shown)


AbeJellinek:
Thanks!

Issue #3621, opened 2026-06-16
NeurIPS translator does not capture abstracts anymore


CarlaSa:
The Abstract is not captured for NeurIPS Links.

# Steps to reproduce
1. Open this link: [Example Paper](https://papers.nips.cc/paper_files/paper/2022/hash/a98846e9d9cc01cfb87eb694d946ce6b-Abstract-Conference.html)
2. Use Zotero connector in Firefox
3. No abstract is saved

# Technical details
In the scrape function the abstract is gotten in this (quite hacky) way:
[Link to lines](https://github.com/zotero/translators/blob/e75590cb70292499673003a740101efcb657c37b/Neural Information Processing Systems.js#L50-L51)
```
function scrape(doc, url) {
	// Unfortunately, the abstract isn't semantically marked at all; this is the best we can do
	let abstract = ZU.xpathText(doc, '//h4[text()="Abstract"]/following-sibling::p[2]');
       ...
```

But NeurIPS changed its website since then and the abstract is not longer found this way. For referece: I looked up the source code from the above mentioned link (ca 2022) on archive.org and it used to be 

```
        <h4>Abstract</h4>
        <p><p>We argue that the theory and practice of diffusion-based generative models ....
```

But now the source code looks like this 
```
        <section class="paper-section">
            <h2 class="section-label">Abstract</h2>
            <p class="paper-abstract"><p>We argue that the theory and practice of diffusion-based generative models  ....
        </section>
```

So the parsing of the abstract should be changed to reflect this new structure.

CarlaSa:
I created #3622 that fixes the issue for me and runs on all test cases.

PR #3622, opened 2026-06-16
Updating NeurIPS file

CarlaSa:
Fix for git issue #3621, so layout crawling the abstract works now.

During this I realized that several valid URLs were not treated correctly ( often for newer papers `...-Abstract-Conference.html` and `...-Abstract.html` are both valid files). So I updated the script to for example handle redirection. 

Finally I added a newer paper (with the `...-Conference.html` notation) which did not parse before in the test cases. I added it in two variations, so the test cases have a better cover over diffenrent URL variations.

### DIFF
--- a/Neural Information Processing Systems.js
+++ b/Neural Information Processing Systems.js
@@ -7,9 +7,9 @@
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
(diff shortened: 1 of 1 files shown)


AbeJellinek:
Thank you!

PR #3623, opened 2026-06-17
Add translator for lovdata.no and delete for brukerhåndboken

sstraume97:
New translator for lovdata.no created and added.

The site Brukerhåndboken has been discontinued and therefore recommends that the translator be deleted.


AbeJellinek:
Need to update deleted.txt

PR #3624, opened 2026-06-22
Fix IOPscience PDF URLs

BadSoyo:
## Summary
- Update the Institute of Physics translator to use the PDF URL advertised by IOPscience article metadata when available
- Keep the existing /pdf URL fallback for article pages without citation_pdf_url
- Add a test case for a current IOPscience article: https://iopscience.iop.org/article/10.1088/1361-665X/ad6812

## Testing
- npm run lint -- 'Institute of Physics.js'

### DIFF
--- a/Institute of Physics.js
+++ b/Institute of Physics.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsbv",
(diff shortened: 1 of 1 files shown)


BadSoyo:
Closing this PR after realizing that the existing translator already constructs the correct /pdf attachment URL for this page. The change is not necessary as submitted.

PR #3627, opened 2026-06-23
PNX: Change xPath to identify Harvard organization

cbaksik:


cbaksik:
I can't figure out why Test 2 reports 'Data mismatch.' I didn't change anything related to what is put into the item object.

AbeJellinek:
It's showing that the translator no longer saves a "HOLLIS Permalink" link attachment on that item. You changed the XPath selector for the data that creates that attachment, so presumably the new selector doesn't work on that test's input, and we need to use the old selector as a fallback.

AbeJellinek:
Right now it's not clear what this fixes, since there are no new test cases. You can add one using Scaffold (Tools → Translator Editor) - paste the XML under Test Input and create a test. When you save, that will also bump `lastUpdated`, which is necessary to get these changes pushed out.

cbaksik:
Ok I think I did it right this time!

AbeJellinek:
Thank you!

PR #3628, opened 2026-07-03
Export the Zotero item key as the RIS ID tag.

michi-zuri:
When exporting RIS files, it may be desired to preserve the Zotero id as the ID. What other programs do with this ID is up to them. I needed this functionality and this is what worked for me.

### DIFF
--- a/RIS.js
+++ b/RIS.js
@@ -17,7 +17,7 @@
 	},
 	"inRepository": true,
 	"translatorType": 3,
-	"lastUpdated": "2026-01-05 18:52:35"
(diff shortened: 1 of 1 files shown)


AbeJellinek:
I don't have any problem with this. @dstillman, seem reasonable?

michi-zuri:
Any news on this? I depend on this "hack" for being able to check if Zotero has a pdf of the article stored. Without the Zotero id this is not possible. The workaround for now would be exporting and importing through csv, which includes the id already.

PR #3629, opened 2026-07-04
SVT Nyheter: Update class names

Pazzaz:
Fixes "SVT Nyheter" by updating class names. One test changed as the page now says that the author is [TT](https://en.wikipedia.org/wiki/TT_News_Agency).


AbeJellinek:
Thanks so much!

PR #3630, opened 2026-07-15
Add PBL Zotero web translator

EwoutH:
Add `PBL.js`: a Zotero web translator for [pbl.nl](https://www.pbl.nl/). Detects individual publication (report) pages and listings (multiple). Extracts title (and subtitle), date, pages, language, report number/type, authors, abstract, topic tag, attaches full-text PDF and snapshot, and sets institution and url. Includes parsing of the site's "Kenmerken" specifications and a test case for a sample publication. Adds translator metadata (target regex, priority, license block, lastUpdated).

- [x] Test case passes
- [x] No errors or warnings when running `npm run lint -- "PBL.js"`

Might be improved in the future with a translator for their English website also: https://www.pbl.nl/en/publications

CC @KevinWittenberg @StefanTroost (I don't know if you guys use Zotero at PBL, but useful anyways for Zotero users trying to cite PBL).

### DIFF
--- a/PBL.js
+++ b/PBL.js
@@ -1,0 +1,225 @@
+{
+	"translatorID": "6debd0c0-902e-44b8-828e-a8f72af124eb",
+	"label": "PBL",
+	"creator": "Ewout ter Hoeven",
+	"target": "^https?://(www\\.)?pbl\\.nl/(publicaties|actueel|zoeken)",
+	"minVersion": "5.0",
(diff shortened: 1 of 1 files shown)


AbeJellinek:
Can we add a test for multiples? And yes, it would be nice to support the `/en/` site if the structure isn't too different. Otherwise looks good.

PR #3631, opened 2026-07-15
Improve KiM translator metadata parsing

EwoutH:
Update the KiM Netherlands Institute for Transport Policy Analysis translator:

- Prefer structured data when available: parse #elastic-content JSON and select the WebPage node from JSON-LD for title, date and language (falling back to page heading)
- Normalize dates with ZU.strToISO, derive language from webPage.inLanguage or doc.lang, and extract authors from elastic.author (comma-separated list).
- Simplify PDF attachment handling and add robust error debug logs for JSON parsing.
- Add English and Dutch test cases and bump lastUpdated.

- [x] Test case passes
- [x] No errors or warnings when running `npm run lint -- "PBL.js"`

### DIFF
--- a/KiM.js
+++ b/KiM.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2026-01-02 16:28:12"
+	"lastUpdated": "2026-07-15 13:55:27"
 }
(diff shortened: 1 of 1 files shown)


AbeJellinek:
Thanks! We should probably reduce the number of PDFs we save (by default) on this site, but looks good for now.

PR #3632, opened 2026-07-16
Add OpenReview translator

jsrdcht:
## Summary

- Add a web translator for OpenReview forum pages
- Import the embedded BibTeX metadata using Zotero's BibTeX translator
- Fall back to citation metadata when BibTeX is unavailable
- Attach the full-text PDF and normalize the canonical URL
- Add test cases for two TMLR articles

## Testing

- `npm run lint -- OpenReview.js`
- Integration testing was attempted, but the local browser environment could not load the Zotero Connector test extension.

### DIFF
--- a/OpenReview.js
+++ b/OpenReview.js
@@ -1,0 +1,266 @@
+{
+	"translatorID": "b1988f2e-e080-42be-8464-38a327100952",
+	"label": "OpenReview",
+	"creator": "Tuo Chen",
+	"target": "^https?://openreview\\.net/forum\\?id=",
+	"minVersion": "5.0",
+	"maxVersion": "",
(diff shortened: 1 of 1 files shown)


jsrdcht:
Hi @AbeJellinek, this translator works in local testing against the two OpenReview paper pages. The CI failures may be caused by OpenReview's Cloudflare challenge: both tests fail during detection, before translation runs.

The translator extracts embedded BibTeX or citation metadata from the fully loaded paper page, so detection cannot succeed if the CI browser receives a challenge page or an incompletely rendered page. Could you please advise how OpenReview translators should be tested in this situation, and review the PR when you have a chance? Thanks!

PR #3636, opened 2026-08-05
Cambridge Core: Don't prefix abstracts of articles with a graphical abstract

mjthoraval:
Abstracts of Cambridge Core articles that have a **graphical abstract** are saved with a stray `, ` in front:

> `, We study the effect of insoluble surfactants on the impact of...`

Such articles have **two** `div.abstract` nodes — the first holds only the figure and has no text. `ZU.xpathText()` joins all matched nodes with its default `", "` delimiter, so the empty first node contributed the separator. This affects a lot of Journal of Fluid Mechanics articles; 188 of the 1367 JFM items in my own library are affected.

The fix takes the first `div.abstract` that actually has text. It deliberately does not pass `delimiter: ''` to `xpathText()`, which would silently concatenate genuinely multi-node abstracts instead.

The text now also goes through `ZU.trimInternal()`. `xpathText()` does no trimming, so abstracts could pick up a trailing space as well — e.g. the abstract of [Micro-bubble morphologies following drop impacts onto a pool surface](https://www.cambridge.org/core/journals/journal-of-fluid-mechanics/article/microbubble-morphologies-following-drop-impacts-onto-a-pool-surface/ED77B4E99987ED9B17A3CAD627492504) is saved 869 characters long, one trailing space more than its text. All seven existing test cases with abstracts are already `trimInternal`-identical, so nothing there changes.

### Verification

The new test case (`10.1017/jfm.2023.224`) **fails on master's code and passes with this change** — I ran the harness on master's code with the new test case added to confirm the guard actually catches the bug. No other case changes state:

| run (local, subscribed network) | failing cases |
|---|---|
| master, as-is (12 cases) | 1, 2, 5, 6, 9, 10, 12 |
| master's code + new test case | 1, 2, 5, 6, 9, 10, 12, **13** |
| this branch (13 cases) | 1, 2, 5, 6, 9, 10, 12 |

Case 13 is the only one whose state changes. Cases 1, 2, 9 and 10 fail locally for an
unrelated, environment-dependent reason described below; on CI they pass, so this branch's
CI shows only 5, 6 and 12 — the same three as master.

### A note on four other cases

I initially also refreshed the `url` of cases 1, 2, 9 and 10, having seen them fail locally: Cambridge redirects `/article/abs/<slug>/<id>` to `/article/<slug>/<id>`, so the saved url didn't match. That was wrong and I've reverted it — the redirect only happens for readers **with access** to the article, so an institutional network gets the canonical form while CI gets the `/abs/` form the test cases already record. The existing URLs are correct; those four only fail on a subscribed network.

### Left alone: the three book/bookSection cases

Cases 5, 6 and 12 fail on master too, for a reason unrelated to this PR, and I'd rather flag it than record it as expected output.

Cambridge's RIS export now emits the DOI with the value itself prefixed:

```
DO  - DOI: 10.1017/9781108770750
```

The connector's bundled schema (version 28) doesn't allow `DOI` on `book` or `bookSection`, so the value is dropped instead of falling back to `Extra` — which is where these test cases expect it (`"extra": "DOI: 10.1017/9781108638210"`). A current desktop client keeps it as a proper `DOI` field (I checked: it imports as `10.1017/9781108770750` either way, so `cleanDOI` copes with the prefix), so saved items are unaffected.

So this looks like a test-environment artifact rather than a user-facing regression, and refreshing those fixtures would bake "no DOI" into the repo. Happy to include a fix if you'd prefer one here — stripping the prefix in this translator would be a one-liner — but it seemed better not to guess at how you'd want the schema pinning handled.

### DIFF
--- a/Cambridge Core.js
+++ b/Cambridge Core.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2024-11-20 15:43:05"
+	"lastUpdated": "2026-08-05 17:31:25"
 }
 
 /*
@@ -146,10 +146,15 @@ async function scrape(doc, url = doc.location.href) {
 		
 		translator.setHandler('itemDone', (_obj, item) => {
 			item.url = url;
-			var abstract = ZU.xpathText(doc,
-				'//div[@class="abstract"]');
-			if (abstract) {
-				item.abstractNote = abstract;
+			// Articles with a graphical abstract have TWO div.abstract nodes:
+			// the first holds only the figure and has no text. xpathText()
+			// joined both with its default ", " delimiter, so those abstracts
+			// began with a stray comma. Use the first node that has text.
+			for (let node of doc.querySelectorAll('div.abstract')) {
+				if (node.textContent.trim()) {
+					item.abstractNote = ZU.trimInternal(node.textContent);
+					break;
+				}
 			}
 			item.title = ZU.unescapeHTML(item.title);
 			item.publisher = ""; // don't grab the publisher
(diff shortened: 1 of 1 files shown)


mjthoraval:
Small correction on my own part, since the first CI run here looked worse than it should have: I'd also refreshed the `url` in cases 1, 2, 9 and 10, having watched them fail locally. That was wrong and is reverted in 94bd249.

`/article/abs/<slug>/<id>` only redirects to `/article/<slug>/<id>` for readers with access to the article, so the saved url differs by network — an institutional connection gets the canonical form, CI gets the `/abs/` form the fixtures already record. The existing URLs are correct, and my change had broken four cases that were passing here.

CI on the current head fails only cases 5, 6 and 12 — the book/bookSection DOI issue described in the description, which fail on master too. Case 13, the one this fix adds, passes.

PR #3639, opened 2026-08-13
Denik CZ: Update for redesigned site

frettie:
The 2018 translator no longer works after the denik.cz redesign:

- Authors: the `meta[property="author"]` tag is gone — now read from JSON-LD (`NewsArticle.author`), still excluding the generic "Redakce".
- Publication title: `og:site_name` is now always `www.denik.cz`, so the regional edition name (e.g. "Třebíčský deník") is taken from JSON-LD `NewsMediaOrganization`.
- Date, section, and tags are now filled from JSON-LD (`datePublished`, `articleSection`, `keywords`) — EM can't find them anymore.
- Listing/search selector `.right h2 a` no longer exists, replaced with `article h2 a`.
- Modernized to async `doWeb` + `requestDocument`, dropped deprecated `ZU.processDocuments`.
- Refreshed test cases: regional article, national article, section listing, and search (all passing).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Denik CZ.js
+++ b/Denik CZ.js
@@ -3,20 +3,20 @@
 	"label": "Denik CZ",
 	"creator": "Jiří Sedláček, Philipp Zumstein",
 	"target": "^https?://[^/]*denik\\.cz",
-	"minVersion": "3.0",
+	"minVersion": "5.0",
 	"maxVersion": "",
 	"priority": 100,
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2018-01-07 09:27:42"
+	"lastUpdated": "2026-08-13 17:04:12"
 }
 
 /*
 	***** BEGIN LICENSE BLOCK *****
 
-	Copyright © 2017 Jiří Sedláček, Philipp Zumstein
-	
+	Copyright © 2017-2026 Jiří Sedláček, Philipp Zumstein
+
 	This file is part of Zotero.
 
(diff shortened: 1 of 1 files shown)


frettie:
The CI failure here is infrastructure, not this translator — the browser tester never starts (5-minute timeout waiting for `#translator-tests-complete`, no per-test output). All four test cases pass locally.

Root cause: the CI clones a fresh zotero-connectors master on every run, and the Safari host-permissions work merged there around Aug 10 changed the signature to `checkIsOnline({active=false, permissionPromptShown=false}={}, tab=null)`. `tools/testTranslators/testTranslators.js` still calls `checkIsOnline(callback)` in callback style, so messaging strips the callback and sends `[undefined]`, which `runtime.sendMessage` serializes to `[null]` — and destructuring `null` throws:

```
TypeError: Cannot read properties of null (reading 'active')
    at Object.checkIsOnline (connector.js)
    at Object.receiveMessage (messaging.js)
```

The tester's init callback never fires, so the page just hangs. Every PR run since Aug 10 fails the same way (e.g. the Lääkärilehti PR), while pushes to master look green only because they skip the browser-test step.

Patching the built connector to tolerate the null (`options || {}` before destructuring) makes the tester work again and all tests here pass. The proper fix belongs in zotero-connectors — either call `checkIsOnline().then(...)` in testTranslators.js or destructure defensively in connector.js.

AbeJellinek:
Thanks!

PR #3641, opened 2026-08-24
Fix scopus search

adam3smith:
Fixes https://forums.zotero.org/discussion/133401/scopus-webpage-snapshot#latest

### DIFF
--- a/Scopus.js
+++ b/Scopus.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsibv",
-	"lastUpdated": "2025-07-29 16:11:13"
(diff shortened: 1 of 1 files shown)


adam3smith (commented):
I wasn't able to generate a stable search URL, so deleting the broken test without replacement

adam3smith:
The test failure is the automated tests, which would be expected given paywall (is there an option to mark them as don't run on CI?)

The fix is trivial, the old URL for searches does no longer seem to exist, everything else looks like it works

AbeJellinek:
> is there an option to mark them as don't run on CI?

Not currently, unfortunately.

Looks good, thank you!

PR #3642, opened 2026-08-25
unAPI: Fallback to default item type if link resolution fails

adomasven:
Preparing unAPI for changes in the Zotero Connector where detection in browser will no longer be allowed to perform HTTP requests. 

https://github.com/zotero/zotero-connectors/pull/632

### DIFF
--- a/unAPI.js
+++ b/unAPI.js
@@ -9,7 +9,7 @@
 	"inRepository": true,
 	"translatorType": 4,
 	"browserSupport": "gcsbv",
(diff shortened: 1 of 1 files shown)


adomasven:
Not sure what to do about the broken tests/CI, since the links no longer work.

AbeJellinek:
I'm fine to merge this now if it's ready, but we should find new test cases.

adomasven:
I've found some new tests and pushed. CI green now.

PR #3643, opened 2026-08-30
Add LibraryThing translator

janbaykara:
I've added a group importer from [LibraryThing](https://www.librarything.com/). Did this so I could pull https://www.librarything.com/catalog/HermesAmaraArchive

### DIFF
--- a/LibraryThing.js
+++ b/LibraryThing.js
@@ -1,0 +1,259 @@
+{
+	"translatorID": "5fee8b6c-8208-4be9-8d3e-dda9fe3c8671",
+	"label": "LibraryThing",
+	"creator": "Jan Baykara",
+	"target": "^https?://(www\\.)?librarything\\.com/(catalog\\.php|catalog_bottom\\.php|catalog/)",
+	"minVersion": "5.0",
+	"maxVersion": "",
+	"priority": 1,
+	"inRepository": true,
+	"translatorType": 4,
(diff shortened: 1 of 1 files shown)


LibraryThing.js
```diff
@@ -0,0 +1,259 @@
...
+		let match = new URL(url).pathname.match(/\/catalog\/[^/]+\/([^/]+)\/?$/i);
+		if (match && match[1] && !/\.php$/i.test(match[1])) {
+			return decodeURIComponent(match[1]);
+		}
+	}
+	catch (e) {}
+	return '';
+}
+
+function originFromUrl(url) {
```
AbeJellinek, line 77: This is always `https://www.librarything.com`, I think.

(Also the `request*()` methods accept relative URLs.)

LibraryThing.js
```diff
@@ -0,0 +1,259 @@
...
+	}
+}
+
+function isbnForBook(book) {
+	return String((book && (book.ISBN || book.ISBN_cleaned)) || '').trim();
+}
+
+function detectWeb(_doc, url) {
+	if (getUserId(url)) {
+		return 'multiple';
```
AbeJellinek, line 150: It looks like individual books have pages - why not support them?

AbeJellinek (commented):
Thanks! A couple comments. I can't see any books in the collection you linked, but this generally seems to work.

PR #3645, opened 2026-09-01
adding mime type to attachment for Harvard permalink

cbaksik:
Attachment for URL was not being added anymore in the Primo translator (Harvard-specific). I added mime/type of text/html so that it will be added again.


AbeJellinek:
Thanks! Fixed some errant spaces before the indentation.

Issue #3325, opened 2024-06-22
Support arXiv HTML pages


adam3smith:
We don't currently detect on arXiv's HTML view. It's debatable if we should import the PDF or the HTML from these, but we should definitely get the right metadata (long-run, we'd want general HTML/PDF/epub preferences configurable, as I discussed with dstillman somewhere on the Zotero forums.

Reported: https://forums.zotero.org/discussion/115441/zotero-translator-cannot-download-the-pdf-from-arxiv-experimental-html-page#latest
Example: https://arxiv.org/html/2406.14514v1

dstillman:
(Attachment type priority is https://github.com/zotero/zotero/pull/4146)

PR #3649, opened 2026-09-08
Support arXiv HTML article URLs

bencevans:
Recognise arXiv /html/ URLs and extract their paper IDs using the existing arXiv metadata lookup. Saving from an HTML article now produces a bibliographic item with a PDF attachment instead of falling back to a webpage.

Implements the association issue mentioned in #3325. The issue also discusses pulling the HTML version, which would be great, but this PR's scoped just to the association.


bencevans:
I'm not too sure why the tests are failing on GitHub, maybe getting blocked? The arXiv tests were passing locally.

gncs:
I am not sure if this related at all, but yesterday the Zotero Connector suddenly stopped working for me on arXiv (Firefox and Chrome).

PR #3650, opened 2026-09-11
Fix PubFactory Journal PDF download

adam3smith:
See https://forums.zotero.org/discussion/133384/only-first-2-pages-of-article-pdf-downloaded-by-zotero#latest


adam3smith:
Tests run for me locally except the https://avmajournals.avma.org/view/journals/javma/javma-overview.xml one which no longer detects.

PR #3651, opened 2026-09-19
Primo 2018: Fix titles in the item selection dialog

tkorner:
`getSearchResults()` reads the title from `.item-title`, which doesn't hold up on either UI Primo VE currently ships.

**Classic UI** (e.g. [swisscovery](https://zhaw.swisscovery.slsp.ch/discovery/search?query=any,contains,Organisationsentwicklung&vid=41SLSP_ZAW:ZHAW&lang=de)): `.item-title` contains a hidden delimiter span (`.media-delimiter.ng-hide`, `aria-hidden="true"`) whose `"; "` still ends up in `textContent`. The existing `replace(/^;/, '')` strips the semicolon but leaves the whitespace, so every title in the dialog is indented.

**NDE UI** (e.g. [Liège](https://explore.lib.uliege.be/discovery/search?query=any,contains,organisation&vid=32ULG_INST:ULIEGE&lang=fr)): there is no `.item-title` in the document at all, so the lookup falls through to `row.parentNode.textContent`. Titles come out as the visually hidden a11y label plus facet text:

```
"Cocher pour sélectionner la notice Organisation  1Livre Organisation  Organisati..."
```

This reads the title from `a.record-title` (NDE) or the `prm-highlight` element inside `.item-title` (classic), falling back to the old behaviour. Verified against both instances above.

The existing Liège test case didn't catch this because it only asserts `"multiple"` and never looks at a title.

Also adds two test cases for swisscovery, the SLSP union catalogue behind essentially every Swiss academic library, which had no coverage. All four tests pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tkorner:
Heads up on the red CI: it's pre-existing, not caused by this change.

The job times out after 300s waiting for `#translator-tests-complete`, with no per-test output in the log. To check whether my change was responsible, I ran a control PR on my own fork containing an *unmodified* `Primo 2018.js` — only a `lastUpdated` bump, original test cases, no code change. It fails identically:

```
Will run tests for translators ["Primo 2018","Idref"]
locator.waitFor: Timeout 300000ms exceeded.
  - waiting for locator('#translator-tests-complete')
```

[Control run](https://github.com/tkorner/translators/actions/runs/35457833975/job/105936164579) · [this PR's run](https://github.com/zotero/translators/actions/runs/35456913581/job/105933726249)

Locally the full set — 3 Primo 2018 tests plus the 3 Idref dependents — passes in 47s:

```
=== Idref === 3/3 succeeded
=== Primo 2018 === 3/3 succeeded
All tests passed
```

So something about these Primo pages hangs on load in the GitHub runner environment specifically. I couldn't narrow it down further from here — happy to dig if it's useful, but it looks like a separate issue from this fix.

One note on scope: I originally included a search-results test case for swisscovery too, but dropped it. `items: "multiple"` never asserts a title, so it wouldn't actually cover the code this PR changes, and it only added load to a job that's already timing out.

